PR #10, opened 2019-06-04
pre-filter fedmsg events before hitting the queue

TomasTomecek:


packit_service/worker/jobs.py
```diff
@@ -49,6 +49,16 @@ def add_to_mapping(kls: Type["JobHandler"]):
     return kls
 
 
+def do_we_process_fedmsg_topic(topic: str) -> bool:
+    """ do we process selected fedmsg topic? """
+    global JOB_NAME_HANDLER_MAPPING
```
jpopelka: [nitpicking]: do we need this when we don't change the value? (we just iterate over it)

packit_service/worker/jobs.py
```diff
@@ -49,6 +49,16 @@ def add_to_mapping(kls: Type["JobHandler"]):
...
+    for kls in JOB_NAME_HANDLER_MAPPING.values():
+        t = getattr(kls, "topic", "this-is-not-a-real-topic")
+        if t == topic:
```
jpopelka: Just wondering if it's more readable for you with the intermediary `t`
(than just `if getattr(kls, "topic", "not-a-real-topic") == topic:`

packit_service/worker/jobs.py
TomasTomecek: I don't know.

packit_service/worker/jobs.py
TomasTomecek: but then we need to catch the exc; ok, let me change

TomasTomecek:
:P

PR #15, opened 2019-06-14
Store task.steve_jobs.process_message results

jpopelka:


packit_service/worker/jobs.py
```diff
@@ -57,6 +65,17 @@ def do_we_process_fedmsg_topic(topic: str) -> bool:
     return topic in PROCESSED_FEDMSG_TOPICS
 
 
+class HandlerResults:
+    class Status(Enum):
```
TomasTomecek: please define it globally so we can import it from other modules as well

jpopelka:
When looking at the task in Flower one can see the input `kwargs` of the task, which contain the whole incoming event. So it AFAIK should be stored somewhere in redis (where else could Flower take it from?), but I can't see it via redis-commander.

jpopelka:
I've added project and trigger to the stored results.
We can add more once we know what we actually need.

packit_service/worker/jobs.py
```diff
@@ -66,6 +67,21 @@ def do_we_process_fedmsg_topic(topic: str) -> bool:
...
+class Status(Enum):
+    success = "success"
+    failure = "failure"
+
+
+class HandlerResults:
+    def __init__(self, status: Status, details: dict = None):
+        super().__init__()
+        self.status = status
+        self.details = details or {}
```
rpitonak: What structure should have this dict? I can see that in case of a failure you are storing `details={"error": {"msg": msg}`. This could be documented here.

rpitonak:
One more question, do we want to store also tasks that we do not process. For example, I tried to send a hook from the github-app installation which is not yet in SteveJobs and result stored is:
```
{"status": "SUCCESS", "result": null, "traceback": null, "children": [], "task_id": "50b002f7-a165-479b-b800-4bb96fe2ca3c", "date_done": "2019-06-27T13:03:35.662240"}
```

packit_service/worker/jobs.py
jpopelka: commented

jpopelka:
> do we want to store also tasks that we do not process

tl;dr; no  :-)

we don't process them because we ignore them so there's no point in storing any info

jpopelka:
I simplified & improved it. PTAL @rpitonak 
If it looks good, go ahead and merge. Thanks.

PR #16, opened 2019-06-18
sandcasterlizing

TomasTomecek:
Requires: https://github.com/packit-service/packit/pull/377
Requires: https://github.com/packit-service/sandcastle/pull/12 [merged]

### DIFF
--- a/packit_service/worker/jobs.py
+++ b/packit_service/worker/jobs.py
@@ -323,7 +323,10 @@ class JobHandler:
             logger.info("volume was not empty")
             logger.debug("content of the volume: %s" % globz)
         for item in globz:
-            shutil.rmtree(item)
+            if item.is_file():
+                item.unlink()
+            else:
+                shutil.rmtree(item)
 
     def clean(self):
         """ clean up the mess once we're done """
--- a/packit_service/worker/jobs.py
+++ b/packit_service/worker/jobs.py
@@ -45,7 +45,7 @@ from packit.exceptions import PackitException, FailedCreateSRPM
 from packit.local_project import LocalProject
 from packit.ogr_services import get_github_project
 from packit.utils import nested_get, get_namespace_and_repo_name
-
+from sandcastle import SandcastleCommandFailed, SandcastleTimeoutReached
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


packit_service/worker/jobs.py
```diff
@@ -295,9 +301,37 @@ def __init__(
...
     def run(self):
         raise NotImplementedError("This should have been implemented.")
 
+    def _clean_workplace(self):
+        logger.debug("remove contents of the PV")
+        p = Path(self.config.actions_handler_work_dir)
+        # remove everything in the volume, but not the volume dir
+        globz = list(p.glob("*"))
+        if globz:
+            logger.info("volume was not empty")
```
phracek: Please include in the info also self.config.actions_handler_work_dir, so it is clear from log what are you talking about.

packit_service/worker/jobs.py
```diff
@@ -323,7 +323,10 @@ def _clean_workplace(self):
             logger.info("volume was not empty")
             logger.debug("content of the volume: %s" % globz)
         for item in globz:
-            shutil.rmtree(item)
+            if item.is_file():
```
phracek: What about if the file is a symlink?

phracek: Sorry I did not read documentation properly (https://docs.python.org/3/library/pathlib.html#pathlib.Path.is_file). Symbolic link is counted.

packit_service/worker/jobs.py
TomasTomecek: I'd say it's not super-useful to show the path since the path is set in a config file created by us, so we know it up front what it is.

But I'd say it's a good thing to log that path at least once.

TomasTomecek: Okay, it doesn't make sense to log the handler_work_dir path explicitly since it will be shown in the log below, you mentioned:
```python
In [1]: p = pathlib.Path("/").glob("*")    

In [2]: list(p)                            
Out[2]: 
[PosixPath('/boot'),
 PosixPath('/dev'),
 PosixPath('/proc'),
 PosixPath('/run'),
 PosixPath('/sys'),
 PosixPath('/root'),
...
```

phracek:
Really big experience to me to test it. Thanks.

PR #20, opened 2019-06-21
install python-kubernetes as an rpm

TomasTomecek:


TomasTomecek:
self-merging so I can deploy

PR #21, opened 2019-06-25
Whitelist

rpitonak:
Needs:

- [x] https://github.com/packit-service/ogr/pull/81
- [x] https://github.com/packit-service/packit/pull/384


packit_service/worker/whitelist.py
```diff
@@ -0,0 +1,202 @@
...
+        """
+        If GitHub username is same as FAS username this method checks if user is packager.
+        User is considered to be packager when he/she has the badge:
+         `If you build it... (Koji Success I)`
+        :param account_login: str, Github username
+        :return: bool
+        """
+
+        url = f"https://badges.fedoraproject.org/user/{account_login}/json"
+        data = requests.get(url)
```
TomasTomecek: trying this now and it's awfully slow if you're a packager:
```
$ time curl https://badges.fedoraproject.org/user/ttomecek/json
JSON
0.05s user 0.03s system 0% cpu 38.078 total
```

How about if we queried koji or src.fp.o instead?

packit_service/worker/whitelist.py
```diff
@@ -0,0 +1,202 @@
...
+            return False
+
+    def approve_account(self, account_name: str) -> bool:
+        """
+        Approve user manually
+        :param account_name: account name for approval
+        :return:
+        """
+        if account_name in self.db:
+            account = self.db[account_name]
```
TomasTomecek: how about doing this right away and return false for keyerror?

packit_service/worker/whitelist.py
```diff
@@ -0,0 +1,202 @@
...
+
+    def remove_account(self, account_name: str) -> bool:
+        """
+        Remove account from whitelist.
+        :param account_name: github login
+        :return:
+        """
+        if account_name in self.db:
+            del self.db[account_name]
+            # TODO: delete all artifacts from copr
```
TomasTomecek: I'd place that code somewhere else, but +1 tracking it

packit_service/worker/jobs.py
```diff
@@ -228,6 +264,19 @@ def process_jobs(
...
+        # check if it is github event and account is on whitelist
+        if (
+            JobTriggerType == JobTriggerType.pull_request
+            or JobTriggerType == JobTriggerType.release
+        ):
+            whitelist = Whitelist()
+
+            if not whitelist.is_approved(project.namespace):
+                logger.error("User is not approved on whitelist!")
+                # TODO let user know that he is not whitelisted?
```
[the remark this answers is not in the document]
TomasTomecek: in the meantime, we should document this, CC @phracek

packit_service/worker/whitelist.py
lachmanfrantisek: Personally, I do not like catching exceptions if it is not needed.

What about this one:

```python
if account_name not in self.db:
    return False
```

TomasTomecek: my point is that what I'm proposing is more efficient: with this implementation we are hitting redis twice

lachmanfrantisek: OK, that's a valid reason.

I've thought you were primary mentioning this:

```python
if something:
    # some work
    return True
else:
    return False
```
:arrow_down: 
```python
if not something:
    return False
# some work
return True
```

packit_service/worker/jobs.py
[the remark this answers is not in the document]
phracek: Here is an issue https://github.com/packit-service/packit.dev/issues/16

packit_service/worker/whitelist.py
```diff
@@ -0,0 +1,202 @@
...
+        """
+
+        url = f"https://badges.fedoraproject.org/user/{account_login}/json"
+        data = requests.get(url)
+        if not data:
+            return False
+        assertions = data.json().get("assertions")
+        if not assertions:
+            return False
+        for item in data.json().get("assertions"):
```
phracek: There could be `assertions` and not calling `data.json().get("assertions")` again.

packit_service/worker/whitelist.py
rpitonak: I am aware of this fact, the question is if we are aiming for speed here.

rpitonak:
Rebased & Fixed. PTAL

rpitonak:
#30 should be fixed by the latest commit

TomasTomecek:
Merging since this is now deployed in prod. I can see 2 unresolved discussions though.

PR #44, opened 2019-07-15
do not validate signatures when testing locally

rpitonak:


packit_service/service/web_hook.py
```diff
@@ -76,7 +77,12 @@ def validate_signature() -> bool:
     """
     if "X-Hub-Signature" not in request.headers:
         # no signature -> no validation
-        return True
+        deployment = os.environ.get("DEPLOYMENT", None)
+        if deployment:
+            if deployment == "dev":  # don't validate signatures when testing locally
+                return True
+
+        return False
```
jpopelka: Could we squash it into:

```python
if "X-Hub-Signature" not in request.headers:
    # don't validate signatures when testing locally
    return os.getenv("DEPLOYMENT") == "dev"
```

jpopelka:
> I'd personally prefer to have a service config where this would be set.

Where what would be set? Can you perhaps create an issue with some more details?

rpitonak:
Thank you for the reviews, PTAL.

PR #56, opened 2019-07-22
Get default copr user from copr config

jpopelka:
See https://github.com/packit-service/packit/pull/401/commits/690eee5bb85d3c1aebc574aaf89d60333a90b04d

### DIFF
--- a/packit_service/worker/jobs.py
+++ b/packit_service/worker/jobs.py
@@ -662,7 +662,7 @@ class GithubCoprBuildHandler(JobHandler):
     #     api = PackitAPI(self.config, self.package_config, local_project)
 
     #     build_id, repo_url = api.run_copr_build(
-    #         owner=self.job.metadata.get("owner") or "packit",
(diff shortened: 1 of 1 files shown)


packit_service/worker/jobs.py
```diff
@@ -736,11 +737,9 @@ def handle_pull_request(self):
             msg = f"There was an error while running the build: {ex}"
             r.report("failure", msg)
             return HandlerResults(success=False, details={"msg": msg})
-        timeout = 60 * 60 * 2
-        # TODO: document this and enforce int in config
```
TomasTomecek: I expect a docs PR now :P

TomasTomecek: on the other hand, the timeout will be obsolete once we start using fedmsg for completed copr builds, instead of polling

TomasTomecek: hence I wouldn't document it

PR #57, opened 2019-07-23
store data about what we do

TomasTomecek:
sending this out so you know what I'm working on; CC @dhodovsk @rpitonak

### DIFF
--- a/packit_service/service/models.py
+++ b/packit_service/service/models.py
@@ -39,7 +39,7 @@ class Model:
                 last_pk = 0
             self.identifier = last_pk + 1
             db[LAST_PK] = self.identifier
-        self.db()[self.identifier] = self.serialize()
+        db[self.identifier] = self.serialize()
 
     def serialize(self):
         """ convert from python data structure into format acceptable by redis """
--- a/packit_service/service/models.py
+++ b/packit_service/service/models.py
@@ -1,0 +1,189 @@
+"""
+Data models for jobs, tasks, builds, etc. The models are mapped to redis.
+"""
+import copy
+import datetime
+from typing import List, Dict, Union
+
+from frambo.dict_in_redis import PersistentDict
+
+from packit_service.service.events import InstallationEvent, WhitelistStatus
+
+
+# if identifier is not set, we need to create primary key ourselves
+# with this variable, we keep track on what's the last PK
+LAST_PK = "last-pk"
+
+
+class Model:
+    """ Abstract representation of a single object passing through p-s """
+
+    # a "table" to store collection of objects
+    table_name: str
+    # unique identifier
+    identifier: Union[int, str] = None
+
+    @classmethod
+    def db(cls) -> PersistentDict:
+        if not cls.table_name:
+            raise RuntimeError("table_name is not set")
+        return PersistentDict(hash_name=cls.table_name)
+
+    def save(self):
+        """ store the current state of the object inside redis """
+        db = self.db()
+        if not self.identifier:
+            last_pk = db[LAST_PK]
+            if last_pk is None:
+                # we are going to insert first item
+                last_pk = 0
+            self.identifier = last_pk + 1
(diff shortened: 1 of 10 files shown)


packit_service/service/models.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class User(Model):
+    """ A GitHub user """
+
+    table_name = "users"
+    username: str
+
+
+class Installation:
```
rpitonak: I know that it is still WIP, but why does `Installation` not inherit from `Model`? And could we use here `events.py` classes somehow?

TomasTomecek: I worked on this before you opened yours, definitely planning on using your work and change this code.

I'm also planning to scope this down and only do tasks, builds and installs

packit_service/service/models.py
```diff
@@ -0,0 +1,174 @@
...
+        data = self.__dict__
+        cp = copy.deepcopy(data)
+        if "table_name" in cp:
+            del cp["table_name"]
+        return cp
+
+    def deserialize(self, inp: Dict):
+        """ reverse operation as serialize """
+        # this is pretty nasty: we could possibly replace this with a serialization
+        # library or an ORM framework
```
rpitonak: +1 for research of alternatives

packit_service/service/models.py
```diff
@@ -0,0 +1,174 @@
...
+        self.__dict__ = inp
+
+    @classmethod
+    def all(cls):
+        """ get all instances """
+        response = []
+        for key, data in cls.db().items():
+            if key == "last_index":
```
rpitonak: what is this constant?

files/recipe.yaml
```diff
@@ -80,10 +80,10 @@
...
       name: git+https://github.com/packit-service/packit.git
       executable: pip3
-  - name: Install ogr from git master
-    pip:
-      name: git+https://github.com/packit-service/ogr.git
-      executable: pip3
+#  - name: Install ogr from git master
+#    pip:
+#      name: git+https://github.com/packit-service/ogr.git
+#      executable: pip3
```
rpitonak: Just because of failing imports?

TomasTomecek: Yes, because of this: https://github.com/packit-service/packit/pull/425

packit_service/service/models.py
TomasTomecek: I did not want to use any library here since it would make this PR gigantic

I also think that it should be fairly easy to migrate to a proper ORM solution in future

packit_service/service/models.py
TomasTomecek: I'll document.

TomasTomecek:
updated

rpitonak:
Tests are still failing, but I am not able to see the logs.

TomasTomecek:
rebased, let's see

TomasTomecek:
wanna see some greens now!

packit_service/service/models.py
```diff
@@ -0,0 +1,189 @@
...
+        """ store the current state of the object inside redis """
+        db = self.db()
+        if not self.identifier:
+            last_pk = db[LAST_PK]
+            if last_pk is None:
+                # we are going to insert first item
+                last_pk = 0
+            self.identifier = last_pk + 1
+            db[LAST_PK] = self.identifier
+        self.db()[self.identifier] = self.serialize()
```
jpopelka:
```suggestion
        db[self.identifier] = self.serialize()
```

TomasTomecek:
Accepted, would wait for CI now. Rado, btw I fear that my PR will make tests in your PR fail. Hopefully it won't be that bad to fix them.

rpitonak:
> Accepted, would wait for CI now. Rado, btw I fear that my PR will make tests in your PR fail. Hopefully it won't be that bad to fix them.

I am aware of this, no problem I will handle it.

Issue #12, opened 2019-06-06
celery should log less


TomasTomecek:
It seems that celery has some crazy logging on by default: complete http response data are being logged:
```
[2019-06-06 12:52:30,789: DEBUG/ForkPoolWorker-1] https://src.fedoraproject.org:443 "GET /modules/pretty-git-prompt/raw/rolling/f/.packit.json HTTP/1.1" 404 3327
[2019-06-06 12:52:30,790: DEBUG/ForkPoolWorker-1] 
<!DOCTYPE html>
<html lang='en'>
<head>
    <meta http-equiv="Content-Type" content="text/html; charset=UTF-8"/>
    <title>Page not found :'( - src.fedoraproject.org</title>
        <link rel="shortcut icon" type="image/vnd.microsoft.icon"
        href="/theme/static/favicon.ico?version=5.5"/>
    <link href="/theme/static/fedora-bootstrap-1.3.0/fedora-bootstrap.min.css?version=5.5"
        type="text/css" rel="stylesheet" />
    <link href="/theme/static/fonts/fonts.css?version=5.5"
        rel="stylesheet" type="text/css" />
    <link href="/theme/static/fonts/hack_fonts/css/hack-extended.min.css?version=5.5"
        type="text/css" rel="stylesheet" />
    <link href="/theme/static/theme.css?version=5.5"
        type="text/css" rel="stylesheet" />

    <link href="/static/pagure.css?version=5.5"
    type="text/css" rel="stylesheet" />
    <link href="/static/vendor/font-awesome/font-awesome.css?version=5.5"
    type="text/css" rel="stylesheet" />
  </head>
  <body id="error">
    
    <!-- start masthead -->
```

would be nice to get rid of it

the action here is to configure logging in celery so that:
* logs from requests python package are ignored (or urllib3, not sure where it's coming from)
* also get rid of these lines in logs: `[2019-07-10 13:22:22,017: DEBUG/MainProcess] pidbox received method enable_events() [reply_to:None ticket:None]`

TomasTomecek:
I'll pay for a dinner if you fix this.

rpitonak:
> I'll pay for a dinner if you fix this.

We should create new label: `reward-included` :D

phracek:
That's a challenge. Wait for a new sprint and I will fix it :)

PR #60, opened 2019-07-25
packit-service config and PR status reports from stg app

rpitonak:
TODO:

- [x] tests 

In the ideal case, this PR will be followed up by:
 * Remove sandcastle from packit completely
 * Packit API is using user config in the constructor, but now we are passing the service config there. This is why there is still `github_token` even if it is not necessary for the service. A solution could be to create an abstract class for these two configs.

Tested here.
https://github.com/packit-service/hello-world/pull/10

### DIFF
--- a/packit_service/config.py
+++ b/packit_service/config.py
@@ -34,7 +34,7 @@ from packit_service.constants import (
     SANDCASTLE_PVC,
     SANDCASTLE_IMAGE,
     SANDCASTLE_DEFAULT_PROJECT,
-    CONFIG_FILE_NAMES,
+    CONFIG_FILE_NAME,
 )
 from packit_service.schema import SERVICE_CONFIG_SCHEMA
 
@@ -45,17 +45,6 @@ class Deployment(enum.Enum):
     dev = "dev"
     stg = "stg"
     prod = "prod"
-
-    @staticmethod
-    def from_str(label):
-        if label == "stg":
-            return Deployment.stg
-        elif label == "prod":
-            return Deployment.prod
-        elif label == "dev":
-            return Deployment.dev
-        else:
-            raise NotImplementedError
 
 
 class Config(BaseConfig):
@@ -111,7 +100,7 @@ class Config(BaseConfig):
         config.github_app_id = raw_dict.get("github_app_id", "")
         config.github_app_cert_path = raw_dict.get("github_app_cert_path", "")
         config.webhook_secret = raw_dict.get("webhook_secret", "")
-        config.deployment = Deployment.from_str(raw_dict.get("deployment", ""))
+        config.deployment = Deployment(raw_dict.get("deployment", ""))
         config.validate_webhooks = raw_dict.get("validate_webhooks", False)
 
         config.command_handler = RunCommandType.local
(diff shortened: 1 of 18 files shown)


packit_service/config.py
```diff
@@ -0,0 +1,167 @@
...
+logger = logging.getLogger(__name__)
+
+
+class Deployment(enum.Enum):
+    dev = "dev"
+    stg = "stg"
+    prod = "prod"
+
+    @staticmethod
+    def from_str(label):
```
TomasTomecek: I don't think this method is needed:
```python
In [1]: from packit_service.config import Deployment                                  

In [2]: Deployment("dev")                  
Out[2]: <Deployment.dev: 'dev'>

In [3]: Deployment("lol")                  
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)    
ValueError: 'lol' is not a valid Deployment
```

packit_service/config.py
```diff
@@ -0,0 +1,167 @@
...
+        )
+        # default project for oc cluster up
+        config.command_handler_k8s_namespace = raw_dict.get(
+            "command_handler_k8s_namespace", SANDCASTLE_DEFAULT_PROJECT
+        )
+
+        return config
+
+    @classmethod
+    def get_service_config(cls) -> "Config":
```
TomasTomecek: I'd say that having this method here is an overkill: I'd say it's just enough to have a default path and make it changeable via an env var.

packit_service/constants.py
```diff
@@ -1 +1,17 @@
 FAQ_URL = "https://packit.dev/packit-as-a-service/#faq"
+SANDCASTLE_WORK_DIR = "/sandcastle"
+SANDCASTLE_IMAGE = "docker.io/usercont/sandcastle"
+SANDCASTLE_DEFAULT_PROJECT = "myproject"
+SANDCASTLE_PVC = "SANDCASTLE_PVC"
+
+CONFIG_FILE_NAMES = [
```
TomasTomecek: ^ just one config name

rpitonak:
Fixes also #12. (This PR is live on stg)

packit_service/worker/jobs.py
```diff
@@ -58,6 +59,18 @@ def process_jobs(self, event: Optional[Any]) -> Dict[str, HandlerResults]:
         """
         handlers_results = {}
         package_config = event.get_package_config()
+
+        if not package_config:
```
TomasTomecek: +1 nice

packit_service/worker/tasks.py
```diff
@@ -19,12 +19,18 @@
...
 
 from packit_service.celerizer import celery_app
 from packit_service.worker.jobs import SteveJobs
 
+logging.getLogger("requests").setLevel(logging.WARNING)
+logging.getLogger("urllib3").setLevel(logging.WARNING)
+logging.getLogger("github").setLevel(logging.WARNING)  # pygithub
+logging.getLogger("ogr").setLevel(logging.WARNING)  # b/c of pagure requests
+logging.getLogger("kubernetes").setLevel(logging.WARNING)
+
```
TomasTomecek: cheapest dinner ever! only 7 lines

rpitonak:
> Didn't review deeply since there is so many changes.
> 
> Is there some changes needed to be done in the deployment?

Indeed, I will open PR in the deployment repository.

rpitonak:
CI is green ✅

rpitonak:
Shall we merge?

PR #63, opened 2019-07-29
Collaborators can trigger Packit-as-a-Service

phracek:
Signed-off-by: Petr "Stone" Hracek <phracek@redhat.com>

This pull request adds function when only `collaborators` for given repo can trigger Packit-as-a-Service.

- [x] test for collaborators is included.

### DIFF
--- a/packit_service/worker/github_handlers.py
+++ b/packit_service/worker/github_handlers.py
@@ -253,7 +253,7 @@ class GithubCoprBuildHandler(AbstractGithubJobHandler):
         r = BuildStatusReporter(self.project, self.event.commit_sha)
         if self.event.github_login not in collaborators:
             msg = "Only collaborators can trigger Packit-as-a-Service"
-            r.reset_status("failure", msg)
+            r.set_status("failure", msg)
             return HandlerResults(success=False, details={"msg": msg})
         try:
             build_id, repo_url = self.api.run_copr_build(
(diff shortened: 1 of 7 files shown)


packit_service/worker/github_handlers.py
```diff
@@ -247,10 +247,14 @@ def handle_pull_request(self):
         default_project_name = (
             f"{self.project.namespace}-{self.project.repo}-{self.event.pr_id}"
         )
+        collaborators = self.project.who_can_merge_pr()
         project = self.job.metadata.get("project") or default_project_name
         owner = self.job.metadata.get("owner") or self.api.copr.config.get("username")
         r = BuildStatusReporter(self.project, self.event.commit_sha)
-
+        if self.event.github_login not in collaborators:
```
TomasTomecek: do we want to make this configurable? so that upstream repos would say that they want to check PRs submitted by anyone?

packit_service/worker/handler.py
```diff
@@ -63,6 +63,12 @@ def report(
             self.commit_sha, state, url, description, "packit/rpm-build"
         )
 
+    def reset_status(self, state: str, description: str):
```
TomasTomecek: nit: shouldn't this be set_status?

packit_service/worker/github_handlers.py
phracek: Personally, I don't want to allow it. Like in `sclorg` case. Some have to call trigger.

packit_service/worker/handler.py
phracek: reset mean, get rid of it to initial state. Will change it.

TomasTomecek: ah, okay, didn't know it would reset the state since the state is an argument

packit_service/worker/github_handlers.py
TomasTomecek: fair enough, but I'm pretty sure we'll receive such request fairly soon

PR #65, opened 2019-07-30
Fix imports from packit/ogr

lachmanfrantisek:
- Stop importing packit.github_service.

### DIFF
--- a/packit_service/service/events.py
+++ b/packit_service/service/events.py
@@ -24,20 +24,19 @@
 This file defines classes for events which are sent by GitHub or FedMsg.
 """
 import enum
+from pathlib import Path
 
+from ogr import PagureService, GithubService
 from ogr.abstract import GitProject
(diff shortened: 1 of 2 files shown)


packit_service/service/events.py
```diff
@@ -58,9 +57,31 @@ class WhitelistStatus(enum.Enum):
 class Event:
     def __init__(self, trigger: JobTriggerType):
         self.trigger: JobTriggerType = trigger
+        self._user_config = None
+
+    @property
+    def user_config(self):
```
phracek:
```
def user_config(self) -> Config:
```

rpitonak: user_config is going to be replaced by service_config in #60

lachmanfrantisek: Do I need to fix that here?

rpitonak: no need, I will do while rebasing my PR.

PR #66, opened 2019-07-30
Make fedpkg clone work in openshift

TomasTomecek:
Fixes https://github.com/packit-service/packit/issues/431

Edit: description on Rado's request
* passwd: fedpkg needs that otherwise it complains that current user doesn't have an entry in /etc/passwd - SCL uses this method in their images
* .ssh - ssh command is very picky about perms of .ssh and id_rsa; stolen from Betka
* I also added a new command which accepts events via CLI and processes them directly for easier local development
* and a gitconfig which makes a nice commit author instead of (not using `git config` b/c we change users and it likely ends up in /opt and not in /home/packit): https://src.fedoraproject.org/rpms/python-ogr/c/07797442e8b27e6ea785b39b4b6d5cfc5f6ef0a1?branch=master

### DIFF
--- a/packit_service/cli/process_message.py
+++ b/packit_service/cli/process_message.py
@@ -1,0 +1,55 @@
+# MIT License
+#
+# Copyright (c) 2018-2019 Red Hat, Inc.
+
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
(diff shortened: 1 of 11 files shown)


packit_service/cli/process_message.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+@click.command("process-message")
+@click.argument("path", nargs=1, required=False)
+def process_message(path):
+    """
+    Accept a message from commandline and process it directly - bypass celery
+
+    Either provide a filename with the message or pipe it:
+      cat event.json | packit-service process-message
```
lachmanfrantisek: Maybe you can use builtin support in click for that:

- docs: https://click.palletsprojects.com/en/7.x/arguments/#file-arguments
- example: https://github.com/pallets/click/blob/master/examples/inout/inout.py

lachmanfrantisek:
```sh
cat event.json | packit-service process-message -
packit-service process-message event.json
```

TomasTomecek:
Rerady for review, I added a description.

PR #76, opened 2019-08-02
whitelisting: fix a bug & don't overwrite records

TomasTomecek:
'ValueError: 'WhitelistStatus.approved_automatically' is not a valid
WhitelistStatus'


TomasTomecek:
glad I added the tests: there was a bug in the former version :D

tests/unit/test_whitelist.py
```diff
@@ -0,0 +1,67 @@
...
+            return None
+
+
+@pytest.fixture()
+def db():
+    return GracefulDict(
+        {
+            "fero": {"status": WhitelistStatus.approved_manually.value},
+            "lojzo": {"status": str(WhitelistStatus.approved_automatically)},
+            "konipas": {"status": WhitelistStatus.waiting.value},
```
rpitonak: konipas +1 😁

PR #88, opened 2019-08-19
Prepare Zuul for the E2E tests for pull-requests

rpitonak:
This PR introduces Zuul pipeline for testing PRs. The workflow is:
1. Openshift cluster is set up on the test node
2. Images (worker and service) are built from the source (PR branch).
3. Images are deployed into an Openshift cluster with fake secrets.
4. `make check_in_container` is executed

This allows us to introduce E2E tests, which does not require to communicate with real services (e.g github and copr).


.zuul.yaml
```diff
@@ -3,7 +3,67 @@
...
+      - files/install-openshift.yaml
+    run: files/zuul-tests.yaml
+    extra-vars:
+      with_testing: true
+      ansible_python_interpreter: /usr/bin/python3
+    nodeset:
+      nodes:
+        - name: test-node
+          label: cloud-fedora-30
+    secrets:
```
morucci: Zuul config elements are not Ansible elements so you cannot use with_items there. Secrets stanza must be a list of Zuul secrets. https://zuul-ci.org/docs/zuul/user/config.html#attr-job.secrets, so something like that:

```
secrets:
  - copr
  - fedora.keytab
  - ...
```
Zuul look at this list and expect to find Zuul secrets with those names.

.zuul.yaml
```diff
@@ -3,7 +3,67 @@
...
-        - noop
+        - packit-service-pre-commit
+        - packit-service-tests
     gate:
       jobs:
-        - noop
+        - packit-service-pre-commit
+        - packit-service-tests
+
+- import_playbook: ./files/zuul/{{ item }}-secret.yaml
```
morucci: .zuul.yaml is just a yaml file, and will not be process by ansible. import_playbook won't work there. Instead you should add the secrets in that file directly. For each repository, Zuul look at a file .zuul.yaml or a directory zuul.d/*.yaml to find the zuul config elements, other files won't be read.

.zuul.yaml
```diff
@@ -3,7 +3,67 @@
...
     check:
       jobs:
-        - noop
+        - packit-service-pre-commit
+        - packit-service-tests
     gate:
       jobs:
-        - noop
+        - packit-service-pre-commit
+        - packit-service-tests
```
morucci: I see you want to use secrets in pre-review jobs. For security reasons Zuul won't let you use secrets defined in a untrusted project. If the job was running is a post-review pipeline like the post pipeline it would have been fine, but with untrusted repositories Zuul reads and load the .zuul.yaml change from the PR then it could be easy to propose a PR with playbook changes and take advantages of the jobs privileges thanks to the secrets. So the secrets and playbooks, jobs that use the secrets should be added in a trusted (or config (this the same)) repository. Zuul does not read/loads the PR changes on the trusted repositories. Set up pre-review jobs to use secrets is not obvious so please read https://www.softwarefactory-project.io/zuul-hands-on-part-5-job-secrets.html it will give insight but it does not cover your use case.

For your use case you need to have the secrets, the playbooks and the jobs that use the secrets defined in a trusted (or config) repository. I advice you to start with a test job that only ensure you have access to the secret in the pre-run or run playbook. Let's create a repository packit-zuul-config and we will register it as a zuul trusted repo in sf-project.io as a starting point. I also need to check with @TristanCacqueray if we should create a Zuul packit tenant from our side.

phracek:
Rebase from master and tests should passed (https://github.com/packit-service/packit-service/pull/89/commits/639e77d7449b0945c94aa5ef786207cda6f66db3)

files/deploy-service.yaml
```diff
@@ -0,0 +1,88 @@
+---
+- name: Deploy packit-service into OpenShift
+  hosts: all
+  tasks:
+  - name: Download the deployment
+    git:
+        repo: 'https://github.com/packit-service/deployment.git'
```
TristanCacqueray: It's better to put that project as a `required-project` instead, then zuul will take care of copying the source.

.zuul.yaml
```diff
@@ -3,7 +3,67 @@
...
+        - copr
+        - fullchain.pem
+        - httpd-packit.conf
+        - id_rsa
+        - id_rsa.pub
+        - packit-service.yaml
+        - private-key.pem
+        - privkey.pem
+        - ssh_config
+        - fedora.keytab
```
TristanCacqueray: Using such secrets in a check job is a security concern. A follow-up PR may be able to change the run task to leak those files. Instead, it seems like this job should be managed in a config project, like that, the job change won't be tested speculatively and it will be possible to run check test with secret (assuming the job doesn't run anything locally from the PR content). We can try something like that, in a project named `config-project` for example:

```yaml
# github.com/packit-service/config-project/zuul.yaml
---
- job:
    name:  packit-deployment-test
    required-projects:
      - packit-service/packit-deployment
    allowed-projects:
      - <name-of-project-that-can-use-the-job-in-check>
    final: true # this prevent job modification
    secrets: []
    run: playbooks/test-deployment.yaml # this file is in a config repository, preventing speculative test
```

Also, since the resources may persist accross job run, we should make sure a [cleanup](https://zuul-ci.org/docs/zuul/user/config.html#attr-job.cleanup-run) is setup.

Alternatively, zuul can provides an openshift project as a test resources if we add your cluster in nodepool.

rpitonak: > Using such secrets in a check job is a security concern. A follow-up PR may be able to change the run task to leak those files. Instead, it seems like this job should be managed in a config project, like that, the job change won't be tested speculatively and it will be possible to run check test with secret (assuming the job doesn't run anything locally from the PR content). We can try something like that, in a project named `config-project` for example:
> 
> ```yaml
> # github.com/packit-service/config-project/zuul.yaml
> ---
> - job:
>     name:  packit-deployment-test
>     required-projects:
>       - packit-service/packit-deployment
>     allowed-projects:
>       - <name-of-project-that-can-use-the-job-in-check>
>     final: true # this prevent job modification
>     secrets: []
>     run: playbooks/test-deployment.yaml # this file is in a config repository, preventing speculative test
> ```
> 

I already created repository here: https://github.com/packit-service/packit-service-zuul

> Also, since the resources may persist accross job run, we should make sure a [cleanup](https://zuul-ci.org/docs/zuul/user/config.html#attr-job.cleanup-run) is setup.
> 

+1

files/deploy-service.yaml
rpitonak: Nice

morucci:
I've proposed to add packit-service-zuul repository into zuul https://softwarefactory-project.io/r/16097 as trusted repo. You can push a simple .zuul.yaml with noop jobs to test the connectivity with Zuul.

morucci:
We won't create a dedicated tenant on Zuul for packit for the moment so you can process as usual (@rpitonak so I don't need the write access to packit-service-zuul repo write access to initialize it)

files/deploy-service.yaml
rpitonak: But in this case do I have to setup zuul on this repo also? https://github.com/packit-service/deployment/

TristanCacqueray: You don't have to setup zuul for the deployment project (though you might want to gate it once there is a test for it), but we need to tell zuul about it, e.g. with https://softwarefactory-project.io/r/16110

When the Zuul application is installed on a project, you don't have to setup zuul.yaml in it, instead you can also configure the pipelines from the packit-service-zuul project.

rpitonak: Ok got it, using which variable I can then access the project in other playbooks? 

E.g, when I add:
```
required-project: packit-service/deployment
```

Where it is going to be cloned?

TristanCacqueray: The project will be added the zuul.projects structure (you can see the variables here in the [zuul-info/inventory.yaml](https://softwarefactory-project.io/logs/88/88/7fe4b2bd187c30fce0fcef20ce110c4827f03e1e/check/packit-service-pre-commit/3113117/zuul-info/inventory.yaml) .
So that would be `{{ ansible_user_dir }}/{{ zuul.projects['github.com/packit-service/deployment'].src_dir }}`

rpitonak:
Ok, right now we need to do tasks in this order:
- files/install-openshift.yaml
- files/packit-service-image.yaml
- then execute playbook from [packit-service/packit-service-zuul](https://github.com/packit-service/packit-service-zuul/blob/master/files/deploy-service.yaml)
- run files/zuul-tests.yaml

Can you help me how can I write `.zuul.yaml` for this use case? @TristanCacqueray @morucci.

TristanCacqueray:
@rpitonak I'm not sure to understand. It seems like this PR is now installing a local openshift cluster on the test instance, but then why would you run that task: https://github.com/packit-service/packit-service-zuul/blob/master/files/deploy-service.yaml#L20-L40 ?

rpitonak:
@TristanCacqueray because we want to be able to run E2E tests and use our staging secrets to execute some actions. In those lines, I am just creating [files structure](https://github.com/packit-service/deployment/tree/master/secrets) which is needed for our deployment playbooks.

TristanCacqueray:
@rpitonak alright, then all the playbooks and tasks that requires secret needs to be managed by the config project packit-service-zuul.  The job in packit-service-zuul shouldn't run the make command as a malicious user could propose a PR on packit-service that makes the make command dump all the secrets installed by zuul.

Instead, the packit-service-zuul job should do the docker build and push the image to the cluster.

Why can't the E2E tests run on the cluster installed in the test node?

rpitonak:
> @rpitonak alright, then all the playbooks and tasks that requires secret needs to be managed by the config project packit-service-zuul. The job in packit-service-zuul shouldn't run the make command as a malicious user could propose a PR on packit-service that makes the make command dump all the secrets installed by zuul.
> 
> Instead, the packit-service-zuul job should do the docker build and push the image to the cluster.

Docker build where the source will be content of the PR on packit-service/packit-service?
> 
> Why can't the E2E tests run on the cluster installed in the test node?

Yes it is one of the ways which we consider to use instead of openhsift cluster.

TristanCacqueray:
Yes, the playbook content and tasks from the config projects are not affected by un-merged (speculative) change, but they can act on the speculative content. e.g. if in packit-service-zuul you have a task that does "docker build ." in zuul.projects["packit-service"].src_dir, then it will build the speculative content of the project. The goal is to prevent speculative content from accessing the secrets.

Perhaps there is still a risk of malicious Dockerfile content to get access to the user home? If that's a concern, then we should run docker build from another instance.
Similarly, once the packit-service image is pushed to the openshift registry, you should be able to run safely "oc apply -f ./..." in the packit-service PR repository.

rpitonak:
@TristanCacqueray. If the workflow would be like this:

1. We set up `oc cluster up` in the Zuul (introduced by this PR)
2. We build a new image with the content of tested PR (e.g `make build` of this repository)
3. We run the [packit-service-zuul](https://github.com/packit-service/packit-service-zuul) job. (This logs shouldn't be visible right?). This job deploys image built in the previous step, with all secrets and deployment configs into OpenShift cluster.
4. In this step, we would have an Openshift cluster with packit-service image deployed and we can run our test suite in a clean environment.

Possible risks to this approach, which cannot be avoided in my opinion, are that if someone alters the code in this repository, he/she can easily print the content of /secrets directory which is mounted to the container once it is deployed to Openshift. Correct me if I am wrong, but even if we have an Openshift instance remotely we cannot prevent this from happening. But we can prevent this (to some extent) by triggering Zuul just by collaborators (which is the default of Zuul I guess).

Would this workflow be possible? Sorry for having this long discussion, but I think I am missing something. We can discuss it on IRC also if it works better for you.

TomasTomecek:
The problem here is also that we don't inject secrets into the container image: all the sensitive data are openshift secrets.

+1 Rado, I feel like that no matter what we do, someone can easily jump in and write a test which prints content of the secrets files.

Honestly, is there even a way around? (apart from triggering CI jobs only by collaborators)

TristanCacqueray:
@rpitonak
1/ there is even a role in zuul-jobs for that:
   https://zuul-ci.org/docs/zuul-jobs/container-roles.html#role-install-openshift and
   https://zuul-ci.org/docs/zuul-jobs/container-roles.html#role-deploy-openshift
3/ zuul job logs are always visible, if a command prints a secret you need to set the no_log task attribute. packit-service-zuul is a config repository, that means its zuul configuration (and the job playbooks) are not loaded speculatively from PR, zuul only execute merged content.
4/ if you have a clean environment with a local cluster, why would you need secrets?

Zuul can not restrict job to collaborators, it will act on any pull request. The job though can have extra task to check the git author, or even query the pr comments to check for a "/ok-to-test" type comment, but that needs to be done at the job layer.

There is also a set of job in zuul-jobs that can be used to do image building in a separate vm with a shared registry for the other job: https://zuul-ci.org/docs/zuul-jobs/docker-image.html#run-an-intermediate-container-registry
That would be the cleanest way of doing that, since the child job can be a trusted playbook which would just pull the image and push it to your external cluster with secret.

Another solution would be to run the deployment test only in the gate, that is, after collaborators review and approve a change.

Using secret in job running as pre-merge test (with zuul) is tricky, there are different strategy that can be used...

TristanCacqueray:
While working on pre tasks, you can add the job attribute `attempt: 1` to prevent the job to be retried 3 times before failing.

files/deployment.yaml
```diff
@@ -0,0 +1,9 @@
+---
+- name: Build packit-service image form current src dir.
+  hosts: all
+  tasks:
+  - name: Build packit-service image from current src dir
+    command: "DEPLOYMENT=zuul make deploy"
+    args:
+      chdir: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/packit-service/deployment'].src_dir }}"
```
TristanCacqueray: It seems like packit-service/deployment isn't part of the required-projects

.zuul.yaml
```diff
@@ -3,7 +3,40 @@
...
+      nodes:
+        - name: test-node
+          label: cloud-fedora-30
+
+- job:
+    name: packit-service-tests
+    parent: base
+    description: Run tests of packit via rpms
+    required-projects:
+      - packit-service/packit-service-zuul
```
TristanCacqueray: i think that should be `packit-service/deployment`

rpitonak:
recheck

rpitonak:
recheck

files/deployment.yaml
```diff
@@ -0,0 +1,80 @@
...
+    command: "openssl req -x509 -new -key {{ deployment_dir }}/secrets/dev/privkey.pem -out {{ deployment_dir }}/secrets/dev/{{ item }} -days 1 -subj '/CN=just.for.the.ci' -passin pass:somepassword -passout pass:anotherpassword -nodes"
+    with_items:
+    - cert.pem
+    - chain.pem
+    - fullchain.pem
+
+  - name: Copy the rest of the secrets from template
+    copy:
+      src: "{{ deployment_dir }}/secrets/template/{{ item }}"
+      dest: "{{ deployment_dir }}/secrets/dev/{{ item }}"
```
morucci: use remote_src: true

rpitonak:
recheck

rpitonak:
recheck

rpitonak:
@morucci @TristanCacqueray can you see what is wrong here? because the file is created in previous step.

TristanCacqueray:
@rpitonak this is because the ansible template modules is looking for file on the executor (localhost). You don't have to do copy + template, the copy content variable is already considered a template and the content should be replaced. Thus remove the template task, and make the copy module write directly to dev.yml

rpitonak:
rebased

files/deployment.yaml
```diff
@@ -0,0 +1,96 @@
+---
+- name: Build packit-service image form current src dir.
+  hosts: all
+  vars:
+    deployment_dir: "{{ ansible_user_dir }}/{{ zuul.projects['github.com/packit-service/deployment'].src_dir }}"
+  tasks:
+  - name: Install packages for deployment
+    dnf:
+      name:
+      - ansible-2.7.10-1.fc30 # because of packit-service deployment
```
TomasTomecek: I fear this will stop working once that build stops being in a compose. I guess we can solve it once that happens.

rpitonak:
How to do encryption is described here: https://github.com/packit-service/deployment/pull/2.

rpitonak:
@TomasTomecek Write-up how it works is in the description of the PR.

TomasTomecek:
@rpitonak nice, thank you! add it to contributing.md now please :)

rpitonak:
PTAL.

PR #92, opened 2019-08-22
check private repos

### LINKED ISSUE #53: disable interacting with private GitHub repos
Since all the services we interact with are open, we shouldn't be able to work with private GitHub repositories - the code might not be open source.

Also add new check to whitelist whether the repo is private or not.

lbarcziova:
Fixes #53


TomasTomecek:
I agree with Jirka that putting this to parser or even SteveJobs makes more sense.

Sorry about the wrong design in the issue, it was really just a brain dump.

Also LGTM the code.

PR #130, opened 2019-09-12
Copr build - actually checkout PR

rpitonak:
This fix is a one-liner.

Relevant code: https://github.com/packit-service/packit/blob/master/packit/local_project.py#L106-L109


TomasTomecek:
awesome, Rado-boss!

PR #148, opened 2019-09-25
We don't need to store passwd file in git, do we?

jpopelka:


docker-compose.yml
```diff
@@ -57,7 +57,7 @@ services:
       - ./secrets/fedora.keytab:/secrets/fedora.keytab:ro,z
       - ./secrets/private-key.pem:/secrets/private-key.pem:ro,z
       #- .:/src:ro,z
-    # user: "123123"
+    user: "123123"
```
phracek: Is this intended? I guess we want an OpenShift user.

Code from betka:
```bash
    export USER_ID=$(id -u)
    export GROUP_ID=$(id -g)
    echo "betka:x:${USER_ID}:${GROUP_ID}:Sync bot from upstream to downstream:${HOME}:/bin/bash" >> "$HOME/passwd"
 

```

jpopelka: Yes, it is intended and it's there to run the container with 'random' user id (which is what Openshift does).
`docker-compose.yml` has nothing to do with OpenShift - I (probably no one else uses it) use `docker-compose` to quickly test service/worker without having to deploy to OpenShift.

Regarding your code from betka - not sure what you want me to show.

What change to you actually request?

PR #150, opened 2019-10-01
Unify config

lachmanfrantisek:
- [x] Change Config to be a subclass of Config from packit.
- [x] Use the new authentication method.
- [x] Get the project objects via `config.get_project(url)`.
~- [ ] backward-compatibility (new authentication)~

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -152,3 +152,14 @@ Once we have jobs defined in `packit` config we are ready to move on to next ste
 
 Thank you for your interest!
 packit team.
+
+### Service configuration
+
+The service configuration is an extension of the user configuration from packit. (`Config` class in [packit/config.py](https://github.com/packit-service/packit/blob/master/packit/config.py).)
+
+To add a new service-related property you need to:
+
+1. Add a property to `ServiceConfig.__init__` in [packit/config.py](https://github.com/packit-service/packit/blob/master/packit/config.py).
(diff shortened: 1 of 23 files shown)


lachmanfrantisek:
recheck

packit_service/service/events.py
```diff
@@ -418,6 +396,4 @@ def get_package_config(self):
         return package_config
 
     def get_project(self) -> GitProject:
```
rpitonak: could we put this method to `AbstractGithubEvent` as it is the same for all of the Github Events?

lachmanfrantisek:
> PS: I still see one item unchecked in PR's TODO list.

That's correct if we do not want to change the structure of the production config.

CONTRIBUTING.md
```diff
@@ -152,3 +152,14 @@ Once we have jobs defined in `packit` config we are ready to move on to next ste
...
 Thank you for your interest!
 packit team.
+
+### Service configuration
+
+The service configuration is an extension of the user configuration from packit. (`Config` class in [packit/config.py](https://github.com/packit-service/packit/blob/master/packit/config.py).)
+
+To add a new service-related property you need to:
+
+1. Add a property to `ServiceConfig.__init__` in [packit/config.py](https://github.com/packit-service/packit/blob/master/packit/config.py).
```
jpopelka: Sorry for being late for the party, but #154

TomasTomecek:
` +210 −327` very nice!

PR #155, opened 2019-10-03
catch exceptions using sentry

TomasTomecek:


TomasTomecek:
> Looks differently then what we use in frambo (`sentry-sdk` vs. `raven`), but I believe you this is the latest-greatest.

> Raven is the legacy Python client for Sentry (getsentry.com) — replaced by sentry-python https://sentry.io

jpopelka:
Should we change also our [frambo code](https://github.com/user-cont/frambo/blob/master/frambo/celery_app.py#L20) then?

TomasTomecek:
> Should we change also our [frambo code](https://github.com/user-cont/frambo/blob/master/frambo/celery_app.py#L20) then?

up to you, ain't touching that

jpopelka:
I was just asking for your opinion, whether you think it's worth changing or we can leave it as it is.

TomasTomecek:
> I was just asking for your opinion, whether you think it's worth changing or we can leave it as it is.

I'd leave it if it works. Especially since we're using the internal instance.

PR #157, opened 2019-10-08
[files/deployment] don't force ansible version to 2.7

jpopelka:


jpopelka:
heh?
```
\| ERROR! no action detected in task. This often indicates a misspelled module name, or incorrect module path.
--
\|
\| The error appears to have been in '/var/opt/rh/rh-python35/lib/zuul/builds/e9c386b7e41f4a098264191644c60d10/untrusted/project_0/github.com/packit-service/packit-service/files/deployment.yaml': line 31, column 5, but may
\| be elsewhere in the file depending on the exact syntax problem.
\|
\| The offending line appears to be:
\|
\|     - fullchain.pem
\|   - name: Generate an OpenSSH keypair with the default values (4096 bits, rsa)
\|     ^ here
```

@rpitonak can you see what's wrong?
`ansible-2.8.5` is installed.
And it works OK when I try the `openssh_keypair` module locally.

TomasTomecek:
works for me locally as well

jpopelka:
No idea why it doesn't work in Zuul so I left the usage of `openssh_keypair` module commented out. PTAL

PR #163, opened 2019-10-09
Add functionality to avoid adding two successful PRcomments in a row

lbarcziova:
#4

### DIFF
--- a/packit_service/worker/fedmsg_handlers.py
+++ b/packit_service/worker/fedmsg_handlers.py
@@ -141,6 +141,20 @@ class CoprBuildEndHandler(FedmsgHandler):
         self.project = self.event.get_project()
         self.package_config = self.event.get_package_config()
 
+    def was_last_build_successful(self):
+        """
+        Check if the last copr build of the PR was succesful
+        :return: bool
+        """
+        comments = self.project.get_pr_comments(pr_id=self.event.pr_id, reverse=True)
+        for comment in comments:
+            if comment.author.startswith("packit-as-a-service"):
+                if "Congratulations!" in comment.comment:
+                    return True
+                return False
+        # if there is no comment from p-s
+        return False
+
     def run(self):
         # get copr build from db
         db = CoprBuildDB()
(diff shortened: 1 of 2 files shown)


packit_service/worker/fedmsg_handlers.py
```diff
@@ -141,6 +141,20 @@ def __init__(self, config: ServiceConfig, job: JobConfig, event: CoprBuildEvent)
         self.project = self.event.get_project()
         self.package_config = self.event.get_package_config()
 
+    def was_last_build_successful(self):
+        """
+        Check if the last copr build of the PR was succesful
```
jpopelka: I think the method actually checks whether ANY (not last) build was successful.

packit_service/worker/fedmsg_handlers.py
```diff
@@ -141,6 +141,20 @@ def __init__(self, config: ServiceConfig, job: JobConfig, event: CoprBuildEvent)
...
+        :return: bool
+        """
+        comments = self.project.get_pr_comments(pr_id=self.event.pr_id, reverse=True)
+        for comment in comments:
+            if comment.author.startswith("packit-as-a-service"):
```
lachmanfrantisek: Nice proposal for extending ogr API: https://github.com/packit-service/ogr/issues/240

packit_service/worker/fedmsg_handlers.py
lachmanfrantisek: It goes through the comments in reverse order and picks the first with `packit-as-a-service` as an author. If it describes success, it returns `True` and `False` otherwise and also `False` when there is no such comment.

I don't see the problem.

jpopelka: you're right, sorry for the fuss

packit_service/worker/fedmsg_handlers.py
```diff
@@ -141,6 +141,20 @@ def __init__(self, config: ServiceConfig, job: JobConfig, event: CoprBuildEvent)
...
+                if "Congratulations!" in comment.comment:
```
TomasTomecek: Honestly, I'd prefer to do this by checking our redis database instead of querying github but the solution is good enough short-term.

PR #164, opened 2019-10-10
Try to add rebase check to pre-commit

lbarcziova:


files/scripts/check-rebase.py
```diff
@@ -0,0 +1,36 @@
...
+    "SHA of the last commit of upstream/master in your branch: {second}. \n"
+    "Please, rebase!"
+)
+
+
+def main():
+    path = os.path.dirname(os.path.abspath(__file__))
+    local_upstream_hash = (
+        subprocess.run(
+            ["git", "log", "upstream/master", "-1", "--format=%H"],
```
TomasTomecek: not everyone may have remote upstream set up

rpitonak:
And the most important thing - this check needs to be in [zuul gating](https://github.com/packit-service/packit-service/blob/fc21624469a18191046631ea5ed5574e0a0a6b20/zuul.d/zuul.yaml#L8)

lachmanfrantisek:
> And the most important thing - this check needs to be in [zuul gating](https://github.com/packit-service/packit-service/blob/fc21624469a18191046631ea5ed5574e0a0a6b20/zuul.d/zuul.yaml#L8)

It should be transitively there (via pre-commit job). (And having that also for each push can be useful as well so one can realize that sooner.)

lachmanfrantisek:
> Are you sure about this? I thought we have to explicitly add `pre-commit-job` to the gating section of `.zuul.yaml`.

Sorry, we are in the wrong repo -- [packit](https://github.com/packit-service/packit/blob/master/.zuul.yaml#L15), [ogr](https://github.com/packit-service/ogr/blob/master/.zuul.yaml#L12) and [requre](https://github.com/packit-service/requre/blob/master/.zuul.yaml#L12) have it already there...

lbarcziova:
I think that fetch is not needed, as git ls-remote should 

> list references in a remote repository

.pre-commit-config.yaml
```diff
@@ -29,3 +29,11 @@ repos:
     hooks:
     -   id: mypy
         args: [--no-strict-optional, --ignore-missing-imports]
+-   repo: local
```
lachmanfrantisek: It would be nice to have a dedicated repo for these scripts so we can use it from multiple repositories => we need to make the project URL parametrizable.

.pre-commit-config.yaml
```diff
@@ -29,3 +29,11 @@ repos:
...
+    hooks:
+    -   id: check-rebase
+        name: Check if rebase is needed
+        entry: ./files/scripts/check-rebase.py
+        language: script
+        types: [python]
```
lachmanfrantisek: I think we can remove this since we have `always_run: true`. (And this may be a reason of the duplication of the output.)

files/scripts/check-rebase.py
```diff
@@ -0,0 +1,58 @@
...
+def main():
+    path = os.path.dirname(os.path.abspath(__file__))
+
+    last_commit_subject = subprocess.run(
+        ["git", "log", "--format=%s", "-1"], capture_output=True, cwd=path
+    ).stdout.decode()
+    print(f"Last commit subject: {last_commit_subject}")
+
+    if "Merge commit" in last_commit_subject:
+        print("Zuul merged the master -- rebase is needed.")
```
lachmanfrantisek: You can use some more verbose message as you have in `WARNING_MSG`. (I don't mean it to be the final version..;-)

lachmanfrantisek:
If you want to move the script to other repo, you need to have a config in that repo:

https://pre-commit.com/#new-hooks

(It will be similar to your current config.)

We only need to improve the script to accept the upstream_url as an argument.

lbarcziova:
@lachmanfrantisek I will move the hook into a new repo https://github.com/packit-service/pre-commit-hooks

Issue #159, opened 2019-10-09
`if service in url`: `argument of type 'NoneType' is not iterable`


TomasTomecek:
Sentry doing hard work:

```
{
    "event": {
        "build": 1051404,
        "chroot": "fedora-31-x86_64",
        "copr": "packit-service-packit-555",
        "ip": "172.25.80.145",
        "owner": "packit",
        "pid": 15609,
        "pkg": "packit",
        "status": 3,
        "topic": "org.fedoraproject.prod.copr.build.start",
        "user": "packit",
        "version": "0.7.1.dev19+g7b8bf00-1.fc30",
        "what": "build start: user:packit copr:packit-service-packit-555 pkg:packit build:1051404 ip:172.25.80.145 pid:15609",
        "who": "backend.worker-63-PC"
    }
}
```

```
TypeError: argument of type 'NoneType' is not iterable
  File "celery/app/trace.py", line 382, in trace_task
    R = retval = fun(*args, **kwargs)
  File "celery/app/trace.py", line 641, in __protected_call__
    return self.run(*args, **kwargs)
  File "packit_service/worker/tasks.py", line 47, in process_message
    return SteveJobs().process_message(event=event, topic=topic)
  File "packit_service/worker/jobs.py", line 218, in process_message
    project
  File "packit/config.py", line 212, in _get_project
    project = get_project(url=url, custom_instances=self.services)
  File "ogr/factory.py", line 83, in get_project
    kls = get_service_class(url=url, service_mapping_update=service_mapping_update)
  File "ogr/factory.py", line 135, in get_service_class
    url=url, service_mapping_update=service_mapping_update
  File "ogr/factory.py", line 118, in get_service_class_or_none
    if service in url:
```

```
ogr/factory.py in get_service_class_or_none at line 118

        mapping.update(_SERVICE_MAPPING)
        if service_mapping_update:
            mapping.update(service_mapping_update)
        for service, service_kls in mapping.items():
-->         if service in url:
                return service_kls
        return None
```

```
Task task.steve_jobs.process_message[4560f82d-b807-4ca2-99c6-1653635a37b3] raised unexpected: TypeError("argument of type 'NoneType' is not iterable")
```

lachmanfrantisek:
It looks like the `url` is None:

```python
  File "packit/config.py", line 212, in _get_project
    project = get_project(url=url, custom_instances=self.services)
```

TomasTomecek:
> It looks like the `url` is None:

[...]

PR #166, opened 2019-10-11
Filter builds not handled by packit as soon as possible

rpitonak:
#159 was caused by fedora messages which come to packit, and they are not saved in Redis. Let's filter them as soon as possible. 

PS: Already deployed on stg.


rpitonak:
I like the idea. PTAL

PR #176, opened 2019-10-17
Add unit tests for CoprBuildEvent parsing

### LINKED ISSUE #175: Unit tests for CoprBuildEvent
Add unit test for `CoprBuildEvent` [here](https://github.com/packit-service/packit-service/blob/master/tests/unit/test_events.py) so we are sure that we parsing it correctly.

lbarcziova:
Fixes #175

### DIFF
--- a/tests/unit/test_events.py
+++ b/tests/unit/test_events.py
@@ -44,6 +44,9 @@ from packit_service.service.events import (
     PullRequestCommentAction,
     IssueCommentEvent,
     IssueCommentAction,
+    CoprBuildEvent,
+    CoprBuildDB,
+    FedmsgTopic,
 )
 from packit_service.worker.parser import Parser
 from tests.spellbook import DATA_DIR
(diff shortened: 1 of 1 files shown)


tests/unit/test_events.py
```diff
@@ -224,6 +237,32 @@ def test_parse_testing_farm_results(self, testing_farm_results):
         assert event_object.copr_repo_name == "packit/packit-service-hello-world-10-stg"
         assert event_object.copr_chroot == "fedora-29-x86_64"
 
+    def test_parse_copr_build_event_start(self, copr_build_results_start):
+        flexmock(CoprBuildDB).should_receive("get_build").and_return()
```
rpitonak: you do not need `and_return()` in this case.

tests/unit/test_events.py
```diff
@@ -91,6 +94,16 @@ def pr_comment_empty_request(self):
         ) as outfile:
             return json.load(outfile)
 
+    @pytest.fixture()
+    def copr_build_results_start(self):
+        with open(DATA_DIR / "fedmsg" / "copr_build_start.json", "r") as outfile:
```
jpopelka: "r" is not needed as it's [default mode](https://docs.python.org/3/library/functions.html#open)

PR #177, opened 2019-10-17
it's not an error that people don't have packit.yaml in their repo

TomasTomecek:


TomasTomecek:
for everyone: our sentry queue is so busy because we set success=False to a lot of situations - if success==False, then we do logger.error(...) later in the code and sentry picks this up and creates an issue for such requests

TomasTomecek:
recheck

PR #178, opened 2019-10-17
don't treat unrelated PR comments as errors

TomasTomecek:


TomasTomecek:
hm, tests are failing; so maybe not mergeit

TomasTomecek:
@rpitonak are our gating tests borked? :( or do we only have 10 minutes to execute them?

rpitonak:
@TomasTomecek they are broken, please merge without Zuul for now.

PR #183, opened 2019-10-23
Packit maintainers are always able to trigger packit as a service

rpitonak:
[Related](https://gitlab.cee.redhat.com/user-cont/packit-service/merge_requests/47)

### DIFF
--- a/packit_service/config.py
+++ b/packit_service/config.py
@@ -22,7 +22,7 @@
 import enum
 import logging
 from pathlib import Path
-from typing import List
+from typing import List, Set
 
 from packit.config import RunCommandType, Config
 from packit.exceptions import PackitException
@@ -63,7 +63,7 @@ class ServiceConfig(Config):
         self.routing_keys: List[str] = []
 
         # List of github users who are allowed to trigger p-s on any repository
-        self.admins: List[str] = []
+        self.admins: Set[str] = set()
 
         # %%% ACTIONS HANDLER CONFIGURATION %%%
         # these values are specific to packit service when we run actions in a sandbox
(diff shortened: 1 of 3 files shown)


packit_service/worker/github_handlers.py
```diff
@@ -241,7 +241,9 @@ def handle_pull_request(self):
 
         collaborators = self.project.who_can_merge_pr()
         r = BuildStatusReporter(self.project, self.event.commit_sha)
-        if self.event.github_login not in collaborators:
+        if (self.event.github_login not in collaborators) and (
+            self.event.github_login not in self.config.admins
```
jpopelka: `collaborators` is already a `set` so you can:
```python
if self.event.github_login not in collaborators | set(self.config.admins):
```
or if you define `config.admins` as `set` then even more simple:
```python
if self.event.github_login not in collaborators | self.config.admins:
```

packit_service/config.py
```diff
@@ -94,6 +97,7 @@ def get_from_dict(cls, raw_dict: dict, validate=True) -> "ServiceConfig":
         config.validate_webhooks = raw_dict.get("validate_webhooks", False)
         config.queue_name = raw_dict.get("queue_name", "")
         config.routing_keys = raw_dict.get("routing_keys", [])
+        config.admins = set(raw_dict.get("admins", []))
```
rpitonak: not sure if this conversion is needed

jpopelka: it is crucial
```python
admins: Set[str] = set()                                                                                                                                                                                                              
admins = ["jirka", "rado"]                                                                                                                                                                                                            
type(admins)                                                                                                                                                                                                                          
list
```

PR #197, opened 2019-10-31
Properly handle failed CoprBuildStartHandler

jpopelka:
Fixes https://sentry.io/organizations/red-hat-0p/issues/1306844592

(or should it actually be a failure?, let me check)

### DIFF
--- a/packit_service/worker/fedmsg_handlers.py
+++ b/packit_service/worker/fedmsg_handlers.py
@@ -145,7 +145,7 @@ class CoprBuildEndHandler(FedmsgHandler):
 
     def was_last_build_successful(self):
         """
-        Check if the last copr build of the PR was succesful
+        Check if the last copr build of the PR was successful
         :return: bool
         """
         comments = self.project.get_pr_comments(pr_id=self.event.pr_id, reverse=True)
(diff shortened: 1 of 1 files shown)


jpopelka:
Ok, let's properly treat this as a failure and investigate the cause in another round.

Ready for Review

packit_service/worker/fedmsg_handlers.py
```diff
@@ -158,15 +158,12 @@ def was_last_build_successful(self):
...
-        db = CoprBuildDB()
-        build = db.get_build(self.event.build_id)
-
+        build = CoprBuildDB().get_build(self.event.build_id)
         if not build:
-            logger.warning(
-                f"Build: {self.event.build_id} is not handled by packit service!"
-            )
-            return
+            # TODO: how could this happen?
```
TomasTomecek: easily: a copr build which is not triggered by packit-service - not an error

jpopelka: > easily

Well the question is what the "not triggered by packit-service" means, because:
- It was a copr project which belongs to packit service because we [process only copr-build messages whose owner is 'packit'](https://github.com/packit-service/packit-service/blob/master/packit_service/fedmsg/consumer.py#L49) (i.e. packit service).
- it probably wasn't triggered by `/packit copr-build` because [the comment handler](https://github.com/packit-service/packit-service/blob/master/packit_service/worker/github_handlers.py#L480) also [stores the build in db](https://github.com/packit-service/packit-service/blob/master/packit_service/worker/copr_build.py#L133) 

How else could it have been triggered? The only other way I see it manually trigger the build in copr?

TomasTomecek: Aha, wasn't aware of that code.

PR #217, opened 2019-11-07
[AbstractGithubEvent] project_url is not optional

jpopelka:
Was trying to investigate (and possibly prevent in future)
https://sentry.io/organizations/red-hat-0p/issues/1318925042

Realized it's probably the other side (testing-farm) problem.
But, at least, fix some inconsistencies.


packit_service/service/events.py
```diff
@@ -397,6 +397,7 @@ def get_package_config(self):
         return package_config
 
 
+# Wait, what? copr build event doesn't sound like github event
```
lachmanfrantisek: Good point!

packit_service/service/events.py
```diff
@@ -133,12 +133,12 @@ def pre_check(self) -> bool:
...
 
 class AbstractGithubEvent(Event):
-    def __init__(self, trigger: JobTriggerType, project_url=None):
+    def __init__(self, trigger: JobTriggerType, project_url: str):
         super().__init__(trigger)
         self.project_url: str = project_url
 
     def get_project(self) -> GitProject:
-        return service_config.get_project(url=self.project_url)
+        return service_config.get_project(self.project_url)
```
lachmanfrantisek: Why this change? Don't you like setting arguments via key?

jpopelka: reverted

PR #227, opened 2019-11-12
[fedmsg] install copr json schemas

jpopelka:
Fixes
```
[WARNING fedora_messaging.message] The schema 'copr.build.end' is not in the schema registry! Either install the package with its schema definition or define a schema. Falling back to the default schema...
```

### DIFF
--- a/files/install-deps.yaml
+++ b/files/install-deps.yaml
@@ -17,4 +17,5 @@
       # web service needs packit just for few bits (see #109), so hopefully we don't need it from master and RPM is enough
       - python3-packit
       - fedora-messaging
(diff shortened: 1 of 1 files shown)


files/install-deps.yaml
```diff
@@ -17,4 +17,5 @@
       # web service needs packit just for few bits (see #109), so hopefully we don't need it from master and RPM is enough
       - python3-packit
       - fedora-messaging
+      - python3-copr-messaging  # json schemas
```
TomasTomecek: -_- don't understand why fedora-messaging rpm is not requiring this

PR #233, opened 2019-11-14
Use FAS to tell if GitHub user is a Fedora packager

### LINKED ISSUE #220: lazy badges.fedoraproject.org/user/{user}/json
We query https://badges.fedoraproject.org/user/{user}/json in [Whitelist._is_packager()](https://github.com/packit-service/packit-service/blob/master/packit_service/worker/whitelist.py#L56) to check whether a user is Fedora packager.

It takes from 5s (for nonexisting user) to over a minute (for users with lots of badges).

```bash
$ time curl https://badges.fedoraproject.org/user/foobarbaz/json
real	0m5.311s
$ time curl https://badges.fedoraproject.org/user/jpopelka/json
real	1m7.064s
```

Wouldn't [Koji API](https://docs.pagure.org/koji/using_the_koji_build_system/#koji-xmlrpc-api) be faster?
```bash
$ koji list-api
getUser(userInfo=None, strict=False)
  description: Return information about a user.

    userInfo may be either a str (Kerberos principal or name) or an int (user id).

    A map will be returned with the following keys:
      id: user id
      name: user name
      status: user status (int), may be null
      usertype: user type (int), 0 person, 1 for host, may be null
      krb_principal: the user's Kerberos principal
```

EDIT: Hint: use [xmlrpc.client](https://docs.python.org/3/library/xmlrpc.client.html#module-xmlrpc.client)

csomh:
Querying badges.fedoraproject.org to tell if a user is a packager in
Fedora is slow: takes seconds for non existing usernames and can take
over a minute for an existing one.

Query FAS instead, and check if the users GitHub username is in the
'packager' group. This is faster and probably also more correct, as FAS
is the central place to manage users in Fedora space.

Add a new property (`fas_password`) to the service configuration to be
used to authenticate with FAS, as `fedora.client.fas2` doesn't support
kerberos.

Resolves #220.


csomh:
Hm, this failure seems to be happening on other PRs too :thinking:

packit_service/config.py
```diff
@@ -58,6 +58,9 @@ def __init__(self):
         self.testing_farm_secret: str = ""
         self.validate_webhooks: bool = True
 
+        # fas.fedoraproject.org needs password to authenticate
+        self.fas_password: Optional[str] = None
+
```
jpopelka: Do we need to add also `fas_user` here and into the schema?

packit_service/worker/whitelist.py
```diff
@@ -39,33 +40,41 @@
...
-        for item in response.json().get("assertions", []):
-            if "Successfully completed a koji build." in item.get("description"):
-                logger.info(f"User {account_login!r} is a packager in Fedora!")
+        for membership in person.get("memberships", []):
+            if membership.get("name") == "packager":
+                logger.info("User %r is a packager in Fedora!", account_login)
                 return True
-        logger.info(f"Cannot verify whether {account_login!r} is a packager in Fedora.")
+
+        logger.info("Cannot verify whether %r is a packager in Fedora.", account_login)
```
jpopelka: There's nothing wrong with you string formatting, it's just that we use the f-strings everywhere (even in logging), so I'd rather see them used also in your PR.

jpopelka:
Yes, sorry for the CI

packit_service/config.py
csomh: `fas_user` is already in [`packit.config.Config`](https://github.com/packit-service/packit/blob/master/packit/config/config.py#L52) and [`packit.schema.USER_CONFIG_SCHEMA`](https://github.com/packit-service/packit/blob/master/packit/schema.py#L73). But I do recognise, that seeing this on it's own can be confusing. I'll update the comment, to make this more obvious.

packit_service/worker/whitelist.py
csomh: Yes, I'm aware that this goes against the conventions of the code-base. Still I think this is better, b/c then string interpolation is left up to `logging`, and it's done only when the log-level requires it (less work). Plus aggregation becomes easier. 

I fail to find the relevant docs to point at right now, but [this post and comment](https://reinout.vanrees.org/weblog/2015/06/05/logging-formatting.html#comment-2063573645) maybe explains it better.

Anyways, if you prefer consistency, just let me know and I'll change them to f-strings.

jpopelka: Yes, we know about that (that was my `(even in logging)` comment supposed to tell you) and we still prefer consistency. I'm just not sure what you mean with 'aggregation'.

csomh: Updated to f-strings.

By aggregation I meant, that for example Sentry would create a single event for `Cannot verify whether %r is a packager in Fedora.` repeating n times. But this might be achievable in other ways, too (~[Server-side fingerprinting](https://docs.sentry.io/data-management/event-grouping/server-side-fingerprinting/)).

PR #251, opened 2019-11-21
document Zuul & playbooks

TomasTomecek:


files/deployment.yaml
```diff
@@ -1,4 +1,5 @@
 ---
+# This file is the entrypoint for Zuul which deploys openshift cluster in Zuul.
```
rpitonak: Small correction, it is actually entry point for deploying packit to openshift cluster. Openshift cluster setup is in https://github.com/packit-service/packit-service/blob/master/files/install-openshift.yaml. Sorry if I said it wrong.

TomasTomecek: oki, will update, thanks!

TomasTomecek:
PTAL (less "Rado" in this PR now)

PR #253, opened 2019-11-22
Remove incorrect/unused/redundant event-related tests parts

jpopelka:


TomasTomecek:
hm, the CI job is timing out

jpopelka:
> hm, the CI job is timing out

We should call you "Lightning" :-) (That's what Rado has been trying to solve since Monday)

TomasTomecek:
but the failure in this job looked different than the one I was seeing before

rpitonak:
> but the failure in this job looked different than the one I was seeing before

Yes because logs are not displayed when the test job fails. It is fixed in #254

PR #258, opened 2019-11-26
Add script to fix enum values

### LINKED ISSUE #212: Fix some enum values in Redis
Since 608545f we store in DB only enum values, not their string representation, for example:
'trigger': 'pull_request' instead of 'trigger': 'JobTriggerType.pull_request'

It'd be nice to change also the older entries, namely the `'trigger'`s in `whitelist` and `github_installation` hash/table.

See https://github.com/packit-service/packit-service/blob/master/packit_service/worker/whitelist.py#L98

lbarcziova:
Script to fix #212

### DIFF
--- a/files/scripts/clean_up_empty_tasks.py
+++ b/files/scripts/clean_up_empty_tasks.py
@@ -3,6 +3,7 @@ from redis import Redis
 import json
 
 
+# https://github.com/packit-service/packit-service/issues/196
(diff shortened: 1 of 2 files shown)


TomasTomecek:
rebase please, all should be green now

TomasTomecek:
and now pre-commit complains about the script you are adding here :)

PR #270, opened 2019-12-02
whitelist: treat admins as whitelisted

TomasTomecek:
\+ random fixes:
* rec-worker: create sandcastle as 0777 (for sake of docker-compose)
* d-c: we no longer have packit.yaml, and docs
* whitelist: polish docstrings
* copr_url_from_event: catch only exceptions from requests

### DIFF
--- a/packit_service/worker/jobs.py
+++ b/packit_service/worker/jobs.py
@@ -113,7 +113,10 @@ class SteveJobs:
                 # check whitelist approval for every job to be able to track down which jobs
                 # failed because of missing whitelist approval
                 whitelist = Whitelist()
-                if not whitelist.check_and_report(event, event.get_project()):
+                github_login = getattr(event, "github_login", None)
+                if github_login and github_login in self.config.admins:
+                    logger.info(f"{github_login} is admin, you shall pass")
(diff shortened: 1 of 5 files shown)


packit_service/worker/jobs.py
```diff
@@ -113,7 +113,10 @@ def process_jobs(self, event: Event) -> Dict[str, HandlerResults]:
                 # check whitelist approval for every job to be able to track down which jobs
                 # failed because of missing whitelist approval
                 whitelist = Whitelist()
-                if not whitelist.check_and_report(event, event.get_project()):
+                project_name = getattr(event, "github_login", None)
+                if project_name and project_name in self.config.admins:
+                    logger.info(f"{project_name} is admin, you shall pass")
+                elif not whitelist.check_and_report(event, event.get_project()):
```
jpopelka: Why we don't do the same also in `process_comment_jobs()`?

TomasTomecek: oh right, good point

TomasTomecek: fix't & rebase't

PR #290, opened 2019-12-06
/packit build alias

lachmanfrantisek:
- Allow '/packit build' to trigger the copr-build as well.

### DIFF
--- a/packit_service/worker/comment_action_handler.py
+++ b/packit_service/worker/comment_action_handler.py
@@ -39,6 +39,7 @@ class CommentAction(enum.Enum):
     copr_build = "copr-build"
     propose_update = "propose-update"
     test = "test"
+    build = "build"
 
 
 COMMENT_ACTION_HANDLER_MAPPING: Dict[CommentAction, Type["CommentActionHandler"]] = {}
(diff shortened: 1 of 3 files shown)


TomasTomecek:
it's `/` right? not `\`

lachmanfrantisek:
> * this needs docs, obviously

:+1: 

> * should we advertise in the build-failed-you-need-retrigger comment this over `copr-build`?

Good point.

- `build`: When there is no `copr-build` defined and if it is built because of the tests.
  - (="I don't want to know anything about copr." workflow)
- `copr-build`: When there is a `copr-build` defined in the jobs.
  - (="I am interested in copr." workflow)
> it's `/` right? not `\`

Thanks, the wrong slash is only in the git message and PR title.

lachmanfrantisek:
- About the docs: @TomasTomecek, you mean `packit.dev` documentation, right?
- About the re-trigger message: Can we change that when there will be support for defining tests without `copr-build` jobs?

TomasTomecek:
>     * `build`: When there is no `copr-build` defined and if it is built because of the tests.
>       
>       * (="I don't want to know anything about copr." workflow)
> 
>     * `copr-build`: When there is a `copr-build` defined in the jobs.
>       
>       * (="I am interested in copr." workflow)

sounds good!

>     * About the docs: @TomasTomecek, you mean `packit.dev` documentation, right?

yep
 
>     * About the re-trigger message: Can we change that when there will be support for defining tests without `copr-build` jobs?

sounds good! would be nice to track that request somewhere

lachmanfrantisek:
> sounds good! would be nice to track that request somewhere

I'll create a new issue.

PR #292, opened 2019-12-06
Data regeneration

jscotka:


jscotka:
> so, simplify doesn't work in the end?

@TomasTomecek simplify itself works as expected, Trouble is that there is no control over reducing, causing, that it lead to case that missig item in tree, caused unable to find some item. I have to think about this mode more, to do it as reliable as possible.

jscotka:
created issue for this issue

PR #297, opened 2019-12-09
WIP: LOGS

### LINKED ISSUE #1: Store and provide logs in packit service
get logs from copr realtime and provide them realtime in a sensible way (show what matters to users, not internal infra stuff people don't understand)

TomasTomecek:
replaces https://github.com/packit-service/packit-service/pull/257

Fixes #1

This PR was split into smaller PRs and needs a rebase now.

### DIFF
--- a/files/install-deps-worker.yaml
+++ b/files/install-deps-worker.yaml
@@ -12,6 +12,8 @@
           - python3-kubernetes # for sandcastle
           - python3-fedora # to access FAS
           - python3-requests
+          - python3-flask # sorry jirko
+          - python3-flask-restplus
           - python3-alembic
           - python3-sqlalchemy
           - python3-psycopg2
--- a/packit_service/config.py
+++ b/packit_service/config.py
@@ -53,6 +53,7 @@ class Deployment(enum.Enum):
 
 class ServiceConfig(Config):
     service_config = None
+    SCHEMA = SERVICE_CONFIG_SCHEMA
 
     def __init__(
         self,
@@ -60,10 +61,9 @@ class ServiceConfig(Config):
         webhook_secret: str = "",
         testing_farm_secret: str = "",
         validate_webhooks: bool = True,
-        admins: list = [],
+        admins: list = None,
         **kwargs,
     ):
-
         super().__init__(**kwargs)
 
         self.deployment = deployment
@@ -78,7 +78,10 @@ class ServiceConfig(Config):
         self.fas_password: Optional[str] = ""
 
         # List of github users who are allowed to trigger p-s on any repository
-        self.admins: Set[str] = set(admins)
+        self.admins: Set[str] = set(admins or [])
+
+        # for flask SERVER_NAME so we can create links to logs
+        self.server_name: str = ""
 
         # for flask SERVER_NAME so we can create links to logs
         self.server_name: str = ""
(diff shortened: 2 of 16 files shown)


TomasTomecek:
recheck

packit_service/service/views.py
```diff
@@ -0,0 +1,56 @@
...
+from flask import Blueprint
+
+from packit_service.worker.copr_db import CoprBuildDB
+
+builds_blueprint = Blueprint("builds", __name__)
+
+
+@builds_blueprint.route(
+    "/build/<string:build_id>/<string:target>/logs", methods=("GET",)
+)
```
jpopelka: What about renaming [copr-builds](https://github.com/packit-service/packit-service/blob/master/packit_service/service/api/copr_builds.py#L36) endpoint to `builds` and adding it there?

packit_service/service/app.py
```diff
@@ -23,11 +23,21 @@
...
 application = Flask(__name__)
 application.register_blueprint(blueprint)
+application.register_blueprint(builds_blueprint)
+
+application.config["SERVER_NAME"] = ServiceConfig.get_service_config().server_name
+application.config["PREFERRED_URL_SCHEME"] = "https"
+
+logger = logging.getLogger("packit_service")
+# no need to thank me, just buy me a beer
+logger.debug(f"URL map = {application.url_map}")
```
jpopelka: How come I see the output even in worker output?

jpopelka: Found it.

files/install-deps-worker.yaml
```diff
@@ -14,6 +14,9 @@
       - python3-kubernetes # for sandcastle
       - python3-fedora # to access FAS
       - python3-requests
+
+      - python3-flask  # sorry jirko
+      - python3-flask-restplus
```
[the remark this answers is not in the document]
jpopelka: You can add it to `setup.cfg` as well since it's a common (to both, server & worker) requirement now.

packit_service/config.py
```diff
@@ -82,6 +83,9 @@ def __init__(self):
         # path to a file where OGR should store HTTP requests
         self.github_requests_log_path: str = ""
 
+        # for flask SERVER_NAME so we can create links to logs
+        self.server_name: str = ""
```
jpopelka: Will you update [deployment/secrets/template/packit-service.yaml](https://github.com/packit-service/deployment/blob/master/secrets/template/packit-service.yaml)?

TomasTomecek: +1 good point

packit_service/service/views.py
TomasTomecek: as we discussed on a meeting: this endpoint is not an API, it's meant for users to look at

files/install-deps-worker.yaml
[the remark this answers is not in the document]
TomasTomecek: > Are you f.... kidding me?
> 
> (you don't need to add `python3-flask` since it's pulled in with `python3-flask-restplus`)

but that's installed only in the service image, not worker, so yeah, I funkin' had to put it here in order to make this PR work

I agree, setup.cfg sounds like a better solution

packit_service/service/views.py
jpopelka: Not sure I can agree as I don't see a difference.
Isn't API meant for users as well to look at?
Can't API endpoint return plain text / HTML or whatever your endpoint returns?

Anyway, your endpoint should at least be re-named `/build/...` -> `/builds/...`

files/install-deps-worker.yaml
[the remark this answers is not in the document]
jpopelka: Ok, I think we are on a different track ...

> but that's installed only in the service image, not worker, so yeah, I funkin' had to put it here in order to make this PR work

I was suggesting to remove only `python3-flask` not `python3-flask-restplus`.

> I agree, setup.cfg sounds like a better solution

Important in my comment was "as well". Please leave `python3-flask-restplus` in `files/install-deps-worker.yaml` so we install RPM, not python package (even there should be no difference)

TomasTomecek:
closing, will reopen

PR #302, opened 2019-12-12
WIP: Add initial version of Marshmallow implementation

lbarcziova:
packit-service/packit#635

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,6 +38,7 @@ setup_requires =
 # The others are in files/install-deps*.yaml
 install_requires =
     celery[redis]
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -38,6 +38,7 @@ setup_requires =
 # The others are in files/install-deps*.yaml
 install_requires =
     celery[redis]
+    marshmallow~=2.0
```
jpopelka: Actually we are not forced to use v2 in `packit-service` because we don't package it into Fedora.
You are free to use v3  :-)

sakalosj:
handled in #320

PR #318, opened 2020-01-09
load service config only once

### LINKED ISSUE #153: Loading config in the config.py
Is there some strong reason why we are running code in the `packit_service.config`?

https://github.com/packit-service/packit-service/blob/5ef755e61c7a3274041a05b073e85b8fed010730/packit_service/config.py#L137

What about using some function for that instead?

In this case, each time we need to import something from the file, it tries to load my user config and if it fails to validate, the code will raise an Exception. (Especially, I cannot run unit tests without having valid config.)

TomasTomecek:
Fixes #153

### DIFF
--- a/packit_service/service/app.py
+++ b/packit_service/service/app.py
@@ -23,11 +23,19 @@
 import logging
 
 from flask import Flask
-
 from packit.utils import set_logging
+
+from packit_service.config import ServiceConfig
 from packit_service.service.api import blueprint
 
 set_logging(logger_name="packit_service", level=logging.DEBUG)
 
 application = Flask(__name__)
 application.register_blueprint(blueprint)
+
+application.config["SERVER_NAME"] = ServiceConfig.get_service_config().server_name
+application.config["PREFERRED_URL_SCHEME"] = "https"
+
+logger = logging.getLogger("packit_service")
+# no need to thank me, just buy me a beer
+logger.debug(f"URL map = {application.url_map}")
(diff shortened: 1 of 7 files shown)


TomasTomecek:
locally service tests now `=== 62 passed, 3 warnings in 3.42 seconds ===`

TomasTomecek:
I don't have the patience to deal with requre tests right now.

packit_service/service/app.py
```diff
@@ -23,11 +23,19 @@
...
 
 application = Flask(__name__)
 application.register_blueprint(blueprint)
+
+application.config["SERVER_NAME"] = ServiceConfig.get_service_config().server_name
+application.config["PREFERRED_URL_SCHEME"] = "https"
+
+logger = logging.getLogger("packit_service")
+# no need to thank me, just buy me a beer
+logger.debug(f"URL map = {application.url_map}")
```
jpopelka: For some strange reason there's `[wsgi:error]` in the output:

```
[Wed Jan 15 08:09:54.088320 2020] [wsgi:error] [pid 19:tid 140426144958208] 08:09:54.088 app.py            DEBUG  URL map = Map([<Rule '/api/testing-farm/results' (OPTIONS, POST) -> api.testing-farm_testing_farm_results>,
[Wed Jan 15 08:09:54.088354 2020] [wsgi:error] [pid 19:tid 140426144958208]  <Rule '/api/webhooks/github' (OPTIONS, POST) -> api.webhooks_github_webhook>,
```

TomasTomecek: I'd say that everything printed by the wsgi app is printed as errors, which is really weird; maybe it can be confiigured in the httpd config

PR #319, opened 2020-01-09
initiate usage of pg

TomasTomecek:
This PR doesn't add any functionality. I'm adding "dead-code" pretty much. On the other hand:
* it should make reviewing other pg-related PRs trivial
* it will be easy for everyone to start moving data from redis to pg

Read more about alembic here: https://alembic.sqlalchemy.org/en/latest/cookbook.html#building-uptodate

### DIFF
--- a/packit_service/service/models.py
+++ b/packit_service/service/models.py
@@ -2,15 +2,22 @@
 Data models for jobs, tasks, builds, etc. The models are mapped to redis.
 """
 import copy
-from typing import List, Dict, Union
+import os
+from typing import List, Dict, Union, TYPE_CHECKING
 
 from persistentdict.dict_in_redis import PersistentDict
+from sqlalchemy import Column, Integer, String, ForeignKey, Text, JSON, create_engine
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import relationship, sessionmaker, Session
 
 from packit_service.service.events import InstallationEvent, WhitelistStatus
 
 # if identifier is not set, we need to create primary key ourselves
 # with this variable, we keep track on what's the last PK
 LAST_PK = "last-pk"
+
+# SQLAlchemy session, get it with `get_sa_session`
+session_instance = None
 
 
 class Model:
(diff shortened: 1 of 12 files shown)


TomasTomecek:
I just tested this locally in a cluster and it works fine.

packit_service/service/models.py
```diff
@@ -145,3 +152,61 @@ def create(
...
+    namespace = Column(String)
+    repo_name = Column(String)
+
+
+class PullRequest(Base):
+    __tablename__ = "pull_requests"
+    id = Column(Integer, primary_key=True)  # our database PK
+    pr_id = Column(
+        String, index=True
+    )  # GitHub PR ID - let's not make this PK since we can't control it
```
lachmanfrantisek: :+1:

packit_service/service/models.py
```diff
@@ -145,3 +152,61 @@ def create(
...
+    project_id = Column(Integer, ForeignKey("github_projects.id"))
+    project = relationship("GitHubProject")
+
+
+JOB_TYPE_SRPM = "SRPM"
+JOB_TYPE_COPR_RPM = "COPR-RPM"
+JOB_TYPE_TFT = "TFT"
```
lachmanfrantisek: What about some enum -- will be probably useful elsewhere.

packit_service/service/models.py
TomasTomecek: bear in mind that the models are really just a first pass, they definitely need a ton of improvements, that's why I didn't want to commit to create and apply real database schema

packit_service/service/models.py
TomasTomecek: as for enums:
* didn't dig into how database engines or SA deal with enums
* I hate when enums are represented as numbers in DB - what the F should know what 2 or 3 does stand for?
* I hope it wouldn't be a big problem to alter the enum in future
* actually I think I hate enums, they bring more problems than benefits

how about I just add your suggestion as a comment to the code?

TomasTomecek:
Thanks for reviews, guys.

packit_service/service/models.py
lachmanfrantisek: >     * I hate when enums are represented as numbers in DB - what the F should know what 2 or 3 does stand for?

You can use a string enum on the top of this. (So the values are strings, not ints.)

> how about I just add your suggestion as a comment to the code?

:+1:

jpopelka:
So what's the state? (other than needs-rebase)

TomasTomecek:
needs-free-time

TomasTomecek:
rebased, let's see

TomasTomecek:
this was the last rebase

PR #324, opened 2020-01-13
No need to configure build for tests

lachmanfrantisek:
- Run the build for the test job as well.
- Restructure the handler/helper classes code for build/test
    - Rename `CoprBuildHandler` to `CoprBuildJobHelper` because it's not related to the `Handler` classes in the code.
    - Create `TestingFarmJobHelper` and share the code with the common superclass of the `CoprBuildJobHelper`.
    - Improve `CoprBuildJobHelper` to not need the copr_build job and calculate the right chroots.
        - Fix the aliases for the testing farm.

### DIFF
--- a/packit_service/worker/github_handlers.py
+++ b/packit_service/worker/github_handlers.py
@@ -23,14 +23,10 @@
 """
 This file defines classes for job handlers specific for Github hooks
 """
-import json
 import logging
-import uuid
 from typing import Union, Any, Optional, List, Callable
 
-import requests
 from ogr.abstract import GitProject
-from ogr.utils import RequestResponse
 from packit.api import PackitAPI
 from packit.config import (
     JobConfig,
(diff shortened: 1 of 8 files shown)


tests/unit/test_copr_build_handler.py
```diff
@@ -0,0 +1,115 @@
...
+                    metadata={"targets": ["fedora-29", "fedora-31"]},
+                ),
+            ],
+            {"fedora-29-x86_64", "fedora-31-x86_64"},
+            {"fedora-29-x86_64", "fedora-31-x86_64"},
+            id="build_without_target&test_with_targets",
+        ),
+    ],
+)
+def test_targets(jobs, build_targets, test_targets):
```
TomasTomecek: these test cases <3

packit_service/worker/github_handlers.py
```diff
@@ -294,6 +296,19 @@ def handle_pull_request(self):
         return handler_results
 
     def run(self) -> HandlerResults:
+        is_copr_build: Callable[
+            [JobConfig], bool
+        ] = lambda job: job.type == JobType.copr_build
+
+        if self.job.job == JobType.tests and any(
+            filter(is_copr_build, self.package_config.jobs)
+        ):
```
TomasTomecek: I'd write this with a for loop, or at least pass the lambda directly to the filter call

but if you and Jirka think this is fine, let's roll

lachmanfrantisek: It's (currently) impossible to type lambda directly... I would be glad to see a better way to satisfy `mypy`...

PR #325, opened 2020-01-13
Skip setting test flags on failed srpm on empty tests

dhodovsk:


packit_service/worker/handler.py
```diff
@@ -116,9 +111,9 @@ def report(
         self,
         state: str,
         description: str,
+        check_names: Union[str, list, None],
```
lachmanfrantisek: Maybe more descriptive:

```suggestion
        check_names: Opotional[Union[str, list]],
```

packit_service/worker/github_handlers.py
```diff
@@ -455,18 +455,18 @@ def run(self) -> HandlerResults:
...
 
                 r.report(
-                    "pending",
-                    "Tests are running ...",
-                    None,
-                    req.json()["url"],
+                    state="pending",
+                    description="Tests are running ...",
+                    build_id=None,
+                    url=req.json()["url"],
```
lachmanfrantisek: Thanks!

packit_service/worker/handler.py
```diff
@@ -190,6 +183,8 @@ def report_rpm_build_failed_because_of_the_srpm_fail(self, build_check_names):
         )
 
     def report_tests_failed_because_of_the_build(self, test_check_names):
+        if not test_check_names:
+            return
```
lachmanfrantisek: What about having this directly in the `report` method so every `report_*` method can use that?

lachmanfrantisek:
> ACK if the failures are unrelated

Mypy has some problem with the code...

TomasTomecek:
and needs rebase :mountain_cableway:

PR #336, opened 2020-01-16
Make error log less variable so sentry can group

dhodovsk:
So we don't have: `{'jobs': {'copr_build_started': None}, 'event': {'trigger': 'commit', 'created_at': 157918113` and similar in event titles.

More info: https://docs.sentry.io/platforms/python/logging/


dhodovsk:
Tests should pass after https://github.com/packit-service/packit-service/pull/335 is merged

Issue #330, opened 2020-01-14
Proposal for the new status workflow


lachmanfrantisek:
# Copr build defined

1. Trigger processed, starting SRPM build
![Screenshot from 2020-01-14 16-47-23](https://user-images.githubusercontent.com/20214043/72359190-217a4900-36ee-11ea-90e4-d40aa277cc5b.png)
2. SRPM created, running RPM build
![Screenshot from 2020-01-14 16-46-20](https://user-images.githubusercontent.com/20214043/72359191-217a4900-36ee-11ea-8ecb-ee33dc2a3671.png)
3.  RPM build passed
![Screenshot from 2020-01-15 15-10-56](https://user-images.githubusercontent.com/20214043/72440519-4e3f6680-37a9-11ea-8f17-76f7b17fc3ab.png)
![Screenshot from 2020-01-14 16-49-52](https://user-images.githubusercontent.com/20214043/72359188-217a4900-36ee-11ea-9e1d-e1a80110534d.png)

## Error during SRPM

![Screenshot from 2020-01-15 14-57-13](https://user-images.githubusercontent.com/20214043/72439513-5d251980-37a7-11ea-919d-4a503fadd9f9.png)
![Screenshot from 2020-01-15 14-55-30](https://user-images.githubusercontent.com/20214043/72439514-5dbdb000-37a7-11ea-9938-612a6638310e.png)

## Error during build

![Screenshot from 2020-01-15 15-09-40](https://user-images.githubusercontent.com/20214043/72440446-20f2b880-37a9-11ea-8651-480543e8e2b0.png)
![Screenshot from 2020-01-15 15-08-50](https://user-images.githubusercontent.com/20214043/72440448-20f2b880-37a9-11ea-8f87-62a9363848b5.png)

lachmanfrantisek:
# Tests defined

1. Trigger processed, starting SRPM build
![Screenshot from 2020-01-15 17-57-08](https://user-images.githubusercontent.com/20214043/72454240-c82f1a00-37c0-11ea-89ff-6bcff3740f45.png)
2. SRPM created, running RPM build
![Screenshot from 2020-01-15 17-58-02](https://user-images.githubusercontent.com/20214043/72454239-c7968380-37c0-11ea-9e6d-9d22b0d6d993.png)
3. RPM build passed, running testing farm
[...]

PR #338, opened 2020-01-17
Simplify statuses

lachmanfrantisek:
- Do not create aditional statuses.
- Run testing farm for one target after copr build-end.

----

Fixes: #330

### DIFF
--- a/packit_service/worker/copr_build.py
+++ b/packit_service/worker/copr_build.py
@@ -382,7 +382,7 @@ class CoprBuildJobHelper(JobHelper):
     def _process_failed_srpm_build(self, ex):
         sentry_integration.send_to_sentry(ex)
         msg = (
-            f"There was an error while creating the SRPM. {MSG_RETRIGGER}\n"
+            f"There was an error while creating SRPM. {MSG_RETRIGGER}\n"
             "\nOutput:"
             "\n```\n"
             f"{ex}"
@@ -400,7 +400,7 @@ class CoprBuildJobHelper(JobHelper):
         else:
             output = ex.output
         msg = (
-            f"There was an error while creating the SRPM. {MSG_RETRIGGER}\n"
+            f"There was an error while creating SRPM. {MSG_RETRIGGER}\n"
             "\nOutput:"
             "\n```\n"
             f"{output}"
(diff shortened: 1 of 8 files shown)


packit_service/worker/copr_build.py
```diff
@@ -396,7 +400,7 @@ def _process_failed_command(self, ex):
         else:
             output = ex.output
         msg = (
-            f"There was an error while creating a SRPM. {MSG_RETRIGGER}\n"
+            f"There was an error while creating the SRPM. {MSG_RETRIGGER}\n"
```
jpopelka: I'd leave it without any article

packit_service/worker/testing_farm.py
```diff
@@ -67,84 +69,119 @@ def __init__(
...
+            ),
+        )
 
-            stg = "-stg" if self.config.deployment == Deployment.stg else ""
-            copr_repo_name = (
-                f"packit/{self.project.namespace}-{self.project.repo}-"
-                f"{self.event.pr_id}{stg}"
+    def run_testing_farm(self, chroot: str) -> HandlerResults:
+        if chroot not in self.tests_chroots:
+            # Who triggered that?
```
jpopelka: naughty, naughty ;-)

packit_service/worker/testing_farm.py
```diff
@@ -67,84 +69,119 @@ def __init__(
...
-            }
+        check_name = PRCheckName.get_testing_farm_check(chroot)
+        pipeline_id = str(uuid.uuid4())
+        logger.debug(f"Pipeline id: {pipeline_id}")
+        payload: dict = {
+            "pipeline": {"id": pipeline_id},
+            "api": {"token": self.config.testing_farm_secret},
+        }
+
+        logger.debug(f"Payload: {payload}")
```
jpopelka: you have one more `logger.debug(payload)` few lines below, do you need both?

packit_service/worker/testing_farm.py
lachmanfrantisek: I don't want to add classical "this should not happen", but it should not really happen. I only want to be sure that we don't break anything in the future...

packit_service/worker/testing_farm.py
lachmanfrantisek: Hmm, I only moved that code..;) Do we want to debug testing farm token?

jpopelka: I don't mind leaking tokens into logs. Whoever can get to the pods logs, can most likely get to secrets too anyway.

I don't understand your note about moving code. (friday, you know)

lachmanfrantisek: I will remove the payload one.

> moved code

(I haven't written those lines. I've only moved that code from other file, so I am not sure about the reason. Maybe some debugging by Rado from the early stages of testing-farm suppport?)

PR #340, opened 2020-01-20
Handle failed status from testing farm

dhodovsk:


packit_service/worker/testing_farm.py
```diff
@@ -167,11 +167,13 @@ def run_testing_farm(self, chroot: str) -> HandlerResults:
             # success set check on pending
             if req.status_code != 200:
                 # something went wrong
-                msg = req.json()["message"]
+                try:
+                    msg = req.json()["message"]
```
lachmanfrantisek: What about this:

```python
if req.json() and "message" in `req.json()`:
    msg = req.json()["message"]
else:
    msg = f"Failed to submit tests: {req.reason}"
```

dhodovsk: Whatever :D

PR #378, opened 2020-01-29
Remove redundant log messages

jpopelka:
[GithubAppInstallationHandler.run()](https://github.com/packit-service/packit-service/blob/master/packit_service/worker/github_handlers.py#L191) calls [whitelist.add_account()](https://github.com/packit-service/packit-service/blob/master/packit_service/worker/whitelist.py#L102) and both log almost the same messages.
Remove those in add_account().

### DIFF
--- a/packit_service/worker/whitelist.py
+++ b/packit_service/worker/whitelist.py
@@ -101,10 +101,10 @@ class Whitelist:
 
     def add_account(self, github_app: InstallationEvent) -> bool:
         """
-        Add account to whitelist, if automatic verification of user
-        (check if user is packager in fedora) fails, account is still inserted in whitelist
-         with status : `waiting`.
-         Then a scripts in files/scripts have to be executed for manual approval
+        Add account to whitelist.
+        Status is set to 'waiting' or to 'approved_automatically'
+        if the account is a packager in Fedora.
+
         :param github_app: github app installation info
         :return: was the account (auto/already)-whitelisted?
         """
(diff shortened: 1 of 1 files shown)


jpopelka:
recheck

packit_service/worker/whitelist.py
```diff
@@ -101,12 +101,11 @@ def get_account(self, account_name: str) -> Optional[dict]:
 
     def add_account(self, github_app: InstallationEvent) -> bool:
         """
-        Add account to whitelist, if automatic verification of user
-        (check if user is packager in fedora) fails, account is still inserted in whitelist
-         with status : `waiting`.
-         Then a scripts in files/scripts have to be executed for manual approval
+        Add account to whitelist.
+        Status is set to approved_automatically' or 'waiting' (in case the account is not in FAS).
```
csomh:
```suggestion
        Status is set to 'waiting' or to 'approved_automatically' if the account is a packager in Fedora.
```

- comment follows the actual order things are going to happen in the code (the reader doesn't have to make the mental inversion).
- the condition in the context of this method is "is packager in Fedora", FAS is an implementation detail of `_is_packager`.

packit_service/worker/whitelist.py
```diff
@@ -101,12 +101,11 @@ def get_account(self, account_name: str) -> Optional[dict]:
...
+
         :param github_app: github app installation info
-        :return: was the account (auto/already)-whitelisted?
+        :return: has the account been (auto/already)-whitelisted?
```
csomh: :nerd_face: 

Okay, I looked up what is the difference between "was" and "has been". But b/c this is a method returning a Boolean flag indicating whether the action (whitelisting) happened or not (i.e. whitelisting was finished), I think "was" would be more correct.

jpopelka:
PTAL

jpopelka:
recheck

csomh:
Hey @jpopelka, please squash the two commits to have a nicer history.

jpopelka:
Sure, done.

@TomasTomecek looks like I'm not the biggest nitpicker in our team anymore :-)  ^^^

jpopelka:
recheck

jpopelka:
Ok, given that the tests had been green before I only squashed the commits, I assume it's ok to merge this even Zuul is broken at the moment.

TomasTomecek:
so, nitpicking competition?

PR #381, opened 2020-02-03
[Zuul] F30 -> F31

jpopelka:


jpopelka:
Ok, so it won't be that easy:
```
Unable to start service docker: Job for docker.service failed because the control process exited with error code.
```

TomasTomecek:
> Ok, so it won't be that easy:
> 
> ```
> Unable to start service docker: Job for docker.service failed because the control process exited with error code.
> ```

Yeah, we need to boot w/ cgroup v1

PR #390, opened 2020-02-11
Fix report in whitelist

lachmanfrantisek:
- Remove code related to static checks.
- Use JobHelper in whitelist to correctly report failure.

Fixes https://github.com/packit-service/packit-service/issues/382

### DIFF
--- a/packit_service/constants.py
+++ b/packit_service/constants.py
@@ -6,12 +6,8 @@ SANDCASTLE_PVC = "SANDCASTLE_PVC"
 
 CONFIG_FILE_NAME = "packit-service.yaml"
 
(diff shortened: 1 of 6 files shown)


TomasTomecek:
recheck

TomasTomecek:
the CI failure is a network flake

lachmanfrantisek:
> (passing the service config looks pretty weird, but... no strong feelings)

- At first, I wanted to pass it in the constructor but it's not needed in other places like tests and CLI script and was requires changes in too many places...
- I agree that this method/class needs refactoring -- we/I can look at it.

PR #398, opened 2020-02-12
All jobs for gate in zuul

lachmanfrantisek:
- Run tests for gate pipeline in zuul.


TomasTomecek:
I just realized that zuul waits for its own local/check before it starts the gating process - which is interesting - nevertheless, we definitely want everything to be checked in the gating process

TomasTomecek:
mine got merged sooner, you probably need to rebase: https://github.com/packit-service/packit-service/pull/319

PR #403, opened 2020-02-13
[Whitelist] approve account if it signed FPCA (packager is not enough)

jpopelka:
Or do we need to check for the `"packager"` as well?

### DIFF
--- a/packit_service/worker/whitelist.py
+++ b/packit_service/worker/whitelist.py
@@ -53,7 +53,7 @@ class Whitelist:
             username=fas_user, password=fas_password
         )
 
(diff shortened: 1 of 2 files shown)


csomh:
This is nowhere mentioned in packit.dev, but it was mentioned during the [DevConf workshop](https://github.com/packit-service/packit-service/wiki/DevConf2020---Workshop#prerequisities), and there it was FPCA, not packager.

And if I remember correctly the motivation mentioned during the workshop was that signing the FPCA serves as a proxy to meet the requirements to build sources in Copr.

jpopelka:
Does it mean that you approve it?

PR #404, opened 2020-02-13
Set environment and custom tags for sentry events

dhodovsk:
* [sentry environment](https://docs.sentry.io/enriching-error-data/environments/?platform=python)
* [configure scope](https://docs.sentry.io/enriching-error-data/scopes/?platform=python#configuring-the-scope)
* [local scope](https://docs.sentry.io/enriching-error-data/scopes/?platform=python#local-scopes)
  * affects events only within the context


csomh:
:tada: 

This is great!

I do understand the "environment" part, but don't really understand why tags will be useful. Would it make sense to capture this info in the commit message for future readers.

dhodovsk:
The reasoning is now present.

PR #320, opened 2020-01-09
Marshmallow schema implementation

sakalosj:
Marshmallow schema implementation

Requires https://github.com/packit-service/packit/pull/657

PR #405, opened 2020-02-14
Set fas_password in service config

dhodovsk:
Fixes: https://github.com/packit-service/packit-service/issues/356

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -385,7 +385,7 @@ The service configuration is an extension of the user configuration from packit.
 
 To add a new service-related property you need to:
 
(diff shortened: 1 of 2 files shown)


csomh:
Wondering if the guidelines for introducing new [Service configuration](https://github.com/packit-service/packit-service/blob/master/CONTRIBUTING.md#service-configuration) values is still valid... Shouldn't this have been updated with the introduction of marshmallow (#320)? /cc @sakalosj 

For me it's not clear at this point how new configuration values should be introduced.

dhodovsk:
I've added a little bit to the doc

PR #414, opened 2020-02-19
fix CI: generate secrets for dashboard

TomasTomecek:


TomasTomecek:
moár errárs

let's see now

PR #420, opened 2020-02-20
logs: set copr build logs url when a build start

TomasTomecek:
with this change, the logs url is set when a build starts, not when it ends

also realised that coprbuildend event worked differently than I thought: so now the status is set correctly in the DB all the time


TomasTomecek:
Tests passed locally, merging since I wouldn't have time to address comments.

PR #426, opened 2020-02-26
Refactor: move handlers to submodule

lachmanfrantisek:
- Move handlers to the `packit_service.worker.handlers` module.


lachmanfrantisek:
I need to fix the importing cycles. We have the code coupled a lot and the `handlers/__init__.py` has a problem with that.

lachmanfrantisek:
> I need to fix the importing cycles. We have the code coupled a lot and the `handlers/__init__.py` has a problem with that.

Should be fixed now. I've added a note about that into `__init__.py`.

PR #443, opened 2020-02-28
Store the date/time of triggering new copr build

### LINKED ISSUE #425: Packit Service API Issues and some suggested additions.
- The packet-service api does not provide the copr build date/time, which is needed for sorting builds by time in relation to packit-service/dashboard#9

-  Also pull request id should be provided alongside other build info in _api/copr-builds/_ (list of all copr builds) instead of having to make new requests to _api/copr-builds/{id}_ (specific build info)

- In some cases _api/copr-builds/{id}_ returns nothing even when _api/copr-builds/_ provides the exact same id. I assume this is unexpected behavior which needs to be fixed. 

Example: In the packet-service **production api**, one of the builds returned by _api/copr-builds/_  (list of all copr builds) is:

` 
 {
    "project": "rpm-software-management-libdnf-768",
    "owner": "packit",
    "chroots": [
      "fedora-rawhide-x86_64",
      "fedora-30-x86_64",
      "fedora-29-x86_64",
      "mageia-cauldron-x86_64",
      "opensuse-tumbleweed-x86_64"
    ],
    "build_id": 997744,
    "status": "failure"
  },
`

However _api/copr-builds/997744_ returns nothing. 

This happens randomly though, for example _api/copr-builds/_ also returns an object with build id 1087767 and in this case _api/copr-builds/1087767_ works perfectly.

IceWreck:
We need to store the time when a build is triggered. (via new PR or comment command )

This is needed for the dashboard as discussed in packit-service/dashboard#9 
Partially fixes #425 

Now, the /api/copr-build looks like:
```
  {
    "project": "IceWreck-hello-world-7",
    "owner": "icewreck",
    "chroots": [
      "fedora-31-x86_64",
      "fedora-30-x86_64",
      "fedora-rawhide-x86_64"
    ],
    "trigger_time": "28-02-2020 19:39:54",
    "build_id": 1259733
  }

```
Right now I've made this addition to the redis models.py which is currently being used. Should I also make the same change to the [sqlalchemy models.py](https://github.com/packit-service/packit-service/blob/master/packit_service/models.py), which I think will be used later and isn't being used right now?


packit_service/service/models.py
```diff
@@ -141,12 +143,14 @@ def create(
         chroots: List[str],
         identifier: int = None,
         save: bool = True,
+        trigger_time: str = None,
```
TomasTomecek: this is not very descriptive, how about we added three fields instead?
* build_submitted_time
* build_start_time
* build_finished_time

packit_service/service/models.py
```diff
@@ -141,12 +143,14 @@ def create(
...
     ):
         b = cls()
         b.identifier = identifier  # generate new ID if None
         b.project = project
         b.owner = owner
         b.chroots = chroots
+        b.trigger_time = datetime.now().strftime("%d-%m-%Y %H:%M:%S")
```
jpopelka: I'd probably store just the timestamp as `int`, i.e.
```python
b.trigger_time: int = int(time.time())
```
One can later create a string from the timestamp according to preferences, for example
```python
datetime.utcfromtimestamp(b.trigger_time).isoformat()
```

IceWreck: I actually removed it just a couple minutes ago. I set it to datetime.utcnow in both postgres and redis, using postgre's datetime field. 

This can be reformatted easily too, though reformatting might be an issue in other languages.

So should I store the timestamp?

IceWreck:
@TomasTomecek @jpopelka  I'm getting this error:

```
psycopg2.ProgrammingError: column copr_builds.build_submitted_time does not exist
LINE 1: ...lds.build_logs_url AS copr_builds_build_logs_url, copr_build...

```

Build are not getting stored. Can you please tell me what is the issue here?

packit_service/service/models.py
sakalosj: i was using it like this, and it worked perfectly
:+1:

IceWreck: > i was using it like this, and it worked perfectly
It works as intended if you only modify the redis model.py (the old commit). 

I then modified the sql alchemy models.py to do the same thing. There is a bug in there which causes the error and then builds aren't stored.

packit_service/service/models.py
```diff
@@ -141,12 +145,14 @@ def create(
...
         identifier: int = None,
         save: bool = True,
+        build_submitted_time: str = None,
     ):
         b = cls()
         b.identifier = identifier  # generate new ID if None
         b.project = project
         b.owner = owner
         b.chroots = chroots
+        b.build_submitted_time = datetime.utcnow()
```
jpopelka: so what type is the `build_submitted_time` now?
`str` or `datetime.datetime` ?

packit_service/service/models.py
jpopelka: I'm fine with storing datetime, no need to store timestamp.

packit_service/service/models.py
IceWreck: Right now its a string. I'll change it to an integer timestamp once I figure out how to fix the postgres problem.

packit_service/models.py
```diff
@@ -154,7 +154,7 @@ class CoprBuild(Base):
     build_logs_url = Column(String)
     # datetime.utcnow instead of datetime.utcnow() because its an argument to the function
     # so it will run when the copr build is initiated, not when the table is made
-    build_submitted_time = Column(DateTime, nullable=False, default=datetime.utcnow)
+    build_submitted_time = Column(DateTime, default=datetime.utcnow)
```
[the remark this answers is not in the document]
IceWreck: Sorry, I thought it was causing the issue so I removed it. Will revert this once I fix the error.

BTW, I just checked, adding any field to the copr_builds table in the postgres models.py file causes this error.

Example, adding `abc = Column(Integer)` inside the table in an unmodified file from master will cause this issue.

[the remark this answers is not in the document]
sakalosj: no need to sorry :), i just thought it was good idea to have it not nullable

yes you are right you have to update db schema once you change db definition in code

@TomasTomecek 
what is exact procedure to perform db schema update?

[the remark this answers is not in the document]
TomasTomecek: generate it using alembic: https://alembic.sqlalchemy.org/en/latest/autogenerate.html

as for nullable: the thing is that we already have this in production with data in the table: once you change schema, you need to figure out what to do with present entries: either make it nullable, set a default, or figure it out during schema upgrade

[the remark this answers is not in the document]
IceWreck: Thanks. 
My current local setup is docker based, so I make changes in packit-service files and then run docker-compose up with --build to test those changes.

Now running autogenerate probably changes some schema permanently, but where do I run it. I can't run it in a regular venv because alembic just says  No module named 'packit_service'.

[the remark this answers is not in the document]
TomasTomecek: autogenerate only creates the schema files which are then meant to be applied by the upgrade command

hence autogen is safe to run

> Is there a way I can run autogenerate in docker packit-service container and then copy the changed files over to the actual code folder ?

yes, that should do

I usually run all this directly on my laptop (no venv, no containers) by starting a psql container and then exposing env vars in shell so alembic knows how to connect

IceWreck:
@TomasTomecek 
I've done the required changes, and tested locally. Everything's working fine. (Checked the api for redis changes and used the psql cli to verify the yet unused psql changes)

I also changed _alembic/env.py_ for easier autogenerates. But it is failing mypy despite being the recommended method in the alembic docs.

packit_service/models.py
```diff
@@ -151,6 +151,11 @@ class CoprBuild(Base):
     web_url = Column(String)
     # url to copr build logs
     build_logs_url = Column(String)
+    # time.time instead of time.time() because its an argument to the function
+    # so it will run when the copr build is initiated, not when the table is made
+    build_submitted_time = Column(Integer, default=time.time)
+    build_start_time = Column(Integer)
+    build_finished_time = Column(Integer)
```
TomasTomecek: what do others think? int or datetime? I personally prefer datetime, I can see that SA supports DT natively https://docs.sqlalchemy.org/en/13/core/type_basics.html#sqlalchemy.types.DateTime

alembic/versions/509652e21254_store_build_timestamp.py
```diff
@@ -0,0 +1,42 @@
...
+    op.add_column(
+        "copr_builds", sa.Column("build_finished_time", sa.Integer(), nullable=True)
+    )
+    op.add_column(
+        "copr_builds", sa.Column("build_start_time", sa.Integer(), nullable=True)
+    )
+    op.add_column(
+        "copr_builds", sa.Column("build_submitted_time", sa.Integer(), nullable=True)
+    )
+    op.drop_column("copr_builds", "logs")
```
TomasTomecek: hm, please don't drop that field

alembic/versions/509652e21254_store_build_timestamp.py
```diff
@@ -0,0 +1,42 @@
...
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "copr_builds", sa.Column("logs", sa.TEXT(), autoincrement=False, nullable=True)
```
TomasTomecek: ^

packit_service/models.py
lachmanfrantisek: If there is no big issue with `datetime`, I would go with that.

alembic/env.py
```diff
@@ -18,7 +19,7 @@
 # for 'autogenerate' support
 # from myapp import mymodel
 # target_metadata = mymodel.Base.metadata
-target_metadata = None
+target_metadata = Base.metadata
```
TomasTomecek: I wonder if it helped mypy to do something like this:
```python
class OurBase(Base):
  metadata: WhateverTypeItIs

target_metadata = OurBase
```

TomasTomecek:
And could also please update the tests in tests_requre/test_db to reflect these changes? Thank you

alembic/versions/509652e21254_store_build_timestamp.py
IceWreck: We have to manually remove this from the generated schema right? Because the _copr_builds_ table in models.py does not have a column called _logs_ but it is in the old schema, which is why autogenerate removed it.

packit_service/models.py
jpopelka: I already said that datetime is perfect.
I had initially suggested timestamp because it's IMHO better than storing string.

IceWreck: @jpopelka  Switched back to datetime in the latest commit. In redis, I'm currently storing the same datetime object as a string, but I think it can be converted back into a datetime object.

alembic/versions/509652e21254_store_build_timestamp.py
TomasTomecek: oh, got it; let's leave that then, it's true that I moved logs from copr build table to the srpm build table

IceWreck:
Finally passed, Yay! 
@TomasTomecek  Done.

packit_service/models.py
TomasTomecek: and we're soon moving away from redis, so I wouldn't worry about it too much

alembic/env.py
```diff
@@ -18,7 +22,12 @@
 # for 'autogenerate' support
 # from myapp import mymodel
 # target_metadata = mymodel.Base.metadata
-target_metadata = None
+
+# issue highlighted in sqlalchemy models.py
+if TYPE_CHECKING:
+    pass
+else:
```
TomasTomecek:
```suggestion
if not TYPE_CHECKING:
```

tests_requre/test_db.py
```diff
@@ -82,6 +82,10 @@ def test_create_copr_build(a_copr_build):
     assert a_copr_build.srpm_build.logs == "asd\nqwe\n"
     assert a_copr_build.target == TARGET
     assert a_copr_build.status == "pending"
+    # Since datetime.utcnow() will return different results in every time its called,
+    # we will check if a_copr_build has build_submitted_time value thats within the past hour
+    time_last_hour = datetime.utcnow() - timedelta(hours=1)
+    assert a_copr_build.build_submitted_time > time_last_hour
```
TomasTomecek: Yes, this is good enough - we just need to know that the datetime was saved in DB, it's valid and it can be retrieved back.

packit_service/models.py
```diff
@@ -151,6 +151,11 @@ class CoprBuild(Base):
     web_url = Column(String)
     # url to copr build logs
     build_logs_url = Column(String)
+    # datetime.utcnow instead of datetime.utcnow() because its an argument to the function
+    # so it will run when the copr build is initiated, not when the table is made
+    build_submitted_time = Column(DateTime, default=datetime.utcnow)
+    build_start_time = Column(DateTime)
+    build_finished_time = Column(DateTime)
```
TomasTomecek: interesting fact is that these default to None:
```python
ipdb> a_copr_build                         
COPRBuild(id=13, pr=PullRequest(id=1, project=GitProject(name=nirvana/lithium)))
ipdb> a_copr_build.build_submitted_time    
datetime.datetime(2020, 3, 5, 9, 17, 24, 689180)
ipdb> a_copr_build.build_start_time        
ipdb> print(a_copr_build.build_start_time) 
None
```

Issue #239, opened 2019-11-15
p-s to react on pushes to an upstream branch


TomasTomecek:
This request is coming from libguestf family of projects:

https://github.com/libguestfs

Their workflow is to review patches on ML and have a tooling to push changes to the github git repos. This project is not using issues nor pull requests. Hence the project requires to have a trigger on push to a branch instead of reacting to PRs

CC @rwmjones

TomasTomecek:
AI: implement this - test pushes to a branch (configurable), set commit status once the build & tests are done
AI: create an issues if something fails (configurable) - stretch goal

TomasTomecek:
@lachmanfrantisek isn't this done?

lachmanfrantisek:
> @lachmanfrantisek isn't this done?

Not yet.

- [ ] Turn on the events in the production github app
- [ ] Finish the database update. Otherwise, only the baby-sit task will return the responses.

TomasTomecek:
good points, thanks for the update!

bocekm:
Hi guys, how does this look like?

lachmanfrantisek:
@bocekm Sorry, we were very busy with the centos-stream in the last two weeks. I would like to finish that in the next week.

lachmanfrantisek:
Finally, we are there!

The workflow and configuration are the same as for pull-requests. Just use `trigger: commit` in the job configuration. (More details can be found in [the documentation](https://packit.dev/docs/configuration/#supported-jobs).)  
  
```yaml  
- job: copr_build  
  trigger: commit  
  metadata:  
    branch: build-branch  
```  
  
The results are shown as commit status. (You can access it with selecting the branch in the `commits` tab on GitHub.)  
  
---  
  
An example build for new commit to the branch can be found here:  
https://github.com/packit-service/hello-world/commits/build-branch  
  
1. Build triggered. Building SRPM.  
![Screenshot from 2020-06-01 13-13-55](https://user-images.githubusercontent.com/20214043/83407731-a4880b80-a411-11ea-89dd-40e6bfcca71b.png)  
2. SRPM successful. Building RPM in Copr.  
![Screenshot from 2020-06-01 13-17-29](https://user-images.githubusercontent.com/20214043/83407730-a3ef7500-a411-11ea-9cdc-09744f9de035.png)  
3. Build successful. Running tests in Testing Farm.  
[...]

PR #448, opened 2020-03-03
Pre-check for jobs and release/push copr triggers

lachmanfrantisek:
- Allow to pre-check the jobs to fail early.
- Use the right configured job when running copr builds.
- Add copr-build for release and push. (related issue: #239)
  - I would like to wait with docs and test it properly on stage.
- Refactor the handlers/events.


packit_service/service/events.py
```diff
@@ -465,7 +499,7 @@ def __init__(
             self.base_repo_namespace = build.get("repo_namespace")
             https_url = build["https_url"]
 
-        super().__init__(trigger=JobTriggerType.commit, project_url=https_url)
+        super().__init__(trigger=JobTriggerType.pull_request, project_url=https_url)
```
lachmanfrantisek: This is the problem I want to discuss on the arch meeting:
- We need to differentiate between jobs and the `part of the config which contains the metadata about the job`.
- We need to support mapping from jobs to the `part of the config which contains the metadata about the job`.

packit_service/worker/build/build_helper.py
```diff
@@ -85,20 +105,6 @@ def api(self) -> PackitAPI:
...
-            return self.event.base_ref
-
-        if isinstance(self.event, CoprBuildEvent):
-            if self.event.ref:
-                # FIXME: return commit sha always once on PG
-                return self.event.ref
-            return self.event.commit_sha
-
-        if isinstance(self.event, PullRequestCommentEvent):
-            return self.event.commit_sha
```
lachmanfrantisek: I've tried to remove this kind of code.

packit_service/service/events.py
```diff
@@ -518,9 +552,10 @@ def pre_check(self):
 
     def get_dict(self, default_dict: Optional[Dict] = None) -> dict:
         d = self.__dict__
-        d.pop("build_pg")
+        build_pg = d.pop("build_pg")
         result = super().get_dict(d)
         result["topic"] = result["topic"].value
+        self.build_pg = build_pg
```
lachmanfrantisek: Honestly, I am not 100% sure what I am doing here -- why we need to edit `__dict__` here just for the get method?
(I understand, that the `build_pg` is not serializable.)

TomasTomecek: this is all just for sake of serialization and the way this is implemented atm

packit_service/worker/handlers/abstract.py
```diff
@@ -112,6 +112,14 @@ def _clean_workplace(self):
             else:
                 shutil.rmtree(item)
 
+    def pre_check(self) -> bool:
+        """
+        Validate the input.
+
+        :return: False if we can skip the run.
```
TomasTomecek: pre_cache -> False mean that run() will not be executed?

lachmanfrantisek: Currently only for the configured jobs (in the `process_jobs` method), not the hardcoded ones in the `process_message`. Shall I use that also there?

packit_service/service/events.py
lachmanfrantisek: Is this change safe? (I was having weird errors in tests because of the missing attribute `build_pg`...)

lachmanfrantisek:
I will merge it to play with that on stage tomorrow. (So I have time to fix potential problems until Friday..;)

Feel free to continue with the review.

packit_service/worker/handlers/github_handlers.py
```diff
@@ -235,87 +235,168 @@ def run(self) -> HandlerResults:
...
+    ):
+        super().__init__(config=config, job=job, event=event)
+
+    def run(self) -> HandlerResults:
+        if isinstance(self.event, GithubPullRequestHandler):
+            collaborators = self.project.who_can_merge_pr()
+            if self.event.github_login not in collaborators | self.config.admins:
+                msg = (
+                    "Only users with write or admin permissions to the repository "
+                    "can trigger Packit-as-a-Service"
```
jpopelka: I've seen lots of those in Hunor's PR. Can we refactor them into a constant? for example `ADMINS_ONLY`

packit_service/worker/handlers/abstract.py
TomasTomecek: I'd love to see this text in the docstring - the present docstring is really sparse.

packit_service/service/events.py
TomasTomecek: should be fine

packit_service/worker/handlers/github_handlers.py
lachmanfrantisek: Fixed in #451

packit_service/worker/handlers/abstract.py
lachmanfrantisek: I've updated the doc string in #451

lachmanfrantisek: I forgot to push that commit, will be in the next one.

PR #450, opened 2020-03-03
Use enum for status states and fix setting of status

lbarcziova:
- implements using of `CommitStatus`enum in statuses
- fixes https://sentry.io/share/issue/15c35a842b2d47aebe971dfa31a6b373/

### DIFF
--- a/packit_service/worker/build/build_helper.py
+++ b/packit_service/worker/build/build_helper.py
@@ -227,7 +227,7 @@ class BaseBuildJobHelper:
 
     def _report(
         self,
-        state: Union[CommitStatus, str],
+        state: CommitStatus,
         description: str,
         url: str = "",
         check_names: Union[str, list, None] = None,
@@ -241,7 +241,7 @@ class BaseBuildJobHelper:
         )
 
     def report_status_to_all(
-        self, description: str, state: Union[CommitStatus, str], url: str = ""
+        self, description: str, state: CommitStatus, url: str = ""
     ) -> None:
         self.report_status_to_build(description, state, url)
         self.report_status_to_tests(description, state, url)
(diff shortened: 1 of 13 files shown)


packit_service/worker/handlers/github_handlers.py
```diff
@@ -40,6 +40,7 @@
 from packit.local_project import LocalProject
 
 from packit_service.config import ServiceConfig, GithubPackageConfigGetter
+from packit_service.constants import GITHUB_FAILURE_STATUS
```
lachmanfrantisek: Please, use the `CommitStatus` enum instead:

https://github.com/packit-service/ogr/blob/63774270a240f9bd7064d33259417e7564a1e1a4/ogr/abstract.py#L536-L542

TomasTomecek:
oh my, so there's more?! thanks for fixing this, I'm dumb

packit_service/worker/handlers/github_handlers.py
lbarcziova: I will refactor it

lachmanfrantisek: Thanks!

packit_service/worker/build/build_helper.py
```diff
@@ -240,7 +240,9 @@ def _report(
             description=description, state=state, url=url, check_names=check_names,
         )
 
-    def report_status_to_all(self, description: str, state: str, url: str = "") -> None:
+    def report_status_to_all(
+        self, description: str, state: Union[CommitStatus, str], url: str = ""
```
lachmanfrantisek: It would be nice to not allow `str` and let the `mypy` find the wrong calls...

lbarcziova: changed

PR #451, opened 2020-03-04
Revert getting job per trigger

lachmanfrantisek:
- Do not filter job-configs per trigger. (To make the stg working.)
  - Then `comments` and other events are not matched.
  - Fixes: https://sentry.io/organizations/red-hat-0p/issues/1546065384/?environment=stg&project=1767823&query=is%3Aunresolved

I'll make a propper fix but need some time for that.


lachmanfrantisek:
I've incorporated your notes from the previous PR and rebased.

PR #452, opened 2020-03-04
Fix ref for pr events and validate push events

lachmanfrantisek:
- Improve the pre-check doc string.
- Use None as a ref for PR events.
  - Fixes https://sentry.io/organizations/red-hat-0p/issues/1546457825
- Validate accounts for push event.
  - Fixes: https://sentry.io/organizations/red-hat-0p/issues/1546506651
- Add the real webhook for push tests.


packit_service/worker/handlers/abstract.py
```diff
@@ -114,9 +114,9 @@ def _clean_workplace(self):
 
     def pre_check(self) -> bool:
         """
-        Validate the input.
+        Validate the input and skip the run if this returns False.
```
csomh: Does documenting this here make sense? The skip itself happens in worker.jobs.process_message() if I understand it correctly. So it's up to the caller of this method to decide what to do with the return value.

packit_service/service/events.py
```diff
@@ -240,7 +241,7 @@ def __init__(
         self.target_repo = target_repo
         self.commit_sha = commit_sha
         self.github_login = github_login
-        self.ref = pr_id
+        self.ref = None
```
csomh: Does this mean that for *PullRequestEvent*s there is no `ref`? Is `ref` a Git-ref?

I'm trying to understand why "Use None as a ref for PR events" is required. Maybe it would be nice to provide more context on this change in the commit message?

packit_service/worker/handlers/abstract.py
TomasTomecek: but "Validate the input" is very sparse and could mean anything, so I requested some more context for this method

lachmanfrantisek: Do you want the same message as we have for events?
```
        Implement this method for those events, where you want to check if event properties are
        correct. If this method returns false during runtime, execution of service code is skipped.
```

> So it's up to the caller of this method to decide what to do with the return value.

Exactly.

packit_service/service/events.py
lachmanfrantisek: - Yes, I wanted to rename that to `git_ref` -- It looks like it would be useful.
- With git-refs for PRs, it's harder. (We are using that ref for checkout and PR needs to be checkout-ed differently.)

The problem was here, that we first checkout `ref` and then the PR:
https://github.com/packit-service/packit/blob/ee02ce960cb6ef42b6cdf30504ac1e7ff2edb56e/packit/local_project.py#L123-L126

packit_service/worker/handlers/abstract.py
csomh: Aha! I see now! Yes, a message similar to the one events have would be much clearer.

packit_service/service/events.py
csomh: I understand now. Thanks for the explanation!

lachmanfrantisek: I've introduced the two properties:
- the one for git checkout
- the second one for identification and presentation purpose (e.g. copr project name, logs, ...)

packit_service/worker/handlers/abstract.py
lachmanfrantisek: I've used the message from events. Is that ok?

lachmanfrantisek:
@csomh @TomasTomecek PTAL

packit_service/worker/handlers/abstract.py
```diff
@@ -114,9 +114,9 @@ def _clean_workplace(self):
 
     def pre_check(self) -> bool:
         """
-        Validate the input.
-
-        :return: False if we can skip the run.
+        Implement this method for those handlers, where you want to check if the properties are
+        correct. If this method returns False during runtime, execution of service code is skipped.
+        :return:
```
csomh: Is this `:return:` still needed?

lachmanfrantisek: Copied from the other one... I'll fix both of them.

packit_service/service/events.py
```diff
@@ -165,6 +165,10 @@ class AbstractGithubEvent(Event, GithubPackageConfigGetter):
     def __init__(self, trigger: JobTriggerType, project_url: str):
         super().__init__(trigger)
         self.project_url: str = project_url
+        self.git_ref: Optional[str] = None  # git ref that can 'git checkout'-ed
```
TomasTomecek:
```suggestion
        self.git_ref: Optional[str] = None  # git ref that can be 'git checkout'-ed
```

packit_service/service/events.py
```diff
@@ -165,6 +165,10 @@ class AbstractGithubEvent(Event, GithubPackageConfigGetter):
...
+        self.identifier: Optional[str] = (
```
TomasTomecek: identifier of the respective object right? PR, issue, release...

lachmanfrantisek:
Thanks for the review.

PR #453, opened 2020-03-04
Split job and jobconfig (service)

lachmanfrantisek:
- Split the JobTriggerType to one related to config and one for actual triggers.
- Requires: https://github.com/packit-service/packit/pull/746

TODO:
- [x] mapping of the `TheJobTriggerType` to `JobConfigTriggerType`.
- [x] tests for config with multiple triggers for one job
- [x] test the build workflow for other build triggers (skipped, since we need to use right trigger for received events from copr.)

### DIFF
--- a/packit_service/service/events.py
+++ b/packit_service/service/events.py
@@ -85,7 +85,8 @@ class TheJobTriggerType(str, enum.Enum):
     commit = "commit"
     installation = "installation"
     testing_farm_results = "testing_farm_results"
-    comment = "comment"
+    pr_comment = "pr_comment"
+    issue_comment = "issue_comment"
 
 
 class TestResult(dict):
@@ -295,7 +296,7 @@ class PullRequestCommentEvent(AbstractGithubEvent):
         comment: str,
         commit_sha: str = "",
     ):
-        super().__init__(trigger=TheJobTriggerType.comment, project_url=https_url)
+        super().__init__(trigger=TheJobTriggerType.pr_comment, project_url=https_url)
         self.action = action
         self.pr_id = pr_id
         self.base_repo_namespace = base_repo_namespace
@@ -344,7 +345,7 @@ class IssueCommentEvent(AbstractGithubEvent):
             str
         ] = "master",  # default is master when working with issues
     ):
-        super().__init__(trigger=TheJobTriggerType.comment, project_url=https_url)
+        super().__init__(trigger=TheJobTriggerType.issue_comment, project_url=https_url)
         self.action = action
         self.issue_id = issue_id
         self.base_repo_namespace = base_repo_namespace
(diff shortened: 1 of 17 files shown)


packit_service/service/events.py
```diff
@@ -75,6 +78,16 @@ class TestingFarmResult(str, enum.Enum):
     running = "running"
 
 
+class TheJobTriggerType(str, enum.Enum):
```
TomasTomecek: that's a funny name

packit_service/worker/handlers/fedmsg_handlers.py
```diff
@@ -124,21 +124,19 @@ def run(self) -> HandlerResults:
 @add_topic
 @add_to_mapping
 class NewDistGitCommitHandler(FedmsgHandler):
-    """ A new flag was added to a dist-git pull request """
+    """ ??? A new flag was added to a dist-git pull request """
```
TomasTomecek: ?

lachmanfrantisek: NewDistGitCommit != new flag was aded to dg pull-request

packit_service/service/events.py
lachmanfrantisek: I will probably revert it back but want to be sure I go through all of the occurrences and decide which enum needs to be there.

packit_service/worker/handlers/fedmsg_handlers.py
TomasTomecek: so just update the docstring then, it was probably just copy-pasted

packit_service/service/events.py
TomasTomecek: oki

packit_service/worker/handlers/fedmsg_handlers.py
lachmanfrantisek: I was not sure and (main reason) had no time to check the code properly to fix it. I will correct that.

packit_service/service/events.py
lachmanfrantisek: Are you ok with merging this as is and revert after I test it properly on stage?

TomasTomecek: I am, that's why I approved

packit_service/mapping.py
```diff
@@ -0,0 +1,22 @@
+from typing import Dict, Optional
```
TomasTomecek: this file is really meta and complicated, but I'm fine with it if you are fine

lachmanfrantisek: I'll add some comments and the licence header.

lachmanfrantisek: I'll add some comments and the licence header.

lachmanfrantisek:
recheck

PR #456, opened 2020-03-05
SA: rollback transaction on exc

TomasTomecek:
Is this a better way of doing sqlalch?

Fixes RED-HAT-0P-2AH

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -23,15 +23,19 @@
 """
 Data layer on top of PSQL using sqlalch
 """
+import logging
 import os
-from typing import TYPE_CHECKING, Optional, Union
+from contextlib import contextmanager
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


packit_service/models.py
```diff
@@ -161,16 +168,10 @@ class CoprBuild(Base):
     data = Column(JSON)
 
     def set_status(self, status: str):
-        session = get_sa_session()
         self.status = status
-        session.add(self)
-        session.commit()
```
dhodovsk: How is this going to propagate into database?

TomasTomecek: `autocommit=True`

every time you change a SA object, it gets automatically committed to the database; see the tests :) which means we no longer need to do `.commit()`

dhodovsk: Neat!

sakalosj:
LGTM

PR #461, opened 2020-03-06
CI: verify the deployment is up and running

TomasTomecek:


files/deployment.yaml
```diff
@@ -112,3 +112,9 @@
       args:
         chdir: "{{ deployment_dir }}"
       become: "{{ as_root }}"
+    - name: Make sure the worker is up and running
+      shell: oc get pods/packit-worker-0 | grep Running
```
jpopelka: Don't the pods need some time to be actually 'Running'?
I'm just wondering whether this will work directly after 'make deploy' without any delay or so.

TomasTomecek: Franta pointed this out as well; you guys are right, gonna fix that in deployment - on Monday, now it's weekendz!

TomasTomecek:
closing since this is now done in deployment repo

PR #462, opened 2020-03-09
Fix sentry in worker

lachmanfrantisek:
- Configure sentry only for the first time.
- Fixes RED-HAT-0P-2AV (I hope.)

### DIFF
--- a/tests/unit/test_utils.py
+++ b/tests/unit/test_utils.py
@@ -1,0 +1,94 @@
+# MIT License
+#
+# Copyright (c) 2018-2019 Red Hat, Inc.
+
+# Permission is hereby granted, free of charge, to any person obtaining a copy
+# of this software and associated documentation files (the "Software"), to deal
+# in the Software without restriction, including without limitation the rights
+# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# copies of the Software, and to permit persons to whom the Software is
+# furnished to do so, subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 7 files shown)


tests/unit/test_utils.py
```diff
@@ -0,0 +1,94 @@
+# MIT License
+#
+# Copyright (c) 2018-2019 Red Hat, Inc.
```
csomh: 2020?

lachmanfrantisek: Please, create a new (good-first?) issue -- it's in all files..;(

lachmanfrantisek:
>     1. `lazy_object_proxy` should become a dependency and installed in `files/install-deps*.yaml`. Maybe it's already indirectly pulled in by something else, still, might be better to be explicit.

It should be already there because of the `packit`. I'll check it once again.

>     2\. I'm not really familiar with lazy initialisation. Is there be any considerable performance impact with this?

Nothing I am aware of -- it only postpones the creation of an object until you need to use it. (No big magic.)

lachmanfrantisek:
@csomh is the `setup.cfg` in packit enough?

https://github.com/packit-service/packit/blob/5fcddebb447a465e2c6deb64d9b5fd643fac2739/setup.cfg#L42

csomh:
> @csomh is the `setup.cfg` in packit enough?
> 
> https://github.com/packit-service/packit/blob/5fcddebb447a465e2c6deb64d9b5fd643fac2739/setup.cfg#L42

As this has become an explicit dependency for `packit-service` would be better to have it here, too. So that things don't break here, if it gets removed from `packit`.

lachmanfrantisek:
> As this has become an explicit dependency for `packit-service` would be better to have it here, too. So that things don't break here, if it gets removed from `packit`.

This is applicable to multiple packit dependencies -- do we really want all of them explicitly marked here as well? (The tests will tell us if we break that, won't they?)

csomh:
> > As this has become an explicit dependency for `packit-service` would be better to have it here, too. So that things don't break here, if it gets removed from `packit`.
> 
> This is applicable to multiple packit dependencies -- do we really want all of them explicitly marked here as well? (The tests will tell us if we break that, won't they?)

packit-service and packit are two separate projects. Implicitly inheriting dependencies from a different project is bad, as it creates a tight coupling between the two.

Are we sure that the coverage of our tests is good enough to catch missing dependencies?

And yes, it's clear that many dependencies of `packit-service` are not explicitly stated as such. That can be fixed as part of a different effort. In this PR we should just make sure not to add yet another item to that list.

lachmanfrantisek:
> Are we sure that the coverage of our tests is good enough to catch missing dependencies?

If we import each file of the code in tests, then yes. (I think we do that.)

(I've added the dependency.)

PR #467, opened 2020-03-09
images: store git HEAD info in the image

TomasTomecek:


TomasTomecek:
> Two extension (can be done later):
> 
>     * Can you please send that info to sentry during setup? https://docs.sentry.io/workflow/releases/?platform=python#configure-sdk (Or/and some sentry tags for that)
> 
>     * Can you please load and log that info during the startup so we can see that in the logs easily.

please create an issue for that, it doesn't seem to be trivial to implement right away

lachmanfrantisek:
> please create an issue for that, it doesn't seem to be trivial to implement right away

Ok, I'll create two good-first issues since it should be easy.

PR #468, opened 2020-03-09
Use project_dir variable in playbooks

lbarcziova:


TomasTomecek:
there 2 PRs in the merge queue atm, so rebasing is a thing :)

csomh:
Why do we need this change?

TomasTomecek:
for zuuuuuul

csomh:
> for zuuuuuul

Why? I tried to search for `project_dir` in the zuul docs, this repo and the Ansible docs, but didn't find anything that would explain what it does. Where does this variable even come from?

lbarcziova:
When we want to run some playbook from another repo, as we are doing [here](https://github.com/packit-service/ogr/blob/8fba377ad134d506d768f6981c6a5d7f54552b29/files/tasks/packit_reverse_dep_test.yaml#L17) , `zuul.project.src_dir` is then unknown and it fails.

TomasTomecek:
@csomh `project_dir` is our invention, everything in `zuul` dict is coming from zuul itself; in this PR, you can see that Laura sets project_dir to be populated with values from zuul and then uses it across playbooks - well done!

and that's literally what needs to be done: have the CI-related playbooks play well in the zuul environment

PR #475, opened 2020-03-09
more tests to exercise /packit comment parsing

TomasTomecek:


TomasTomecek:
> I think we should also add some cases where the comment ends or begins with a space, that is what caused the value error earlier today. (Now fixed)

good point, will add more then

Issue #465, opened 2020-03-09
copr build for packit repo was not triggered


TomasTomecek:
Why?
```
[2020-03-09 09:20:26,410: INFO/MainProcess] Received task: task.steve_jobs.process_message[497c6f98-152f-4b39-b24b-15e5fe6a2dc7]  
GitHub PR#753 'opened' event.
[2020-03-09 09:20:26,412: INFO/ForkPoolWorker-1] GitHub PR#753 'opened' event.
Target repo: packit-service/packit.
[2020-03-09 09:20:26,413: INFO/ForkPoolWorker-1] Target repo: packit-service/packit.
[2020-03-09 09:20:28,688: DEBUG/ForkPoolWorker-1] Found a config file '.packit.yaml' on ref '3d3a952a82b930d4c6bd635b55a0ff84a8ef91e0' of the packit-service/packit repository.
[2020-03-09 09:20:28,913: DEBUG/ForkPoolWorker-1] Package config:
{
    "synced_files": [
        {
            "src": "fedora-tests/",
            "dest": "tests/"
        }
    ],
    "upstream_package_name": "packitos",
    "upstream_project_url": "https://github.com/packit-service/packit ",
    "actions": {
        "create-archive": [
            "python3 setup.py sdist --dist-dir .",
            "sh -c 'echo packitos-$(python3 setup.py --version).tar.gz'"
        ],
        "get-current-version": [
            "python3 setup.py --version"
        ]
    },
    "jobs": [
        {
            "job": "propose_downstream",
            "trigger": "release",
            "metadata": {
                "dist-git-branch": "fedora-all"
            }
        },
        {
            "job": "sync_from_downstream",
            "trigger": "commit"
        },
        {
            "job": "copr_build",
            "trigger": "pull_request",
            "metadata": {
                "targets": [
                    "fedora-all"
                ]
            }
        },
        {
            "job": "tests",
            "trigger": "pull_request",
            "metadata": {
                "targets": [
                    "fedora-stable"
                ]
            }
        }
    ]
}
[2020-03-09 09:20:28,913: WARNING/ForkPoolWorker-1] Path to spec file is not set.
TomasTomecek is admin, you shall pass
[2020-03-09 09:20:28,917: INFO/ForkPoolWorker-1] TomasTomecek is admin, you shall pass
09:20:28.917 jobs.py           DEBUG  Running handler: <class 'packit_service.worker.handlers.github_handlers.PushGithubCoprBuildHandler'>
[...]

PR #476, opened 2020-03-10
Rework mapping

lachmanfrantisek:
- Make the mapping of handlers to events work and easier to use.
  - Testing farm results and Copr events are handled as other events.
  - Mapping support multiple triggers for one job and get's a correct config part.
- Fixes: #465

TODO:
- [x] More tests (more combinations of triggers and configs).
- [x] Check the commenting.
- [x] Add docs for the decorators.
- [ ] Remove internal job type (e.g. `copr_end`) from the `JobType` enum in packit. (After this PR, it can contain only the config-relevant values.)

### DIFF
--- a/packit_service/worker/handlers/abstract.py
+++ b/packit_service/worker/handlers/abstract.py
@@ -41,9 +41,9 @@ from packit_service.worker.result import HandlerResults
 
 logger = logging.getLogger(__name__)
 
-JOB_REQUIRED_MAPPING: Dict[JobType, Set[Type["JobHandler"]]] = defaultdict(set)
+MAP_REQUIRED_JOB_TO_HANDLERS: Dict[JobType, Set[Type["JobHandler"]]] = defaultdict(set)
 
-MAP_EVENT_TRIGGER_TO_HANDLER: Dict[
+MAP_EVENT_TRIGGER_TO_HANDLERS: Dict[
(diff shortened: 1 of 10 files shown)


lachmanfrantisek:
The commenting part is still WIP -- I need to check/fix all the tests since I've changed the handling of the comment events. (Now, they should be handled in the same way as other events.)

packit_service/worker/handlers/abstract.py
```diff
@@ -25,32 +25,80 @@
...
 from packit_service.config import ServiceConfig
+from packit_service.sentry_integration import push_scope_to_sentry
 from packit_service.service.events import Event, TheJobTriggerType
 from packit_service.worker.result import HandlerResults
-from packit_service.sentry_integration import push_scope_to_sentry
 
 logger = logging.getLogger(__name__)
 
-JOB_NAME_HANDLER_MAPPING: Dict[JobType, Type["JobHandler"]] = {}
+JOB_REQUIRED_MAPPING: Dict[JobType, Set[Type["JobHandler"]]] = defaultdict(set)
```
jpopelka: Just wondering why this is named differently from the other `MAP_SOMETHING_TO_SOMETHING`, but I don't mind.

lachmanfrantisek:
recheck

Issue #459, opened 2020-03-06
Move from redis to psql as a data store


TomasTomecek:
# Plan

## Stage 1
* [ ] store data in psql and redis in parallel, read data from psql, fall back to redis
  * [x] **required** change REST API code to do this [p-s/service/api](https://github.com/packit-service/packit-service/tree/master/packit_service/service/api) #479
    * assigned to @IceWreck 
  * [x] **required** turn whitelist into psql (schema needs to be created) [p-s/worker/whitelist](https://github.com/packit-service/packit-service/blob/master/packit_service/worker/whitelist.py)
    * assigned to @IceWreck 
  * [ ] store installations-related data into psql (also needs schema and design) [p-s/worker/handlers/github](https://github.com/packit-service/packit-service/blob/master/packit_service/worker/handlers/github_handlers.py)
    * assigned to @rishavanand
  * [x] store testing farm data in psql [p-s/worker/handlers/testing-farm](https://github.com/packit-service/packit-service/blob/master/packit_service/worker/handlers/testing_farm_handlers.py) (done by @lachmanfrantisek in #502 )
  * [x] store information about all tasks/jobs executed in p-s into psql (needs schema and design)
    (done by @lbarcziova in #520)
  * [ ] make sure that data we store in redis are a **subset** of what we put in psql
    * assigned to @rishavanand
  * [ ] get a bottle of champagne and prepare for p2
    * assigned to @rishavanand

## Stage 2
* [ ] create a script to read data from redis and transform them into sqlalchemy objects and save them into psql
  * make the script idempotent - can be run multiple times on the same data set without creating duplicate entries
  * remove the data from redis
  * DON'T RUN THE SCRIPT IN PROD JUST YET

## Stage 3
* [ ] once everything is migrated into psql, get rid off all the redis-related code
* [ ] deploy the code to prod
* [ ] run the script from Stage 2 on prod

---

* needs-design mean that you should first propose a solution (mainly the schema) to the maintainers and get it ack'd before working on the code itself

please update this as needed

IceWreck:
Hey, I can work on atleast parts of Stage 1, though after 9th March (the first two checkboxes)

grapheo12:
I am also interested in this. Can we work together? @TomasTomecek  @IceWreck

grapheo12:
@TomasTomecek I have an action plan in mind for this:

[...]

PR #479, opened 2020-03-11
Rewrite the copr-builds part of the API using PostgresSQL instead of Redis

IceWreck:
Completes the first checkbox of #459 

This is completely back-compatible with the previous redis API, although I added more information than before because we need it for the dashboard.

Some previously used info was not stored in psql so I added fields for them and modified the code to store them.

_/api/copr-builds_ now returns a list of objects where one item looks like:
```
{
    "project": "IceWreck-hello-world-9",
    "owner": "icewreck",
    "repo_name": "hello-world",
    "build_id": "1300479",
    "status": "pending",
    "chroots": [
      "fedora-31-x86_64",
      "fedora-rawhide-x86_64",
      "fedora-30-x86_64"
    ],
    "build_submitted_time": "11/03/2020 10:09:48",
    "repo_namespace": "IceWreck",
    "web_url": "https://copr.fedorainfracloud.org/coprs/icewreck/IceWreck-hello-world-9/build/1300479/"
}
```

And the specific copr build info returns everything we know about said build.
Screenshot_2020-03-11 https packit abifog com

Also moving to postgres fixed the issue where specific copr build info was not shown (stored?) unless fedora messaging was working.


TomasTomecek:
@IceWreck don't worry about the tests now, I broke the deployment, fixing it right now

packit_service/models.py
```diff
@@ -169,6 +169,12 @@ class CoprBuild(Base):
     build_submitted_time = Column(DateTime, default=datetime.utcnow)
     build_start_time = Column(DateTime)
     build_finished_time = Column(DateTime)
+    # git url
+    https_url = Column(String)
```
TomasTomecek: 2 things:
1. I don't see any code which would ingest this
2. git web url is not directly related to a copr build, can we move that field to the GitProject model?

Could you also put an example data to the comment? we have so many URLs that it's so easy to get lost

packit_service/models.py
```diff
@@ -169,6 +169,12 @@ class CoprBuild(Base):
...
+    # project name as shown in copr
+    project_name = Column(String)
+    owner = Column(String)
```
TomasTomecek: nice!

packit_service/service/api/copr_builds.py
```diff
@@ -41,22 +41,50 @@ class CoprBuildsList(Resource):
...
-        resp.headers["Content-Range"] = f"copr-builds {first + 1}-{last}/{len(result)}"
-        resp.headers["Content-Type"] = "application/json"
-        return resp
+        """ List all Copr builds. """
+
+        # Return relevant info thats concise
+        # Usecases like the packit-dashboard copr-builds table
+
+        with get_sa_session() as session:
+            builds_list = session.query(CoprBuild).all()
```
TomasTomecek: could you please create a method on the CoprBuild class which would do what you need and not import get_sa_session here? I'd love to have all the database logic in models.py so we don't duplicate code

packit_service/service/api/copr_builds.py
```diff
@@ -41,22 +41,50 @@ class CoprBuildsList(Resource):
...
+                            "%d/%m/%Y %H:%M:%S"
+                        ),
+                        "repo_namespace": build.pr.project.namespace,
+                        "web_url": build.web_url,
+                    }
+                    # same_buildid_builds are copr builds created due to the same trigger
+                    # multiple identical builds are created which differ only in target
+                    # so we merge them into one
+                    same_buildid_builds = session.query(CoprBuild).filter(
+                        CoprBuild.build_id == build.build_id
```
TomasTomecek: I think this should be `.filter(build_id=build.build_id)`

IceWreck: I'll check if that works, but _CoprBuild.build_id_ is working too.

packit_service/models.py
IceWreck: > I don't see any code which would ingest this

I saw a fixme in the code somewhere where they were creating this again from available data and it said to replace it with an field in psql so I created this field.

We can then change the code later.

packit_service/service/api/copr_builds.py
TomasTomecek: It can't work, because this expression is evaluated before the filter method gets executed so you're passing True/False to the filter function

packit_service/models.py
TomasTomecek: > > I don't see any code which would ingest this
> 
> I saw a fixme in the code somewhere where they were creating this again from available data and it said to replace it with an field in psql so I created this field.
> 
> We can then change the code later.

right, oki

TomasTomecek:
tests should now be passing once again, please rebase

packit_service/service/api/copr_builds.py
IceWreck: Oh. Then it probably worked for me because I only had three builds (same build_id, different targets)  
And then it combined all of them.

IceWreck:
@TomasTomecek I updated the tests for the new columns in postgres. 

Should I also do something about the new CoprBuild classmethods I created instead of using `get_sa_session` ?
They only make queries to fetch data like the preexisting `get_by_id` and `get_by_build_id` methods, which do not have any tests either.

IceWreck:
> Do you think that we need tests for some new code? I'll leave the answer to you :)

They only return queries so I guess we can leave them for now

IceWreck:
About this:

> git web url is not directly related to a copr build, can we move that field to the GitProject model?

This was already there before I made changes so if I change stuff then previous data in the stg/prod database will be deleted.

Maybe I can leave it there and create another one for new data but then the old data won't be retrievable.

TomasTomecek:
> About this:
> 
> > git web url is not directly related to a copr build, can we move that field to the GitProject model?
> 
> This was already there before I made changes so if I change stuff then previous data in the stg/prod database will be deleted.
> 
> Maybe I can leave it there and create another one for new data but then the old data won't be retrievable.

I'd say that we don't have such a change in stg/prod, it was done locally only on your machine in the previous iteration, so we should be good with the way it's done right now in this PR.


> > Do you think that we need tests for some new code? I'll leave the answer to you :)
> 
> They only return queries so I guess we can leave them for now

...and I thought I taught you something already :(

Yes, please add new tests for the two new methods.

IceWreck:
> > About this:
> > > git web url is not directly related to a copr build, can we move that field to the GitProject model?
> > 
> > 
> > This was already there before I made changes so if I change stuff then previous data in the stg/prod database will be deleted.
> > Maybe I can leave it there and create another one for new data but then the old data won't be retrievable.
> 
> I'd say that we don't have such a change in stg/prod, it was done locally only on your machine in the previous iteration, so we should be good with the way it's done right now in this PR.
> 

I think we are talking about different things here. You told me to move the git web url thing I placed in CoprBuilds and I changed the https_url location.
Then I misunderstood and confused it with web_url  which has been here since the beginning of this file https://github.com/packit-service/packit-service/blame/master/packit_service/models.py#L164

> ...and I thought I taught you something already :(

Yes, please add new tests for the two new methods.

Sheesh! Sorry, will do.

TomasTomecek:
> I think we are talking about different things here. You told me to move the git web url thing I placed in CoprBuilds and I changed the https_url location.
> Then I misunderstood and confused it with web_url which has been here since the beginning of this file https://github.com/packit-service/packit-service/blame/master/packit_service/models.py#L164

Oh you bet, I have no idea what we're talking about here.

I think the current code is good:
```
op.add_column("git_projects", sa.Column("https_url", sa.String(), nullable=True))
```
Yes, that's what we want, to have the url in git_projects.

CoprBuild has web_url with a link to a build web UI. That change is live now.

So I think we're good here. Please let me know if there is anything I'm missing here.

IceWreck:
@TomasTomecek 
Added tests for those two methods and added another for an old method which didn't have tests.
Also did the changes you requested earlier today.

PR #487, opened 2020-03-12
Permission problem is not sentry issue

lachmanfrantisek:
- Use successful result for permission problems. (= We don't want to create an issue in sentry for those events.)


lachmanfrantisek:
@lbarcziova thanks

PR #490, opened 2020-03-13
periodically check the status of a copr build

### LINKED ISSUE #334: Don't rely on fedora-messaging to give us results when builds are done
Once we submit a build, we have a new celery task, which would check the build periodically so that we wouldn't need to count on f-m. And then, either f-m or celery would update the commit status and our database.

_Originally posted by @TomasTomecek in https://github.com/packit-service/packit-service/issues/25#issuecomment-573585712_

TomasTomecek:
TODO:
* [x] write tests
* [x] merge https://github.com/packit-service/packit-service/pull/497

Fixes #334


packit_service/worker/tasks.py
```diff
@@ -40,3 +49,43 @@
 @celery_app.task(name="task.steve_jobs.process_message")
 def process_message(event: dict, topic: str = None) -> Optional[dict]:
     return SteveJobs().process_message(event=event, topic=topic)
+
+
+@celery_app.task(
+    bind=True,
+    name="task.babysit_copr_build",
+    # 120 minutes
+    default_retry_delay=120 * 60,
```
lachmanfrantisek: This is the time when we start this baby-sit task for the first time? (Any then retry each minute?) (i.e. 120, 1, 1, 1, ...)


I would try some exponential sequence as Dominika suggested. (e.g. 1, 2, 4, 8, 16, ... ) or, a constant + exponential sequence (e.g. 10, 1, 2, 4, 8, ... )

[examples are in minutes and mean time to the next check]

packit_service/worker/build/copr_build.py
```diff
@@ -177,6 +178,13 @@ def run_copr_build(self) -> HandlerResults:
         self.copr_build_model.build_id = build_metadata.copr_build_id
         self.copr_build_model.save()
 
+        # release the hounds!
+        celery_app.send_task(
+            "task.babysit_copr_build",
+            args=(build_metadata.copr_build_id,),
+            countdown=60,
+        )
```
lachmanfrantisek: So we have one-baby-sit task for all chroots?

packit_service/models.py
```diff
@@ -204,7 +204,7 @@ def get_all(cls) -> Optional[Iterable["CoprBuild"]]:
 
     # Returns all builds with that build_id, irrespective of target
     @classmethod
-    def get_all_build_id(
+    def get_all_by_build_id(
```
lachmanfrantisek: How it is possible to get multiple builds for one `build_id`?

packit_service/worker/tasks.py
```diff
@@ -40,3 +49,43 @@
...
+
+        # copr doesn't tell status of how a build in the chroot went:
+        #   https://bugzilla.redhat.com/show_bug.cgi?id=1813227
+        for build in builds:
+            event = CoprBuildEvent(
+                topic=str(FedmsgTopic.copr_build_finished),
+                build_id=build_id,
+                build={},
+                chroot=build.target,
+                status=1 if build_copr.state == COPR_SUCC_STATE else 2,
```
lachmanfrantisek: Some constants for the `1` and `2`?

TomasTomecek: this is coming from copr client, I can do constants

packit_service/models.py
TomasTomecek: because in copr, one build_id is built in different chroots (terrible schema)
we have a line in DB for every chroot (better schema)

packit_service/worker/build/copr_build.py
TomasTomecek: yes, that's how copr werks

packit_service/worker/tasks.py
TomasTomecek: I'm not too sold on the exponential backoff

packit_service/models.py
lachmanfrantisek: Thanks for the clarification.

lachmanfrantisek: Thanks for the clarification.

TomasTomecek:
two requre tests broke

lachmanfrantisek:
> two requre tests broke

Let's check on other PRs if it fails everywhere.

It may be caused by an API change in pygithub or Github. 

```
github.GithubException.GithubException: 401 {"message":"'Issuer' claim ('iss') must be an Integer","documentation_url":"https://developer.github.com/v3"}
```

sakalosj:
LGTM :+1:

TomasTomecek:
I'm definitely cursed - "simple" PR turned into... touch and change everything

Makefile
```diff
@@ -38,6 +38,33 @@ check_in_container: test_image
 		-v $(CURDIR)/files/packit-service.yaml:/root/.config/packit-service.yaml \
 		$(TEST_IMAGE) make check
 
+# This is my target so don't touch it! :) How to:
```
lachmanfrantisek: ?

files/deployment.yaml
```diff
@@ -29,52 +29,71 @@
       command: oc whoami -t
       register: kubeconfig_token
       become: "{{ as_root }}"
+
+    # path_to_secrets = this is what we get from the outside
+    # internal_secrets_path = this is what we'll use here
+    - set_fact: secrets_path={{ deployment_dir }}/secrets/dev
```
lachmanfrantisek: Can you please send the deployment changes in another PR? (This started to be really unreviewable.)

packit_service/service/events.py
```diff
@@ -188,7 +188,9 @@ def __init__(self, trigger: TheJobTriggerType, project_url: str):
         )
 
     def get_project(self) -> GitProject:
-        return ServiceConfig.get_service_config().get_project(url=self.project_url)
+        project = ServiceConfig.get_service_config().get_project(url=self.project_url)
+        logger.debug(f"GitService = {project.service.__dict__}")
```
lachmanfrantisek: This is very dangerous -- print the `str` representation instead so we don't log service secrets.

TomasTomecek: dangerous where? that's literally the point here: print the secrets so we know what's the setup

I mean, I can drop that commit since the issue is resolved - it's just these log lines helped me resolve it

thinking... it's true that we share srpm logs with people and this could potentially be printed in there, I'll drop it

Makefile
TomasTomecek: so... the target for running tests in a container is not good enough for me and I keep changing it all the time when working on p-s: so I decided to start my own target (which everyone can use)

files/deployment.yaml
TomasTomecek: yes

packit_service/service/events.py
lachmanfrantisek: The `str` version contains the first and last character of the tokens to be sure we are using the right tokens...

files/deployment.yaml
lachmanfrantisek: Thanks!

Makefile
lachmanfrantisek: Oh, it understood that as "don't touch it and don't try to use it"..;)

packit_service/worker/tasks.py
```diff
@@ -40,3 +53,56 @@
 @celery_app.task(name="task.steve_jobs.process_message")
 def process_message(event: dict, topic: str = None) -> Optional[dict]:
     return SteveJobs().process_message(event=event, topic=topic)
+
+
+@celery_app.task(
+    bind=True,
+    name="task.babysit_copr_build",
+    # 120 minutes
+    default_retry_delay=120 * 60,
```
csomh: Why is this set to 120 minutes? `countdown` in the `retry()` bellow will override this to be 60 seconds, and when the task is sent, it's also 60 seconds.

Another note: isn't doing this work every minute too often? The cadence is very close to what we would get just listening to the message bus. I would go with 5 minutes, but I can imagine even 15 minutes would be fine. Longer times seem to be a better fit for what this functionality would try to achieve: pick up the leftovers.

How many times will this retry? [3](https://docs.celeryproject.org/en/stable/reference/celery.app.task.html#celery.app.task.Task.max_retries)?

csomh: Oh, and how about making all these values part of the configuration, so that we don't have to touch the code, to change them?

TomasTomecek: yeah, this is a good point

so I changed it to use the exponential backoff, current setup:

do the first check in 120s, then next one in 60s, next 120s, 240s, 480s, etc.

would that be okay?

(btw after we merge this, I'm stoked to see how it goes on stg and will fix all the issues - which should be much easier to do and review once we see it in action)

also, I set max retries to 7

thanks for such a thorough review

TomasTomecek:
please don't merge this before fixing https://github.com/packit-service/packit-service/issues/504

we probably need to deploy to prod tomorrow, we can merge this PR after we fix the prod TFT problems

TomasTomecek:
recheck

packit_service/worker/tasks.py
```diff
@@ -40,3 +53,56 @@
...
 def process_message(event: dict, topic: str = None) -> Optional[dict]:
     return SteveJobs().process_message(event=event, topic=topic)
+
+
+@celery_app.task(
+    bind=True,
+    name="task.babysit_copr_build",
+    retry_backoff=60,  # retry again in 60s, 120s, 240s, 480s...
+    retry_backoff_max=60 * 60 * 8,  # is 8 hours okay? gcc/kernel build really long
+    max_retries=7,
```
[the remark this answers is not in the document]
TomasTomecek: I was actually checking the code itself, and they are picking it from these kwargs, so, this* should* work

TomasTomecek:
looking fwd fixing stg tomorrow!

PR #491, opened 2020-03-16
[test] test requre test

lachmanfrantisek:
- Test requre test on master because of the #490


TomasTomecek:
it werks, so my PR broke them

lachmanfrantisek:
> it werks, so my PR broke them

Probably.

PR #493, opened 2020-03-16
Fix error in API when old data exists.

IceWreck:
https://stg.packit.dev/api/copr-builds Is returning internal server error in staging since #479 was deployed.

This is because the database has rows of old data where fields like build_submitted_time are empty. (thanks @csomh )

Fixed that.

### DIFF
--- a/packit_service/service/api/copr_builds.py
+++ b/packit_service/service/api/copr_builds.py
@@ -29,10 +29,21 @@ from flask_restplus import Namespace, Resource
 from packit_service.service.api.parsers import indices, pagination_arguments
 from packit_service.models import CoprBuild
 
+from typing import Union
+
+
 logger = getLogger("packit_service")
 
 
 ns = Namespace("copr-builds", description="COPR builds")
+
+
+def optional_time(datetime_object) -> Union[str, None]:
+    """Returns a string if argument is a datetime object."""
+    if datetime_object is None:
+        return None
+    else:
+        return datetime_object.strftime("%d/%m/%Y %H:%M:%S")
 
 
 @ns.route("")
@@ -57,9 +68,7 @@ class CoprBuildsList(Resource):
                     "build_id": build.build_id,
                     "status": build.status,
                     "chroots": [],
-                    "build_submitted_time": build.build_submitted_time.strftime(
-                        "%d/%m/%Y %H:%M:%S"
-                    ),
+                    "build_submitted_time": optional_time(build.build_submitted_time),
                     "repo_namespace": build.pr.project.namespace,
                     "web_url": build.web_url,
                 }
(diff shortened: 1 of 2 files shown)


packit_service/service/api/copr_builds.py
```diff
@@ -35,6 +35,13 @@
 ns = Namespace("copr-builds", description="COPR builds")
 
 
+def if_time_exists(datetime_object):
+    if datetime_object:
```
csomh: Although this is completely functional, just for the sake of being explicit, you could write `if datetime_object is not None`.

packit_service/service/api/copr_builds.py
```diff
@@ -35,6 +35,13 @@
...
+        return datetime_object.strftime("%d/%m/%Y %H:%M:%S")
+    else:
+        return ""
```
csomh: Would returning `None` here work?

`None` is a very good value to have, when you want to express a missing value. Much better then an empty string.

packit_service/service/api/copr_builds.py
```diff
@@ -57,9 +64,7 @@ def get(self):
                     "build_id": build.build_id,
                     "status": build.status,
                     "chroots": [],
-                    "build_submitted_time": build.build_submitted_time.strftime(
-                        "%d/%m/%Y %H:%M:%S"
-                    ),
+                    "build_submitted_time": if_time_exists(build.build_submitted_time),
```
csomh: Instead of `if_time_exists` you could call the function `optional_time`, so that this line reads: "build_submitted_time equals optional_time".

(This might be personal preference, but just by reading `if_time_exists` it's hard for me to guess what this function could do).

packit_service/service/api/copr_builds.py
```diff
@@ -35,6 +35,13 @@
 ns = Namespace("copr-builds", description="COPR builds")
 
 
+def if_time_exists(datetime_object):
```
csomh: Would be nice to cover this function with some unit tests, and add type annotation.

packit_service/service/api/copr_builds.py
IceWreck: Yes, will do.

packit_service/service/api/copr_builds.py
IceWreck: Sure!

IceWreck:
@csomh Done!

tests/unit/test_api.py
```diff
@@ -0,0 +1,11 @@
+from packit_service.service.api.copr_builds import optional_time
+from datetime import datetime
+
+
+def test_optional_time():
+    time_object = datetime.utcnow()
+    nonetype_object = None
```
csomh: Let's try to use [pytest.mark.parametrize](https://docs.pytest.org/en/latest/parametrize.html) to prepare these input values and also the expect values bellow. With that there would be 2 separate tests executed during the test run, with 2 different inputs and expectations.

This would be better, b/c you want to have your tests as atomic as possible, so that they provide you with as much information as possible, even when they fail.

Right now if the first assert fails, the test run will never get to the second assert, so you would first need to fix the first issue, before being able to tell, if the the other one works. In many cases this will slow you down.

packit_service/service/api/copr_builds.py
```diff
@@ -29,12 +29,23 @@
...
+from typing import Union
+
+
 logger = getLogger("packit_service")
 
 
 ns = Namespace("copr-builds", description="COPR builds")
 
 
+# Returns a string if argument is a datetime object
```
csomh: Please make this a docstring instead of a comment.

IceWreck:
@csomh Did the requested changes.

PR #497, opened 2020-03-17
deployment: process path_to_secrets actually + random commits

TomasTomecek:
* check-inside-os: correct playbook name
* make: new target for local in-container testing
* make: drop unused target check-within-os

I can rename, change or merge the new target - don't really care

This PR does:
* actually provides our secrets/dev/ during local requre regeneration
* creates new make target to run requre tests w/o openshift (!!!)

### DIFF
--- a/files/deployment.yaml
+++ b/files/deployment.yaml
@@ -29,52 +29,71 @@
       command: oc whoami -t
       register: kubeconfig_token
       become: "{{ as_root }}"
-    - name: Creates zuul secrets directory
+
+    # path_to_secrets = this is what we get from the outside
+    # internal_secrets_path = this is what we'll use here
+    - set_fact: internal_secrets_path={{ deployment_dir }}/secrets/dev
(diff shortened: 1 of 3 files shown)


files/deployment.yaml
```diff
@@ -29,52 +29,71 @@
       command: oc whoami -t
       register: kubeconfig_token
       become: "{{ as_root }}"
+
+    # path_to_secrets = this is what we get from the outside
+    # internal_secrets_path = this is what we'll use here
```
jpopelka: pardon my ignorance but where here? (I don't see it anywhere)

TomasTomecek: /facepalm

TomasTomecek: corrected it now and even improved the logic around dirs

TomasTomecek:
recheck
some race condition or what, precommit is passing for me locally

TomasTomecek:
recheck

PR #501, opened 2020-03-17
Move Whitelist from Redis to Postgres

IceWreck:
None of the redis code has been touched. I've added the postgres code on top of it so everythings done twice, once for either database.

Manual approval and deletion via the script works. 

I've created methods for automatic approval and added comments/TODOs for it as well but someone else will have to do the final part because I have no way of testing if someone is in the fedora packager's group.

The unit tests for the functions in worker/whitelist.py are failing because I've added sqlalchemy related code to them and _make check_in_container_ does not support postgres code.
@TomasTomecek  How do I fix this? Just move that specific tests file over to the tests_requre folder alongside tests_db.py?
Edit: Done


packit_service/worker/whitelist.py
```diff
@@ -146,6 +161,22 @@ def is_approved(self, account_name: str) -> bool:
...
+            return True
+
+            # Can also do the following like in redis but seems pointless in this case ??
+
+            # db_status = DBWhitelist.get_account(account_name).status
+            # s = WhitelistStatus(db_status)
+            # return (
+            #     s == WhitelistStatus.approved_automatically
+            #     or s == WhitelistStatus.approved_manually
+            # )
```
TomasTomecek: yeah, no need to

packit_service/models.py
```diff
@@ -299,6 +299,52 @@ def __str__(self):
         return self.__repr__()
 
 
+class Whitelist(Base):
+    __tablename__ = "whitelist"
+    id = Column(Integer, primary_key=True)
+    account_name = Column(String, index=True)
+    status = Column(String, index=True)
```
TomasTomecek: would be nice to do statuses as enum like @lachmanfrantisek is doing it in #502

TomasTomecek:
> The unit tests for the functions in worker/whitelist.py are failing because I've added sqlalchemy related code to them and _make check_in_container_ does not support postgres code.
> @TomasTomecek How do I fix this? Just move that specific tests file over to the tests_requre folder alongside tests_db.py?

yeah, either move them or flexmock the db calls

>     * blacklist

no need to, we havent needed blacklist so far

>     * whitelist part of the API

I'd say that whitelist is our internal information and shouldn't be in the api

packit_service/worker/whitelist.py
IceWreck: I actually forgot about accounts that will exist in db, but will have waiting status. So I guess this wasn't pointless after all

IceWreck:
@TomasTomecek 
I've made the changes.

tests_requre/test_db.py
```diff
@@ -245,3 +272,30 @@ def test_get_by_build_id(multiple_copr_builds):
...
+    assert new_whitelist_entry.account_name == "Rayquaza"
+
+
+def test_get_account(multiple_whitelist_entries):
+    assert Whitelist.get_account("Rayquaza").status == "approved_manually"
+    assert Whitelist.get_account("Rayquaza").account_name == "Rayquaza"
+    assert Whitelist.get_account("Deoxys").status == "waiting"
+    assert Whitelist.get_account("Deoxys").account_name == "Deoxys"
+    assert Whitelist.get_account("Solgaleo").status == "waiting"
+    assert Whitelist.get_account("Solgaleo").account_name == "Solgaleo"
```
TomasTomecek: finally we are getting some more interesting names in the tests! thank you

tests_requre/test_db.py
```diff
@@ -129,6 +131,31 @@ def multiple_copr_builds():
...
+
+# Create new whitelist entry
+@pytest.fixture()
+def new_whitelist_entry():
+    with get_sa_session() as session:
+        session.query(Whitelist).delete()
+        yield Whitelist.add_account(account_name="Rayquaza", status="approved_manually")
+    clean_db()
+
+
```
TomasTomecek: awesome, tests!!

TomasTomecek:
@IceWreck please run pre-commit on the changes: it seems that requre wants to sanitize some response files

tests_requre/test_db.py
IceWreck: Thanks!

IceWreck:
recheck

IceWreck:
recheck

PR #510, opened 2020-03-18
Fix event argument for handlers

lachmanfrantisek:
- Use name 'event' for all handlers to match the call in jobs.
- Fixes RED-HAT-0P-2BK


lachmanfrantisek:
recheck

lachmanfrantisek:
recheck

PR #515, opened 2020-03-23
CentOS integration initial work

sakalosj:


Makefile
```diff
@@ -18,9 +18,15 @@ worker: files/install-deps-worker.yaml files/recipe-worker.yaml
 # Make sure you have latest docker.io/usercont/packit:prod prior to running this
 worker-prod: files/install-deps-worker.yaml files/recipe-worker.yaml
 	$(CONTAINER_ENGINE) build --rm -t $(WORKER_IMAGE_PROD) -f Dockerfile.worker.prod .
+
 worker-prod-push: worker-prod
 	$(CONTAINER_ENGINE) push $(WORKER_IMAGE_PROD)
 
+# in case of need change ip this is default so whould work, also be sure docker is logged to local oc registry 'docker login 172.30.1.1:5000'
+worker-local-push: worker
+	$(CONTAINER_ENGINE) tag $(WORKER_IMAGE) 172.30.1.1:5000/packit-stg/packit-worker:dev && \
```
TomasTomecek:
```suggestion
	$(CONTAINER_ENGINE) tag $(WORKER_IMAGE) 172.30.1.1:5000/packit-stg/packit-worker:dev
```
no need to, make will take care of the serialization

packit_service/worker/centos/events.py
```diff
@@ -0,0 +1,157 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+# SOFTWARE.
+
+"""
+This file defines classes for events which are sent by GitHub or FedMsg.
```
TomasTomecek, line 24:
```suggestion
This file defines classes for events which are coming from CentOS Pagure - "git.centos.org".
```

packit_service/worker/centos/parser.py
```diff
@@ -0,0 +1,140 @@
...
+    PullRequestCommentAction,
+    PushPagureEvent,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class CentosEventParser:
+    """
+    Class responsible for parsing received from CentOS stream
```
TomasTomecek, line 17:
```suggestion
    Class responsible for parsing events received from CentOS infrastructure
```

packit_service/worker/centos/parser.py
```diff
@@ -0,0 +1,140 @@
...
+        try:
+            event_object = self.event_mapping[git_topic](event)
+        except KeyError:
+            logger.info(f"Event type {git_topic} is not processed")
+            return None
+
+        return event_object
+
+    @staticmethod
+    def _pull_request_event(event, action):
```
TomasTomecek, line 58:
```suggestion
    def _pull_request_event(event: dict, action: str):
```

packit_service/worker/centos/parser.py
```diff
@@ -0,0 +1,140 @@
...
+            base_repo_name,
+            base_ref,
+            target_repo,
+            https_url,
+            commit_sha,
+            pagure_login,
+        )
+
+    def _pull_request_comment(
+        self, event: dict, action
```
TomasTomecek, line 89:
```suggestion
        self, event: dict, action: str
```

packit_service/worker/handlers/centosmsg_handlers.py
```diff
@@ -0,0 +1,156 @@
...
+        self,
+        config: ServiceConfig,
+        job_config: JobConfig,
+        event: PullRequestPagureEvent,
+    ):
+        super().__init__(config=config, job_config=job_config, event=event)
+
+    def run(self) -> HandlerResults:
+        collaborators = self.project.who_can_merge_pr()
+        if self.event.user_login not in collaborators | self.config.admins:
```
TomasTomecek, line 105: I'd say that we don't need to check for this - we can build everything

packit_service/worker/handlers/github_handlers.py
```diff
@@ -213,12 +213,14 @@ def __init__(
...
-                "are accepted."
-            )
+        # Should be not required as this test is done in precheck implementation
+        #
+        # if not isinstance(event, (PullRequestEvent, PushGitHubEvent, ReleaseEvent)):
+        #     raise PackitException(
+        #         "Unknown event, only "
+        #         "PullRequestEvent, ReleaseEvent, and PushGitHubEvent "
+        #         "are accepted."
+        #     )
```
TomasTomecek, line 223: feel free to remove the code then

packit_service/worker/reporting.py
```diff
@@ -65,6 +66,10 @@ def report(
     def set_status(
         self, state: CommitStatus, description: str, check_name: str, url: str = "",
     ):
+        # required because pagure api doesnt accept, empty url
+        if isinstance(self.project, PagureProject):
+            url = "http://no.url"
```
TomasTomecek, line 71: hm? we should find a better default: e.g. "https://wiki.centos.org/Manuals/ReleaseNotes/CentOSStream"

lachmanfrantisek: Or just empty one -- we already use empty string for GitHub.

sakalosj: it will be ideal, but pagure api doesnt support empty string and is checking if is valid string

TomasTomecek: Franta, read the comment at line 69 :P

sakalosj: fixed

packit_service/worker/handlers/github_handlers.py
sakalosj: fixed

packit_service/worker/handlers/centosmsg_handlers.py
sakalosj: fixed

packit_service/worker/centos/parser.py
sakalosj: fixed

packit_service/worker/centos/parser.py
sakalosj: fixed

packit_service/worker/centos/parser.py
sakalosj: fixed

packit_service/worker/centos/events.py
sakalosj: fixed

packit_service/service/events.py
```diff
@@ -635,3 +641,119 @@ def get_copr_build_logs_url(event: CoprBuildEvent) -> str:
         f"{event.project_name}/{event.chroot}/"
         f"{event.build_id:08d}-{event.pkg}/builder-live.log.gz"
     )
+
+
+class AbstractPagureEvent(Event, PagurePackageConfigGetter):
```
csomh, line 646: I cannot find anything "abstract" in this class? What does the name try to suggest?

packit_service/worker/handlers/centosmsg_handlers.py
```diff
@@ -0,0 +1,144 @@
...
+            logger.info(
+                "Skipping build for testing. The COPR build is defined in the config."
+            )
+            return False
+        return True
+
+
+@use_for(job_type=JobType.copr_build)
+@use_for(job_type=JobType.build)
+class PagurePullRequestCoprBuildHandler(AbstractPagureCoprBuildHandler):
```
csomh, line 89: Is this class intended to be responsible for starting builds when a pull-request is filed/updated in CentOS infra? If yes, could we call it simply `PullRequestHandler`? This name is too long, and it has too many details that can be interfered from the context or is irrelevant at this level of abstraction.

packit_service/service/events.py
```diff
@@ -635,3 +641,119 @@ def get_copr_build_logs_url(event: CoprBuildEvent) -> str:
...
+        self.identifier = str(pr_id)
+        self.git_ref = None  # pr_id will be used for checkout
+
+    def get_dict(self, default_dict: Optional[Dict] = None) -> dict:
+        result = super().get_dict()
+        result["action"] = result["action"].value
+        return result
+
+
+class PullRequestPagureEvent(AddPullRequestDbTrigger, AbstractPagureEvent):
```
csomh, line 720: A diff with `PullRequestEvent` shows very little difference. Why is the reason behind this code duplication?

This is also a question for the other `*Event` classes added.

```diff
@@ -1,4 +1,4 @@
-class PullRequestEvent(AddPullRequestDbTrigger, AbstractGithubEvent):
+class PullRequestPagureEvent(AddPullRequestDbTrigger, AbstractPagureEvent):
     def __init__(
         self,
         action: PullRequestAction,
@@ -9,7 +9,7 @@ class PullRequestEvent(AddPullRequestDbTrigger, AbstractGithubEvent):
         target_repo: str,
         https_url: str,
         commit_sha: str,
-        github_login: str,
+        user_login: str,
     ):
         super().__init__(trigger=TheJobTriggerType.pull_request, project_url=https_url)
         self.action = action
@@ -19,7 +19,7 @@ class PullRequestEvent(AddPullRequestDbTrigger, AbstractGithubEvent):
         self.base_ref = base_ref
         self.target_repo = target_repo
         self.commit_sha = commit_sha
-        self.github_login = github_login
+        self.user_login = user_login
         self.identifier = str(pr_id)
         self.git_ref = None  # pr_id will be used for checkout

@@ -31,8 +31,7 @@ class PullRequestEvent(AddPullRequestDbTrigger, AbstractGithubEvent):
     def get_package_config(self) -> Optional[PackageConfig]:
         package_config: PackageConfig = self.get_package_config_from_repo(
             project=self.get_project(),
-            reference=self.base_ref,
-            pr_id=self.pr_id,
+            reference=self.commit_sha,
             fail_when_missing=False,
         )
         if not package_config:
```

packit_service/config.py
```diff
@@ -188,3 +190,24 @@ def get_package_config_from_repo(
                     logger.debug(f"Created issue for invalid packit config: {i.url}")
             raise ex
         return package_config
+
+
+class PagurePackageConfigGetter:
```
csomh, line 195: What is the purpose if this class.

I both see `packit.yaml` and `*.spec` here? How do these 2 come together? Is spec-file the same as a config (packit.yaml) file?

packit_service/worker/parser.py
```diff
@@ -505,3 +509,131 @@ def parse_copr_event(event) -> Optional[CoprBuildEvent]:
...
+            comment,
+        )
+
+    def _push_event(self, event: dict) -> PushPagureEvent:
+        logger.debug("Parsing git.receive (git push) event.")
+
+        return PushPagureEvent(
+            repo_namespace=event["repo"]["namespace"],
+            repo_name=event["repo"]["name"],
+            # pagure dont return git_ref, how to handle this?
```
TomasTomecek, line 634: branch name is good enough

packit_service/worker/reporting.py
lachmanfrantisek: OK, I understand empty as `None`. Sorry

packit_service/worker/parser.py
sakalosj: ok comment removed

packit_service/config.py
sakalosj: Please in checkpackit/config/package_config.py check get_package_config_from_repo() for implmentation this one is simplified version because of limited Pagure api capabilities.

Because of it is github api dependent, is imposible to use it for Pagure. Didnt want to spread "mess" in other repos becuase:
1. we need something what is working
2. there other open question for arch meeting which can possible affect how to implement this in clean way

packit_service/service/events.py
sakalosj: Those differences are crucial and it is most natural and also clean way to implement it in with current design.
I agree that there is unwanted duplication, but there will be required redesign to avoid it and it is out of  scope of this task

packit_service/worker/handlers/centosmsg_handlers.py
sakalosj: I have followed current architecture. 

There are different implementation how we are getting configs for GitHub and Pagure because of different api capabilities. 
Event.get_package_config() and AbstractGithubJobHandler/AbstractPagureJobHandler (via GithubPackageConfigGetter/PagurePackageConfigGetteris) GitService type dependent, thus preventing of using more general class. I think this breaking of SRP, i will raise arch day topic for this.

packit_service/service/events.py
sakalosj: this class is abstract - it should not be used create instances, should be used only for inheritance.

Also I followed existing architecture approach, so in case you have objection please add topic to planned arch day.

packit_service/service/events.py
```diff
@@ -635,3 +641,119 @@ def get_copr_build_logs_url(event: CoprBuildEvent) -> str:
...
+        self.git_ref: Optional[str] = None  # git ref that can be 'git checkout'-ed
+        self.identifier: Optional[str] = (
+            None  # will be shown to users -- e.g. in logs or in the copr-project name
+        )
+
+    def get_project(self) -> GitProject:
+        return ServiceConfig.get_service_config().get_project(url=self.project_url)
+
+
+class PushPagureEvent(AbstractPagureEvent):
```
lachmanfrantisek, line 659: Those event classes look very similar to Github ones:
- It would be nice to merge them.
- We can make a subclass just for the differentiation without any code change if we really want to differentiate that.
  -  Since we have the `https_url`, we can still be able to get service easily.

packit_service/service/events.py
lachmanfrantisek: This discussion is similar to my previous comment -- if it is easy to do it here, I would merge them. If it requires big changes in the other code, let's leave it as is for now.

packit_service/worker/jobs.py
```diff
@@ -179,8 +181,13 @@ def process_jobs(self, event: Event) -> Dict[str, HandlerResults]:
             # failed because of missing whitelist approval
             whitelist = Whitelist()
             github_login = getattr(event, "github_login", None)
```
lachmanfrantisek, line 183: Can we rename the `github_login` to `user_login` as well to have the same API?

sakalosj:
@csomh 

* we agreed with @lachmanfrantisek, (maybe also @TomasTomecek  was there, not sure) that we will go using old way (current architecture), also we agreed that any findings and suggestions will be raised in architecture document and discussed inarchitecture
* limitation of current architecture is not argument is fact
* feel free to suggest how it should be done

TomasTomecek:
@csomh thanks for the thorough and honest write-up

I agree this PR adds technical debt to the codebase. The question here is:
1. Do we want to address it in this PR?
2. Do we want to merge and create a follow-up PR with the tech debt reduced and more tests added?

Nevertheless, someone needs to think and come up with a solution to refactor the codebase so things are more clear - who's gonna do that? when and how?

My main worry is that if we keep polishing this PR, it will take weeks to merge and we'll get "behind the schedule" also adding frustration to Jano since he'd need to rebase and keep on iterating. Doing subsequent PRs with smaller changes should be easier to review and we should see the benefits immediately, rather than reviewing `+1000 -1000`.

Hence, my preference would be to:
1. Merge it and NOT deploy it in the current window (for sake of safety)
2. Think about the refactoring and come up with a concrete plan (this to be done Monday/Tuesday)
3. Execute, merge, deploy to stg
4. Test it on stg and hit next prod deploy window with things already polished
5. Party :partying_face:

packit_service/service/events.py
sakalosj: There is different ConfigGetter class implementation and this should be object of refactoring - this shouldn't be part of Event (its breaking SRP and creating tight coupling). This needs discussion and also i think it is out of scope of this PR.

packit_service/worker/jobs.py
sakalosj: fixed

packit_service/service/events.py
sakalosj: reply to previous comment applies also here:
There is different ConfigGetter class implementation and this should be object of refactoring - this shouldn't be part of Event (its breaking SRP and creating tight coupling). This needs discussion and also i think it is out of scope of this PR.
This mentioned above this should be discussed

lachmanfrantisek: Since you now have `user_login` also in GitHub events, what about moving the common code to the new superclass?

These classes can inherit from the superclass and config getter class. We can definitely refactor later, but this deduplication should be really easy.

PR #520, opened 2020-03-25
Save task results in psql instead of redis

lbarcziova:
- Celery saves the task result data in psql instead of redis
- We create own table with the results (without Celery information, e.g.traceback) - API uses this table


packit_service/celerizer.py
```diff
@@ -41,9 +42,10 @@ def celery_app(self):
             redis_url = "redis://{host}:{port}/{db}".format(
                 host=redis_host, port=redis_port, db=redis_db
             )
+            postgres_url = f"db+{get_pg_url()}"
 
             # http://docs.celeryproject.org/en/latest/reference/celery.html#celery.Celery
-            self._celery_app = Celery(backend=redis_url, broker=redis_url)
+            self._celery_app = Celery(backend=postgres_url, broker=redis_url)
```
TomasTomecek: wow, it's really that simple? :O no need to change backend type?

packit_service/service/api/tasks.py
```diff
@@ -67,6 +68,14 @@ def get(self):
                     data["result"]["event"] = Event.ts2str(event)
                 tasks.append(data)
 
+        for task in TaskResult.get_all():
+            data = task.to_dict()
```
TomasTomecek: wow, I didn't know that SA has such method, neat!

packit_service/models.py
```diff
@@ -356,6 +357,45 @@ def __str__(self):
...
+                session.add(task_result)
+                return task_result
+            else:
+                task_result = cls(task_id)
+                task_result.status = status
+                task_result.result = result
+                task_result.traceback = traceback
+                task_result.date_done = date_done
+                session.add(task_result)
+                return task_result
```
lachmanfrantisek: I know it's still WIP, but small improvement:

```suggestion
            if task_result is None:
                task_result = cls(task_id)
            task_result.status = status
            task_result.result = result
            task_result.traceback = traceback
            task_result.date_done = date_done
            session.add(task_result)
            return task_result
```

packit_service/service/api/tasks.py
lbarcziova: actually I think it is a method from celery task model

packit_service/celerizer.py
```diff
@@ -41,9 +42,10 @@ def celery_app(self):
             redis_url = "redis://{host}:{port}/{db}".format(
                 host=redis_host, port=redis_port, db=redis_db
             )
+            postgres_url = f"db+{get_pg_url()}"
```
TomasTomecek: could you please add a comment explaining what the db+ prefix stands for?

packit_service/models.py
```diff
@@ -356,6 +353,38 @@ def __str__(self):
         return self.__repr__()
 
 
+class TaskResult(Task):
```
TomasTomecek: we should inherit from Base actually and define the fields ourselves

I recall that @jpopelka did not like this in redis that we would store the data twice, but on the other hand: one copy is handled by celery and the other one by us to use - alternatively we could write code which would retrieve data from the celery tables, not sure if we want that

my preferred solution is to have two copies and full control of the data, rather then rely on celery

jpopelka: Sorry guys, I have no idea why we need this and have no connection to even read the docs so feel free to do whatever you like. If it's just about having some redundant data, well, ok, disk space is cheaper than memory (redis), so, go ahead.

TomasTomecek: Oki, we can fight about this on the next architecture meeting :))

packit_service/celerizer.py
lbarcziova: added

PR #521, opened 2020-03-26
Minor API Bugfixes

IceWreck:
- Fixed the incorrect git repo link in the API
- Newer builds are shown first, so that we dont have to go to the last page to view recent builds.

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -30,7 +30,7 @@ from contextlib import contextmanager
 from datetime import datetime
 from typing import TYPE_CHECKING, Optional, Union, Iterable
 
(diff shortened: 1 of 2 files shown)


packit_service/models.py
```diff
@@ -202,7 +202,7 @@ def get_by_id(cls, id_: int) -> Optional["CoprBuild"]:
     @classmethod
     def get_all(cls) -> Optional[Iterable["CoprBuild"]]:
         with get_sa_session() as session:
-            return session.query(CoprBuild).all()
+            return session.query(CoprBuild).order_by(desc(CoprBuild.id)).all()
```
TomasTomecek: ordering by ID is yucky, but it's better than what we have now and since not all the builds have datetime attribs in prod, I'd say this is good enough

Issue #523, opened 2020-03-27
Copr URLs are wrong for groups


lachmanfrantisek:
The copr URLs for groups (starting with `@`) are different.

> @sturivny, @lachmanfrantisek, @TomasTomecek, building in the custom copr works. That's great!
> 
> The log linked by packit for the RPM build however points to an invalid copr build URL.
>
> The log: https://prod.packit.dev/copr-build/3500/logs
> The invalid link in the log: https://copr.fedorainfracloud.org/coprs/@oamg/convert2rhel/build/1321726/
> The link that works: https://copr.fedorainfracloud.org/coprs/g/oamg/convert2rhel/build/1321726/

_Originally posted by @bocekm in https://github.com/oamg/convert2rhel/pull/27#issuecomment-604972368_

> @sturivny, @lachmanfrantisek, @TomasTomecek
Also, the testing farm is trying to install the convert2rhel build from an invalid copr repo URL.
> Testing farm log: https://console-testing-farm.apps.ci.centos.org/pipeline/1bb7bded-36e2-4b9b-b34f-47333aaf418e
> It tries to install the pkg from: https://copr.fedorainfracloud.org/coprs/@oamg/oamg-convert2rhel-27/repo/epel-6/oamg-convert2rhel-27-epel-6.repo
It should be installing it from: https://copr.fedorainfracloud.org/coprs/g/oamg/convert2rhel/repo/epel-6/group_oamg-convert2rhel-epel-6.repo

_Originally posted by @bocekm in https://github.com/oamg/convert2rhel/pull/27#issuecomment-604974158_

csomh:
The PRs above should have fixed this issue (and I should have closed it long ago), but still copr build URLs are invalid in the logs, though the repo URLs seem to be fine ([logs](https://console-testing-farm.apps.ci.centos.org/pipeline/859d1ad0-5615-4af1-a34e-07c7555d39ad))

```
[18:05:15] [+] [cruncher] [/plans/smoke] [prepare] Installing builds from copr
# cd /etc/yum.repos.d && curl -LO https://copr.fedorainfracloud.org/coprs/@oamg/convert2rhel/repo/epel-6/convert2rhel-epel-6.repo
% Total % Received % Xferd Average Speed Time Time Time Current
Dload Upload Total Spent Left Speed
127 382 127 382 0 0 952 0 --:--:-- --:--:-- --:--:-- 9794
[...]

PR #524, opened 2020-03-30
Use the configured COPR project when triggering Testing Farm

csomh:
Before this, we always referred to the default COPR project when
triggering Testing Farm. This caused tests to fail in repos which choose
to build in their own COPR namespace and project.

Relates to #523.

### DIFF
--- a/packit_service/worker/testing_farm.py
+++ b/packit_service/worker/testing_farm.py
@@ -135,7 +135,7 @@ class TestingFarmJobHelper(CoprBuildJobHelper):
             "artifact": {
                 "repo-name": self.project.repo,
                 "repo-namespace": self.project.namespace,
-                "copr-repo-name": f"{self.job_owner}/{self.default_project_name}",
+                "copr-repo-name": f"{self.job_owner}/{self.job_project}",
                 "copr-chroot": chroot,
                 "commit-sha": self.event.commit_sha,
                 "git-url": self.event.project_url,


csomh:
Okay, so this took a little longer and ended up also making a few tweaks for how to run tests locally, this is why there are 7 (seven) commits.

The first three are related to the subject of this PR.
The following three are a few Makefile tweaks - I can brake this out in a separate PR, if you say so.
The last is to get rid of some pytest warnings - I can brake this out in a separate PR, if you say so.

See individual commit messages for the motivation of these changes.

packit_service/worker/testing_farm.py
```diff
@@ -116,36 +137,8 @@ def run_testing_farm(self, chroot: str) -> HandlerResults:
...
-        logger.debug(f"Pipeline id: {pipeline_id}")
-
-        test_run_model = TFTTestRunModel.create(
-            pipeline_id=pipeline_id,
-            commit_sha=self.event.commit_sha,
-            status=TestingFarmResult.new,
-            target=chroot,
-            web_url=None,
-            trigger_model=self.event.db_trigger,
-        )
```
TomasTomecek, line 129: where did this code go?

csomh: facepalm

tests/unit/test_testing_farm.py
```diff
@@ -20,31 +20,35 @@
...
-    TestingFarmResultsEvent,
-    TestingFarmResult,
-    TestResult,
-)
-from packit_service.worker.handlers import TestingFarmResultsHandler
+
+# These names are definately not nice, still they help with making classes
+# whose names start with Testing* or Test* to become invisible for pytest,
+# and so stop the test discovery warnings.
+from packit_service.service.events import TestingFarmResultsEvent as TFResultsEvent
```
lachmanfrantisek, line 33: Good catch. You can rename the actual class.

TomasTomecek: I'd rename it to TFT :)

csomh: I actually like the current name of the class: it's much easier to read and understand than TF-something.

csomh: Sorry, no, not this time, no more renaming :)

These aliases are local to this test, if you really think renaming these classes is useful then lets do this in some other PR. But as I said above, the current names are perfectly fine from a readability point of view.

csomh:
recheck

lachmanfrantisek:
Probably some flexmocking problem:

```
[1m    def checkout_ref(self, ref: str):[0m
[1m        """ Check out selected ref in the git repo; equiv of git checkout -B ref """[0m
[1m        logger.info(f"Checking out ref {ref}.")[0m
[1m>       if ref not in self.git_repo.branches:[0m
[1m[31mE       AttributeError: 'NoneType' object has no attribute 'branches'[0m
```

csomh:
> Probably some flexmocking problem:
> 

Yeah, I spent the last 2 hours looking into this, but I'm still not able to tell what's the issue, and I'm unable to reproduce this locally. (I've updated/rebuilt all my local images, made sure everything is freshly pulled.)

lachmanfrantisek:
@csomh I'll take a look if you want, it looks familiar to me.

csomh:
> @csomh I'll take a look if you want, it looks familiar to me.

That would be sooooo great! :hugs:

I've updated this to only include the fix commit, and removed all the others. Let's see how the tests are doing. 

The rest of the commits you can pull it from

`https://github.com/csomh/packit-service test-curstom-copr-project`

lachmanfrantisek:
> That would be sooooo great! hugs

lachmanfrantisek:
Sorry, wrong button.

It passes for me locally as well -- I need to find the right version of dependencies to reproduce it.

PR #525, opened 2020-03-30
Add flexmock to the service

lachmanfrantisek:
We need to use flexmock to solve some problem in alembic during the database upgrade.


csomh:
/me wonders what "some problem" was :thinking:

lachmanfrantisek:
> /me wonders what "some problem" was

See the comment in the code:
https://github.com/packit-service/packit-service/blob/1a6a314aee67f6c7e2ef5302e946538c79bd4925/alembic/versions/d90948124e46_add_tables_for_triggers_koji_and_tests.py#L26-L29

(We already discuss that with Tomáš and didn't find any better solution.)

PR #526, opened 2020-03-30
Add script to move data from redis to psql

lbarcziova:


TomasTomecek:
Thinking out loud: what would be a better solution: have it as a standalone script, or have a separate migration for it and utilize alembic?

https://github.com/packit-service/packit-service/pull/502/files#diff-9311d455327ddfdbc8aca25513b84428R275

@lbarcziova @lachmanfrantisek WDYT?

lachmanfrantisek:
If you look at the previous migration (with the trigger data migration), it should be easy to migrate data from redis in the similar way.

alembic/versions/c6250555a36c_.py
```diff
@@ -0,0 +1,370 @@
...
+    DateTime,
+    ForeignKey,
+    Enum,
+    orm,
+)
+from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import Session, relationship
+from sqlalchemy.types import PickleType
+
+from packit_service.service.models import Installation, CoprBuild
```
lachmanfrantisek, line 32: Please, copy also these models to be sure we are using the right version. (Some skeleton should be enough.)

lbarcziova:
@lachmanfrantisek thanks! I will try to test it with the data as you mentioned

lbarcziova:
recheck

alembic/versions/c6250555a36c_.py
```diff
@@ -0,0 +1,370 @@
...
+        if account is not None:
+            account.status = status
+            session.add(account)
+            return account
+        else:
+            account = cls()
+            account.account_name = account_name
+            account.status = status
+            session.add(account)
+            return account
```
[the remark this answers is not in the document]
lbarcziova: Oh, I just copied these lines from the code in models, will change it.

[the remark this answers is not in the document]
lachmanfrantisek: (You can simplify them in the real code as well..)

lachmanfrantisek:
> Also, what's the point of copying the models instead of importing them?

If you do the migration in the future, you will not have the corresponding version of the model -- we need the model that fits this state of the database.

TomasTomecek:
> > Also, what's the point of copying the models instead of importing them?
> 
> If you do the migration in the future, you will not have the corresponding version of the model -- we need the model that fits this state of the database.

true, that's clever

lbarcziova:
Can we merge this and try it on stage?

alembic/versions/c6250555a36c_.py
```diff
@@ -0,0 +1,409 @@
...
+depends_on = None
+
+if TYPE_CHECKING:
+    Base = object
+else:
+    Base = declarative_base()
+
+
+# Redis models
+class Model:
```
lachmanfrantisek, line 48: Can we rename that to `RedisModel`?

(I was a bit confused about the following classes what models are they -- postgres x redis.)

lachmanfrantisek: Maybe also rename the following following classes so the upgrade code can be more readable.

lbarcziova:
@lachmanfrantisek thanks! I will rename the classes and merge

PR #528, opened 2020-03-31
Add a psql model for Github installations

lbarcziova:


packit_service/models.py
```diff
@@ -761,3 +771,48 @@ def get_by_pipeline_id(cls, pipeline_id: str) -> Optional["TFTTestRunModel"]:
                 .filter_by(pipeline_id=pipeline_id)
                 .first()
             )
+
+
+class InstallationModel(Base):
+    __tablename__ = "github_installations"
+    installation_id = Column(Integer, primary_key=True)
```
TomasTomecek, line 778: please don't make this a PK since we don't control the value - it would also be nice if our IDs started from 1 and not getting random 6 digits numbers

packit_service/models.py
```diff
@@ -761,3 +771,48 @@ def get_by_pipeline_id(cls, pipeline_id: str) -> Optional["TFTTestRunModel"]:
...
+    account_login = Column(String)
+    account_id = Column(Integer)
+    account_url = Column(String)
+    account_type = Column(String)
```
TomasTomecek, line 782: these 4 are related to the person who installed the app, right? if so, can you make a comment about this info here please?

packit_service/models.py
```diff
@@ -761,3 +771,48 @@ def get_by_pipeline_id(cls, pipeline_id: str) -> Optional["TFTTestRunModel"]:
...
+    created_at = Column(DateTime, default=datetime.utcnow)
+    repositories = Column(ARRAY(String))
```
TomasTomecek, line 784: It would be awesome to connect here to the GitProject table, but is it an overkill?

packit_service/models.py
```diff
@@ -761,3 +771,48 @@ def get_by_pipeline_id(cls, pipeline_id: str) -> Optional["TFTTestRunModel"]:
...
+    sender_id = Column(Integer)
+    sender_login = Column(String)
```
TomasTomecek, line 786: please document what are these

packit_service/models.py
lbarcziova: I was thinking about this too, I can change it

packit_service/models.py
lbarcziova: added

packit_service/models.py
lbarcziova: added

packit_service/models.py
```diff
@@ -771,3 +781,61 @@ def get_by_pipeline_id(cls, pipeline_id: str) -> Optional["TFTTestRunModel"]:
...
+    account_login = Column(String)
+    account_id = Column(Integer)
+    account_url = Column(String)
+    account_type = Column(String)
+    # information about user who installed the app into 'account'
+    sender_id = Column(Integer)
+    sender_login = Column(String)
+
+    created_at = Column(DateTime, default=datetime.utcnow)
+    repositories = Column(ARRAY(Integer, ForeignKey("git_projects.id")))
```
TomasTomecek, line 799: wow, this looks crazy, doing the relation from this side

But I'd say it makes sense. Just that I never did this myself.

packit_service/models.py
```diff
@@ -771,3 +781,61 @@ def get_by_pipeline_id(cls, pipeline_id: str) -> Optional["TFTTestRunModel"]:
...
+
+    @classmethod
+    def get_project(cls, repo: str):
+        return GitProjectModel.get_or_create(repo.split("/")[0], repo.split("/")[1])
```
TomasTomecek, line 803:
```suggestion
        namespace, repo_name = repo.split("/")
        return GitProjectModel.get_or_create(namespace, repo_name)
```
:)

PR #530, opened 2020-04-01
Support different triggers in copr build view

lachmanfrantisek:
- Support different triggers in the copr-build view.
- Added tests for view that use database.


lachmanfrantisek:
We definitely need to design solid API tests...

PR #533, opened 2020-04-03
Add a unit test checking the payload to trigger Testing Factory

csomh:
This is to finish the work from #524.

### DIFF
--- a/tests/integration/test_listen_to_fedmsg.py
+++ b/tests/integration/test_listen_to_fedmsg.py
@@ -303,7 +303,10 @@ def test_copr_build_end_testing_farm(copr_build_end, copr_build_pr):
             }
         )
     )
-    flexmock(CoprBuildJobHelper).should_receive("job_owner").and_return("some-owner")
+    flexmock(TestingFarmJobHelper).should_receive("job_owner").and_return("some-owner")
+    flexmock(TestingFarmJobHelper).should_receive("job_project").and_return(
+        "foo-bar-123-stg"
+    )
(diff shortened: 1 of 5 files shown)


tests/integration/test_listen_to_fedmsg.py
```diff
@@ -303,7 +303,10 @@ def test_copr_build_end_testing_farm(copr_build_end, copr_build_pr):
             }
         )
     )
-    flexmock(CoprBuildJobHelper).should_receive("job_owner").and_return("some-owner")
+    flexmock(TestingFarmJobHelper).should_receive("job_owner").and_return("some-owner")
```
csomh, line 306: These 3 integration tests got broken by the unit test added for `_trigger_payload()`. This manifested as follows:
- running the unit tests first (with the new one included) and then the integration tests made these 3 tests to fail;
- running the integration tests before the unit tests made all tests pass;
- removing the new unit test made all tests pass.

Flexmocking `TestingFarmJobHelper` here instead of `CoprBuildJobHelper` helped fixing the above.

But I wasn't able to figure out what the side effect (introduced maybe by the new unit test -- not sure) was.

If you can tell, please let me know. :confused: 

@lachmanfrantisek @TomasTomecek please take another look.

lachmanfrantisek: Not sure. It's too tricky in that amount of fixtures and mocks.

Just two general ideas:

- Import mocking (same applies for flexmock and requre) is hard to revert
- The order of fixture execution can cause some problems (trouble-makers are situations when some fixtures are executed when not needed or duplicated mocking of the same thing).

PR #534, opened 2020-04-03
Log Package Versions

### LINKED ISSUE #478: Log version(s) at the start
- [ ] Log version of packit-service and dependant libraries (packit, ogr, ...) at the start of the service/worker.
- [ ] Log that information at the start of each task.
- [ ] Share as much code as possible.

IceWreck:
Fixes #478

### DIFF
--- a/packit_service/log_versions.py
+++ b/packit_service/log_versions.py
@@ -5,11 +5,10 @@ from packit_service import __version__ as ps_version
 from sqlalchemy import __version__ as sqlal_version
 from flask_restx import __version__ as restx_version
 
-from flask import __version__ as flask_version  # type: ignore
-
 # Mypy errors out with Module 'flask' has no attribute '__version__'.
 # Python can find flask's version but mypy cannot.
 # So we use "type: ignore" to cause mypy to ignore that line.
+from flask import __version__ as flask_version  # type: ignore
 
 
 logger = logging.getLogger(__name__)
--- a/packit_service/log_versions.py
+++ b/packit_service/log_versions.py
@@ -1,0 +1,44 @@
+import logging
+
+from ogr import __version__ as ogr_version
+from packit_service import __version__ as ps_version
+from sqlalchemy import __version__ as sqlal_version
+from flask_restx import __version__ as restx_version
+
+from flask import __version__ as flask_version  # type: ignore
+
+# Mypy errors out with Module 'flask' has no attribute '__version__'.
+# Python can find flask's version but mypy cannot.
+# So we use "type: ignore" to cause mypy to ignore that line.
+
+
+logger = logging.getLogger(__name__)
+
+
+def log_package_versions(package_versions: list):
+    """It does the actual logging. Input is a list of tuples having pkg name and version."""
+    log_string = "\nPackage Versions:"
+    for name, version in package_versions:
+        log_string += f"\n* {name} {version}"
+    logger.info(log_string)
+
+
+def log_job_versions():
+    """Log essential package versions before running a job."""
+    package_versions = [
+        ("OGR", ogr_version),
+        ("Packit Service", ps_version),
+        ("SQL Alchemy", sqlal_version),
+        # NOTE: Can't log packit's version for now because it does not provide one.
+    ]
+    log_package_versions(package_versions)
+
+
+def log_service_versions():
+    """Log versions of packages used in the service."""
+    package_versions = [
+        ("Flask", flask_version),
+        ("Flask RestX", restx_version),
+        ("Packit Service", ps_version),
+    ]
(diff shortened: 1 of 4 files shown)


jpopelka:
Check the alerts from lgtm-com above, please

packit_service/log_versions.py
```diff
@@ -0,0 +1,42 @@
...
+from sqlalchemy import __version__ as sqlal_version
+from flask_restx import __version__ as restx_version
+from flask import __version__ as flask_version
+
+# from packit import __version__ as packit_version
+
+logger = logging.getLogger(__name__)
+
+
+class LogPackageVersions:
```
csomh, line 14: As far as I see, nothing in the code bellow required having a class. Let's keep things simple and use only functions, without having a class.

packit_service/log_versions.py
```diff
@@ -0,0 +1,46 @@
...
+# from packit import __version__ as packit_version
+
+from ogr import __version__ as ogr_version
+from packit_service import __version__ as ps_version
+from sqlalchemy import __version__ as sqlal_version
+from flask_restx import __version__ as restx_version
+from flask import __version__ as flask_version  # type: ignore
+
+
+# For some reason mypy has an issue with flask's __version__ even though it works completely fine.
```
csomh, line 12: This comment is about a line which is 3 lines above. Usually comments tend to belong to lines that follow. Please move this above the line that belongs to.

"some reason" and "works completely fine" contradict each other, and in this sentence they are pretty vague: it's not clear what the `mypy` error is the `type: ignore` tries to get rid of. It would help future readers to clarify that.

packit_service/log_versions.py
```diff
@@ -0,0 +1,46 @@
...
+# log package versions before running a job
+def log_versions_before_job():
+    package_versions = [
+        ("OGR", ogr_version),
+        ("Packit Service", ps_version),
+        ("SQL Alchemy", sqlal_version),
+    ]
+    log_package_versions(package_versions)
+
+    # NOTE: Can't log packit's version for now because it does not provide one.
```
csomh, line 36: This comment is also out of context, it would be much more useful to put it in the `package_versions` list above.

packit_service/log_versions.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+# For some reason mypy has an issue with flask's __version__ even though it works completely fine.
+# The "type: ignore" comment causes mypy to ignore that line.
+
+
+logger = logging.getLogger(__name__)
+
+
+# does the actual logging
```
csomh, line 19: These comments above the function signatures seem to be better suited as doc-strings.

packit_service/log_versions.py
```diff
@@ -0,0 +1,46 @@
...
+def log_package_versions(package_versions: list):
+    log_string = "\nPackage Versions:"
+    for name, version in package_versions:
+        log_string += f"\n* {name} {version}"
+    logger.info(log_string)
+
+
+# log package versions before running a job
+def log_versions_before_job():
```
csomh, line 28:
```suggestion
def log_job_versions():
```

packit_service/log_versions.py
```diff
@@ -0,0 +1,46 @@
...
+        ("Packit Service", ps_version),
+        ("SQL Alchemy", sqlal_version),
+    ]
+    log_package_versions(package_versions)
+
+    # NOTE: Can't log packit's version for now because it does not provide one.
+
+
+# log versions of packages used in the service part
+def log_version_during_service():
```
csomh, line 40:
```suggestion
def log_service_versions():
```

packit_service/log_versions.py
IceWreck: Sorry, I should've been clearer. I'll make these changes today and make the comments more detailed.

IceWreck:
@csomh Done! I still had to leave one line between the import and the comment  because black doesn't let me use it. 
BTW, zuul pre-commit is failing with Internal Server Error.

csomh:
> @csomh Done! I still had to leave one empty line between the import and the comment because black doesn't let me fill that line

I meant something like this:

```python
from ogr import __version__ as ogr_version
from packit_service import __version__ as ps_version
from sqlalchemy import __version__ as sqlal_version
from flask_restx import __version__ as restx_version

# Mypy errors out with Module 'flask' has no attribute '__version__'.
# Python can find flask's version but mypy cannot.
# So we use "type: ignore" to cause mypy to ignore that line.
from flask import __version__ as flask_version  # type: ignore
```

> BTW, zuul pre-commit is failing with Internal Server Error.

Yes, [prettier has issues](https://github.com/packit-service/packit/issues/784).

TomasTomecek:
recheck

PR #542, opened 2020-04-07
Small fixes and improvements in models.py

lbarcziova:
I was testing my migration script and found some bugs in representations (calling non-existing attributes)

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -345,7 +345,7 @@ class JobTriggerModel(Base):
             )
 
     def __repr__(self):
-        return f"JobTriggerModel(id={self.pr_id}, type={self.type}, trigger_id={self.trigger_id})"
+        return f"JobTriggerModel(type={self.type}, trigger_id={self.trigger_id})"
 
     def __str__(self):
         return self.__repr__()
@@ -638,16 +638,12 @@ class WhitelistModel(Base):
     def add_account(cls, account_name: str, status: str):
         with get_sa_session() as session:
             account = cls.get_account(account_name)
-            if account is not None:
-                account.status = status
-                session.add(account)
-                return account
-            else:
+            if not account:
                 account = cls()
                 account.account_name = account_name
-                account.status = status
-                session.add(account)
-                return account
+            account.status = status
+            session.add(account)
+            return account
 
     @classmethod
     def get_account(cls, account_name: str) -> Optional["WhitelistModel"]:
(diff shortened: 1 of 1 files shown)


packit_service/models.py
```diff
@@ -840,3 +836,9 @@ def create(cls, event):
                 ]
                 session.add(installation)
             return installation
+
+    def __repr__(self):
+        return f"InstallationModel(id={self.id}, account={self.account_login})"
+
+    def __str__(self):
+        return self.__repr__()
```
jpopelka, line 844: `__str__` is I think not needed if it's the same `__repr__`.
From [docs](https://docs.python.org/3/reference/datamodel.html#object.__repr__):
"If a class defines `__repr__()` but not `__str__()`, then `__repr__()` is also used when an “informal” string representation of instances of that class is required."

packit_service/models.py
```diff
@@ -245,7 +245,9 @@ def get_or_create(
             return git_branch
 
     def __repr__(self):
-        return f"GitBranchModel(id={self.pr_id}, name={self.name},  project={self.project})"
+        return (
+            f"GitBranchModel(id={self.id}, name={self.name},  project={self.project})"
```
lachmanfrantisek, line 249: We probably don't need the private keys in that.

lbarcziova:
@lachmanfrantisek I had not noticed that, sorry

packit_service/models.py
lbarcziova: It is used for every class in models, so can I remove it everywhere?

lachmanfrantisek:
> @lachmanfrantisek I had not noticed that, sorry

No need to sorry, I discovered that during the work on some fix and haven't created an issue. The simplification of `create*` methods should be still valid. I hope it will be merged soon so you can rebase.

packit_service/models.py
TomasTomecek: @jpopelka you're idd correct - `__repr__` is better and we don't need `__str__`:
```python
class C:
    a = 1
    b = "2"

    def __str__(self):
        return f"C(a={self.a}, b={self.b})"


class D:
    a = 1
    b = "2"

    def __repr__(self):
        return f"D(a={self.a}, b={self.b})"


print(f"str(C()): {str(C())}")
print(f"str(C()): {C()}")
print(f"str(D()): {str(D())}")
print(f"str(D()): {D()}")
print(f"repr(C()): {repr(C())}")
print(f"repr(D()): {repr(D())}")
```

```
str(C()): C(a=1, b=2)
str(C()): C(a=1, b=2)
str(D()): D(a=1, b=2)
str(D()): D(a=1, b=2)
repr(C()): <__main__.C object at 0x7fae2cce9fd0>
repr(D()): D(a=1, b=2)
```

PR #544, opened 2020-04-08
dont send deployment.dev events to sentry

TomasTomecek:
I was getting annoyed by the nasty tracebacks lately.

### DIFF
--- a/packit_service/sentry_integration.py
+++ b/packit_service/sentry_integration.py
@@ -23,6 +23,7 @@ import logging
 from contextlib import contextmanager
 from os import getenv
 
+from packit_service.config import ServiceConfig, Deployment
 from packit_service.utils import only_once
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


csomh:
@lachmanfrantisek @lbarcziova @dhodovsk @sakalosj was any of you checking these sentry alerts from dev envs? is it okay if disabling it is hardcoded, or better to be  configurable instead?

TomasTomecek:
> @lachmanfrantisek @lbarcziova @dhodovsk @sakalosj was any of you checking these sentry alerts from dev envs? is it okay if disabling it is hardcoded, or better to be configurable instead?

Thanks, we probably should have this discussion before merging. As I wrote in the OP, I got frustrated from seeing ugly errors which are not happening in the official deployments.

Will add this to arch mtg.

packit_service/sentry_integration.py
```diff
@@ -77,10 +78,11 @@ def configure_sentry(
 
 
 def send_to_sentry(ex):
-    # so that we don't have to have sentry sdk installed locally
-    import sentry_sdk
+    if ServiceConfig.get_service_config().deployment != Deployment.dev:
```
lachmanfrantisek, line 81: Cannot this be handled in the sentry-setup?

TomasTomecek: Please elaborate.

lachmanfrantisek: We have a place in code where we setup entry:
- If this "filtering" can be moved there.
- Check if sentry itself cannot handle this "filtering".

PR #549, opened 2020-04-09
Rewrite API calls to get data from postgres, remove getting info from…

lbarcziova:
… redis

TODO:

- [x] get data only from postgres
- [x]  fix the tests
- [x] check that the API works


rishavanand:
Hey @lbarcziova 
Will this PR also resolve the `500 internal server error` of the [/api/copr-builds](https://stg.packit.dev/api/copr-builds) endpoint? If not I'll create a new issue.
The same error is thrown even on the local deployment but only for the `/api/copr-builds/<build-id>` endpoint.

PR #550, opened 2020-04-09
Fix the migration script behaviour when pr ID cannot be parsed from t…

lbarcziova:
…he project name

E.g. when project name is `packit-service-hello-world-build-branch-stg`


TomasTomecek:
can you actually check on stg if it's already deployed?

lbarcziova:
When starting the container, `alembic upgrade head` fails each time, so I think there is no need for a new migration

jpopelka:
`stg` looks OK now, thanks!

PR #560, opened 2020-04-14
Fix bug in api/copr-builds

lbarcziova:


packit_service/service/api/copr_builds.py
```diff
@@ -121,7 +121,7 @@ def get(self, id):
             if project:
                 build_dict["repo_namespace"] = project.namespace
                 build_dict["repo_name"] = project.repo_name
-                build_dict["pr_id"] = build.trigger.get_trigger_object().pr_id
+                build_dict["pr_id"] = build.job_trigger.trigger_id
```
lachmanfrantisek, line 124: This does not look like `pr_id` -- it's a database `id` of the specific trigger object (like `GitBranchModel`, `PullRequestModel`).

If you expect `build` for pull-request -- so the trigger object is `PullRequestModel`, you can use this:

```suggestion
                build_dict["pr_id"] = build.job_trigger.get_trigger_object().pr_id
```

But you need to check, that you get `PullRequestModel` and not e.g. `GitBranchModel`.

---

`job_trigger` is only a mapper between `CoprBuildModel` and specific trigger like `PullRequestModel`, `GitBranchModel`, `ProjectReleaseModel`, ...

packit_service/models.py
```diff
@@ -390,6 +390,12 @@ def get_project(self) -> Optional[GitProjectModel]:
             return None
         return trigger_object.project
 
+    def get_pr_id(self) -> Optional[int]:
+        trigger_object = self.job_trigger.get_trigger_object()
+        if isinstance(trigger_object, PullRequestModel):
+            return trigger_object.pr_id
+        return None
```
lachmanfrantisek, line 397: :+1:

tests_requre/database/test_models.py
```diff
@@ -95,6 +95,7 @@ def test_create_copr_build(clean_before_and_after, a_copr_build_for_pr):
     # we will check if a_copr_build has build_submitted_time value that's within the past hour
     time_last_hour = datetime.utcnow() - timedelta(hours=1)
     assert a_copr_build_for_pr.build_submitted_time > time_last_hour
+    assert a_copr_build_for_pr.get_pr_id() == 342
```
lachmanfrantisek, line 98: Maybe also some `None` scenario for other triggers?

lbarcziova: added

TomasTomecek:
(I actually wanted a test for the API itself - never mind, will ask in the next PR related to API :D )

csomh:
What was the bug about?

lbarcziova:
Getting `pr_id` value was called wrongly

PR #561, opened 2020-04-14
Optionally Disable Sentry

### LINKED ISSUE #558: make it configurable to send events to sentry
This is a follow-up to https://github.com/packit-service/packit-service/pull/544

We should have a new packit-service.yaml config option to configure if we want to send events to sentry.

IceWreck:
Fixes #558 

- Adding `disable_sentry: True` to _secrets/packit-service.yaml_ disables sentry. 
- If this line is not present or `disable_sentry: False` is set then events will be sent to sentry like normal.

### DIFF
--- a/packit_service/sentry_integration.py
+++ b/packit_service/sentry_integration.py
@@ -24,8 +24,10 @@ from contextlib import contextmanager
 from os import getenv
 
 from packit_service.utils import only_once
+from packit_service.config import ServiceConfig
 
 logger = logging.getLogger(__name__)
+config = ServiceConfig.get_service_config()
 
 
 @only_once
@@ -35,6 +37,8 @@ def configure_sentry(
     flask_integration: bool = False,
     sqlalchemy_integration: bool = False,
 ) -> None:
+    """Sentry Configuration. Called once for each container."""
+
     logger.debug(
         f"Setup sentry for {runner_type}: "
         f"celery_integration={celery_integration}, "
@@ -42,9 +46,17 @@ def configure_sentry(
         f"sqlalchemy_integration={sqlalchemy_integration}"
     )
 
-    secret_key = getenv("SENTRY_SECRET")
-    if not secret_key:
+    if config.disable_sentry:
         return
+
+    secret_key = getenv("SENTRY_SECRET")
+
+    if not secret_key:
+        err_msg = (
+            "\n* Sentry is enabled but no key has been provided."
+            "\n* Please add 'disable_sentry: True' to packit-service.yaml to disable it."
+        )
+        raise NoSentryKeyError(err_msg)
 
     # so that we don't have to have sentry sdk installed locally
     import sentry_sdk
@@ -100,3 +112,9 @@ def push_scope_to_sentry():
 
         with sentry_sdk.push_scope() as scope:
             yield scope
+
+
+class NoSentryKeyError(Exception):
+    """Raise when sentry is enabled but key has not been provided."""
+
+    pass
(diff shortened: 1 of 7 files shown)


packit_service/sentry_integration.py
```diff
@@ -42,10 +44,13 @@ def configure_sentry(
         f"sqlalchemy_integration={sqlalchemy_integration}"
     )
 
+    if config.disable_sentry:
+        return
+
     secret_key = getenv("SENTRY_SECRET")
+
     if not secret_key:
         return
```
csomh, line 53: Now that I see these two together, isn't having yet another config flag to disable Sentry superfluous, when not setting the secret would achieve the same result?

Maybe this issue could be solved by updating the docs/tools for local deployments.

TomasTomecek: I'd say this is the right thing to do: if disable_sentry is False and the key is not set, we'd get an error right away - without the config option, we should get an error once we want to send an event to sentry, which is pretty late

csomh: > I'd say this is the right thing to do: if disable_sentry is False and the key is not set, we'd get an error right away...

```
if not secret_key:
    return
```

I see  no error here, am I missing something?

TomasTomecek: I actually wrote that sentence without looking at the code :D

@IceWreck could you please update the code by raising an exc if the secret_key is not set? Thank you.

packit_service/service/app.py
```diff
@@ -37,12 +38,18 @@
...
+    # SENTRY_SECRET is not passed to the service pod or container
+    # https://github.com/packit-service/deployment/blob/master/openshift/packit-service.yml.j2
+
+    # configure_sentry(
+    #     runner_type="packit-service",
+    #     celery_integration=True,
+    #     sqlalchemy_integration=True,
+    #     flask_integration=True,
+    # )
+
```
TomasTomecek, line 52: care to elaborate? do we want this change?

TomasTomecek: oh, this is about the httpd pod, got it now

IceWreck:
recheck

IceWreck:
@TomasTomecek I guess zuul is erroring out because the sentry key is not provided in zuul and disable_sentry is set to false by default. This is probably raising NoSentryKeyError and zuul tries again and again until it hits the retry limit.

We can do multiple things to fix this
- We can just change the default behaviour to `disable_sentry: true  ` but this would mean that everyone who is using sentry right now will have to change their config file to enable sentry.
-  Or someone will have to change the zuul packit-service.yaml file and set the disable sentry flag to true but then every other open PR will have to rebase or they'll get a marshmellow error in zuul, (they have to rebase anyway)
- We can just remove the error and replace it with a warning in the logs to either disable sentry or provide the key

TomasTomecek:
@IceWreck you're absolutely correct; looks like we need to disable sentry in the template p-s.yaml config in the deployment repo: https://github.com/packit-service/packit-service/blob/master/files/deployment.yaml#L82

IceWreck:
I did the change [there](https://github.com/packit-service/deployment/pull/77#pullrequestreview-394374680). 
So that will have to be merged first and then this PR will have to be merged and in the inbetween time zuul tests won't work for existing PRs (marshmellow errors)

csomh:
rechek

csomh:
recheck

IceWreck:
> Build failed.

Wait, I'll fix this

PR #564, opened 2020-04-15
JobMetadataConfig.dist_git_branch has been renamed to dist_git_branches

jpopelka:
Tests are expected to fail until https://github.com/packit-service/packit/pull/797 is merged

### DIFF
--- a/packit_service/worker/handlers/github_handlers.py
+++ b/packit_service/worker/handlers/github_handlers.py
@@ -24,7 +24,7 @@
 This file defines classes for job handlers specific for Github hooks
 """
 import logging
-from typing import Union, Any, Optional, List, Callable
+from typing import Union, Any, Optional, Callable, Set
 
 from ogr.abstract import GitProject, CommitStatus
 from packit.api import PackitAPI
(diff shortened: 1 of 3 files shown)


jpopelka:
recheck

packit_service/worker/handlers/github_handlers.py
```diff
@@ -521,13 +520,13 @@ def pre_check(self) -> bool:
         if not valid:
             return False
 
-        configured_branch = (
-            self.koji_build_helper.job_build.metadata.dist_git_branch or "master"
+        configured_branches = (
+            self.koji_build_helper.job_build.metadata.branch or "master"
         )
-        if configured_branch != self.event.git_ref:
+        if self.event.git_ref != configured_branches:
```
lachmanfrantisek, line 526: This looks like you can configure multiple branches, but it can be only one `str` for now. (I don't get the reason for renaming to `configured_branches`.)

lachmanfrantisek: Do we want to support also a list of branch-names? That's not a bad idea.

jpopelka: Oh, I see. Thats a long story, I'll rather create a follow-up PR :-)

jpopelka: Will be resolved in #568

Issue #489, opened 2020-03-12
introduce tests for the API


TomasTomecek:
I did not have the heart to ask @IceWreck to write API tests for his PR https://github.com/packit-service/packit-service/pull/479

But we should start testing the API code as well. I'd start with the calls which are migrated to psql.

lachmanfrantisek:
The Schemathesis can be helpful for some validation and for testing the deployment (see  #368 ). I am not sure if we can use it for writting custom tests.

TomasTomecek:
my expectation here is functional tests: store some objects in DB and make sure you're getting the expected data in the API responses

PR #567, opened 2020-04-16
Tests for the API and other changes

IceWreck:
- [x] Modify API to give build status for every chroot
- [x] Tests for the API (partially done)
- [x] replace/supplement existing flexmock test for the logs view with actual/db based test


packit_service/service/api/copr_builds.py
```diff
@@ -60,7 +60,8 @@ def get(self):
                     "project": build.project_name,
                     "owner": build.owner,
                     "build_id": build.build_id,
-                    "status": build.status,
+                    "status": build.status,  # Legacy, remove later.
+                    "status_per_chroot": {},
```
TomasTomecek, line 64: I wouldn't personally mind breaking the API here and just set the dict to status

others may have a different opinion

IceWreck: I'll just build tests for the entire API. It was long overdue anyways #489

IceWreck:
I wanted to use pytests's (official) flask plugin but it isn't in dnf so I had change pytest's source from dnf to pip. (pytest from dnf but pytest-flask from pip don't work together)

Now the API tests are mostly working but an openshift integration test broke.

TomasTomecek:
I'd say the fedpkg test data need to be just regenerated.

IceWreck:
> I'd say the fedpkg test data need to be just regenerated.

Worked. Thanks.

packit_service/service/api/copr_builds.py
IceWreck: > I wouldn't personally mind breaking the API here and just set the dict to status
> 
> others may have a different opinion

So should I remove status and chroots, both are redundant if status_per_chroot exists.

TomasTomecek: since no one from the team raised an opinion here, let's keep it as it is now

PR #569, opened 2020-04-17
Report build status in Pagure

csomh:
When a build for a PR from a Pagure instance (git.centos.org, for example) starts or ends, update the flags on the PR, to make the check status visible on the PR page.

### DIFF
--- a/packit_service/worker/build/build_helper.py
+++ b/packit_service/worker/build/build_helper.py
@@ -80,7 +80,9 @@ class BaseBuildJobHelper:
         ] = event
         self.pr_id = (
             self.event.pr_id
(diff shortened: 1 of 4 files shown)


csomh:
@lachmanfrantisek @TomasTomecek please take a look.

I leave this as a draft as I'm still figuring out how to test this, but so far it seems this is all that is needed. Unless I missed something, which might easily happen.

Here is what this change intends to achieve:
- let StatusReporter be aware of the PR id if there is one (in which case it will be None).
- after setting the commit status flags, get the PR object and if it has a `set_flag` method, use that to set the status on the PR, too. (This will require ogr 0.11.2 to work).

The PR id should be already in the database.

csomh:
Added unit tests. Please label with *mergeit* if you still think it's okay.

PR #571, opened 2020-04-20
Revert "Optionally Disable Sentry"

csomh:
Reverts packit-service/packit-service#561

'packit-service.yaml' is shared between the service and worker pods, and I cannot see a quick way to disable sentry in one of them but keep it enabled in the other one.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -44,8 +44,6 @@ services:
       - redis
       - postgres
     environment:
-      # Sentry secret is passed via an env var, not the secrets folder.
-      # SENTRY_SECRET: https://<key>@<organization>.ingest.sentry.io/<project>
       DEPLOYMENT: dev
       REDIS_SERVICE_HOST: redis
       APP: packit_service.worker.tasks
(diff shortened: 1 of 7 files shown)


TomasTomecek:
dang

what would be the proper solution here? shall we discuss on arch meeting tomorrow?

csomh:
So the behaviour before this change was made was to disable Sentry, if the `SENTRY_SECRET` env var was not set. Basically, this env var worked both as a flag to turn on/off Sentry integration and to provide the token for it.

To be honest, I would find this behaviour sufficient, and I don't see the value in adding a separate flag for the turn on/off aspect. But I might be missing something.

From my point of view if we would like to avoid folks accidentally setting up Sentry integration in their local environments (which can happen only if they have access to the secrets), we should update the setup scripts/docs for these environments to avoid it, instead of adding this complication in the code (which needs tests and has to be considered during deployments).

Though if we see value adding a separate flag for the turn on/off aspect, then we should make sure, that either:
1. This flag is read from an env var, similar to `SENTRY_SECRET`.
2. We update our deployment playbooks to separate the configuration of the workers from the service (currently they share `packit-service.yaml`, amongst other things) in order to be able to configure the flag (and the secret) from the config file.

Whichever we choose, configuring the flag and the secret should be consistent: either set both by config file of set both both by env vars, otherwise we might have issues with setting them in tandem in the future, too.

PR #574, opened 2020-04-20
Add logging into the migration script, remove getting info via Copr API

lbarcziova:


alembic/versions/c6250555a36c_.py
```diff
@@ -43,6 +43,8 @@
 else:
     Base = declarative_base()
 
+logger = logging.getLogger(__name__)
```
TomasTomecek, line 46: I fear this would not log anything since the logger is not set up anywhere here; unless alembic is smart enough to set up the logger for us

PR #578, opened 2020-04-21
Fix getting of start and end time in CoprBuildEnd event

### LINKED ISSUE #575: Trying to get datetime from None when processing Copr event
Sentry Issue: [RED-HAT-0P-2EA](https://sentry.io/organizations/red-hat-0p/issues/1621355440/?referrer=github_integration)

```
TypeError: an integer is required (got type NoneType)
  File "packit_service/worker/tasks.py", line 55, in process_message
    task_results: dict = SteveJobs().process_message(event=event, topic=topic)
  File "packit_service/worker/jobs.py", line 332, in process_message
    jobs_results = self.process_jobs(event_object)
  File "packit_service/worker/jobs.py", line 195, in process_jobs
    handlers_results[job.type.value] = handler.run_n_clean()
  File "packit_service/worker/handlers/abstract.py", line 109, in run_n_clean
    return self.run()
  File "packit_service/worker/handlers/fedmsg_handlers.py", line 287, in run
    build_start_time = datetime.utcfromtimestamp(self.event.started_on)
```

lbarcziova:
Fixes #575 

- There is no info about the start and end time in the fedmsg event. In this PR we set the start and end time to time when processing CoprBuildStart and CoprBuildEnd event (not using Copr API calls to avoid possible loss of the message when Copr API is down)

- The copr logs url is set during CoprBuildStart, so it is not needed to set it in the CoprBuildEnd

### DIFF
--- a/packit_service/worker/handlers/fedmsg_handlers.py
+++ b/packit_service/worker/handlers/fedmsg_handlers.py
@@ -24,7 +24,6 @@
 This file defines classes for job handlers specific for Fedmsg events
 """
 
-from copr.v3 import Client as CoprClient
 import logging
 from typing import Type, Optional
 
@@ -239,6 +238,7 @@ class CoprBuildEndHandler(FedmsgHandler):
             logger.info(msg)
             return HandlerResults(success=True, details={"msg": msg})
 
+        build.set_end_time(datetime.utcnow())
         url = get_log_url(build.id)
 
         # https://pagure.io/copr/copr/blob/master/f/common/copr_common/enums.py#_42
@@ -284,17 +284,6 @@ class CoprBuildEndHandler(FedmsgHandler):
             chroot=self.event.chroot,
         )
         build.set_status(PG_COPR_BUILD_STATUS_SUCCESS)
-
-        copr_client = CoprClient.create_from_config_file()
-        chroot_build = copr_client.build_chroot_proxy.get(
-            self.event.build_id, self.event.chroot
-        )
-        started_on = chroot_build.started_on
-        ended_on = chroot_build.ended_on
-
-        build_start_time = datetime.utcfromtimestamp(started_on) if started_on else None
-        build_end_time = datetime.utcfromtimestamp(ended_on) if ended_on else None
-        build.set_start_end_time(build_start_time, build_end_time)
 
         if (
             self.build_job_helper.job_tests
@@ -353,6 +342,7 @@ class CoprBuildStartHandler(FedmsgHandler):
             logger.warning(msg)
             return HandlerResults(success=False, details={"msg": msg})
 
+        build.set_start_time(datetime.utcnow())
         url = get_log_url(build.id)
         build.set_status("pending")
         copr_build_logs = get_copr_build_logs_url(self.event)
--- a/packit_service/worker/handlers/fedmsg_handlers.py
+++ b/packit_service/worker/handlers/fedmsg_handlers.py
@@ -238,7 +238,12 @@ class CoprBuildEndHandler(FedmsgHandler):
             logger.info(msg)
             return HandlerResults(success=True, details={"msg": msg})
 
-        build.set_end_time(datetime.utcnow())
+        end_time = (
+            datetime.utcfromtimestamp(self.event.timestamp)
+            if self.event.timestamp
+            else None
+        )
+        build.set_end_time(end_time)
         url = get_log_url(build.id)
 
         # https://pagure.io/copr/copr/blob/master/f/common/copr_common/enums.py#_42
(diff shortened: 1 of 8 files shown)


csomh:
Hmmmm :thinking: 

Reaching out to Copr while handling the message for the build-end seems fragile: if Copr API is down, there is a risk of loosing a message. I'm wondering if it makes sense to assume this risk just to keep statistics (iirc, these timestamps are only for the dashboard).

I'm not sure, if this would make sense, but couldn't the build start and end times be the times when the build start and end messages are received? Or better: use the timestamps of those messages?

TomasTomecek:
> Hmmmm thinking
> 
> Reaching out to Copr while handling the message for the build-end seems fragile: if Copr API is down, there is a risk of loosing a message. I'm wondering if it makes sense to assume this risk just to keep statistics (iirc, these timestamps are only for the dashboard).
> 
> I'm not sure, if this would make sense, but couldn't the build start and end times be the times when the build start and end messages are received? Or better: use the timestamps of those messages?

+1

would be nice to wrap the code which interacts with the API in try/except and default to current time

we should also have some code in CoprBuildStartHandler to set the start time

lbarcziova:
@csomh @TomasTomecek thanks for the feedback! I will split the setting of the start and end time and default will be the current time of CoprBuildStart and CoprBuildEnd events

csomh:
> would be nice to wrap the code which interacts with the API in try/except and default to current time

How about dropping that code for good and don't rely on Copr API for this data? Yes, message timestamps and Copr timestamps might be somewhat different, though that shouldn't be an issue.

lbarcziova:
@TomasTomecek @csomh https://fedora-fedmsg.readthedocs.io/en/latest/topics.html#id265 is this the fedmsg that we use? If yes, we could possibly use the timestamp directly from the message

csomh:
> @TomasTomecek @csomh https://fedora-fedmsg.readthedocs.io/en/latest/topics.html#id265 is this the fedmsg that we use? If yes, we could possibly use the timestamp directly from the message

Yes, this is it. [Here](https://apps.fedoraproject.org/datagrepper/id?id=2020-50a4642d-a81d-4e5f-8ac2-961b5b88b798&is_raw=true&size=extra-large) is a recent example from datagrepper.

Unfortunately to message these handlers are working with is not the real message. That one is handled in `packit-service-fedmsg`, and it's only the `body` and the `topic` that is forwarded in the celery task. So the timestamp first would need to be saved there ([similar to `topic`](https://github.com/packit-service/packit-service-fedmsg/blob/master/packit_service_fedmsg/consumer.py#L89)) and then could be used in this code.

packit_service/worker/handlers/fedmsg_handlers.py
```diff
@@ -238,6 +238,7 @@ def run(self):
             logger.info(msg)
             return HandlerResults(success=True, details={"msg": msg})
 
+        build.set_end_time(datetime.utcnow())
```
csomh, line 241: With `utcnow`, if the celery task waits in the queue for a worker, the timestamp will be quite far from reality.

I'm okay to merge this as it is now, but should get the timestamp field in the message and use that one in the future.

Do you need help with that?

lbarcziova: Do you mean to change it firstly in the `packit-service-fedmsg` repo as you described above and use that value?

csomh: Yes, that's what I meant.

packit_service/worker/handlers/fedmsg_handlers.py
```diff
@@ -238,6 +238,12 @@ def run(self):
             logger.info(msg)
             return HandlerResults(success=True, details={"msg": msg})
 
+        end_time = (
+            datetime.utcfromtimestamp(self.event.timestamp)
+            if self.event.timestamp
+            else None
+        )
+        build.set_end_time(end_time)
```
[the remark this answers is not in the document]
TomasTomecek: Just verified that psql is fine with it:
```
a_copr_build_for_pr.set_end_time(None)                                                                                                                             
assert a_copr_build_for_pr.build_finished_time is None  
```
could you please add it in tests_requre/test_models? so we're sure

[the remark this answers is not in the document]
lbarcziova: >I debate what's worse: have no time or slightly incorrect time - if we did now() here (a task may not be processed right away - what Hunor raised)

It should not be None as we will set the timestamp in packit-service-fedmsg, I added it just to be sure this won't cause failures
>could you please add it in tests_requre/test_models? so we're sure

added

Issue #556, opened 2020-04-13
Add support for Pagure (pagure.io and other pagure instances)


Conan-Kudo:
A number of projects that would benefit from Packit CI are hosted on [pagure.io](https://pagure.io). For example:

* [ironthree projects](https://pagure.io/projects/ironthree/%2A)
* [`rust2rpm`](https://pagure.io/fedora-rust/rust2rpm)
* [`go2rpm`](https://pagure.io/GoSIG/go2rpm)
* [`rpkg`](https://pagure.io/rpkg)
* [FreeIPA](https://pagure.io/freeipa)
* [SSSD](https://pagure.io/SSSD/sssd)
* [389ds](https://pagure.io/389-ds-base)

Also, with the upcoming forge by the FSF based on Pagure coming, it'd be great if projects hosted there could leverage Packit too. And naturally, other instances hosted by people as well...

Edit @TomasTomecek: adding more to the list:
* [STR](https://pagure.io/standard-test-roles)
* Fedora kernel (?)
* [~~Copr~~](https://pagure.io/copr/copr) (moved to GitHub)

---

### Outcome of the Packit planning for Q1 of '23

* Priority: **won't do** (reasons and update on the current state of this epic in the [comment below](https://github.com/packit/packit-service/issues/556#issuecomment-1385225514))
* Benefits that we expect: more users of our service, common and more available automation for the users
* Consumers of those benefits: summed up in the description of this issue

ignatenkobrain:
rust2rpm won't use it because packit is pretty much useless for these cases where spec is not stored upstream.

TomasTomecek:
> rust2rpm won't use it because packit is pretty much useless for these cases where spec is not stored upstream.

you can set up packit to fetch spec from any URL: e.g. rawhide; no need to store it in upstream

https://packit.dev/faq/#how-can-i-download-rpm-spec-file-if-it-is-not-part-of-upstream-repository

TomasTomecek:
@sakalosj @lachmanfrantisek @csomh guys, given we are adding pagure support to packit right now, how much work do you expect this to be?

lachmanfrantisek:
@TomasTomecek Depends on what jobs we want to support. For the same functionality as we have on GitHub:

- [ ] we need to listen for new events on that instance
- [ ] parsing of the events (should be done by Jano)
  - [ ] we are missing some features in Pagure API (e.g. file listing)
  - [ ] Jano is probably working only on PRs (not Release and branch push)
[...]

PR #582, opened 2020-04-21
add parsing of centos pull-request comments and updates

sakalosj:
Added parsing of centos pull-request comments and updates.


packit_service/worker/jobs.py
```diff
@@ -271,6 +279,13 @@ def process_comment_jobs(
                 success=True, details={"msg": "Account is not whitelisted!"}
             )
 
+        # VERY UGLY
+        # TODO: REFACTOR !!!
+        if handler_kls == GitHubPullRequestCommentCoprBuildHandler and isinstance(
+            event, PullRequestCommentPagureEvent
+        ):
+            handler_kls = PagurePullRequestCommentCoprBuildHandler
+
```
sakalosj, line 288: this is very ugly, but enough for my testing and probably can stay here till refactoring.
any suggestions are welcome

lachmanfrantisek: Some nicer way would require lists in `MAP_COMMENT_ACTION_TO_HANDLER` and making some logic on top of it.

I would use this approach for now and discuss after summit -- I'll add a note to our refactor agenda about this.

lachmanfrantisek: I think we already discuss that, but do we want to use whitelist for centos-stream? Can we move this above the whitelisting code?

packit_service/worker/handlers/centosmsg_handlers.py
```diff
@@ -142,3 +147,35 @@ def pre_check(self) -> bool:
...
+    event: PullRequestCommentPagureEvent
+
+    def __init__(self, config: ServiceConfig, event: PullRequestCommentPagureEvent):
+        super().__init__(config=config, event=event)
+        self.config = config
+        self.event = event
+        self.project: GitProject = event.get_project()
+        # Get the latest pull request commit
+        self.event.commit_sha = self.project.get_pr(self.event.pr_id).head_commit
+        self.event.base_ref = self.event.commit_sha
```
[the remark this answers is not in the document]
sakalosj: followed [github implementation](https://github.com/packit-service/packit-service/blob/d86f25073029cd243767aeeb3e85d22f60c03ddb/packit_service/worker/handlers/github_handlers.py#L572-L592) counterpart here. will fix it

packit_service/worker/jobs.py
sakalosj: yes u are right, will fix it

packit_service/worker/handlers/centosmsg_handlers.py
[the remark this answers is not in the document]
lachmanfrantisek: Thanks!

TomasTomecek:
uhhh, is github having issues again? I cannot reply to the inline comments

> I think we already discuss that, but do we want to use whitelist for centos-stream? Can we move this above the whitelisting code?

We don't need whitelist for the Stream.

lachmanfrantisek:
> uhhh, is github having issues again? I cannot reply to the inline comments

Neither can I...;(

packit_service/worker/handlers/centosmsg_handlers.py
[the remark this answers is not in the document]
sakalosj: fixed in parser

packit_service/worker/jobs.py
sakalosj: fixed  - moved logic to Whitelist to avoid whitelisting

packit_service/worker/whitelist.py
```diff
@@ -190,8 +191,17 @@ def check_and_report(
         :param project: GitProject
         :return:
         """
+
+        # whitelist checks dont apply to CentOS (Pagure)
+        if isinstance(
+            event,
+            (PushPagureEvent, PullRequestPagureEvent, PullRequestCommentPagureEvent),
+        ):
+            logger.info("Centos (Pagure) events dont require whitelist checks")
```
lachmanfrantisek, line 200: We need to think about this more when working on the Pagure support as upstream (#556).

sakalosj: yes, i am aware of that, and in future we can achieve CentOS stream events exclusion using some Event attribute 

e.g. our centos listener is adding topic field to message dictionary which is sent for processing. it consists of sourceurl+gitevent -> sourceurl can be used for this purpose 

i think that at this point, there is no reason to implement it.

sakalosj: should i implement it in this PR?

packit_service/worker/handlers/centosmsg_handlers.py
```diff
@@ -142,3 +147,33 @@ def pre_check(self) -> bool:
...
+    type = CommentAction.copr_build
+    triggers = [TheJobTriggerType.pr_comment]
+    event: PullRequestCommentPagureEvent
+
+    def __init__(self, config: ServiceConfig, event: PullRequestCommentPagureEvent):
+        super().__init__(config=config, event=event)
+        self.config = config
+        self.event = event
+        self.project: GitProject = event.get_project()
+        # Get the latest pull request commit
```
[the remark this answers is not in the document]
sakalosj: fixed

packit_service/worker/whitelist.py
lachmanfrantisek: Not need to solve that here.

PR #586, opened 2020-04-23
CentOS related changes

lachmanfrantisek:
- Merge the handlers for github and centos.
  - It makes us troubles with mapping to have another set of (very similar) classes. (e.g. required builds for test configured are problematic)
  - Most of the code can be shared between those groups of classes.

- Move package_config and project handling to the events.    
    - Sharing of the implementation.
    - The required info is in the Event classes.
    - Handlers do not need to differ when the package_config needs to be loaded differently.
    - Events can more easily load config from the PR branch.

TODO:
- [x] load package from PR
- [x] add tests
- [ ] merge also the comment handlers (partially)


tests/unit/test_jobs.py
```diff
@@ -59,10 +58,10 @@
                 JobConfig(
                     type=JobType.copr_build,
                     trigger=JobConfigTriggerType.pull_request,
-                    metadata={},
```
jpopelka, line 62: Whoops, sorry for not changing this in #548.

You don't have to specify the `metadata` argument at all, cause it will be [set to JobMetadataConfig()](https://github.com/packit-service/packit/blob/master/packit/config/job_config.py#L122) anyway.

lachmanfrantisek: No problem. It was not used. I'll remove the empty ones.

I am just wondering why mypy does not tell us about that. (PyCharm shows that problem.)

lachmanfrantisek:
Let's merge and improve in the follow-up PR(s).

PR #589, opened 2020-04-28
Do not use forks in github

lachmanfrantisek:
Do not use event.base_project in Github. As a GitHub app, we cannot authenticate to the fork projects.

Requires: https://github.com/packit-service/packit/pull/818

TODO:
- [x] Add tests for actual config loading. (We mock the getter in all tests.)

### DIFF
--- a/packit_service/config.py
+++ b/packit_service/config.py
@@ -24,7 +24,6 @@ import logging
 from pathlib import Path
 from typing import Set, Optional
 
-import yaml
 from yaml import safe_load
 
 from ogr.abstract import GitProject
@@ -33,7 +32,6 @@ from packit.config import (
     Config,
     get_package_config_from_repo,
     PackageConfig,
-    parse_loaded_config,
 )
 from packit.exceptions import PackitException, PackitConfigException
 from packit_service.constants import (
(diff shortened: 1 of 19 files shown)


lachmanfrantisek:
recheck

packit_service/config.py
```diff
@@ -146,25 +144,28 @@ def get_service_config(cls) -> "ServiceConfig":
         return cls.service_config
 
 
-class PackageConfigGetterForGithub:
+class GeneralPackageConfigGetter:
```
TomasTomecek, line 147: We should also have SergeantPackageConfigGetter

jpopelka: aye aye captain ;-)

TomasTomecek:
recheck

lachmanfrantisek:
It has some problem with the centos-config loading.

Please, wait for https://github.com/packit-service/packit/pull/818

packit_service/config.py
sakalosj: i suggest BasePackgeConfigGetter, "Base" is used at least in ogr for similar, so to keep naming convention

csomh: If it's a naming discussion, I must join!

Would calling it only `PackageConfigGetter` be good enough?

lachmanfrantisek: Oh no. Naming...;(

My notes regardning the suggestion. (But I can pick any of these if you want.)

> `BasePac...`
- In OGR we use it between abstract class and the real implementation to share the code. Here, we are using this class as well.

> `PackageCo...` 
- This makes sense to me when we don't have any other implementation. (Like we have for Pagure now.)

lachmanfrantisek: OK, as I am thinking about that. I'll merge those two classes to `PackageConfigGetter` and solve the difference in another way.

lachmanfrantisek:
The PR in packit was merged and the code was a bit simplified.

PTAL

PR #592, opened 2020-04-29
Add a script for removing redis data

lbarcziova:


TomasTomecek:
Before we run this in prod. Could we retrieve all the data first, as json and store them in some internal repo? Just to be perfectly safe in case something goes wrong.

WDYT?

lbarcziova:
> Edit: oh, so this is only for "celery-task-meta-*", I guess we don't need to back up those.

This should be also for clearing whole db 1 (if I am right, there we store copr-builds, installations, and whitelist). We can store it in some repo although we have these data already in postgres

TomasTomecek:
> > Edit: oh, so this is only for "celery-task-meta-*", I guess we don't need to back up those.
> 
> This should be also for clearing whole db 1 (if I am right, there we store copr-builds, installations, and whitelist). We can store it in some repo although we have these data already in postgres

My point here is that I'm trying to be really careful since we are talking about removing data.

If you feel like we have all the data in psql, safe and sound, I'm going to trust you and we can merge this and run it in prod.

Maybe one thing: do have the DBs (redis, psql) in sync since we had that problem with the data migration the other day?

lbarcziova:
@TomasTomecek You are right, it is definitely safer to make a copy of the data before we run it in prod.

> Maybe one thing: do have the DBs (redis, psql) in sync since we had that problem with the data migration the other day?

I will check the whitelist, other tables should be synced.

lbarcziova:
@TomasTomecek anyway, can we merge this to have it in prod?

PR #596, opened 2020-04-30
Optimize httpd memory usage 

### LINKED ISSUE #594: Apache memory consumption
`packit-service` pod consumes 200Mib even before it starts to do anything.

Could [the config](https://github.com/packit-service/deployment/blob/master/openshift/cmap-packit-httpd-conf.yml.j2) be tweaked [somehow to lower it](https://www.google.cz/search?q=lower+apache+memory+usage)?

jpopelka:
With these settings, the container's memory consumption doesn't go over 80MiB.

Fixes #594

### DIFF
--- a/files/packit-httpd.conf
+++ b/files/packit-httpd.conf
@@ -25,6 +25,21 @@ LogLevel wsgi:debug
 ErrorLog /dev/stderr
 TransferLog /dev/stdout
 
+# These directives are here to lower the memory usage.
+# For their description see https://httpd.apache.org/docs/2.4/mod/mpm_common.html
+# Note, Fedora's httpd is compiled with event MPM
+# https://httpd.apache.org/docs/2.4/mod/event.html
+# Number of child server PROCESSES created at startup
+StartServers 1
+# Number of THREADS created by each child server/process
+ThreadsPerChild 6
+# Minimum number of idle threads available to handle request spikes
+MinSpareThreads 2
+# Maximum number of idle threads
+MaxSpareThreads 2
+# Maximum number of connections that will be processed simultaneously (others are queued)
+MaxRequestWorkers 8
+
 <VirtualHost *:8443>
     SSLEngine on
     SSLCertificateFile /secrets/fullchain.pem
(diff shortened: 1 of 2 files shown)


files/packit-httpd.conf
```diff
@@ -25,6 +25,21 @@ LogLevel wsgi:debug
 ErrorLog /dev/stderr
 TransferLog /dev/stdout
 
+# These directives are here to lower the memory usage.
+# For their description see https://httpd.apache.org/docs/2.4/mod/mpm_common.html
+# Note, Fedora's httpd is compiled with event MPM
+# https://httpd.apache.org/docs/2.4/mod/event.html
+# Number of child server PROCESSES created at startup
+StartServers 1
```
csomh, line 33: From the docs:

> The StartServers directive sets the number of child server processes created on startup. As the number of processes is dynamically controlled depending on the load, (see MinSpareThreads, MaxSpareThreads, MinSpareServers, MaxSpareServers) **there is usually little reason to adjust this parameter**.

jpopelka: Yes, **usually**, if you have normal/big load it makes no sense to set the initial value, because httpd will increase it later, **but** if we have a minimal load, setting it to minimal value (`1` in our case) IMHO makes `httpd` to not start more than necessary if you'll never need it (and `httpd` probably won't increase it due to our minimal load).

jpopelka:
> Do we have an understanding/expectation how the service will behave differently with these values?

It shouldn't behave differently, just consume less memory.

> why is it an issue Apache consuming 200M on start? Don't we have enough memory available?

We have enough, but it's expensive.

jpopelka:
>     * packit-service-pre-commit : FAILURE in 2m 45s

Looks like requre issue

```
  File "/home/zuul-worker/.cache/pre-commit/repo0zxvjxwp/py_env-python3/lib/python3.7/site-packages/requre/utils.py", line 29, in <module>
    from _pytest.python import Function
ModuleNotFoundError: No module named '_pytest'
```

It succeeded in the `check` pipeline, merging manually.

jpopelka:
> Looks like requre issue

https://github.com/packit-service/requre/pull/99

PR #601, opened 2020-05-04
Refactor the "is private repository" check

csomh:
Use the "project" property of "event_object" instead of a local
variable. This way, the project is cached and handlers will not need to
recreate them once again.

Also: don't do this only for GithubProjects. GitLab and Pagure projects
have the `is_private()` method implemented in OGR, so it is safe to skip
the type check and call `is_private()` whenever it's possible to tell
the project from the event.

Should cause #595 to happen less.


csomh:
recheck

PR #602, opened 2020-05-05
Use correct log levels

jpopelka:
https://docs.celeryproject.org/en/stable/reference/celery.bin.worker.html#cmdoption-celery-worker-l

Or do we want to use DEBUG everywhere?


jpopelka:
> debug logs contain the pidbox spam, right?

yes

> this sets celery's logging system, not our logging system, where we totally want debug

yes, we set them to DEBUG [here](https://github.com/packit-service/packit-service/blob/master/packit_service/worker/tasks.py#L49) and it works with this change

PR #606, opened 2020-05-06
[api/copr-builds] fix displaying of not-None start and finished time

lbarcziova:
These values, as well as `build_submitted_time` already, should be processed by `optional_time` function, not None value (datetime type) can't be displayed in API


lbarcziova:
recheck

PR #607, opened 2020-05-06
Use packit-service-fedmsg:stg in Docker Compose

csomh:
The `latest` tag is going to get deprecated once the image built from
the master branch is going to be tagged with `stg`, while the one built
from the stable branch is going to be tagged with `prod`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -96,7 +96,7 @@ services:
 
   fedora-messaging:
     container_name: fedora-messaging
-    image: usercont/packit-service-fedmsg:latest
+    image: docker.io/usercont/packit-service-fedmsg:stg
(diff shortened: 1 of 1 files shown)


docker-compose.yml
```diff
@@ -96,7 +96,7 @@ services:
 
   fedora-messaging:
     container_name: fedora-messaging
-    image: usercont/packit-service-fedmsg:latest
+    image: docker.io/usercont/packit-service-fedmsg:prod
```
jpopelka, line 99: Since docker-compose is used for local development, I'd rather see `:stg` here, but this is OK too.

csomh: Changed.

PR #613, opened 2020-05-11
Implement get_project() in InstallationEvent

lbarcziova:
I was checking the `events.py` file and all the events except `InstallationEvent` inherit from `AbstractForgeIndependentEvent`, which has implemented project() and get_project() methods, therefore this could solve the problem.

### DIFF
--- a/packit_service/service/events.py
+++ b/packit_service/service/events.py
@@ -497,6 +497,13 @@ class InstallationEvent(Event):
         result["status"] = result["status"].value
         return result
 
+    @property
+    def project(self):
+        return self.get_project()
+
+    def get_project(self):
+        return None
+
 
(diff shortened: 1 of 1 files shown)


csomh:
I'm wondering why `InstallationEvent` is different from all other events? Shouldn't it also inherit from `AbstractForgeIndependentEvent`?

lbarcziova:
> I'm wondering why InstallationEvent is different from all other events? Shouldn't it also inherit from AbstractForgeIndependentEvent?

I don't know the original reason, but `InstallationEvent` doesn't have any `project_url` (which we can make instance of project from), so it really doesn't make sense to inherit these methods from `AbstractForgeIndependentEvent`

PR #618, opened 2020-05-19
Koji build follow-up: correct url, state, target and more logging

lachmanfrantisek:
- Fix `state` and `url` for build status.
- Pick the correct target.
- Add missing test file. (Forgot to add when restructuring the commits in the first PR.)

### DIFF
--- a/packit_service/worker/build/koji_build.py
+++ b/packit_service/worker/build/koji_build.py
@@ -91,7 +91,7 @@ class KojiBuildJobHelper(BaseBuildJobHelper):
         for chroot in self.build_chroots:
 
             try:
-                build_id, web_url = self.run_build()
+                build_id, web_url = self.run_build(target=chroot)
             except Exception as ex:
                 sentry_integration.send_to_sentry(ex)
                 # TODO: Where can we show more info about failure?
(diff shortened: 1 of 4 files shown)


packit_service/worker/build/koji_build.py
```diff
@@ -136,12 +148,21 @@ def run_build(
...
-        )
+        try:
+            out = self.api.up.koji_build(
+                scratch=self.is_scratch,
+                nowait=True,
+                koji_target=target,
+                srpm_path=self.srpm_path,
+            )
+        except PackitCommandFailedError as ex:
+            logger.warning(
```
csomh, line 159: Wouldn't make more sense to log an error if we raise an error?

lachmanfrantisek: This code is located in packit -- we already do it but without the stout/stderr.

(I'll take a look at it.)

lachmanfrantisek:
I've also fixed the used target and added a test for it. Please, PTAL

packit_service/worker/build/koji_build.py
lachmanfrantisek: OK, I understood your comment as why we don't do this in a place the original error is raised.

PR #619, opened 2020-05-19
API for koji builds

lachmanfrantisek:
- Add API endpoints for koji builds.
- Add 'submitted' to the build views.

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -28,7 +28,7 @@ import logging
 import os
 from contextlib import contextmanager
 from datetime import datetime
-from typing import TYPE_CHECKING, Optional, Union, Iterable, Dict, Type
+from typing import TYPE_CHECKING, Optional, Union, Iterable, Dict, Type, Any
 
 from sqlalchemy import (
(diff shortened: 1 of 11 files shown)


packit_service/models.py
```diff
@@ -537,6 +537,35 @@ def set_build_logs_url(self, build_logs: str):
     def get_project(self) -> GitProjectModel:
         return self.job_trigger.get_trigger_object().project
 
+    @property
+    def api_structure(self) -> Dict[str, Any]:
```
csomh, line 541: This seems to me to be a serialisation exercise: are there other ways to do this, without having to write (and then maintain) ~30 lines of code?

lachmanfrantisek: - I wanted to have it in one place (unlike the koji build implementation).
- I wanted to control what's there.
  - Not all attributes make sense.
  - Handling with triggers needs to be done manually for now. (We can use `trigger` and nested serialization instead.) 
- It makes sense to do this change for all at once and have some design when providing payload with nested objects.

PR #622, opened 2020-05-21
Use the alias list from Packit in the tests

csomh:
Instead of hard-coding the values of targets in the tests, use the
ALIASES list from Packit and calculate values based on that.

This prevents the tests breaking when the list in Packit is updated.


csomh:
recheck

PR #640, opened 2020-05-29
override package_config from job.overrides

TomasTomecek:
Fixes packit-service/packit-service#418
Fixes https://github.com/packit-service/packit/issues/371

TODO:
* [x] merge packit-service/packit#848
* [x] new tests
* [x] get unblocked on insanity https://github.com/packit-service/packit/pull/853
* [x] fix old tests
* [ ] docs https://github.com/packit-service/packit.dev/pull/132
* [ ] be awesome

### DIFF
--- a/packit_service/worker/handlers/fedmsg_handlers.py
+++ b/packit_service/worker/handlers/fedmsg_handlers.py
@@ -143,6 +143,7 @@ class NewDistGitCommitHandler(FedmsgHandler):
 
 @add_topic
 @use_for(job_type=JobType.copr_build)
+@use_for(job_type=JobType.build)
 @required_by(job_type=JobType.tests)
 class CoprBuildEndHandler(FedmsgHandler):
     topic = "org.fedoraproject.prod.copr.build.end"
@@ -268,6 +269,7 @@ class CoprBuildEndHandler(FedmsgHandler):
 
 @add_topic
 @use_for(job_type=JobType.copr_build)
+@use_for(job_type=JobType.build)
 @required_by(job_type=JobType.tests)
 class CoprBuildStartHandler(FedmsgHandler):
     topic = "org.fedoraproject.prod.copr.build.start"
--- a/packit_service/worker/jobs.py
+++ b/packit_service/worker/jobs.py
@@ -23,10 +23,15 @@
 """
 We love you, Steve Jobs.
 """
+import datetime
 import logging
-from typing import Optional, Dict, Union, Type, Set, List, Any
+from copy import deepcopy
+from typing import Any
+from typing import Optional, Dict, Union, Type, Set, List
 
 from packit.config import JobType, PackageConfig, JobConfig
+from packit.constants import DATETIME_FORMAT
+
 from packit_service.config import ServiceConfig
 from packit_service.log_versions import log_job_versions
 from packit_service.models import PullRequestModel
@@ -100,7 +105,7 @@ def get_handlers_for_event(
 
 
 def get_config_for_handler_kls(
-    handler_kls: Type[JobHandler], event: Event, package_config: PackageConfig
+    handler_kls: Type[Handler], event: Event, package_config: PackageConfig
 ) -> List[JobConfig]:
     """
     Get a JobConfig relevant to event and the handler class.
@@ -179,6 +184,8 @@ class SteveJobs:
             logger.warning(f"There is no handler for {event.trigger} event.")
             return handlers_results
 
+        pristine_package_config = deepcopy(event.package_config.__dict__)
+
         for handler_kls in handler_classes:
             job_configs = get_config_for_handler_kls(
                 handler_kls=handler_kls,
@@ -206,8 +213,16 @@ class SteveJobs:
                     config=self.config, job_config=job_config, event=event
                 )
                 if handler.pre_check():
-                    handlers_results[job_config.type.value] = handler.run_n_clean()
-            # don't break here, other handlers may react to the same event
+                    event.package_config.set_job_overrides(job_config)
+                    # we can have multiple job definitions with the same type
+
+                    current_time = datetime.datetime.now().strftime(DATETIME_FORMAT)
+                    result_key = f"{job_config.type.value}-{current_time}"
+                    handlers_results[result_key] = handler.run_n_clean()
+                    # once the handler is done, we need to reset package_config
+                    # to the former content because not every job may override the same values
+                    event.package_config.__dict__ = deepcopy(pristine_package_config)
+                # we want to run handlers for all possible jobs, not just the first one
 
         return handlers_results
 
@@ -247,7 +262,7 @@ class SteveJobs:
             PullRequestCommentPagureEvent,
             IssueCommentEvent,
         ],
-    ) -> HandlerResults:
+    ) -> Dict[str, HandlerResults]:
 
         msg = f"comment '{event.comment}'"
         packit_command, pr_comment_error_msg = self.find_packit_command(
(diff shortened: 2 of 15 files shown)


csomh:
Better commit message, please, please, please :stuck_out_tongue_closed_eyes:

TomasTomecek:
> Better commit message, please, please, please stuck_out_tongue_closed_eyes

How?

csomh:
> > Better commit message, please, please, please stuck_out_tongue_closed_eyes
> 
> How?

For example: https://chris.beams.io/posts/git-commit/

What I'm missing here an explanation for why this change is happening.

TomasTomecek:
> Works for me.

I know the solution is not very pretty, but this is the simplest & elegant solution I was able to come up with.

packit_service/worker/handlers/fedmsg_handlers.py
```diff
@@ -143,6 +143,7 @@ def run(self) -> HandlerResults:
 
 @add_topic
 @use_for(job_type=JobType.copr_build)
+@use_for(job_type=JobType.build)
```
lachmanfrantisek, line 146: So this will solve #650?

packit_service/worker/jobs.py
```diff
@@ -206,8 +213,16 @@ def process_jobs(self, event: Event) -> Dict[str, HandlerResults]:
...
-            # don't break here, other handlers may react to the same event
+                    event.package_config.set_job_overrides(job_config)
+                    # we can have multiple job definitions with the same type
+
+                    current_time = datetime.datetime.now().strftime(DATETIME_FORMAT)
+                    result_key = f"{job_config.type.value}-{current_time}"
+                    handlers_results[result_key] = handler.run_n_clean()
+                    # once the handler is done, we need to reset package_config
+                    # to the former content because not every job may override the same values
+                    event.package_config.__dict__ = deepcopy(pristine_package_config)
```
lachmanfrantisek, line 224: Would be nice to avoid reverting and don't touch the event's config from handlers.

I know it's a completely different way, but what about this:

- Each handler will work only with `JobConfig`.
- JobConfig will have the same attributes as the `Config`. Each value can be set on the job level but will use the global one if not set.

packit_service/worker/handlers/fedmsg_handlers.py
lachmanfrantisek: OK, that's a different problem.

TomasTomecek: tests brought this up - jobs were not being matched b/c of JobType.build vs. JobType.copr_build

packit_service/worker/jobs.py
TomasTomecek: You are asking me to rewrite the whole thing from scratch. I understand the proposed implementation is not ideal, but is the simplest & most efficient way I was able to come up with.

Honestly, I don't understand the benefits your solution brings, aside from a ton of refactoring in p-s and adding support for such a thing in packit.

TomasTomecek:
>     * I am not against this implementation but I am worried that this will bring us a lot of headaches when debugging.

We already have those, to some extent. Luckily we have very good tests, sentry and logging so it's not that bad. So please clarify what exactly is troubling you.

packit_service/worker/jobs.py
lachmanfrantisek: > You are asking me to rewrite the whole thing from scratch.

No, it's either too late or outside of the label-friday scope.

> Honestly, I don't understand the benefits your solution brings, aside from a ton of refactoring in p-s and adding support for such a thing in packit.

- :heavy_plus_sign: cleaner work with the config from handlers
  - You have only one object with all needed info. That can help with the task refactor.
  - You can easily see what value comes from what place.
- :heavy_plus_sign: transparent, you can match `package_config` object with the one from the file (unlike this one) anytime.
- :heavy_plus_sign: overwritting possible without the need for dedicated field => better UX
- :heavy_plus_sign: sharing common values across multiple jobs (e.g. `targets`).
- :heavy_minus_sign: a lot of work to make it properly

It's not doable short term => It's much better to have your implementation than nothing.

lachmanfrantisek:
> We already have those, to some extent. Luckily we have very good tests, sentry and logging so it's not that bad. So please clarify what exactly is troubling you.

It looks a bit hacky. Few concerns:
- You can't see the original config from any place => hard to match the file content in upstream.
- We need to do this rewrite-reverse magic each time we want to use it.
  - We can forget to apply it. We can forget to revert.

But as I wrote above, It works and it's ready now -- so, I am not against.

csomh:
> > An approach like this will cause us trouble in the future and slow us down.
> 
> Without objective arguments this sentence has zero value to me.

Okay, fair enough, let me elaborate.

These concerns are analogous to the issue you filed for having to do schema changes in two places (packit-service/packit#744).

In this PR *some* config manipulation is happening in `process_jobs()`, namely part of the override functionality is implemented here. But this method, just by its name is not meant for that. So this will result in the following:

- confusion (extremely likely): anyone reading this code will get confused why override (configuration related) functionality is present here. Anyone reviewing changes in this part of code will get confused why override functionality is present here.
- unexpected test failures (less likely, for now): changes introduced in configuration might lead to troubles in the config manipulation in `process_jobs()`. A failure like this is unlikely to be an expected one by any contributor: they just wanted to do a change in config, so changed config related code, and now this other (seemingly unrelated part of) code is failing. Fixing this failure will take additional time.
- design rigidity (100%): refactoring (breaking up) `process_jobs()` becomes harder, b/c one not only hast to account for the base functionality of the method but also for this additional piece of unrelated functionality. Refactoring configuration handling becomes harder, as the functionality is spread around unrelated modules.
- harder debugging (extremely likely): for config related issues suddenly one has to look at multiple places.

I see all of the above as a considerable risk which will lead to more work required to do changes, more difficult to figure out issues and ultimately causing frustration and at some point leading to some refactor work (which at that point is going to be much more difficult b/c of all the things which get build on top of this).

lachmanfrantisek:
> Now that I cooled off a bit I wouldn't say it's gonna be that insane. On packit's side, we'd ideally reuse definitions of PackageConfig and jammed them in JobConfig somehow - this is how I started but it turned to be moving too many moving pieces, not that much of work really. I should have been more patient I guess.

Sorry. I originally meant that as a wild idea for future..;) (Since It looked like if we didn't merge this one, we would not have anything in the next year...;-)

That would be really nice if you can make in that way!

> I'm sorry but I don't understand. Do you mean the "overrides" field?

:+1:

>> * heavy_plus_sign sharing common values across multiple jobs (e.g. `targets`).
>
> I don't follow how this is related to the PR.

The counter-idea makes this easily possible. But it will not be so hard also with this implementation.

TomasTomecek:
> Sorry. I originally meant that as a wild idea for future..;) (Since It looked like if we didn't merge this one, we would not have anything in the next year...;-)

yup, we won't probably





> But this method, just by its name is not meant for that.

:roll_eyes:

The code can be reorganized and put into functions with proper names if needed.

>     * confusion (extremely likely): anyone reading this code will get confused why override (configuration related) functionality is present here. Anyone reviewing changes in this part of code will get confused why override functionality is present here.

"Anyone reviewing changes ... will get confused ..." - how can you make a statement like that - can you read people's minds?

I understand the implementation is pretty wonky - but in total it's 9 lines. Once you see all the lines (30 seconds of time?) and read associated comments I'd say it's pretty clear what's happening and why it's happening.

Since @lbarcziova +1'd this - is the implementation of overrides really so confusing and worrying for you? Or would you be able to live with it and accept the risks Hunor is talking about?

>     * unexpected test failures (less likely, for now): changes introduced in configuration might lead to troubles in the config manipulation in `process_jobs()`. A failure like this is unlikely to be an expected one by any contributor: they just wanted to do a change in config, so changed config related code, and now this other (seemingly unrelated part of) code is failing. Fixing this failure will take additional time.

Yes, fixing failures take time, welcome to software engineering. But this is literally our job: write code, make tests, track down bugs, resolve them. And if you look at how complex the project is at this point, I don't understand how the override mechanics is adding such a step-up complexity.

>     * design rigidity (100%): refactoring (breaking up) `process_jobs()` becomes harder, b/c one not only hast to account for the base functionality of the method but also for this additional piece of unrelated functionality. Refactoring configuration handling becomes harder, as the functionality is spread around unrelated modules.

This is a good point.

>     * harder debugging (extremely likely): for config related issues suddenly one has to look at multiple places.

Which is already the case: we have config code in p-s and packit repositories - how is this PR making it any different?

Also, this is why we have tests and IDEs to help us.

> I see all of the above as a considerable risk which will lead to more work required to do changes, more difficult to figure out issues and ultimately causing frustration and at some point leading to some refactor work (which at that point is going to be much more difficult b/c of all the things which get build on top of this).

Fair, your decision, hence I closed this PR and proposed reverting the overrides changes - I don't have the bandwidth doing it your way.

As Franta pointed out, it's either this PR or no PR (from me).

PR #641, opened 2020-05-29
Restructure the log pages

csomh:
Log pages for Copr builds had a lazy structure, often causing confusion
for our users. There were cases when it was not obvious for the user if
the SRPM or the Copr build failed.

Improve this by linking SRPM build logs instead of including them. Also
use Flask templating instead of returning raw HTML and add a touch of a
style.


packit_service/service/templates/build_logs.html
```diff
@@ -0,0 +1,21 @@
+{% extends "layout.html" %}
+<!-- prettier-ignore -->
```
lachmanfrantisek, line 2: What is the problem with `prettier` here?

Wouldn't be easier to ignore the whole file instead?

packit_service/service/views.py
```diff
@@ -72,35 +65,23 @@ def _get_build_logs_for_build(
...
-        f"Submitted: {models.optional_time(build.build_submitted_time)}<br>"
-        f"State: {build.status}<br><br>"
-        f'Build web interface URL: <a href="{build.web_url}">{build.web_url}</a><br>'
+    return render_template(
+        "build_logs.html",
+        project=project,
+        title_identifier=title_identifier,
+        build_description=build_description,
+        build=build,
+        build_submitted_time=models.optional_time(build.build_submitted_time),
```
lachmanfrantisek, line 74: I am missing this info in the templates.

packit_service/service/views.py
```diff
@@ -72,35 +65,23 @@ def _get_build_logs_for_build(
...
-    response = (
-        "<html><head>"
-        f"<title>{build_description} {project.namespace}/{project.repo_name}:"
-        f" {title_identifier}</title></head><body>"
-        f"{build_description} ID: {build.build_id}<br>"
-        f"Submitted: {models.optional_time(build.build_submitted_time)}<br>"
-        f"State: {build.status}<br><br>"
-        f'Build web interface URL: <a href="{build.web_url}">{build.web_url}</a><br>'
+    return render_template(
+        "build_logs.html",
```
TomasTomecek, line 69: With this change it seems to me this view is now build_info rather than build_logs - would it make sense to rename it?

packit_service/service/templates/build_logs.html
csomh: Prettier is missing [Jinja Template](https://github.com/prettier/prettier/issues/5754) support. These ignore lines disable Prettier until the next HTML-tag to keep the template in a readable form.

It's a pain, but I wouldn't disable Prettier for the whole file, as there is still some value in uniform formatting for the parts which are not ignored.

packit_service/service/views.py
csomh: Oh, yes, I forgot about this...

The reason it's not in the template is b/c I couldn't come up with a good reason why it would be useful for the users to be there. The information is already on the Copr-build page we link, and that page gives more context on the build. So I just didn't see the point in a lonely build submitted timestamp on our page :)

My idea is to keep our log pages as minimal, clean and low-maintenance as possible.

Should we keep this timestamp? If no, I'll clean up here, if yes, I'll add to the template.

packit_service/service/views.py
csomh: Definitely :+1:

csomh: This requires changing some parts of the routing too, to make it consistent all over the place.

Next Friday.

TomasTomecek: or we can merge this as it is, and do the change as a new PR next friday :)

packit_service/service/templates/build_logs.html
lachmanfrantisek: > These ignore lines disable Prettier until the next HTML-tag to keep the template in a readable form.

I know and that was my point if it makes sense to use it for templates.

packit_service/service/views.py
lachmanfrantisek: - I -- as a user -- would not like to click multiple times to get that basic information.
  - And in copr, you don't have the reverse link to get to our build page.
- This page should be accessible in long-term -- github-status can change and copr as well. (Outage, removed temporary project,...)
- You can easily compare multiple builds for one PR.
- Having basic information can help us solve potential problems more easily.
  - Users can send us this link => The content is stable and we can read everything needed. (That relates to the second one: PR can change, copr can change.)
  - That's why I added that.
- I would try to use more styling instead of data cleanup to make the page cleaner.

packit_service/service/templates/build_logs.html
csomh: > I know and that was my point if it makes sense to use it for templates.

If you knew, why did you ask what the problem is? :confused: 

As I said above, I still see value in keeping Prettier enabled by default, and disable it only when it would otherwise make things very ugly.

lachmanfrantisek: > If you knew, why did you ask what the problem is? confused

Sorry, I haven't read your comment properly and my comment does not make sense also to me... My bad.

> As I said above, I still see value in keeping Prettier enabled by default, and disable it only when it would otherwise make things very ugly.

:+1: That's what I was interested in. Thanks for the explanation.

packit_service/service/views.py
csomh: Here is proposal to make this consistent all over in `../service/views.py`:

- rename `build_logs.html` to `build_info.html` as you suggested
- rename and redirect routes from `/copr-build/<id>/logs` to `/copr-build/<id>`. The same for `/koji-build`
- rename function names to match the above

What do you think?

TomasTomecek: SGTM

csomh:
@TomasTomecek @lachmanfrantisek please take another look!

Changes:
- resized the logo to have a smaller size (20K -> 6K)
- Added the time stamp when builds for individual targets are started
- renamed `build_logs` to `build_info` everywhere where appropriate
- updated requre tests

Additionally when a build id is not found, we now return 404 together with the message.

PR #643, opened 2020-06-01
Use list for copr targets

lachmanfrantisek:
- Use list for copr targets.
- Type annotations added to the packit code: https://github.com/packit-service/packit/pull/849


packit_service/worker/build/copr_build.py
```diff
@@ -218,7 +218,7 @@ def run_build(
 
         self.api.copr_helper.create_copr_project_if_not_exists(
             project=self.job_project,
-            chroots=self.build_targets,
+            chroots=list(self.build_targets),
```
csomh, line 221: Why wasn't this issue caught prior to going to prod? Shouldn't this be the kind of thing to be noticed by mypy?

PR #648, opened 2020-06-02
Fix get_config_for_handler_kls

lachmanfrantisek:
- Return list of matching job_configs instead of the first one.

### DIFF
--- a/packit_service/worker/jobs.py
+++ b/packit_service/worker/jobs.py
@@ -24,12 +24,12 @@
 We love you, Steve Jobs.
 """
 import logging
-from typing import Optional, Dict, Union, Type, Set
+from typing import Optional, Dict, Union, Type, Set, List
 
 from packit.config import JobType, PackageConfig, JobConfig
 from packit_service.config import ServiceConfig
-from packit_service.models import PullRequestModel
 from packit_service.log_versions import log_job_versions
+from packit_service.models import PullRequestModel
 from packit_service.service.events import (
     PullRequestCommentGithubEvent,
     IssueCommentEvent,
@@ -98,7 +98,7 @@ def get_handlers_for_event(
 
 def get_config_for_handler_kls(
     handler_kls: Type[JobHandler], event: Event, package_config: PackageConfig
-) -> Optional[JobConfig]:
+) -> List[JobConfig]:
     """
     Get a JobConfig relevant to event and the handler class.
 
@@ -107,22 +107,36 @@ def get_config_for_handler_kls(
     :param package_config: we pick the JobConfig from this package_config instance
     :return: JobConfig
     """
-    for job in package_config.jobs:
-        if job.type in MAP_HANDLER_TO_JOB_TYPES[handler_kls] and (
-            (
-                event.db_trigger
-                and event.db_trigger.job_config_trigger_type == job.trigger
-            )
-            or is_trigger_matching_job_config(trigger=event.trigger, job_config=job)
-        ):
-            return job
+    matching_jobs = []
+    jobs_that_can_be_triggered = []
 
-        required_handlers = MAP_REQUIRED_JOB_TO_HANDLERS[job.type]
-        for pos_handler in required_handlers:
+    for job in package_config.jobs:
+        if (
+            event.db_trigger and event.db_trigger.job_config_trigger_type == job.trigger
+        ) or is_trigger_matching_job_config(trigger=event.trigger, job_config=job):
+            jobs_that_can_be_triggered.append(job)
+
+    matching_job_types = MAP_HANDLER_TO_JOB_TYPES[handler_kls]
+    for job in jobs_that_can_be_triggered:
+        if job.type in matching_job_types and job not in matching_jobs:
+            matching_jobs.append(job)
+
+    if matching_jobs:
+        return matching_jobs
+
+    # The job was not configured but let's try required ones.
+    # e.g. we can use `tests` configuration when running build
+    for job in jobs_that_can_be_triggered:
+        matching_required_job_types = MAP_REQUIRED_JOB_TO_HANDLERS[job.type]
+        for pos_handler in matching_required_job_types:
             for trigger in pos_handler.triggers:
-                if trigger == event.trigger:
-                    return job
-    return None
+                if (
+                    is_trigger_matching_job_config(trigger=trigger, job_config=job)
+                    and job not in matching_jobs
+                ):
+                    matching_jobs.append(job)
+
+    return matching_jobs
 
 
 class SteveJobs:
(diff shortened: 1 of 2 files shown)


packit_service/worker/jobs.py
```diff
@@ -107,22 +107,36 @@ def get_config_for_handler_kls(
     :param package_config: we pick the JobConfig from this package_config instance
     :return: JobConfig
     """
+    matching_jobs = []
```
jpopelka, line 110:
```suggestion
    matching_jobs = set()
```

packit_service/worker/jobs.py
```diff
@@ -107,22 +107,36 @@ def get_config_for_handler_kls(
...
+            event.db_trigger and event.db_trigger.job_config_trigger_type == job.trigger
+        ) or is_trigger_matching_job_config(trigger=event.trigger, job_config=job):
+            jobs_that_can_be_triggered.append(job)
 
-        required_handlers = MAP_REQUIRED_JOB_TO_HANDLERS[job.type]
-        for pos_handler in required_handlers:
+    matching_job_types = MAP_HANDLER_TO_JOB_TYPES[handler_kls]
+    for job in jobs_that_can_be_triggered:
+        if job.type in matching_job_types and job not in matching_jobs:
+            matching_jobs.append(job)
```
jpopelka, line 122:
```suggestion
matching_jobs = {job for job in jobs_that_can_be_triggered if job.type in matching_job_types}
```

packit_service/worker/jobs.py
```diff
@@ -98,7 +98,7 @@ def get_handlers_for_event(
 
 def get_config_for_handler_kls(
     handler_kls: Type[JobHandler], event: Event, package_config: PackageConfig
-) -> Optional[JobConfig]:
+) -> List[JobConfig]:
```
jpopelka, line 101:
```suggestion
) -> Set[JobConfig]:
```

packit_service/worker/jobs.py
```diff
@@ -107,22 +107,36 @@ def get_config_for_handler_kls(
...
+        for pos_handler in matching_required_job_types:
             for trigger in pos_handler.triggers:
-                if trigger == event.trigger:
-                    return job
-    return None
+                if (
+                    is_trigger_matching_job_config(trigger=trigger, job_config=job)
+                    and job not in matching_jobs
+                ):
+                    matching_jobs.append(job)
```
jpopelka, line 137:
```suggestion
                if is_trigger_matching_job_config(trigger=trigger, job_config=job):
                    matching_jobs.add(job)
```

packit_service/worker/jobs.py
lachmanfrantisek: I already have the set there, but I am not sure if we don't want to preserve order. Can this be useful? Now, I can't remember the usecase for that...;)

packit_service/worker/jobs.py
TomasTomecek: This won't work b/c job is not hashable (been there, done that)

TomasTomecek: ...so Franta would need to put job.type in the set, not the job itself

or fix this in packit first

lachmanfrantisek: Can we go with the current solution then?

TomasTomecek: fine with me

jpopelka: go ahead

lachmanfrantisek: OK I'll add some more test cases and merge.

packit_service/worker/jobs.py
```diff
@@ -98,7 +98,7 @@ def get_handlers_for_event(
 
 def get_config_for_handler_kls(
     handler_kls: Type[JobHandler], event: Event, package_config: PackageConfig
-) -> Optional[JobConfig]:
+) -> List[JobConfig]:
     """
     Get a JobConfig relevant to event and the handler class.
```
csomh, line 103: This is not "a JobConfig" but multiple ones. Please update the doc-string to reflect the change in the return type and explain whether the order of the list has any meaning or not.

packit_service/worker/jobs.py
csomh, line 110: This variable is used ~10 lines bellow for the first time. How about declaring it just before that for-loop?

lachmanfrantisek: I don't see much value in moving this one. This is the one used across the whole function and is returned and e.g. `matching_job_types` and `matching_required_job_types` are created before the loop where they are used. I treat those a bit differently.

packit_service/worker/jobs.py
lachmanfrantisek: > This is not "a JobConfig" but multiple ones.

Thanks, I'll update that.

> Please update the doc-string to reflect the change in the return type and explain whether the order of the list has any meaning or not.

It's a list => ordering has meaning. (I am using `set` where the order is not important.) I'll make a note to the doc as well.

lachmanfrantisek:
> Sorry for being late to the party and ruining it. Could you please explain what the actual problem was that this PR tries to fix?

@csomh The function didn't return what it was supposed to return. And this PR is fixing that.

- return list instead of one value
- get the correct values (that's it; I don't think we need to describe each weird combination that was fixed)

I know that I am a better programmer then English writer but other two team members did not have any problem with the understanding so it was probably not so misleading. I am doing my best to make everything as clear as possible, but I would like to spend reviews on the implementation and not on the commit messages and documentation. (And since we do agile: *Working software over comprehensive documentation*..;-)

csomh:
> > Sorry for being late to the party and ruining it. Could you please explain what the actual problem was that this PR tries to fix?
> 
> @csomh The function didn't return what it was supposed to return. And this PR is fixing that.

What was the function supposed to return? Where did this expectation come from? How should I have found out about what the correct expectation was?

> 
> * return list instead of one value

Why?

> * get the correct values (that's it; I don't think we need to describe each weird combination that was fixed)

Is it possible describe what "correct values" are?

> 
> I know that I am a better programmer then English writer but other two team members did not have any problem with the understanding so it was probably not so misleading. 

Probably they are much smarter than I am. Good for them.

> I am doing my best to make everything as clear as possible, but I would like to spend reviews more on the implementation than on the commit messages and documentation. (And since we do agile: _Working software over comprehensive documentation_..;-)

I don't think an ask to clearly explain the problem when submitting a PR is that much of a thing. And I also don't think it would require extensive knowledge of English. You were able to figure out the issue and come up with a solution, so why wouldn't you share your findings instead of pulling folks into a guess-game about what was wrong? If folks don't have to second-guess the problem, you might even receive more meaningful feedback during the review.

This is communication 101, which is far from "comprehensive documentation".

lachmanfrantisek:
As we discussed elsewhere, I thought you actually understood it and was asking for more clear commit-messages and docs.

I've sent a new PR with more precise docstrings for this method and the `get_handlers_for_event` one (https://github.com/packit-service/packit-service/pull/652) to help this a bit.

PR #652, opened 2020-06-03
Document get_handlers_for_event and get_config_for_handler_kls

lachmanfrantisek:
- Improve docstrings for get_handlers_for_event and get_config_for_handler_kls.

### DIFF
--- a/packit_service/worker/jobs.py
+++ b/packit_service/worker/jobs.py
@@ -76,9 +76,16 @@ def get_handlers_for_event(
     """
     Get all handlers that we need to run for the given event.
 
+    We need to return all handler classes that:
+    - can react to the given event AND
+    - are configured in the package_config (either directly or as a required job)
+
+    Examples of the matching can be found in the tests:
+    ./tests/unit/test_jobs.py:test_get_handlers_for_event
+
     :param event: event which we are reacting to
     :param package_config: for checking configured jobs
-    :return: set of handler instances
+    :return: set of handler instances that we need to run for given event and user configuration
     """
     handlers: Set[Type[JobHandler]] = set()
     classes_for_trigger = MAP_EVENT_TRIGGER_TO_HANDLERS[event.trigger]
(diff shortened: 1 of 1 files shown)


lachmanfrantisek:
recheck

packit_service/worker/jobs.py
```diff
@@ -76,9 +76,16 @@ def get_handlers_for_event(
...
+    - can react to the given event
+    - are configured in the package_config (either directly or as a required job)
+
+    Examples of the matching can be found in the tests:
+    ./tests/unit/test_jobs.py:test_get_handlers_for_event
+
     :param event: event which we are reacting to
     :param package_config: for checking configured jobs
-    :return: set of handler instances
+    :return: set of handler instances that we need to run for given event and user's configuration
```
csomh, line 88:
```suggestion
    :return: set of handler instances that we need to run for given event and user configuration
```

packit_service/worker/jobs.py
```diff
@@ -76,9 +76,16 @@ def get_handlers_for_event(
     """
     Get all handlers that we need to run for the given event.
 
+    We need to return all handler classes that:
+    - can react to the given event
+    - are configured in the package_config (either directly or as a required job)
```
csomh, line 81: Make it explicit that the relationship between these two is 'or':

```suggestion
    We need to return all handler classes that either:
    - can react to the given event OR
    - are configured in the package_config (either directly or as a required job)
```

packit_service/worker/jobs.py
```diff
@@ -103,12 +113,23 @@ def get_config_for_handler_kls(
     handler_kls: Type[JobHandler], event: Event, package_config: PackageConfig
 ) -> List[JobConfig]:
     """
-    Get a JobConfig relevant to event and the handler class.
+    Get a list of JobConfigs relevant to event and the handler class.
+
+    We need to find all job configurations that:
+    - can be run by the given handler class
+    - that matches the trigger of the event
```
csomh, line 120: Make it explicit that this is 'or':

```suggestion
    We need to find all job configurations that either:
    - can be run by the given handler class OR
    - that matches the trigger of the event
```

packit_service/worker/jobs.py
lachmanfrantisek: Actually, there should  be `AND`..;-)

I'll make it clear.

csomh: I inferred the 'or' from 

```python
for job in package_config.jobs:
        if (
            event.db_trigger and event.db_trigger.job_config_trigger_type == job.trigger
        ) or is_trigger_matching_job_config(trigger=event.trigger, job_config=job):
```

What did I miss?

lachmanfrantisek: But we are going only through the configured jobs.

Those two checks are both about triggers.

csomh: Duh! Of course. Thanks for clarifying.

PR #653, opened 2020-06-03
Do not search for comments when no pr_id

lachmanfrantisek:
- Rename `was_last_build_successful` and don't fail when no pr_id.
- Fixes https://github.com/packit-service/packit-service/issues/646 but probably not the real cause.


lachmanfrantisek:
I am going to merge manually since regular tests passed.

Thanks for the review.

PR #654, opened 2020-06-03
Set job_config for babysit task

lachmanfrantisek:
- Set the matching job_config when running babysit task for copr_build.
- Fixes the real cause of https://github.com/packit-service/packit-service/issues/646

TODO:
- [x] Tests
- [x] Check why BuildHelper does not calculate the job configs properly and if we cannot use `get_config_for_handler_kls` for that.
  - `get_config_for_handler_kls` uses `handler` => we can't use it
  - It's better to se correct values `BuildHelper` property `job_build` from outside. (The better solution cal be done after `JobType` cleanup.)

### DIFF
--- a/packit_service/worker/jobs.py
+++ b/packit_service/worker/jobs.py
@@ -150,6 +150,9 @@ def get_config_for_handler_kls(
     matching_job_types = MAP_HANDLER_TO_JOB_TYPES[handler_kls]
     for job in jobs_that_can_be_triggered:
         if (
+            # Check if the job matches any job supported by the handler.
+            # The function `are_job_types_same` is used
+            # because of the `build` x `copr_build` aliasing.
             any(are_job_types_same(job.type, type) for type in matching_job_types)
             and job not in matching_jobs
         ):
--- a/packit_service/worker/jobs.py
+++ b/packit_service/worker/jobs.py
@@ -41,7 +41,10 @@ from packit_service.service.events import (
     TheJobTriggerType,
     PullRequestCommentPagureEvent,
 )
-from packit_service.trigger_mapping import is_trigger_matching_job_config
+from packit_service.trigger_mapping import (
+    is_trigger_matching_job_config,
+    are_job_types_same,
+)
 from packit_service.worker.handlers import (
     CoprBuildEndHandler,
     CoprBuildStartHandler,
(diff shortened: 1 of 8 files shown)


lachmanfrantisek:
@TomasTomecek I've added some (a little bit lazy) tests, PTAL

tests/integration/test_babysit.py
```diff
@@ -0,0 +1,127 @@
...
+                            repo_name="repo_name",
+                            namespace="the-namespace",
+                            project_url="https://github.com/the-namespace/repo_name",
+                        ),
+                        pr_id=5,
+                        job_config_trigger_type=JobConfigTriggerType.pull_request,
+                    )
+                )
+                .mock(),
+            )
```
TomasTomecek, line 96: you sure are adventurous, I didn't even know you can do this with flexmock

packit_service/worker/jobs.py
```diff
@@ -142,7 +145,10 @@ def get_config_for_handler_kls(
 
     matching_job_types = MAP_HANDLER_TO_JOB_TYPES[handler_kls]
     for job in jobs_that_can_be_triggered:
-        if job.type in matching_job_types and job not in matching_jobs:
+        if (
+            any(are_job_types_same(job.type, type) for type in matching_job_types)
```
TomasTomecek, line 149: this is starting to be more and more complicated: if you still have this particular change in your mind, could you please comment in the code here what's happening?

lachmanfrantisek: Still same story about `build` x `copr_build`...;( I'll make a note here.

tests/integration/test_babysit.py
lachmanfrantisek: Yes, but I would not consider this as a nice and clean test-case. (But don't want to spend too much time on it now.)

packit_service/worker/jobs.py
lachmanfrantisek: > this is starting to be more and more complicated

Sadly, yes. Those aliases and requires don't help us much. I hope we can refactor this once we clear the enums a bit.

lachmanfrantisek: Comment added.

lachmanfrantisek:
The tests should be fixed now. (My global config fixture discovered some problems.)

@TomasTomecek PTAL

lachmanfrantisek:
Last time... Now, it does not check the urls generated by server since it can change in absence/presence of the port.

lachmanfrantisek:
@TomasTomecek finally, the tests are passing.

Can we merge and move the `stable` branch, please?

PR #657, opened 2020-06-03
Koji reporing

lachmanfrantisek:
- [x] Add KojiBuildEvent class for the messages about start/stop of the koji build.
- [x] Parse the koji events.
- [x] Unit tests for parsing.
- [x] Database tests for parsing.
- [x] Add other info to the events
- [x] Implement handlers
- [x] test for handlers
----

- Requires: https://github.com/packit-service/packit-service-fedmsg/pull/13
~~- Handlers will be done in the next PR.~~

### DIFF
--- a/packit_service/service/events.py
+++ b/packit_service/service/events.py
@@ -263,7 +263,10 @@ class AbstractForgeIndependentEvent(Event):
     def pr_id(self) -> Optional[int]:
         return self._pr_id
 
-    def get_project(self) -> GitProject:
+    def get_project(self) -> Optional[GitProject]:
+        if not self.project_url and not self.db_trigger:
+            return None
+
         return ServiceConfig.get_service_config().get_project(
             url=self.project_url or self.db_trigger.project.project_url
         )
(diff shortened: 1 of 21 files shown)


packit_service/service/events.py
```diff
@@ -786,6 +889,25 @@ def get_copr_build_logs_url(event: CoprBuildEvent) -> str:
     )
 
 
+def get_koji_build_logs_url(event: KojiBuildEvent) -> Optional[str]:
+    if not event.rpm_build_task_id:
+        return None
+
+    return (
+        f"https://kojipkgs.fedoraproject.org//work/tasks/"
```
TomasTomecek, line 897: we should soon prioritize the distinction b/w koji instances so this would work for centos as well

packit_service/trigger_mapping.py
```diff
@@ -40,6 +41,11 @@
 }
 
 
+@deprecated(
+    reason="Should be avoided since it is hardcoded. "
+    "We can use info from database instead for events without the real trigger "
+    "(e.g. copr/koji/tests results)."
+)
 def is_trigger_matching_job_config(
```
TomasTomecek, line 49: would be nice to drop this function in future then

lachmanfrantisek:
I've added tests and fixed few problems.

PR #660, opened 2020-06-03
process properly multiple jobs with the same type

TomasTomecek:
This is a follow-up patch to #648 to address the problem when the
project defines multiple jobs with the same type. We address several
issues in here:

* comment jobs are processed correctly (only a first one was processed
  before)
* HandlerResults are not being overwritten with the same job type
* all build-related handlers have aliasing (build/copr-build) set up

### DIFF
--- a/tests/spellbook.py
+++ b/tests/spellbook.py
@@ -24,8 +24,12 @@
 A book with our finest spells
 """
 from pathlib import Path
-
+from typing import Any
 
 TESTS_DIR = Path(__file__).parent
 DATA_DIR = TESTS_DIR / "data"
 SAVED_HTTPD_REQS = DATA_DIR / "http-requests"
+
+
+def first_dict_value(a_dict: dict) -> Any:
+    return a_dict[next(iter(a_dict))]
--- a/tests/unit/test_steve.py
+++ b/tests/unit/test_steve.py
@@ -38,6 +38,7 @@ from packit_service.constants import SANDCASTLE_WORK_DIR
 from packit_service.service.db_triggers import AddReleaseDbTrigger
 from packit_service.worker.jobs import SteveJobs
 from packit_service.worker.whitelist import Whitelist
+from tests.spellbook import first_dict_value
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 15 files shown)


tests/spellbook.py
```diff
@@ -24,8 +24,12 @@
...
-
+from typing import Any
 
 TESTS_DIR = Path(__file__).parent
 DATA_DIR = TESTS_DIR / "data"
 SAVED_HTTPD_REQS = DATA_DIR / "http-requests"
+
+
+def first_dict_value(a_dict: dict) -> Any:
+    return list(a_dict.values())[0]
```
lachmanfrantisek, line 35:
```suggestion
    return next(iter(a_dict))
```

... I was interested if this can be done in a better way..;)
https://www.geeksforgeeks.org/python-get-the-first-key-in-dictionary/

lachmanfrantisek, line 35: I was really interested if this can be done in a better way, but [this](https://www.geeksforgeeks.org/python-get-the-first-key-in-dictionary/) speaks only about keys...

This may be more efficient:
```suggestion
def first_dict_value(a_dict: dict) -> Any:
    return a_dict[next(iter(a_dict))]
```

TomasTomecek: that's right, definitely more efficient since we don't need to construct the list actually, thank you

tests/unit/test_steve.py
```diff
@@ -85,5 +86,7 @@ def test_process_message(event):
     )
     flexmock(Whitelist, check_and_report=True)
     results = SteveJobs().process_message(event)
-    assert "propose_downstream" in results["jobs"]
+    j = first_dict_value(results["jobs"])
+    assert "propose_downstream" in list(results["jobs"].keys())[0]
```
lachmanfrantisek, line 90: Small nit-pick:
```suggestion
    assert "propose_downstream" in next(iter(results["jobs"]))
```

TomasTomecek:
@_@

lachmanfrantisek:
> @_@

Sorry, how I could I overtake you? Your job started much sooner.

TomasTomecek:
> > @_@
> 
> Sorry, how I could I overtake you? Your job started much sooner.

your code was more fancy so zuul prioritized it

PR #661, opened 2020-06-04
[files/install-deps(-worker).yaml] install python3-celery & python3-redis

jpopelka:
They'd be pip installed anyway, but better be explicit.

Fixes https://sentry.io/organizations/red-hat-0p/issues/1707007212


sakalosj:
LGTM

PR #666, opened 2020-06-05
Add global service config

lachmanfrantisek:
- Add global auto-used fixtures to overwrite service config and don't use the local one.


lachmanfrantisek:
@jpopelka PTAL

Issue #633, opened 2020-05-26
'GET /api/copr-builds' yields nothing in prod


TomasTomecek:
```
tt at hops ~ curl -s 'https://prod.packit.dev/api/copr-builds?page=1&per_page=50'
tt at hops ~ curl -s 'https://prod.packit.dev/api/copr-builds'
tt at hops ~   
```

is there a bug in the API code?

csomh:
Looking at the logs, it seems that the process (or thread) crashed and then is restarted whenever the I call this endpoint.

The OpenShift monitoring graphs show a peak in CPU whenever this happens.

Are resources enough?

lachmanfrantisek:
The call is more complicated since it tries to merge all chroots:

https://github.com/packit-service/packit-service/blob/dcc01a0ab55a1a4caa63be9522bc9257ddcbe0bb/packit_service/service/api/copr_builds.py#L56-L89

csomh:
Maybe we shouldn't load all 22000+ copr builds into memory at once and have something line `CoprBuildModel.get(first, last)` instead?

IceWreck:
So instead of doing pagination provided by Flask RestX, which loads everything into memory and then serves info about 10 or 20 builds at a time, we should fetch those 10 or 20 from the DB.

csomh:
> So instead of doing pagination provided by Flask RestX, which loads everything into memory and then serves info about 10 or 20 builds at a time, we should fetch those 10 or 20 from the DB ?

Yes. Afaik, the data loaded is not cached between API calls, so what is happening now, is that the whole data is loaded in memory, but only a very small part of it is used.

> 
> That seems way more efficient but one (minor) issue with doing that is that we don't know how many builds to fetch from the DB to get a fixed number (say 10 or 20) of builds in the API after merging chroots.

Could this be achieved with a 'grouped by' in `CoprBuildModel`? Such a thing seems to me to have a better place in the database layer than in the API.

IceWreck:
> Could this be achieved with a 'grouped by' in CoprBuildModel? Such a thing seems to me to have a better place in the database layer than in the API.

👍 I'll work on this while modifying the API.

csomh:
@IceWreck finished this and it's merged, but let's wait until this reaches prod and close after that.

csomh:
This works now. :tada:

TomasTomecek:
nice! big ♥ to @IceWreck

PR #674, opened 2020-06-09
Merge chroots and do pagination (API) in SQL instead of python

IceWreck:
#633 
- right now, we load info about every build ever into memory and then serve info about 10 or 20 builds at a time (pagination). this repeats again when we want the next twenty builds
- we make queries to fetch "all builds" and "same build_id builds" and then merge them into one. and then we do this for every single build (above mentioned point)
- as a result /copr-builds is just timing out on the prod api

This PR replaces all of that by doing this in SQL. Now if we need the 10 latest builds with unique build_ids, we fetch that directly. Chroots are merged in SQL, not python.

- [x] Find a way to add repo_namespace and repo_name as well
- [x] Tests

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -41,11 +41,13 @@ from sqlalchemy import (
     desc,
     JSON,
     create_engine,
+    func,
     Boolean,
 )
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import sessionmaker, Session, relationship
 from sqlalchemy.types import PickleType, ARRAY
+from sqlalchemy.dialects.postgresql import array as psql_array
 
 from packit.config import JobConfigTriggerType
 from packit_service.constants import WHITELIST_CONSTANTS
@@ -440,6 +442,32 @@ class CoprBuildModel(Base):
         with get_sa_session() as session:
             return session.query(CoprBuildModel).order_by(desc(CoprBuildModel.id)).all()
 
+    @classmethod
+    def get_merged_chroots(
+        cls, first: int, last: int
+    ) -> Optional[Iterable["CoprBuildModel"]]:
+        """Returns a list of unique build ids with merged status, chroots
+        Details:
+        https://github.com/packit-service/packit-service/pull/674#discussion_r439819852
+        """
+        with get_sa_session() as session:
+            builds = (
+                session.query(
+                    # We need something to order our merged builds by,
+                    # so set new_id to be min(ids of to-be-merged rows)
+                    func.min(CoprBuildModel.id).label("new_id"),
+                    # Select identical element(s)
+                    CoprBuildModel.build_id,
+                    # Merge chroots and statuses from different rows into one
+                    func.array_agg(psql_array([CoprBuildModel.target])).label("target"),
+                    func.array_agg(psql_array([CoprBuildModel.status])).label("status"),
+                )
+                .group_by(CoprBuildModel.build_id)  # Group by identical element(s)
+                .order_by(desc("new_id"))[first:last]
+            )
+
+            return builds
+
     # Returns all builds with that build_id, irrespective of target
     @classmethod
     def get_all_by_build_id(
@@ -454,7 +482,7 @@ class CoprBuildModel(Base):
     # returns the build matching the build_id and the target
     @classmethod
     def get_by_build_id(
-        cls, build_id: Union[str, int], target: str
+        cls, build_id: Union[str, int], target: str = None
     ) -> Optional["CoprBuildModel"]:
         if isinstance(build_id, int):
             # PG is pesky about this:
@@ -463,11 +491,10 @@ class CoprBuildModel(Base):
             #   You might need to add explicit type casts.
             build_id = str(build_id)
         with get_sa_session() as session:
-            return (
-                session.query(CoprBuildModel)
-                .filter_by(build_id=build_id, target=target)
-                .first()
-            )
+            query = session.query(CoprBuildModel).filter_by(build_id=build_id)
+            if target:
+                query = query.filter_by(target=target)
+            return query.first()
 
     @classmethod
     def get_or_create(
--- a/packit_service/service/api/copr_builds.py
+++ b/packit_service/service/api/copr_builds.py
@@ -20,7 +20,6 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 from http import HTTPStatus
-from itertools import islice
 from json import dumps
 from logging import getLogger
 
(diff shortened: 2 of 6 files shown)


IceWreck:
The problem with adding repo_namespace and repo_name here is that they are not columns in CoprBuildModel but present in a linked table. We usually get them by `build.get_project().repo_name ` but i'm not sure how to get them here

csomh:
> The problem with adding repo_namespace and repo_name here is that they are not columns in CoprBuildModel but present in a linked table. We usually get them by `build.get_project().repo_name ` but i'm not sure how to get them here.
> 
> CoprBuildModel links to JobTriggerModel which then links to PullRequestModel and GitProjectModel. What does JobTriggerModel do? How do I use sql (or sqlalchemy) to access repo_name or pr_id from CoprBuildModel?

Gimme some time to look into these questions for you.

csomh:
Update: we discussed to step back and don't try to do everything in a single query, but first get the Copr builds, then complement that information with repo data in subsequent calls. This should result in a maximum of 50+1 SQL queries as per the current pagination limits, which should still be better than the current performance.

IceWreck:
This did not turn out as well as we had hoped but should be more efficient that the previous implementation.

Since CoprBuildModel is not directly related to PullRequestModel (like it was till last month) I cannot find a way to fetch pr_id or repo_name from a build via SQL.

So we have to make separate queries to fetch this info for each build rather than do it all in one single query.

packit_service/models.py
```diff
@@ -440,6 +442,27 @@ def get_all(cls) -> Optional[Iterable["CoprBuildModel"]]:
...
             return session.query(CoprBuildModel).order_by(desc(CoprBuildModel.id)).all()
 
+    @classmethod
+    def get_merged_chroots(
+        cls, first: int, last: int
+    ) -> Optional[Iterable["CoprBuildModel"]]:
+        with get_sa_session() as session:
+            builds_list = (
+                session.query(
+                    # Merged row will keep the id of any one row
```
csomh, line 452: Please make a note here, that this is needed b/c `CoprBuildModel.build_id` is a string and we cannot order the way we want based on that.
(Though checking reveals, that Copr V3 API returns build_id as an integer, so not sure, why it's not saved as such... maybe @TomasTomecek recalls?)

packit_service/models.py
```diff
@@ -440,6 +442,27 @@ def get_all(cls) -> Optional[Iterable["CoprBuildModel"]]:
...
+                    func.min(CoprBuildModel.id).label("new_id"),
+                    # Add all the fields which MUST be the same (repeat in group_by as well)
```
csomh, line 454: Is this comment still relevant?

packit_service/models.py
```diff
@@ -463,11 +486,16 @@ def get_by_build_id(
...
+            if target:
+                return (
+                    session.query(CoprBuildModel)
+                    .filter_by(build_id=build_id, target=target)
+                    .first()
+                )
+            else:
+                return (
+                    session.query(CoprBuildModel).filter_by(build_id=build_id).first()
+                )
```
csomh, line 498: You can use the `Query` object returned by `session.query()` and do multiple rounds of filtering. If my understanding is correct, SQLAlchemy does the query only once, when the scope of the session is about to end.

```python
            query = session.query(CoprBuildModel).filter_by(build_id=build_id)
            if target:
                query = query.filter_by(target=target)
            return query.first()
```

Also, I would rather set the default of `target` to `None` to allow calling this without a target.

packit_service/models.py
```diff
@@ -440,6 +442,27 @@ def get_all(cls) -> Optional[Iterable["CoprBuildModel"]]:
...
+                session.query(
+                    # Merged row will keep the id of any one row
+                    func.min(CoprBuildModel.id).label("new_id"),
+                    # Add all the fields which MUST be the same (repeat in group_by as well)
+                    CoprBuildModel.build_id,
+                    # Merge chroots and statuses into one row
+                    func.array_agg(psql_array([CoprBuildModel.target])).label("target"),
+                    func.array_agg(psql_array([CoprBuildModel.status])).label("status"),
+                )
+                .group_by(CoprBuildModel.build_id,)
```
csomh, line 460:
```suggestion
                .group_by(CoprBuildModel.build_id)
```

packit_service/models.py
```diff
@@ -440,6 +442,27 @@ def get_all(cls) -> Optional[Iterable["CoprBuildModel"]]:
         with get_sa_session() as session:
             return session.query(CoprBuildModel).order_by(desc(CoprBuildModel.id)).all()
 
+    @classmethod
+    def get_merged_chroots(
+        cls, first: int, last: int
+    ) -> Optional[Iterable["CoprBuildModel"]]:
+        with get_sa_session() as session:
+            builds_list = (
```
csomh, line 450: Keep naming simple :)

```suggestion
            builds = (
```

packit_service/service/api/copr_builds.py
```diff
@@ -51,42 +50,29 @@ def get(self):
...
-                    build_dict["chroots"].append(sbid_build.target)
-                    # Get status per chroot as well
-                    build_dict["status_per_chroot"][
-                        sbid_build.target
-                    ] = sbid_build.status
-
-                checklist.append(int(build.build_id))
-                result.append(build_dict)
+        for build in CoprBuildModel.get_merged_chroots(first, last):
+
```
csomh, line 56: Delete this empty line.

tests_requre/database/test_models.py
```diff
@@ -107,20 +107,38 @@ def test_copr_build_get_pr_id(
...
 
+def test_get_merged_chroots(clean_before_and_after, too_many_copr_builds):
+    # fetch 10 merged groups of builds
+    builds_list = CoprBuildModel.get_merged_chroots(10, 20)
+    assert len(builds_list) == 10
+    # two merged chroots so two statuses
+    assert len(builds_list[0].status) == 2
+    assert len(builds_list[0].target) == 2
+    assert builds_list[1].status[0][0] == "success"
+    assert builds_list[2].target[0][0] == "fedora-42-x86_64"
```
csomh, line 118: Fedora 42, the answer to all the questions! Can't wait :smile:

packit_service/models.py
IceWreck: Added a comment there.

Since we are merging rows here, we need to tell SQL about which column (out of three or four ie the number of targets) we want to keep.

We want the id in our query result to keep have the id of any one row, so we order them by the MIN function. Earlier this week, when where we were trying to get every single thing in one query, we used the MIN function to even get identical stuff like
```

# since theyre identical, just get any one web_url and call the new query column as web_url 
func.min(CoprBuildModel.web_url).label("web_url"),

```
MIN and MAX, for some reason also works on strings, even identical ones.


But now that we had to offload build details to a seperate query, we just dont need to do that.

packit_service/models.py
IceWreck: No, I'll replace it with something more suitable

packit_service/models.py
IceWreck: > You can use the Query object returned by session.query() and do multiple rounds of filtering

Thanks

> Also, I would rather set the default of target to None to allow calling this without a target.

Will do

PR #675, opened 2020-06-09
Change the handler classes code to have more separate attributes

lbarcziova:
This PR prepares the code to be able to create Celery tasks - handlers accept the events as dicts and don't use event as object, have separate attribute for project, package config

**TODO:**

- [ ]  make sure no info is lost, all attributes are accessed correctly
- [x] make sure each handler has valid `db_trigger`
- [x] make sure that each event dict contains `project_url` if it was present in event (to get the project in handler)
- [x] fix the tests
- [ ] reduce attributes if possible
- [x] change newly added code (for now handler for KojiBuild)

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -1027,23 +1027,20 @@ class InstallationModel(Base):
             return session.query(InstallationModel).all()
 
     @classmethod
-    def create(cls, event: dict):
+    def create(cls, event):
         with get_sa_session() as session:
-            account_login = event.get("account_login")
-            installation = cls.get_by_account_login(account_login)
+            installation = cls.get_by_account_login(event.account_login)
             if not installation:
                 installation = cls()
-                installation.account_login = account_login
-                installation.account_id = event.get("account_id")
-                installation.account_url = event.get("account_url")
-                installation.account_type = event.get("account_type")
-                installation.sender_login = event.get("sender_login")
-                installation.sender_id = event.get("sender_id")
-                installation.created_at = datetime.fromtimestamp(
-                    event.get("created_at")
-                )
+                installation.account_login = event.account_login
+                installation.account_id = event.account_id
+                installation.account_url = event.account_url
+                installation.account_type = event.account_type
+                installation.sender_login = event.sender_login
+                installation.sender_id = event.sender_id
+                installation.created_at = event.created_at
                 installation.repositories = [
-                    cls.get_project(repo).id for repo in event.get("repositories")
+                    cls.get_project(repo).id for repo in event.repositories
                 ]
                 session.add(installation)
             return installation
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -202,6 +202,11 @@ class PullRequestModel(Base):
             type=JobTriggerModelType.pull_request, trigger_id=self.id
         ).copr_builds
 
+    @classmethod
+    def get_by_id(cls, id_: int) -> Optional["PullRequestModel"]:
+        with get_sa_session() as session:
+            return session.query(PullRequestModel).filter_by(id=id_).first()
+
     def __repr__(self):
         return f"PullRequestModel(id={self.pr_id}, project={self.project})"
 
@@ -234,6 +239,11 @@ class IssueModel(Base):
                 issue.project_id = project.id
                 session.add(issue)
             return issue
+
+    @classmethod
+    def get_by_id(cls, id_: int) -> Optional["IssueModel"]:
+        with get_sa_session() as session:
+            return session.query(IssueModel).filter_by(id=id_).first()
 
     def __repr__(self):
         return f"IssueModel(id={self.issue_id}, project={self.project})"
@@ -268,6 +278,11 @@ class GitBranchModel(Base):
                 git_branch.project_id = project.id
                 session.add(git_branch)
             return git_branch
+
+    @classmethod
+    def get_by_id(cls, id_: int) -> Optional["GitBranchModel"]:
+        with get_sa_session() as session:
+            return session.query(GitBranchModel).filter_by(id=id_).first()
 
     def __repr__(self):
         return f"GitBranchModel(name={self.name},  project={self.project})"
@@ -362,6 +377,11 @@ class ProjectReleaseModel(Base):
                 project_release.commit_hash = commit_hash
                 session.add(project_release)
             return project_release
+
+    @classmethod
+    def get_by_id(cls, id_: int) -> Optional["ProjectReleaseModel"]:
+        with get_sa_session() as session:
+            return session.query(ProjectReleaseModel).filter_by(id=id_).first()
 
     def __repr__(self):
         return (
@@ -1007,20 +1027,23 @@ class InstallationModel(Base):
             return session.query(InstallationModel).all()
 
     @classmethod
-    def create(cls, event):
+    def create(cls, event: dict):
         with get_sa_session() as session:
-            installation = cls.get_by_account_login(event.account_login)
+            account_login = event.get("account_login")
+            installation = cls.get_by_account_login(account_login)
             if not installation:
                 installation = cls()
-                installation.account_login = event.account_login
-                installation.account_id = event.account_id
-                installation.account_url = event.account_url
-                installation.account_type = event.account_type
-                installation.sender_login = event.sender_login
-                installation.sender_id = event.sender_id
-                installation.created_at = event.created_at
+                installation.account_login = account_login
+                installation.account_id = event.get("account_id")
+                installation.account_url = event.get("account_url")
+                installation.account_type = event.get("account_type")
+                installation.sender_login = event.get("sender_login")
+                installation.sender_id = event.get("sender_id")
+                installation.created_at = datetime.fromtimestamp(
+                    event.get("created_at")
+                )
                 installation.repositories = [
-                    cls.get_project(repo).id for repo in event.repositories
+                    cls.get_project(repo).id for repo in event.get("repositories")
                 ]
                 session.add(installation)
             return installation
--- a/packit_service/service/events.py
+++ b/packit_service/service/events.py
@@ -114,6 +114,24 @@ class TheJobTriggerType(str, enum.Enum):
     copr_end = "copr_end"
 
 
+class EventType(str, enum.Enum):
+    installation = "InstallationEvent"
+    copr_build = "CoprBuildEvent"
+    koji_build = "KojiBuildEvent"
+    tft_result = "TestingFarmResultEvent"
+    release = "ReleaseEvent"
+    push_gh = "PushGithubGHEvent"
+    pull_request_gh = "PullRequestGHEvent"
+    pull_request_comment_gh = "PullRequestCommentGHEvent"
+    issue_comment = "IssueCommentEvent"
+    push_pagure = "PushPagureEvent"
+    pull_request_comment_pg = "PullRequestCommentPagureEvent"
+    pull_request_pg = "PullRequestPagureEvent"
+    push_gl = "PushGitlabEvent"
+    merge_request_gl = "MergeRequestGitlabEvent"
+    merge_request_comment_gl = "MergeRequestCommentGitlabEvent"
+
+
 class TestResult(dict):
     def __init__(self, name: str, result: TestingFarmResult, log_url: str):
         dict.__init__(self, name=name, result=result, log_url=log_url)
@@ -175,7 +193,9 @@ class Event:
         d = default_dict or self.__dict__
         d = copy.deepcopy(d)
         # whole dict have to be JSON serializable because of redis
+        d["event_type"] = self.__class__.__name__
         d["trigger"] = d["trigger"].value
+        d["trigger_id"] = self.db_trigger.id if self.db_trigger else None
         d["created_at"] = int(d["created_at"].timestamp())
         return d
 
@@ -955,14 +975,6 @@ class CoprBuildEvent(AbstractForgeIndependentEvent):
         result["topic"] = result["topic"].value
         self.build = build
         return result
-
-
-def get_copr_build_logs_url(event: CoprBuildEvent) -> str:
-    return (
-        f"https://copr-be.cloud.fedoraproject.org/results/{event.owner}/"
-        f"{event.project_name}/{event.chroot}/"
-        f"{event.build_id:08d}-{event.pkg}/builder-live.log.gz"
-    )
 
 
 def get_koji_build_logs_url(event: KojiBuildEvent) -> Optional[str]:
--- a/packit_service/service/events.py
+++ b/packit_service/service/events.py
@@ -870,6 +870,16 @@ class KojiBuildEvent(AbstractForgeIndependentEvent):
                 return None  # With Github app, we cannot work with fork repo
         return self.project
 
+    def get_dict(self, default_dict: Optional[Dict] = None) -> dict:
+        result = super().get_dict()
+        result["state"] = result["state"].value
+        result["old_state"] = result["old_state"].value
+        result["commit_sha"] = self.commit_sha
+        result["pr_id"] = self.pr_id
+        result["git_ref"] = self.git_ref
+        result["identifier"] = self.identifier
+        return result
+
 
 class CoprBuildEvent(AbstractForgeIndependentEvent):
     build: Optional[CoprBuildModel]
@@ -980,25 +990,6 @@ class CoprBuildEvent(AbstractForgeIndependentEvent):
         result["topic"] = result["topic"].value
         self.build = build
         return result
-
-
-def get_koji_build_logs_url(event: KojiBuildEvent) -> Optional[str]:
-    if not event.rpm_build_task_id:
-        return None
-
-    return (
-        f"https://kojipkgs.fedoraproject.org//work/tasks/"
-        f"{event.rpm_build_task_id % 10000}/{event.rpm_build_task_id}/build.log"
-    )
-
-
-def get_koji_rpm_build_web_url(event: KojiBuildEvent) -> Optional[str]:
-    if not event.rpm_build_task_id:
-        return None
-
-    return (
-        f"https://koji.fedoraproject.org/koji/taskinfo?taskID={event.rpm_build_task_id}"
-    )
 
 
 class AbstractPagureEvent(AbstractForgeIndependentEvent):
--- a/packit_service/service/events.py
+++ b/packit_service/service/events.py
@@ -197,6 +197,9 @@ class Event:
         d["trigger"] = d["trigger"].value
         d["trigger_id"] = self.db_trigger.id if self.db_trigger else None
         d["created_at"] = int(d["created_at"].timestamp())
+        d["project_url"] = d.get("project_url") or (
+            self.db_trigger.project.project_url if self.db_trigger else None
+        )
         return d
 
     @property
@@ -629,6 +632,7 @@ class IssueCommentEvent(AddIssueDbTrigger, AbstractGithubEvent):
     def get_dict(self, default_dict: Optional[Dict] = None) -> dict:
         result = super().get_dict()
         result["action"] = result["action"].value
+        result["tag_name"] = self.tag_name
         return result
 
 
--- a/packit_service/service/events.py
+++ b/packit_service/service/events.py
@@ -141,6 +141,62 @@ class TestResult(dict):
         )
 
 
+class EventData:
+    """
+    Class to represent the data which are common for handlers and comes from the original event
+    """
+
+    def __init__(
+        self,
+        event_type: str,
+        trigger: TheJobTriggerType,
+        user_login: str,
+        trigger_id: int,
+        project_url: str,
+        tag_name: Optional[str],
+        git_ref: Optional[str],
+        pr_id: Optional[int],
+        commit_sha: Optional[str],
+        identifier: Optional[str],
+    ):
+        self.event_type = event_type
+        self.trigger = trigger
+        self.user_login = user_login
+        self.trigger_id = trigger_id
+        self.project_url = project_url
+        self.tag_name = tag_name
+        self.git_ref = git_ref
+        self.pr_id = pr_id
+        self.commit_sha = commit_sha
+        self.identifier = identifier
+
+    @classmethod
+    def from_event_dict(cls, event: dict):
+        event_type = event.get("event_type")
+        trigger = TheJobTriggerType(event.get("trigger"))
+        user_login = event.get("user_login")
+        trigger_id = event.get("trigger_id")
+        project_url = event.get("project_url")
+        tag_name = event.get("tag_name")
+        git_ref = event.get("git_ref")
+        pr_id = event.get("pr_id")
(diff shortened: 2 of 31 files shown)


packit_service/service/events.py
```diff
@@ -114,6 +114,27 @@ class TheJobTriggerType(str, enum.Enum):
...
+    issue_comment = "IssueCommentEvent"
+    push_pagure = "PushPagureEvent"
+    pull_request_comment_pg = "PullRequestCommentPagureEvent"
+    pull_request_pg = "PullRequestPagureEvent"
+    pull_request_label_pg = "PullRequestLabelPagureEvent"
+    dist_git_event = "DistGitEvent"
+    issue_comment_gl = "IssueCommentGitlabEvent"
+    merge_request_comment_gl = "MergeRequestCommentGitlabEvent"
+    merge_request_gl = "MergeRequestGitlabEvent"
+    push_gl = "PushGitlabEvent"
```
csomh, line 135: - Remove 'Event', as it's mostly redundant (keep only for dist-git, as that one doesn't sound event-like at all without it :) ).
- avoid unnecessary abbreviations (i.e.: 'pg' is already used in other places to note 'postgres')
- change word order, to sound more natural (i.e. to the question "What type of event is this?" one naturally would answer: "a release" or "a GitLab merge request" and so on).
- capitalise "GitLab" as it should be 

```suggestion
    installation = "Installation"
    copr_build = "CoprBuild"
    koji_build = "KojiBuild"
    tft_result = "TestingFarmResult"
    release = "Release"
    github_push = "GitHubPush"
    github_pull_request = "GitHubPullRequest"
    github_pull_request_comment = "GitHubPullRequestComment"
    issue_comment = "IssueComment"
    pagure_push = "PagurePush"
    pagure_pull_request_comment = "PagurePullRequestComment"
    pagure_pull_request = "PagurePullRequest"
    pagure_pull_request_label = "PagurePullRequestLabel"
    dist_git_event = "DistGitEvent"
    gitlab_issue_comment = "GitLabIssueComment"
    gitlab_merge_request_comment = "GitLabMergeRequestComment"
    merge_request = "GitLabMergeRequest"
    gitlab_push = "GitLabPush"
```

packit_service/worker/build/build_helper.py
```diff
@@ -65,33 +56,23 @@ def __init__(
...
-            PullRequestGithubEvent,
-            PullRequestPagureEvent,
-            PullRequestCommentGithubEvent,
-            PullRequestCommentPagureEvent,
-            CoprBuildEvent,
-            PushGitHubEvent,
-            ReleaseEvent,
-            KojiBuildEvent,
-        ],
+        event: dict,
```
lachmanfrantisek, line 59: Cannot we have some class representation for this?

With general dictionaries we are losing the benefit of static typing checks.

packit_service/service/events.py
lbarcziova: You're right, these are better, I can definitely change the suggested enum names, but in case of the enum values, they are the same as the event class names, since it is used when creating dict from event object as:
`d["event_type"] = self.__class__.__name__` and then
` self.event_type = EventType(event.get("event_type"))`

packit_service/worker/handlers/abstract.py
```diff
@@ -161,12 +174,57 @@ class JobHandler(Handler):
...
+                self._db_trigger = GitBranchModel.get_by_id(self.trigger_id)
+        return self._db_trigger
+
+    @property
+    def project(self) -> Optional[GitProject]:
+        if not self._project:
+            self._project = self.get_project()
+        return self._project
+
+    def get_project(self) -> Optional[GitProject]:
```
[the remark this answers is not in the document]
lbarcziova: OK, I will merge them and and leave only `project`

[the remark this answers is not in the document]
lachmanfrantisek: I see the point when using inheritance. I would only add a docstring, that describe the workflow:
- `get_project` will cause project instantiation each-time
- `project` property is lazzy and will run `get_project` when called for the first time

packit_service/worker/handlers/comment_action_handler.py
```diff
@@ -76,26 +70,17 @@ def add_to_comment_action_mapping_with_name_inner(
...
     ):
-        super().__init__(config)
-        self.event: Union[
-            PullRequestCommentGithubEvent,
-            IssueCommentEvent,
-            PullRequestCommentPagureEvent,
-        ] = event
-        self.job = job
+        super().__init__(package_config, job_config, event)
+        self.identifier = event.get("pr_id") or event.get("issue_id")
```
lachmanfrantisek, line 83: This is the reason we should use classes instead of dicts.

packit_service/service/events.py
lachmanfrantisek: > They are the same as the event class names

That's my point. Do we need this enum at all? Cannot we work only with the class name?

lachmanfrantisek: As a bonus. If we can avoid that enum, we can use things like `if self.event_type == GithubPullRequest.__name__`.

csomh: I see, didn't realise this. Let's improve the enum names then, if it's not too complicated.

lbarcziova: I just wanted to avoid comparing it directly like `self.event_type == "GithubPullRequest"`, but this way `if self.event_type == GithubPullRequest.__name__` of comparing looks good, so I can remove the enum then

packit_service/models.py
```diff
@@ -891,20 +911,23 @@ def get_all(cls) -> Optional[Iterable["InstallationModel"]]:
...
-                installation.created_at = event.created_at
+                installation.account_login = account_login
+                installation.account_id = event.get("account_id")
+                installation.account_url = event.get("account_url")
+                installation.account_type = event.get("account_type")
+                installation.sender_login = event.get("sender_login")
+                installation.sender_id = event.get("sender_id")
+                installation.created_at = datetime.fromtimestamp(
+                    event.get("created_at")
+                )
```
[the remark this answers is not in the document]
lbarcziova: Most of the attributes are not needed in the rest of InstallationHandler code, but yes,
we can:
1. parse all these attributes from dict in handler code to be able to use them here
2. create the db record when the InstallationEvent is parsed from webhook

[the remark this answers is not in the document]
lachmanfrantisek: > create the db record when the InstallationEvent is parsed from webhook

Since this requires a database action, I would try to avoid any unnecessary action during parsing.

packit_service/worker/build/build_helper.py
lbarcziova: So would be creating a class which would represent data needed by helpers a better option? Something like 
```python
class BuildHelperData:
    def __init__(
        self,
        event: dict,
    ):
        self.pr_id = event.get("pr_id")
        self.git_ref = event.get("git_ref")
        self.commit_sha = event.get("commit_sha")
        self.trigger = (
            TheJobTriggerType(event.get("trigger")) if event.get("trigger") else None
        )
```

packit_service/worker/handlers/abstract.py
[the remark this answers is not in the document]
lbarcziova: > I see the point when using inheritance.

It does make sense in the Events, as subclasses can have their own `get_project` implementation, and the `project` property is implemented in the Event class and uses `get_project`, but since here is only one implementation for `get_project`, I would merge it as Hunor suggested.

packit_service/worker/build/build_helper.py
lachmanfrantisek: Yes if the set of attributes differ to the original event. (Otherwise, it makes sense to reinitialize the original object.)

It allows us to better define and name this context we are forwarding.

packit_service/models.py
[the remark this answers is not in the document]
lbarcziova: changed

packit_service/worker/build/build_helper.py
lbarcziova: created

lbarcziova:
`packit-service-tests : FAILURE in 8m 04s` 1 failed test, because of #685

packit_service/worker/handlers/abstract.py
```diff
@@ -161,12 +174,47 @@ class JobHandler(Handler):
...
+            elif self.trigger == TheJobTriggerType.release:
+                self._db_trigger = ProjectReleaseModel.get_by_id(self.trigger_id)
+            elif self.trigger in TheJobTriggerType.push:
+                self._db_trigger = GitBranchModel.get_by_id(self.trigger_id)
+        return self._db_trigger
+
+    @property
+    def project(self) -> Optional[GitProject]:
+        if not self._project and self.project_url:
+            self._project = self.config.get_project(url=self.project_url)
```
csomh, line 216: `self.config` is a `ServiceConfig`, right? And that one does not have `get_project()`. Or am I missing something?

lbarcziova: > self.config is a ServiceConfig, right?

yes

> And that one does not have get_project(). Or am I missing something?
 
ServiceConfig inherits from Config in packit and it has method get_project()

csomh: Oh, okay, so that's what I missed :)

packit_service/worker/whitelist.py
```diff
@@ -87,22 +87,23 @@ def _signed_fpca(self, account_login: str) -> bool:
         logger.info(f"Cannot verify whether {account_login!r} signed FPCA.")
         return False
 
-    def add_account(self, event: InstallationEvent) -> bool:
+    def add_account(self, account_login: str, sender_login: str) -> bool:
         """
         Add account to whitelist.
         Status is set to 'waiting' or to 'approved_automatically'
         if the account is a packager in Fedora.
         :param event: Github app installation info
```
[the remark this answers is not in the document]
lbarcziova: updated

packit_service/worker/build/build_helper.py
```diff
@@ -34,26 +34,50 @@
...
     are_job_types_same,
 )
 from packit_service.worker.reporting import StatusReporter
+from packit_service.service.events import TheJobTriggerType
 from sandcastle import SandcastleTimeoutReached
 
 logger = logging.getLogger(__name__)
 
 
+class BuildHelperMetadata:
```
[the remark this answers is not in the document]
lbarcziova: done

packit_service/worker/handlers/comment_action_handler.py
```diff
@@ -76,26 +70,16 @@ def add_to_comment_action_mapping_with_name_inner(
...
-        config: ServiceConfig,
-        event: Union[
-            PullRequestCommentGithubEvent,
-            IssueCommentEvent,
-            PullRequestCommentPagureEvent,
-        ],
-        job: JobConfig,
+        package_config: PackageConfig,
+        job_config: Optional[JobConfig],
+        event: dict,
```
TomasTomecek, line 80: I'll be honest that I don't like this - passing dict, that will make things harder to debug since it's dict and anything can be inside, not sure if we could pass real objects here which have defined the (de-)serialization process

this comment is just my observation, no need to take any action

packit_service/worker/handlers/fedmsg_handlers.py
```diff
@@ -78,20 +74,17 @@ def add_topic(kls: Type["FedmsgHandler"]):
...
-
 class FedmsgHandler(JobHandler):
     """ Handlers for events from fedmsg """
 
     topic: str
 
-    def __init__(self, config: ServiceConfig, job_config: JobConfig, event: Event):
-        super().__init__(config=config, job_config=job_config, event=event)
+    def __init__(
+        self, package_config: PackageConfig, job_config: JobConfig, event: dict
```
TomasTomecek, line 83: nice, package_config being passed directly and not being part of the event

lbarcziova:
recheck

packit_service/worker/build/build_helper.py
```diff
@@ -34,26 +34,55 @@
...
+        git_ref: Optional[str] = None,
+        commit_sha: Optional[str] = None,
+        # identifier of the respective event (PR, release) used e.g. in copr project name
+        identifier: Optional[str] = None,
+    ):
+        self.trigger = trigger
+        self.pr_id = pr_id
+        self.git_ref = git_ref
+        self.commit_sha = commit_sha
+        self.identifier = identifier
```
lachmanfrantisek, line 66: It looks like all of these can be calculated from the `db_trigger` object so:
- Name can be something related to trigger.
- We can (?in the future) potentially use the id of the database object only. But I am not sure about this now.

packit_service/worker/handlers/abstract.py
```diff
@@ -81,11 +90,15 @@ def _add_to_mapping(kls: Type["JobHandler"]):
...
+    api: Optional[PackitAPI] = None
+    local_project: Optional[LocalProject] = None
+    _config: Optional[ServiceConfig] = None
 
-    def __init__(self, config: ServiceConfig):
-        self.config: ServiceConfig = config
-        self.api: Optional[PackitAPI] = None
-        self.local_project: Optional[LocalProject] = None
+    @property
+    def config(self) -> ServiceConfig:
```
lachmanfrantisek, line 98: I would appreciate if we can be specific here. I am always wondering about type when I see only `config`.

```suggestion
    def service_config(self) -> ServiceConfig:
```

packit_service/worker/handlers/abstract.py
```diff
@@ -161,12 +174,47 @@ class JobHandler(Handler):
...
+                TheJobTriggerType.pr_label,
+            ):
+                self._db_trigger = PullRequestModel.get_by_id(self.trigger_id)
+            elif self.trigger == TheJobTriggerType.issue_comment:
+                self._db_trigger = IssueModel.get_by_id(self.trigger_id)
+            elif self.trigger == TheJobTriggerType.release:
+                self._db_trigger = ProjectReleaseModel.get_by_id(self.trigger_id)
+            elif self.trigger in TheJobTriggerType.push:
+                self._db_trigger = GitBranchModel.get_by_id(self.trigger_id)
+        return self._db_trigger
```
lachmanfrantisek, line 211: I was a bit lazy last time when I create a similar if-else but we should share this code across handlers.
(Later if you don't want to do it (here).)

Same for other trigger-type-related properties.

packit_service/worker/handlers/comment_action_handler.py
lachmanfrantisek: @lbarcziova you created a new class for those metadata, didn't you?

Cannot be that enhanced to server also there?

I would like to avoid dicts for the same reason as I wrote in build-helpers: typing, debugging,...

packit_service/worker/handlers/fedmsg_handlers.py
```diff
@@ -142,20 +136,64 @@ def run(self) -> HandlerResults:
...
             # we use branch on purpose so we get the latest thing
             # TODO: check if rev is HEAD on {branch}, warn then?
-            dist_git_branch=self.distgit_event.branch,
+            dist_git_branch=self.branch,
             upstream_branch="master",  # TODO: this should be configurable
         )
         return HandlerResults(success=True, details={})
 
 
+class CoprBuildHandler(FedmsgHandler):
```
lachmanfrantisek, line 145: Maybe add something about results to the name? (So we easily differentiate the ones that trigger build and ones that update the status.)

packit_service/worker/handlers/fedmsg_handlers.py
```diff
@@ -142,20 +136,64 @@ def run(self) -> HandlerResults:
...
+                str(self.copr_event.build_id), self.copr_event.chroot
+            )
+        return self._build
+
+    @property
+    def db_trigger(self) -> Optional[AbstractTriggerDbType]:
+        if not self._db_trigger:
+            self._db_trigger = self.build.job_trigger.get_trigger_object()
+        return self._db_trigger
+
```
lachmanfrantisek, line 188: :+1:

packit_service/worker/handlers/fedmsg_handlers.py
lbarcziova: what do you think about `CoprBuildStateHandler`?

packit_service/worker/build/build_helper.py
lbarcziova: can I rename it for now to something like `TriggerMetadata` and later (in some next PR) try to use only DB objects?

packit_service/worker/handlers/abstract.py
lbarcziova: I just wanted to be consistent as in `SteveJobs` class there it is also named as `config`, so should I change it everywhere?

packit_service/worker/handlers/abstract.py
```diff
@@ -161,12 +174,47 @@ class JobHandler(Handler):
...
+        event: dict,
     ):
-        super().__init__(config)
+        self.package_config: PackageConfig = package_config
         self.job_config: Optional[JobConfig] = job_config
+        self.event_type: str = event.get("event_type")
+        self.trigger: TheJobTriggerType = TheJobTriggerType(event.get("trigger"))
+        self.user_login: str = event.get("user_login")
+        self.trigger_id: int = event.get("trigger_id")
+        self.project_url: str = event.get("project_url")
```
lbarcziova, line 188: >  @lbarcziova you created a new class for those metadata, didn't you?
Cannot be that enhanced to server also there?
I would like to avoid dicts for the same reason as I wrote in build-helpers: typing, debugging,...

@lachmanfrantisek I don't know if I understand how you mean it, in the JobHandler there are these attributes needed almost by each handler parsed from the dict

packit_service/worker/build/build_helper.py
lachmanfrantisek: Yes, can be. I am even not sure now if we want to forward that info via database. It's for another discussion.

packit_service/worker/handlers/abstract.py
lachmanfrantisek: Other PRs authors' will not like us during rebase..;)

What about changing this one here and the one in `SteveJobs` in a separate PR so other's can rebase more easily?

packit_service/worker/handlers/fedmsg_handlers.py
lachmanfrantisek: Can be. OR `CoprBuildReportHandler` to match the Koji one? Or `AbstractCoprBuildReportHandler` to also show that it needs to be inherited.

packit_service/worker/handlers/abstract.py
lachmanfrantisek: I have the same problems with using dictionary representation as I have with it in helper classes:
- debugging is harder
- types cannot be (easily) checked
- code is less readable (a lot of `get` noise and strings can be mistyped)
- autocompletion does not work (and other IDE helpers)

Two possible solutions:
- Reuse and enhance the metadata class you already created.
- Create a new class that will handle those values.

lbarcziova: I got your point, I was just thinking that now each handler in init calls the super(JobHandler) init method so in JobHandler, the common attributes (event_type, trigger_id, trigger, user_login, project_url) are parsed (and optionally some specific attributes in own init method are parsed), therefore no simple way came to my mind to get rid of the dictionary totally.

Would something like this make sense? Having new class with the common attributes as I wrote above, having classmethod in which the parsing would be done as in the new TriggerMetadata class and then in the handlers it would look like this:
```python
class ProposeDownstreamHandler(JobHandler):
    type = JobType.propose_downstream
    triggers = [TheJobTriggerType.release]

    def __init__(
        self,
        package_config: PackageConfig,
        job_config: Optional[JobConfig],
        handler_data: HandlerData,
        tag_name: str,
    ):
        super().__init__(
            package_config=package_config, job_config=job_config, data=handler_data,
        )
        self.tag_name = tag_name
        
    @classmethod
    def from_event_dict(
        cls,
        package_config: PackageConfig,
        job_config: Optional[JobConfig],
        event: dict,
    ):
        tag_name = event.get("tag_name")
        return ProposeDownstreamHandler(
            package_config=package_config,
            job_config=job_config,
            handler_info=HandlerData.from_event_dict(event)
            tag_name=tag_name
        )
```

lachmanfrantisek: I see that it's not the same case.

- If the attributes are mostly the same, I would add everything to the `HandlerData` and make the specific ones optional. An instance of `HandlerData` can then be instantiated before handler and send as an instance (not a dictionary) to the `__init__` of the handler. (We will need to fight with the 'Optional` there.)
- If the attributes differ a lot we can have multiple types of `HandlerData` (a lot of work), or leave it as it is now. (I am not so against now..;)
- Or, we can implement that as `kwargs` that will be forwarded to the `__init__` of handlers. (You will send  `**event` when instantiating the handler.)

lbarcziova: So, I created a class for the common attributes and since more of the TriggerData attributes were required there, I removed the TriggerData class. Nevertheless, for some handlers (e.g.Koji/CoprReportHandler), there is also event dict passed, as the attributes there differ from the most of the handlers, but in the following PR, where the tasks will be implemented, these specific attributes can be parsed before handler instantiation as the handlers would be called separately by tasks.

TomasTomecek:
Tests in Franta's PR passed - was he just lucky or a rebase or recheck is needed here?

lbarcziova:
recheck

packit_service/worker/build/__init__.py
```diff
@@ -5,7 +5,7 @@
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
 # in the Software without restriction, including without limitation the rights
-# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+# to use, copy, modiffy, merge, publish, distribute, sublicense, and/or sell
```
lachmanfrantisek, line 8: Probably a typo...
```suggestion
# to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
```

packit_service/worker/build/babysit.py
```diff
@@ -94,7 +93,11 @@ def check_copr_build(build_id: int) -> bool:
         )
 
         for job_config in job_configs:
+            event_dict = event.get_dict()
             CoprBuildEndHandler(
-                ServiceConfig.get_service_config(), job_config=job_config, event=event,
+                package_config=event.package_config,
+                job_config=job_config,
+                data=EventData.from_event_dict(event_dict),
+                event=event_dict,
```
lachmanfrantisek, line 101: Can we rename `event` to `event_dict` to make it clear? OR encapsulate the raw `event_dict` to `EventData`?

packit_service/worker/build/__init__.py
lbarcziova: yes, thanks

packit_service/worker/build/babysit.py
lbarcziova: going to change it

lachmanfrantisek: Thanks!

PR #676, opened 2020-06-09
let handlers use jobconfig only

### LINKED ISSUE #393: copr builds: add a way to customize SRPMs (the build process)
freenode IRC conversation:
```
13:25 <martstyk> Hello Packit gods. Question here. Can I create different builds accordingly to different build targets in COPR? For example in F30 i will pass additional steps into srpm
13:42 <ttomecek> martstyk, short answer: no, long answer: what are you trying to achieve? :)
13:43 <ttomecek> there is no way to customize copr builds now: I'd say you are asking for different spec files per copr build job, or maybe some custom commands being ran for different targets
13:43 <martstyk> ttomecek, Currently we have this in spec %{!?_without_static:%global with_static 1}
13:44 <martstyk> w/o --without static you will be building all assets
13:44 <martstyk> which is not needed for Fedora, .. as I can link everything dynamically
13:44 <martstyk> I guess I will make new spec file only for Fedora + Packit Service
13:45 <martstyk> Until now I was able to ship same spec to build targets from RHEL4 - Fedora Rawhide
13:53 <ttomecek> martstyk, so you want to perform two builds? "with_static 0" and "with_static 1" based on the target platform? (rhel vs fedora)
13:54 <martstyk> more like targets
13:54 <martstyk> Because I can do EL8 w/ with static 0
13:55 <ttomecek> this means to either override the spec file ( = run custom command per target), or supply overrides for rpmbuild to define the variables
13:55 <martstyk> yup      
```

We need a way in packit.yaml for people to customize stuff (spec file most of the time) for a given copr build job. We can either do it by running a user defined command, and/or pass custom variables  to the rpmbuild command run.

The question is, how to make this available per PR and also via CLI.

### LINKED ISSUE #418: Allow `downstream_package_name` to be specified per-job
Some packages have different names on different Fedora/EPEL releases (for example, a newer version of a library on EPEL 7 might be named in the form `packagename$version` while in Fedora it is just `packagename`.

It should be possible to do the following:
```
specfile_path: libmodulemd.spec
upstream_package_name: libmodulemd
downstream_package_name: libmodulemd
upstream_tag_template: libmodulemd-{version}
current_version_command:
  - ./.packit_version.sh

synced_files:
  - .packit.yml
  - .make_packit_specfile.sh
  - libmodulemd.spec

actions:
    post-upstream-clone: ./.make_packit_specfile.sh

jobs:
- job: copr_build
  trigger: pull_request
  metadata:
    targets: fedora-all

- job: propose_downstream
  trigger: release
  metadata:
    dist-git-branch: fedora-all

- job: propose_downstream
  trigger: release
  metadata:
    dist-git-branch: epel-7
    downstream_package_name: libmodulemd2
```

TomasTomecek:
* [x] finish code
* [x] fix tests
* [x] get a beer

With this insane refactor, handlers would only use jobconfig and not
packageconfig.

Fixes #418
Fixes #393

I'm pretty sure this will conflict with #675 a lot.

### DIFF
--- a/packit_service/worker/build/build_helper.py
+++ b/packit_service/worker/build/build_helper.py
@@ -25,22 +25,23 @@ from pathlib import Path
 from typing import Union, List, Optional, Tuple, Set
 
 from kubernetes.client.rest import ApiException
-
 from ogr.abstract import GitProject, CommitStatus
 from packit.api import PackitAPI
-from packit.config import PackageConfig, JobType, JobConfig
+from packit.config import JobType, JobConfig
+from packit.config.package_config import PackageConfig
 from packit.local_project import LocalProject
 from packit.utils import PackitFormatter
+from sandcastle import SandcastleTimeoutReached
+
 from packit_service import sentry_integration
 from packit_service.config import ServiceConfig, Deployment
 from packit_service.models import SRPMBuildModel
+from packit_service.service.events import EventData
 from packit_service.trigger_mapping import (
     is_trigger_matching_job_config,
     are_job_types_same,
 )
 from packit_service.worker.reporting import StatusReporter
-from packit_service.service.events import EventData
-from sandcastle import SandcastleTimeoutReached
 
 logger = logging.getLogger(__name__)
 
@@ -58,10 +59,11 @@ class BaseBuildJobHelper:
         project: GitProject,
         metadata: EventData,
         db_trigger,
-        job: Optional[JobConfig] = None,
+        job_config: JobConfig,
     ):
         self.config: ServiceConfig = config
-        self.package_config: PackageConfig = package_config
+        self.job_config = job_config
+        self.package_config = package_config
         self.project: GitProject = project
         self.db_trigger = db_trigger
         self.msg_retrigger: Optional[str] = ""
@@ -75,14 +77,8 @@ class BaseBuildJobHelper:
         self._build_check_names: Optional[List[str]] = None
         self._srpm_model: Optional[SRPMBuildModel] = None
         self._srpm_path: Optional[Path] = None
-
-        # lazy properties, current job by default
-        self._job_build = (
-            job if job and are_job_types_same(job.type, self.job_type_build) else None
-        )
-        self._job_tests = (
-            job if job and are_job_types_same(job.type, self.job_type_test) else None
-        )
+        self._job_tests: Optional[JobConfig] = None
+        self._job_build: Optional[JobConfig] = None
 
     @property
     def local_project(self) -> LocalProject:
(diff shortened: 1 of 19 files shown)


packit_service/worker/build/build_helper.py
```diff
@@ -132,102 +124,20 @@ def api_url(self) -> str:
         )
 
     @property
-    def configured_build_targets(self) -> Set[str]:
+    def configured_targets(self) -> Set[str]:
```
lachmanfrantisek, line 127: You are removing the whole logic with targets?

We still need to know both build and test targets.
- We need to set a commit status for both variants.
- We are defaulting with targets to other-one.

I am not saying this needs to be done in this class, but somewhere, this needs to be calculated.

TomasTomecek:
I had to start from scratch, lol

packit_service/worker/handlers/github_handlers.py
```diff
@@ -569,7 +570,7 @@ def run(self) -> HandlerResults:
 
         api = PackitAPI(
             config=self.config,
-            package_config=self.package_config,
+            package_config=self.job_config,
```
lbarcziova, line 573: This was a little bit confusing when I read it for the first time (maybe we could rename the attribute later in packit)

tests/unit/test_whitelist.py
```diff
@@ -297,16 +297,15 @@ def test_check_and_report(
...
-    flexmock(PullRequestGithubEvent).should_receive("get_package_config").and_return(
-        flexmock(
-            jobs=[
-                JobConfig(
-                    type=JobType.tests,
-                    trigger=JobConfigTriggerType.pull_request,
-                    metadata=JobMetadataConfig(targets=["fedora-rawhide"]),
-                )
-            ],
+    jobs = [
```
csomh, line 300: This seems to be a list of `JobConfig`s. So maybe call this `job_configs`?

packit_service/worker/whitelist.py
```diff
@@ -224,17 +230,19 @@ def check_and_report(
...
-                        package_config=event.get_package_config(),
-                        project=project,
-                        metadata=EventData.from_event_dict((event.get_dict())),
-                        db_trigger=event.db_trigger,
-                    )
-                    msg = "Account is not whitelisted!"  # needs to be shorter
-                    job_helper.report_status_to_all(
-                        description=msg, state=CommitStatus.error, url=FAQ_URL
-                    )
+                    for job in job_configs:
```
csomh, line 233:
```suggestion
                    for job_config in job_configs:
```

Calling things what they are, helps understanding code :yum:

packit_service/worker/handlers/github_handlers.py
TomasTomecek: agreed, `common_config` could be a better name

lachmanfrantisek: Or, use `service_config` for service config and `config` for the general one. (We discussed the expliciteness of `service_config` in her last PR..)

TomasTomecek:
Thanks for the review, merging.

@lbarcziova sorry for the rebase hell :angel:

Issue #185, opened 2019-10-24
even better API 


jpopelka:
- remove traling `/` (e.g. `/api/copr-builds/` -> `/api/copr-builds`)
- total items count ([[1]](https://stackoverflow.com/questions/3715981/what-s-the-best-restful-method-to-return-total-number-of-items-in-an-object), [[2]](https://stackoverflow.com/questions/5393558/how-should-i-implement-a-count-verb-in-my-restful-web-service))
- timestamps -> datetime strings

PR #187, opened 2019-10-24
Even better API

jpopelka:
Fixes #185

PR #678, opened 2020-06-15
Add API endpoint for fetching testing farm results

### LINKED ISSUE #636: `/tests` endpoint to provide information on test jobs

IceWreck:
Fixes #636.

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -854,6 +854,18 @@ class TFTTestRunModel(Base):
             self.web_url = web_url
             session.add(self)
 
+    def get_project(self) -> Optional[GitProjectModel]:
+        trigger_object = self.job_trigger.get_trigger_object()
+        if not trigger_object:
+            return None
+        return trigger_object.project
+
+    def get_pr_id(self) -> Optional[int]:
+        trigger_object = self.job_trigger.get_trigger_object()
+        if isinstance(trigger_object, PullRequestModel):
+            return trigger_object.pr_id
+        return None
+
     @classmethod
     def create(
         cls,
@@ -887,6 +899,13 @@ class TFTTestRunModel(Base):
                 .filter_by(pipeline_id=pipeline_id)
                 .first()
             )
+
+    @classmethod
+    def get_range(cls, first: int, last: int) -> Optional[Iterable["TFTTestRunModel"]]:
+        with get_sa_session() as session:
+            return session.query(TFTTestRunModel).order_by(desc(TFTTestRunModel.id))[
+                first:last
+            ]
 
 
 class InstallationModel(Base):
--- a/packit_service/service/api/testing_farm.py
+++ b/packit_service/service/api/testing_farm.py
@@ -22,8 +22,9 @@
 
 import logging
 from http import HTTPStatus
+from json import dumps
 
-from flask import request
+from flask import request, make_response
 
 try:
     from flask_restx import Namespace, Resource, fields
@@ -33,6 +34,8 @@ except ModuleNotFoundError:
 from packit_service.celerizer import celery_app
 from packit_service.config import ServiceConfig
 from packit_service.service.api.errors import ValidationFailed
+from packit_service.models import TFTTestRunModel
+from packit_service.service.api.parsers import indices, pagination_arguments
 
 logger = logging.getLogger("packit_service")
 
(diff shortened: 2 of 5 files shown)


packit_service/service/api/testing_farm.py
```diff
@@ -134,3 +137,40 @@ def validate_testing_farm_request():
         msg = "Invalid testing farm secret provided"
         logger.warning(msg)
         raise ValidationFailed(msg)
+
+
+@ns.route("/get-results")
+class TestingFarmResultsList(Resource):
+    @ns.expect(pagination_arguments)
+    @ns.response(HTTPStatus.PARTIAL_CONTENT, "Testing Farm Results list follows")
+    def get(self):
```
csomh, line 146: I think it would make more sense to make this part of `TestingFarmResults`, instead of having a separate route for it.

packit_service/service/api/testing_farm.py
```diff
@@ -134,3 +137,40 @@ def validate_testing_farm_request():
...
+
+            project = tf_result.get_project()
+            if project:
+                result_dict["repo_namespace"] = project.namespace
+                result_dict["repo_name"] = project.repo_name
+
+            result.append(result_dict)
+
+        resp = make_response(dumps(result), HTTPStatus.PARTIAL_CONTENT)
+        resp.headers["Content-Range"] = f"copr-builds {first + 1}-{last}/{len(result)}"
```
csomh, line 173: From a quick search on this header, this seems to be for transferring big files (so that clients can continue a download), but in this case we talk pagination. So I think this is somewhat misleading HTTP-talk :)

I'll look more into it.

packit_service/service/api/testing_farm.py
```diff
@@ -134,3 +137,40 @@ def validate_testing_farm_request():
...
+                "pipeline_id": tf_result.pipeline_id,
+                "ref": tf_result.commit_sha,
+                "status": tf_result.status,
+                "target": tf_result.target,
+                "web_url": tf_result.web_url,
+                "pr_id": tf_result.get_pr_id(),
+            }
+
+            project = tf_result.get_project()
+            if project:
```
csomh, line 166: When will this happen?

packit_service/models.py
```diff
@@ -873,6 +885,13 @@ def get_by_pipeline_id(cls, pipeline_id: str) -> Optional["TFTTestRunModel"]:
                 .first()
             )
 
+    @classmethod
+    def get_all(cls, first: int, last: int) -> Optional[Iterable["TFTTestRunModel"]]:
```
csomh, line 889: This is using `first` and `last` so it will never be really `all`, is it? :smile: 

How about `get_range()` or something similar?

packit_service/service/api/testing_farm.py
IceWreck: TestingFarmResults is a POST webhook at /testing-farm/results

Its not mentioned here but /get-results means /testing-farm/get-results

packit_service/models.py
IceWreck: Sure

packit_service/service/api/testing_farm.py
IceWreck: I think almost always. If project is a check to see if its not None / exists. It might be redundant because a Test Result will always have a git project associated with it

packit_service/service/api/testing_farm.py
csomh: Some Git-archeology reveals that using the `Content-Range` header was started in #187, which aimed to fix #185. The later makes some references to some Stackoverflow questions about REST pagination.

Following links from there, it seems to me, that the original intention was to do pagination as described in [this blog post](http://otac0n.com/blog/2012/11/21/range-header-i-choose-you.html), but implementation wasn't really carried out all the way.

Anyways... this doesn't have to be fixed in this PR, but please use `{first+1}-{last}/*` in the answer or get a count of all the test results if that's not to expensive.

Also: the unit here should be `test-results` instead of `copr-builds`.

packit_service/service/api/testing_farm.py
csomh: Yeah, I know that, but usually in a REST API one can use the same endpoint to both save data in the service or retrieve it, and this is the effect that we need here: POST to /testing-farm/results will save results, GET to /testing-farm/results will list them.

packit_service/service/api/testing_farm.py
csomh: Okay, sorry, wrong question :) I wanted to ask when 'project' will be not true-ish :)

So if this is expected to be redundant, then let's remove it.

packit_service/service/api/testing_farm.py
IceWreck: Oh, sorry. I thought you meant something else.

PR #681, opened 2020-06-16
Count GitHub webhook calls

csomh:
Set up Prometheus Client to serve the `/metrics` endpoint, and
instrument the GitHub webhook to count the number of calls.

Use a label called 'result' to mark the different outcomes of these
calls.

### DIFF
--- a/packit_service/service/app.py
+++ b/packit_service/service/app.py
@@ -25,6 +25,8 @@ from os import getenv
 
 from flask import Flask
 from lazy_object_proxy import Proxy
+from werkzeug.middleware.dispatcher import DispatcherMiddleware
+from prometheus_client import make_wsgi_app as prometheus_app
 from packit.utils import set_logging
 
 from packit_service.config import ServiceConfig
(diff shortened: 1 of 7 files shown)


packit_service/service/app.py
```diff
@@ -64,8 +66,10 @@ def get_flask_application():
     return app
 
 
-application = Proxy(get_flask_application)
+packit_as_a_service = Proxy(get_flask_application)
 
+# Make Prometheus Client serve the /metrics endpoint
+application = DispatcherMiddleware(packit_as_a_service, {"/metrics": prometheus_app()})
```
TomasTomecek, line 72: I naively thought it's /api/metrics; had to look here to realize it's really just /metrics

dhodovsk: same here :D

csomh: :see_no_evil: sorry guys, I'm not even sure if `/api/metrics` would have been possible, and my logic was that `/metrics` is not really part of the application API (but I'm aware that that's debatable).

PR #689, opened 2020-06-18
fix requre tests after refactoring

jscotka:

PR #690, opened 2020-06-18
Add CORS Headers to the API

IceWreck:
Right now we dont have CORS headers so essentially we do not allow API requests from other websites. This doesnt affect curl/requests/etc since they ignore them. However browsers do respect them so theyre blocking requests from javascript.

According to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSMissingAllowOrigin), its okay for public APIs to wildcard allow API requests from anywhere.

### DIFF
--- a/packit_service/service/api/copr_builds.py
+++ b/packit_service/service/api/copr_builds.py
@@ -76,6 +76,7 @@ class CoprBuildsList(Resource):
         resp = make_response(dumps(result), HTTPStatus.PARTIAL_CONTENT)
         resp.headers["Content-Range"] = f"copr-builds {first + 1}-{last}/*"
         resp.headers["Content-Type"] = "application/json"
+        resp.headers["Access-Control-Allow-Origin"] = "*"
 
         return resp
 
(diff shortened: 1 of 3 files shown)


IceWreck:
Does anyone know why this is failing?

```
ERROR tests_requre/openshift_integration/test_copr.py - AttributeError: type ...
ERROR tests_requre/openshift_integration/test_fedpkg.py - AttributeError: typ...
```

lbarcziova:
> Does anyone know why this is failing?

It should be fixed by #689 (the 2 tests are failing because of the refactoring in packit-service/requre)

IceWreck:
> > Does anyone know why this is failing?
> 
> It should be fixed by #689 (the 2 tests are failing because of the refactoring in packit-service/requre)

Okay, thanks.

IceWreck:
> OPTION requests and handle Access-Control-Request-* headers in the future or there is no need for that?

TBT, I did not know that this was a thing. But I looked it up and I dont think we need to do that. This is a simple request and browser based Javascript is able to fetch from packit-service API after just adding _Access-Control-Allow-Origin: *_  header in the API.

From what Ive understood  preflighted requests  aren't necessary for simple GET requests that fetch data

PR #698, opened 2020-06-24
Fix mistaken attributes in PagurePullRequestLabelHandler after refactor

lbarcziova:


lbarcziova:
I can look into it and add some more test cases in following PR

PR #704, opened 2020-06-29
Create Celery tasks to run handlers

lbarcziova:
- have one processing task (parses webhook to an event, makes checks - whitelist, package config presence,..) and then tasks for running handlers
- rename HandlersResults to general TaskResults (since these are results also from `process_message` task, not only handler tasks), change what is included in `details` in processing task
- create a function for each Celery task running specific handler
- get rid of event dict in `__init__` of handler classes (parse additional attributes in `run...handler` functions if there are any)
- for handlers which need original Event object move methods for parsing the event to events.py (Installation, KojiBuild, CoprBuild)

### DIFF
--- a/packit_service/service/events.py
+++ b/packit_service/service/events.py
@@ -719,10 +719,38 @@ class InstallationEvent(Event):
         self.sender_login = sender_login
         self.status = status
 
+    @classmethod
+    def from_event_dict(cls, event: dict):
+        account_type = event.get("account_type")
+        account_login = event.get("account_login")
+        sender_login = event.get("sender_login")
+        account_id = event.get("account_id")
+        account_url = event.get("account_url")
+        sender_id = event.get("sender_id")
+        created_at = event.get("created_at")
+        installation_id = event.get("installation_id")
+        repositories = event.get("repositories")
+
+        return InstallationEvent(
+            installation_id=installation_id,
+            account_login=account_login,
+            account_id=account_id,
+            account_url=account_url,
+            account_type=account_type,
+            created_at=created_at,
+            repositories=repositories,
+            sender_id=sender_id,
+            sender_login=sender_login,
+        )
+
     def get_dict(self, default_dict: Optional[Dict] = None) -> dict:
         result = super().get_dict()
         result["status"] = result["status"].value
         return result
+
+    @property
+    def package_config(self):
+        return None
 
     @property
     def project(self):
@@ -926,6 +954,26 @@ class KojiBuildEvent(AbstractForgeIndependentEvent):
                 self._identifier = self.commit_sha
         return self._identifier
 
+    @classmethod
+    def from_event_dict(cls, event: dict):
+        build_id = event.get("build_id")
+        state = KojiBuildState(event.get("state")) if event.get("state") else None
+        old_state = (
+            KojiBuildState(event.get("old_state")) if event.get("old_state") else None
+        )
+        start_time = event.get("start_time")
+        rpm_build_task_id = event.get("rpm_build_task_id")
+        completion_time = event.get("completion_time")
+
+        return KojiBuildEvent(
+            build_id=build_id,
+            state=state,
+            old_state=old_state,
+            rpm_build_task_id=rpm_build_task_id,
+            start_time=start_time,
+            completion_time=completion_time,
+        )
+
     def get_base_project(self) -> Optional[GitProject]:
         if self.pr_id is not None:
             if isinstance(self.project, PagureProject):
@@ -1059,6 +1107,28 @@ class CoprBuildEvent(AbstractForgeIndependentEvent):
             return None
         return cls(
             topic, build_id, build, chroot, status, owner, project_name, pkg, timestamp
+        )
+
+    @classmethod
+    def from_event_dict(cls, event: dict):
+        topic = event.get("topic")
+        project_name = event.get("project_name")
+        owner = event.get("owner")
+        build_id = event.get("build_id")
+        chroot = event.get("chroot")
+        timestamp = event.get("timestamp")
+        pkg = event.get("pkg")
+        status = event.get("status")
+
+        return CoprBuildEvent.from_build_id(
+            topic=topic,
+            build_id=build_id,
+            chroot=chroot,
+            status=status,
+            owner=owner,
+            project_name=project_name,
+            pkg=pkg,
+            timestamp=timestamp,
         )
 
     def pre_check(self):
@@ -1233,6 +1303,10 @@ class PullRequestLabelPagureEvent(AddPullRequestDbTrigger, AbstractPagureEvent):
         self.git_ref = None  # pr_id will be used for checkout
         self.commit_sha = commit_sha
         self.labels = labels
+
+    @property
+    def package_config(self):
+        return None
 
     def get_dict(self, default_dict: Optional[Dict] = None) -> dict:
         result = super().get_dict()
--- a/packit_service/service/events.py
+++ b/packit_service/service/events.py
@@ -735,26 +735,16 @@ class InstallationEvent(Event):
 
     @classmethod
     def from_event_dict(cls, event: dict):
-        account_type = event.get("account_type")
-        account_login = event.get("account_login")
-        sender_login = event.get("sender_login")
-        account_id = event.get("account_id")
-        account_url = event.get("account_url")
-        sender_id = event.get("sender_id")
-        created_at = event.get("created_at")
-        installation_id = event.get("installation_id")
-        repositories = event.get("repositories")
-
         return InstallationEvent(
-            installation_id=installation_id,
-            account_login=account_login,
-            account_id=account_id,
-            account_url=account_url,
-            account_type=account_type,
-            created_at=created_at,
-            repositories=repositories,
-            sender_id=sender_id,
-            sender_login=sender_login,
+            installation_id=event.get("installation_id"),
+            account_login=event.get("account_login"),
+            account_id=event.get("account_id"),
+            account_url=event.get("account_url"),
+            account_type=event.get("account_type"),
+            created_at=event.get("created_at"),
+            repositories=event.get("repositories"),
+            sender_id=event.get("sender_id"),
+            sender_login=event.get("sender_login"),
         )
 
     def get_dict(self, default_dict: Optional[Dict] = None) -> dict:
@@ -970,22 +960,17 @@ class KojiBuildEvent(AbstractForgeIndependentEvent):
 
     @classmethod
     def from_event_dict(cls, event: dict):
-        build_id = event.get("build_id")
-        state = KojiBuildState(event.get("state")) if event.get("state") else None
-        old_state = (
-            KojiBuildState(event.get("old_state")) if event.get("old_state") else None
-        )
-        start_time = event.get("start_time")
-        rpm_build_task_id = event.get("rpm_build_task_id")
-        completion_time = event.get("completion_time")
-
         return KojiBuildEvent(
-            build_id=build_id,
-            state=state,
-            old_state=old_state,
-            rpm_build_task_id=rpm_build_task_id,
-            start_time=start_time,
-            completion_time=completion_time,
+            build_id=event.get("build_id"),
+            state=KojiBuildState(event.get("state")) if event.get("state") else None,
+            old_state=(
+                KojiBuildState(event.get("old_state"))
+                if event.get("old_state")
+                else None
+            ),
+            rpm_build_task_id=event.get("rpm_build_task_id"),
+            start_time=event.get("start_time"),
+            completion_time=event.get("completion_time"),
         )
 
     def get_base_project(self) -> Optional[GitProject]:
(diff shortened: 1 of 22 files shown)


tests/integration/test_release_event.py
```diff
@@ -50,10 +52,19 @@ def test_dist_git_push_release_handle(github_release_webhook):
...
+    assert processing_results["details"]["event"]["trigger"] == "release"
+
+    event_dict = processing_results["details"]["event"]
+    package_config = processing_results["details"]["package_config"]
+    job = processing_results["details"]["matching_jobs"][0]
+
+    results = run_propose_downstream_handler(
+        package_config=package_config, event=event_dict, job_config=job,
+    )
+    assert first_dict_value(results)["success"]
```
lbarcziova, line 67: is there some better way to test this?

packit_service/worker/jobs.py
```diff
@@ -232,26 +231,28 @@ def process_jobs(self, event: Event) -> Dict[str, HandlerResults]:
...
             # we want to run handlers for all possible jobs, not just the first one
             for job_config in job_configs:
-                logger.debug(f"Running handler: {str(handler_kls)} for {job_config}")
-                event_dict = event.get_dict()
-                handler = handler_kls(
-                    package_config=event.package_config,
-                    job_config=job_config,
-                    data=EventData.from_event_dict(event_dict),
+                send_handler_task(
+                    task_name=handler_kls.task_name, event=event, job=job_config
```
jpopelka, line 242: Not questioning your design, just wondering:
- have you seen [group primitive of Celery](https://docs.celeryproject.org/en/stable/userguide/canvas.html#groups)?
- would using it give us any benefit?
- could it perhaps help us to get rid of our `TaskResults`, by using their [Group Results](https://docs.celeryproject.org/en/stable/userguide/canvas.html#group-results)?

lbarcziova: I have already seen it, but I will recheck it, if it can give us some benefit

lbarcziova: So, I do not consider Group Results really beneficial, we can get the success/failure (in Celery context) of the tasks by waiting for finishing the tasks and waiting for tasks within a task is not recommended in docs. In our `TaskResults` we can add additional info which describes the task. The difference in using group and the current implementation would be that group (tasks executing matching jobs) "Creates a group of tasks to be executed in parallel."  and in the current implementation for all matching jobs is `send_task` executed.

jpopelka: > we can get the success/failure (in Celery context) of the tasks by waiting for finishing the tasks and waiting for tasks within a task is not recommended in docs

Are you talking about the current state or about the possible 'group workflow'?

> In our `TaskResults` we can add additional info which describes the task.

I'm sorry, I confused `TaskResults`with `task_results`/`jobs_results`. I think you can ignore me then.

> group (tasks executing matching jobs) "Creates a group of tasks to be executed in parallel."

and I was thinking that that might be what we actually want, i.e. let Celery run them in parallel and collect results once all are done. But maybe not :-)

> in the current implementation for all matching jobs is `send_task` executed

yes, initially we were using `send_task`, because we were sending the task (there was only one) from service to workers and because the task was not registered in the service we were calling it by name (see Tip [here](https://docs.celeryproject.org/en/stable/userguide/calling.html#example)). But now when we send task from worker to worker(s), we can use any other mechanism (like described in [Groups](https://docs.celeryproject.org/en/stable/userguide/canvas.html#groups)), can't we?

Anyway, I don't want to argue here, I'll let this on you since you're now the Celery guru and my map of the related code is heavily outdated. I just want to suggest that when we are doing such a big change we can investigate whether there already isn't a better mechanism in Celery.

lbarcziova: > Are you talking about the current state or about the possible 'group workflow'?

About 'group workflow', in the current after sending the handlers tasks we end the processing task.

> and I was thinking that that might be what we actually want, i.e. let Celery run them in parallel and collect results once all are done. But maybe not :-)

Yes, it would be the best if we could collect the results once all are done, but if I understand the docs correctly, we would need to wait in the processing task for finishing all of the tasks.

> yes, initially we were using send_task, because we were sending the task (there was only one) from service to workers and because the task was not registered in the service we were calling it by name (see Tip here). But now when we send tasks from worker to worker(s), we can use any other mechanism (like described in Groups), can't we?

thanks for the explanation, now I understand it...so yes, then we can use other method

> Anyway, I don't want to argue here, I'll let this on you since you're now the Celery guru and my map of the related code is heavily outdated. I just wanted to suggest that when we are doing such a big change we can investigate whether there already isn't a better mechanism in Celery.

No arguing, I am glad for your feedback

csomh:
Should `CoprBuildEndHandler.run()` create tasks for starting the testing jobs?

packit_service/worker/tasks.py
```diff
@@ -74,3 +113,226 @@ def babysit_copr_build(self, build_id: int):
...
+        package_config=PackageConfigSchema().load_config(package_config),
+        job_config=JobConfigSchema().load_config(job_config),
+        data=EventData.from_event_dict(event),
+        koji_event=KojiBuildEvent.from_event_dict(event),
+    )
+    return get_handlers_task_results(handler.run_job(), event)
+
+
+def get_handlers_task_results(results: dict, event: dict):
+    return {"job": results, "event": event}
```
csomh, line 338: Why is the event dict included here?

lbarcziova: I can remove it, but then the results would look like:
`{'copr_build-20200701101150167006': {'success': True, 'details': {}}}`
which doesn't tell much in my opinion.

csomh: No need to remove, but maybe add a comment to explain that it is included to provide more context in the result.

lbarcziova:
> Should CoprBuildEndHandler.run() create tasks for starting the testing jobs?

For constistency it should, I changed it, but the tests should be better, now it is more workarounded

packit_service/worker/jobs.py
```diff
@@ -459,17 +445,39 @@ def process_message(
...
+        return processing_results or get_processing_results(event=event_object, jobs=[])
 
-        task_results = {"jobs": jobs_results, "event": event_object.get_dict()}
 
-        for v in jobs_results.values():
-            if not (v and v["success"]):
-                logger.warning(task_results)
-                logger.error(v["details"]["msg"])
-        return task_results
+def get_signature(task_name: str, event: Event, job: Optional[JobConfig]):
```
jpopelka, line 456: What about having this as a `JobHandler` class method?

Then you'd use it as
```python
  signatures = [handler_kls.get_signature(event=event, job=job) for job in jobs]
  group(signatures).apply_async()
```
instead of
```python
  signatures = [
      get_signature(task_name=handler_kls.task_name, event=event, job=job)
      for job in jobs
  ]
  group(signatures).apply_async()
```

packit_service/service/events.py
```diff
@@ -719,11 +733,39 @@ def __init__(
...
+            installation_id=installation_id,
+            account_login=account_login,
+            account_id=account_id,
+            account_url=account_url,
+            account_type=account_type,
+            created_at=created_at,
+            repositories=repositories,
+            sender_id=sender_id,
+            sender_login=sender_login,
+        )
```
lachmanfrantisek, line 758: I am thinking if we can avoid this manual forwarding.

At least, what about getting rid of the one-time variables and fill the attributes directly? Using `get` should be safe:

```suggestion
        return InstallationEvent(
            installation_id=event.get("installation_id"),
            account_login=event.get("account_login"),
            account_id=event.get("account_id"),
            account_url=event.get("account_url"),
            account_type=event.get("account_type"),
            created_at=event.get("created_at"),
            repositories=event.get("repositories"),
            sender_id=event.get("sender_id"),
            sender_login=event.get("sender_login"),
        )
```

Or what about expanding the dictionary? Are there any unused values?


```suggestion
        return InstallationEvent(**event)
```

packit_service/service/events.py
```diff
@@ -1061,6 +1123,28 @@ def from_build_id(
...
+
+        return CoprBuildEvent.from_build_id(
+            topic=topic,
+            build_id=build_id,
+            chroot=chroot,
+            status=status,
+            owner=owner,
+            project_name=project_name,
+            pkg=pkg,
+            timestamp=timestamp,
```
lachmanfrantisek, line 1145: Same question as above. Can we get rid of those one-time variables and, ideally, key hardcoding?

packit_service/worker/handlers/abstract.py
```diff
@@ -210,5 +212,26 @@ def project(self) -> Optional[GitProject]:
...
+        if self.pre_check():
+            current_time = datetime.now().strftime(DATETIME_FORMAT)
+            result_key = f"{job_type.value}-{current_time}"
+            job_results[result_key] = self.run_n_clean()
+            logger.debug("Job finished!")
+
+            for v in job_results.values():
+                if not (v and v["success"]):
+                    logger.warning(job_results)
+                    logger.error(v["details"]["msg"])
```
lachmanfrantisek, line 232: It's maybe copied from another place, but I would prefer longer names..;)

```suggestion
            for result in job_results.values():
                if not (result and result["success"]):
                    logger.warning(job_results)
                    logger.error(result["details"]["msg"])
```

---

Do we really want to log the whole `job_results` before each nonsuccessful result?

```python
logger.warning(job_results)
```

packit_service/worker/jobs.py
lachmanfrantisek, line 456: Can we type those objects?

packit_service/worker/jobs.py
```diff
@@ -459,17 +445,39 @@ def process_message(
...
+        task_name,
+        kwargs={
+            "package_config": dump_package_config(event.package_config),
+            "job_config": dump_job_config(job),
+            "event": event.get_dict(),
+        },
+    )
+
+
+def get_processing_results(event: Event, jobs: List[JobConfig], success: bool = True):
```
lachmanfrantisek, line 475:
```suggestion
def get_processing_results(event: Event, jobs: List[JobConfig], success: bool = True) -> TaskResults:
```

packit_service/worker/tasks.py
```diff
@@ -74,3 +113,240 @@ def babysit_copr_build(self, build_id: int):
     """ check status of a copr build and update it in DB """
     if not check_copr_build(build_id=build_id):
         self.retry()
+
+
+# tasks for running the handlers
+@celery_app.task(name="task.run_copr_build_start_handler")
```
lachmanfrantisek, line 119: Does it make sense to have enum for that? Or, what about using handler class names?

Just thinking if we can use fewer strings and make that more type-able...

packit_service/worker/tasks.py
```diff
@@ -74,3 +113,240 @@ def babysit_copr_build(self, build_id: int):
...
+        package_config=load_package_config(package_config),
+        job_config=load_job_config(job_config),
+        data=EventData.from_event_dict(event),
+        tests=tests,
+        result=result,
+        pipeline_id=pipeline_id,
+        log_url=log_url,
+        copr_chroot=copr_chroot,
+        message=message,
+    )
```
lachmanfrantisek, line 240: Same as above. Please, avoid one-time variables and, ideally, avoid getters at all.

packit_service/worker/tasks.py
```diff
@@ -74,3 +113,240 @@ def babysit_copr_build(self, build_id: int):
...
+    handler = PagurePullRequestLabelHandler(
+        package_config=load_package_config(package_config),
+        job_config=load_job_config(job_config),
+        data=EventData.from_event_dict(event),
+        labels=labels,
+        action=action,
+        base_repo_owner=base_repo_owner,
+        base_repo_name=base_repo_name,
+        base_repo_namespace=base_repo_namespace,
+    )
```
lachmanfrantisek, line 335: Same here.

packit_service/worker/jobs.py
lbarcziova: thanks, changed

packit_service/service/events.py
lbarcziova: using `InstallationEvent(**event)` would need to pop more attributes from the event dict before, so I changed it at least without using the one-time variables

packit_service/worker/handlers/abstract.py
lbarcziova: yes, this was copied, but changed now

packit_service/worker/tasks.py
lbarcziova: I am not sure what to do about this, I have not found any elegant way of getting rid of the strings :thinking:

csomh: Task names need to be strings. We could leave them out and [change the automatic naming behaviour](https://docs.celeryproject.org/en/stable/userguide/tasks.html#changing-the-automatic-naming-behavior) so we have names that makes sense for us, though honestly I'm not sure the benefits this effort would bring us worth it.

lachmanfrantisek: We can have string enums if we want.

My concerns about strings:

- hard to follow across the code (find usage,...)
- hard to rename (I know sed, but still.)
- fewer benefits from type-checkers

lbarcziova: `Task names need to be strings`

Yes, I meant it to get rid of hardcoding the strings, sorry. 

`I'm not sure the benefits this effort would bring us worth it.`
+1

lbarcziova: @lachmanfrantisek do you mean something like this?
```python3
class TaskName(str, enum.Enum):
    copr_build_start_handler = "task.run_copr_build_start_handler"
   ...
```

lachmanfrantisek: @lbarcziova looks good!

> Task names need to be strings

I am not against strings as types but against the need of writing the same strings multiple times. Sorry if I didn't make that clear at the start.

lbarcziova: @lachmanfrantisek enum added, PTAL

lbarcziova:
recheck

lbarcziova:
recheck

TomasTomecek:
YAY!

PR #707, opened 2020-06-29
Use thread-local sqlalchemy session

### LINKED ISSUE #706: API: use thread-local SQLAlchemy Sessions
While developing the Dashboard, we hit this issue where multiple API calls happening at the same time would cause Packit Service threads to crash with a traceback like bellow:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/flask/app.py", line 1950, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.7/site-packages/flask/app.py", line 1936, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/usr/local/lib/python3.7/site-packages/flask_restx/api.py", line 375, in wrapper
    resp = resource(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/flask/views.py", line 89, in view
    return self.dispatch_request(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/flask_restx/resource.py", line 44, in dispatch_request
    resp = meth(*args, **kwargs)
  File "/usr/local/lib/python3.7/site-packages/packit_service/service/api/copr_builds.py", line 55, in get
    for build in CoprBuildModel.get_merged_chroots(first, last):
  File "/usr/local/lib/python3.7/site-packages/packit_service/models.py", line 544, in get_merged_chroots
    return builds
  File "/usr/lib64/python3.7/contextlib.py", line 119, in __exit__
    next(self.gen)
  File "/usr/local/lib/python3.7/site-packages/packit_service/models.py", line 83, in get_sa_session
    session_instance.commit()
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/orm/session.py", line 1042, in commit
    self.transaction.commit()
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/orm/session.py", line 508, in commit
    t[1].commit()
  File "/usr/lib64/python3.7/site-packages/sqlalchemy/engine/base.py", line 1770, in commit
    raise exc.InvalidRequestError("This transaction is inactive")
sqlalchemy.exc.InvalidRequestError: This transaction is inactive
```

What seems to happen here is `commit()` being called twice on the same transaction. This will not work in an SQLAlchemy world.

The way this can happen is that Packit Service currently is using a single session object created by `get_sa_session()`. But [SQLAlchemy session are not thread-safe](https://docs.sqlalchemy.org/en/13/orm/session_basics.html#is-the-session-thread-safe), so sharing a single session between multiple Flask App threads it's not a smart thing to do.

SQLAlchemy though has a way to create [thread-local sessions](https://docs.sqlalchemy.org/en/13/orm/contextual.html#contextual-thread-local-sessions). Packit Service should use these in order to avoid the issue above.

@IceWreck will you have time to look into making this change?

IceWreck:
This may fix #706 or break everything else

[The official flask docs](https://flask.palletsprojects.com/en/1.1.x/patterns/sqlalchemy/?highlight=sqlalchemy#declarative)  are doing something like this.

It does fix that issue but may have unwanted side effects

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -45,7 +45,7 @@ from sqlalchemy import (
     Boolean,
 )
 from sqlalchemy.ext.declarative import declarative_base
-from sqlalchemy.orm import sessionmaker, Session, relationship
+from sqlalchemy.orm import sessionmaker, Session, relationship, scoped_session
 from sqlalchemy.types import PickleType, ARRAY
 from sqlalchemy.dialects.postgresql import array as psql_array
 
@@ -66,24 +66,20 @@ def get_pg_url() -> str:
     )
 
 
+engine = create_engine(get_pg_url())
+ScopedSession = scoped_session(sessionmaker(bind=engine))
+
+
 @contextmanager
 def get_sa_session() -> Session:
     """ get SQLAlchemy session """
-    # we need to keep one session for all the operations b/c SA objects
-    # are bound to this session and we can use them, otherwise we'd need
-    # add objects into all newly created sessions:
-    #   Instance <PullRequest> is not bound to a Session; attribute refresh operation cannot proceed
-    global session_instance
-    if session_instance is None:
-        engine = create_engine(get_pg_url())
-        Session = sessionmaker(bind=engine)
-        session_instance = Session()
+    session = ScopedSession()
     try:
-        yield session_instance
-        session_instance.commit()
+        yield session
+        session.commit()
     except Exception as ex:
         logger.warning(f"Exception while working with database: {ex!r}")
-        session_instance.rollback()
+        session.rollback()
         raise
 
 


IceWreck:
Surprisingly 😲, tests passed.

@TomasTomecek your comment is still bothering me

```python
@contextmanager
def get_sa_session() -> Session:
    """ get SQLAlchemy session """
    # we need to keep one session for all the operations b/c SA objects
    # are bound to this session and we can use them, otherwise we'd need
    # add objects into all newly created sessions:
    #   Instance <PullRequest> is not bound to a Session; attribute refresh operation cannot proceed
    global session_instance
    if session_instance is None:
        engine = create_engine(get_pg_url())
```

packit_service/models.py
```diff
@@ -76,8 +76,7 @@ def get_sa_session() -> Session:
     global session_instance
```
csomh, line 76: Due to this line I would say we'll still have a global session shared between threads, so it won't be thread local :)

TomasTomecek:
LGTM if tests pass :)

packit_service/models.py
csomh: How about storing the session in [Flask.g](https://flask.palletsprojects.com/en/1.1.x/appcontext/#storing-data)?

IceWreck: I dont understand why this needs to be global but it fixed that issue, i tested it multiple times.

sakalosj: it should not be global, scoped session is already handling it - it will always return same session in same thread, regardless how many times it is called.

https://docs.sqlalchemy.org/en/13/orm/contextual.html

sakalosj: also some useful FAQ
https://docs.sqlalchemy.org/en/13/orm/session_basics.html#session-faq-whentocreate

IceWreck: >  scoped session is already handling it - it will always return same session in same thread, regardless how many times it is called.

thanks

IceWreck: on removing the global declaration, I get 

is not bound to a Session; attribute refresh operation cannot proceed (Background on this error at: http://sqlalche.me/e/bhk3),

csomh: Can we have the following outside `get_sa_sesssion()`?

```python
engine = create_engine(get_pg_url())
Session = scoped_session(sessionmaker(bind=engine))
```

(not sure if this would matter...)

IceWreck:
Okay @csomh @sakalosj @TomasTomecek 
this seems to be working

```
engine = create_engine(get_pg_url())
session_instance = scoped_session(sessionmaker(bind=engine))

@contextmanager
def get_sa_session() -> Session:
    """ get SQLAlchemy session """
    # This comment may or maynot be outdated

    # we need to keep one session for all the operations b/c SA objects
    # are bound to this session and we can use them, otherwise we'd need
    # add objects into all newly created sessions:
    #   Instance <PullRequest> is not bound to a Session; attribute refresh operation cannot proceed
    try:
        yield session_instance
        session_instance.commit()
    except Exception as ex:
        logger.warning(f"Exception while working with database: {ex!r}")
        session_instance.rollback()
        raise


```

csomh:
@IceWreck please try:

```python
engine = create_engine(get_pg_url())
Session = scoped_session(sessionmaker(bind=engine))

@contextmanager
def get_sa_session() -> Session:
    """ get SQLAlchemy session """
    session = Session()
    try:
        yield session
        session.commit()
    except Exception as ex:
        logger.warning(f"Exception while working with database: {ex!r}")
        session.rollback()
        raise
```

IceWreck:
> @IceWreck please try:

Yep, working well.

csomh:
> Yep, working well.

:tada: Let's update this PR then with this code and see what the tests do :)

PR #713, opened 2020-07-03
Gitlab events dont require whitelist checks

shreyaspapi:


csomh:
> @packit-service/the-packit-team Do we want that?

This should be fine fine. If I'm not wrong this would simplify the process of adding new users, right?

shreyaspapi:
> @shreyaspapi What was your reason for disabling whitelist for GitLab?

I didn't feel the need of it, as you said currently we add user's webhook-token to our config. We don't really require the whitelist that this moment.

> The `.DS_Store` is probably unnecessary here.

Yeah! sorry, will remove this

> This should be fine fine. If I'm not wrong this would simplify the process of adding new users, right?
@csomh Yes exactly,  when beginning with the work on authenticating webhook tokens we can discuss onboarding flow and revert it later.

shreyaspapi:
Rebased!

PR #717, opened 2020-07-07
Breaking image dependency: pacit-service-worker <- packit

sakalosj:
Docker and ansible configs updated, so dependencies are installed, 
without requirement to have them in base container image.

  Deployment related configuration was improved, so same branch for
all internal projects installed form github repositories will be used.
  This is achieved using docker build hook (files/docker/hooks)

### DIFF
--- a/Dockerfile.worker
+++ b/Dockerfile.worker
@@ -1,6 +1,6 @@
 # Celery worker which runs tasks (packit) from the web service
 
-FROM docker.io/usercont/packit
+FROM fedora:32
 
 ENV LANG=en_US.UTF-8 \
     ANSIBLE_PYTHON_INTERPRETER=/usr/bin/python3 \
@@ -9,9 +9,11 @@ ENV LANG=en_US.UTF-8 \
     HOME=/home/packit \
     PS_PATH=/src-packit-service
 
+
 # Ansible doesn't like /tmp
 COPY files/install-deps-worker.yaml $PS_PATH/files/
-RUN cd $PS_PATH/ \
+RUN dnf install -y ansible \
+    && cd $PS_PATH/ \
     && ansible-playbook -vv -c local -i localhost, files/install-deps-worker.yaml \
     && dnf clean all
 
--- a/Dockerfile.worker.prod
+++ b/Dockerfile.worker.prod
@@ -1,7 +1,7 @@
 # Celery worker which runs tasks (packit) from the web service
 # The only difference between this and Dockerfile.worker is the ":prod" on first line
 
-FROM docker.io/usercont/packit:prod
+FROM fedora:32
 
 ENV LANG=en_US.UTF-8 \
     ANSIBLE_PYTHON_INTERPRETER=/usr/bin/python3 \
@@ -10,9 +10,11 @@ ENV LANG=en_US.UTF-8 \
     HOME=/home/packit \
     PS_PATH=/src-packit-service
 
+
 # Ansible doesn't like /tmp
 COPY files/install-deps-worker.yaml $PS_PATH/files/
-RUN cd $PS_PATH/ \
+RUN dnf install -y ansible \
+    && cd $PS_PATH/ \
     && ansible-playbook -vv -c local -i localhost, files/install-deps-worker.yaml \
     && dnf clean all
 
--- a/files/docker/Dockerfile
+++ b/files/docker/Dockerfile
@@ -1,0 +1,38 @@
+# Image for the web service (httpd), for celery worker see Dockerfile.worker
+
+FROM fedora:32
+
+ARG SOURCE_BRANCH
+
+ENV LANG=en_US.UTF-8 \
+    ANSIBLE_PYTHON_INTERPRETER=/usr/bin/python3 \
+    ANSIBLE_STDOUT_CALLBACK=debug \
+    USER=packit \
+    HOME=/home/packit
+
+RUN dnf install -y ansible
+
+COPY files/install-deps.yaml /src/files/
+
+RUN cd /src/ \
+    && ansible-playbook -vv -c local -i localhost, files/install-deps.yaml \
+    && dnf clean all
+
+COPY setup.py setup.cfg files/recipe.yaml files/tasks/httpd.yaml files/tasks/common.yaml files/packit.wsgi files/run_httpd.sh files/setup_env_in_openshift.sh files/packit-httpd.conf /src/
+# setuptools-scm
+COPY .git /src/.git
+COPY packit_service/ /src/packit_service/
+
+RUN cd /src/ \
+    && git rev-parse HEAD >/.packit-service.git.commit.hash \
+    && git show --quiet --format=%B HEAD >/.packit-service.git.commit.message \
+    && ansible-playbook -vv -c local -i localhost, recipe.yaml
+
+# no need to rm /src, it will stay in the image anyway
+
+COPY alembic.ini /src/
+COPY alembic/ /src/alembic/
+
+EXPOSE 8443
+
+CMD ["/usr/bin/run_httpd.sh"]
--- a/files/docker/README.md
+++ b/files/docker/README.md
@@ -1,0 +1,1 @@
+For details about build configuration please check [deployment documentation](https://github.com/packit-service/deployment#image-build-process)
--- a/files/install-deps-worker.yaml
+++ b/files/install-deps-worker.yaml
@@ -1,7 +1,15 @@
 ---
 - name: Install dependencies for packit-service worker.
   hosts: all
+  vars:
+    # added default value to not break functionality during implementing
+    # https://github.com/packit-service/packit-service/pull/717
+    # have to be removed once implemented
+    source_branch: "{{ lookup('env', 'SOURCE_BRANCH') or 'master' }}"
   tasks:
+    - name: Check source_branch value
+      debug:
+        msg: source_branch is set to {{ source_branch }}
     - name: Install all RPM/python packages needed to run packit-service worker.
       dnf:
         name:
@@ -28,7 +36,7 @@
           # provides dnf builddep, required for packit dep installation from spec file
           - dnf-plugins-core
         state: present
-    - name: Install dependencies for pakcit
+    - name: Install dependencies for packit
       block:
         - name: Download script setupcfg2rpm.py
           get_url:
@@ -37,7 +45,7 @@
             mode: "0744"
         - name: Download packit spec file
           get_url:
-            url: https://raw.githubusercontent.com/packit-service/packit/master/packit.spec
+            url: https://raw.githubusercontent.com/packit-service/packit/{{ source_branch }}/packit.spec
             dest: ./packit.spec
         - name: Install packit RPM build dependencies from packit.spec
           shell: dnf builddep packit.spec -y
@@ -50,9 +58,9 @@
           become: true
         - name: Download packit setup.cfg
           get_url:
-            url: https://raw.githubusercontent.com/packit-service/packit/master/setup.cfg
+            url: https://raw.githubusercontent.com/packit-service/packit/{{ source_branch }}/setup.cfg
             dest: ./packit_setup.cfg
-        - name: Install depndencies provided by setupcfg2rpm
+        - name: Install dependencies provided by setupcfg2rpm
           shell: dnf install $(./setupcfg2rpm.py packit_setup.cfg)
           become: true
     - name: Install dependencies from setup.cfg as rpm packages for ogr
@@ -60,9 +68,10 @@
         # setupcfg2rpm.py was downloaded in previous block
         - name: Download ogr setup.cfg
           get_url:
+            # ogr has only master branch
             url: https://raw.githubusercontent.com/packit-service/ogr/master/setup.cfg
             dest: ./ogr_setup.cfg
-        - name: Install dependencies provided by setupcfg2rpm
+        - name: Install ogr dependencies provided by setupcfg2rpm
           shell: dnf install $(./setupcfg2rpm.py ogr_setup.cfg)
           args:
             warn: no
@@ -74,11 +83,12 @@
           - sentry-sdk
           - flask-restx
         executable: pip3
-    # to avoid unwanted pip packages installations, because of not installed rpms
-    - name: Install pip deps with --no-deps
+    # --no-deps: to fail instead of installing from PyPI when we forget to add some dependency to packit.spec
+    - name: pip install packit & ogr with --no-deps
       pip:
         name:
-          - git+https://github.com/packit-service/packit.git
+          - git+https://github.com/packit-service/packit.git@{{ source_branch }}
+          # ogr has only master branch
           - git+https://github.com/packit-service/ogr.git
         executable: pip3
         extra_args: --no-deps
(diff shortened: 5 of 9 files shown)


Dockerfile.worker
```diff
@@ -9,6 +9,8 @@ ENV LANG=en_US.UTF-8 \
     HOME=/home/packit \
     PS_PATH=/src-packit-service
 
+RUN dnf install -y ansible && dnf clean all -y
```
jpopelka, line 12: Don't clean the `dnf` cache here, otherwise, the next step (`ansible-playbook files/install-deps-worker.yaml`) will need to download all the metadata again.

```suggestion
RUN dnf install -y ansible
```

Dockerfile.worker.prod
```diff
@@ -10,6 +10,8 @@ ENV LANG=en_US.UTF-8 \
     HOME=/home/packit \
     PS_PATH=/src-packit-service
 
+RUN dnf install -y ansible && dnf clean all -y
```
jpopelka, line 13: same as above

files/install-deps-worker.yaml
```diff
@@ -22,11 +22,26 @@
...
+      become: true
     - name: Install pip deps
       pip:
         name:
           - git+https://github.com/packit-service/sandcastle.git
           - sentry-sdk==0.14.4
           - flask-restx
         executable: pip3
+    - name: Install packit from github with --no-deps
+      # ansible pip doesnt have --no-deps switch
```
jpopelka, line 46: And why do we actually need the `--no-deps`? If all the deps were installed in previous steps, then they won't be installed in this step anyway. Or is that so much faster with `--no-deps`?

files/install-deps-worker.yaml
```diff
@@ -22,11 +22,26 @@
...
+          - python3-pip
+          - make
         state: present
+    - name: Download packit spec file
+      get_url:
+        url: https://raw.githubusercontent.com/packit-service/packit/master/packit.spec
+        # this should be improved - zuul is using var peoject_dir - can be used also in non-zuul
+        # runs - if not set -> set?, will be there any zuul runs conflicts?
+        dest: ./packit.spec
+    - name: Build packit RPM dependencies based on packit.spec
```
jpopelka, line 35:
```suggestion
    - name: Install packit RPM build dependencies
```

But I think we said the `BuildRequires` are not enough and we want to install `Requires`. @csomh suggested:
```sh
dnf install $(rpmspec -q --requires packit.spec | grep -v 'packit')
```

Dockerfile.worker
lachmanfrantisek: Or merge those two `RUN` commands to have it in one layer and not saved in the final image. (And move the `COPY` before that.)

files/install-deps-worker.yaml
sakalosj: we agreed that we will not install any deps via pip and this way we will avoid to install any forgotten depndecy, but if this is ok i also prefer to put it to existing ansible pip install  block

Dockerfile.worker
sakalosj: fixed

Dockerfile.worker.prod
sakalosj: fixed

files/install-deps-worker.yaml
sakalosj: fixed

files/install-deps-worker.yaml
```diff
@@ -22,11 +22,26 @@
...
+          - python3-pip
+          - make
         state: present
+    - name: Download packit spec file
+      get_url:
+        url: https://raw.githubusercontent.com/packit-service/packit/master/packit.spec
+        # this should be improved - zuul is using var peoject_dir - can be used also in non-zuul
+        # runs - if not set -> set?, will be there any zuul runs conflicts?
+        dest: ./packit.spec
+    - name: Install packit RPM build dependencies
```
jpopelka, line 35:
```suggestion
    - name: Install packit RPM dependencies
```

files/install-deps-worker.yaml
jpopelka: Ok, I probably missed that discussion, because I don't remember it.
I don't know what's `'forgotten dependency'`, my view is that:
(as I already said in previous comment) If all deps are installed in the previous `Install packit RPM dependencies` task, then this `pip install` would NOT install any other dependencies.
And if it installs any, then it means that something is missing in the previous step.

sakalosj: you are right, will implement it as you suggested

lachmanfrantisek: But isn't it good if it fails instead of the installation from PyPI when forget to add something to the spec?

We will not get that info when silently installing missing deps from PyPI.

jpopelka: Ok, that's a valid point. Then I'm fine with `--no-deps` as long as the reason for using it is documented in the playbook.

files/install-deps-worker.yaml
```diff
@@ -2,6 +2,15 @@
...
   hosts: all
   tasks:
+    - name: Download packit spec file
+      get_url:
+        url: https://raw.githubusercontent.com/packit-service/packit/master/packit.spec
+        # this should be improved - zuul is using var peoject_dir - can be used also in non-zuul
+        # runs - if not set -> set?, will be there any zuul runs conflicts?
+        dest: ./packit.spec
+    - name: Install packit RPM build dependencies
+      command: dnf install $(rpmspec -q --requires packit.spec | grep -v 'packit')
```
jpopelka, line 12: Thanks, but there's actually needed more than this:
See https://github.com/packit-service/research/pull/52#discussion_r448998138 and `rpm -q --requires python3-packit`.

Maybe the ['local packit rpm build' idea](https://github.com/packit-service/research/pull/52#discussion_r449002033) would be the easiest to implement after all. Have you tried that?

files/install-deps-worker.yaml
```diff
@@ -22,11 +31,15 @@
...
+          - python3-pip
+          - make
         state: present
     - name: Install pip deps
       pip:
         name:
           - git+https://github.com/packit-service/sandcastle.git
           - sentry-sdk==0.14.4
           - flask-restx
+          - git+https://github.com/packit-service/packit.git
```
jpopelka, line 44: As [Hunor pointed out in the research](https://github.com/packit-service/research/pull/52#discussion_r448181457) we want to install packit from `master` (branch) when building for stage and from `stable` when building for prod.

This playbook (`install-deps-worker.yaml`) is run by both, `Dockerfile.worker` and `Dockerfile.worker.prod` so we need to somehow pass the information about whether we want packit from `master` or `stable` into this playbook and act according to it.

The same applies to installing packit dependencies.

Dockerfile.worker
```diff
@@ -1,6 +1,6 @@
 # Celery worker which runs tasks (packit) from the web service
 
-FROM docker.io/usercont/packit
+FROM fedora:31
```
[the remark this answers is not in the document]
sakalosj: fixed

Dockerfile.worker.prod
```diff
@@ -1,7 +1,7 @@
 # Celery worker which runs tasks (packit) from the web service
 # The only difference between this and Dockerfile.worker is the ":prod" on first line
 
-FROM docker.io/usercont/packit:prod
+FROM fedora:31
```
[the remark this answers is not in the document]
sakalosj: fixed

sakalosj:
recheck

files/install-deps-worker.yaml
sakalosj: i tried local build -  it required to install build deps using `dnf builddep file.spec`
so i cloned packit-service/packit and installed builddeps  using `dnf builddep file.spec`,  build rpm using `rpmbuild -ba --build-in-place packit.spec --define "_rpmdir ."`  then i was comparing what `dnf install $(rpmspec -q --requires packit.spec| grep -v packit)` and `rpm -q --requires python3-packit` want to install using locally build rpms:
`dnf install $(rpmspec -q --requires packit.spec| grep -v packit)` wanted to install `python3-packit dependencies` + fedpkg, python, rpmdevtools.

Not 100% sure if all of this additional packages are required (fedpkg, python, rpmdevtools), but in any case this is most straightforward approach:

`dnf builddep file.spec`
`dnf install $(rpmspec -q --requires packit.spec| grep -v packit)`

this 2 commands installed all require dependencies. 

WDYT?

jpopelka: I see the following in `rpm -q --requires python3-packit`:
```
python3.8dist(munch)
python3.8dist(pygithub)
python3.8dist(requests)
```
they are not part of `dnf builddep packit.spec` or `rpmspec -q --requires packit.spec`.

I'm wondering whether we should add them as `BuildRequires`?

sakalosj: i just checked that 
`python3.8dist(munch)`
`python3.8dist(pygithub)`
`python3.8dist(requests)`
are installed with `dnf builddep file.spec`

jpopelka: They are pulled in by other modules:
`pygithub` -> `python3-ogr`, `munch`&`requests` -> `python3-copr`
```
repoquery -q --requires python3-copr
repoquery -q --requires python3-ogr
```

I see [munch](https://github.com/packit-service/packit/blob/master/packit/copr_helper.py#L30) and [requests](https://github.com/packit-service/packit/blob/master/packit/distgit.py#L31) used in code, so I'd add those two.
I don't see `pygithub` used in code so I'd try to remove it from [setup.cfg](https://github.com/packit-service/packit/blob/master/setup.cfg#L34).

sakalosj: created PR for this https://github.com/packit-service/packit/pull/899

files/install-deps-worker.yaml
sakalosj: implemented

files/docker/Dockerfile
```diff
@@ -0,0 +1,38 @@
+# Image for the web service (httpd), for celery worker see Dockerfile.worker
+
+FROM fedora:31
```
csomh, line 3: 32, please :)

files/docker/Dockerfile
```diff
@@ -0,0 +1,38 @@
+# Image for the web service (httpd), for celery worker see Dockerfile.worker
```
csomh, line 1: How are these files different from the ones in the repository root? Is the build configuration need to be updated to use these files?

files/docker/Dockerfile
sakalosj: sorry :)

files/docker/Dockerfile
sakalosj: this change require hooks directory in Dockerfile location. To avoid messy project root dir i have decided to move Dockerfiles and hooks to dedicated location. 

Because of dockerhub autobuilds depends on locations of Dockerfile's, this is required to perform "smooth"  transition. Once this PR is merged, i will configure dockerhub to consume Dockerfiles in new location, test it, document it  and then remove the ones from project root

files/install-deps-worker.yaml
sakalosj: implemented with --no-deps

files/install-deps-worker.yaml
```diff
@@ -22,11 +24,66 @@
           - python3-bugzilla # python-bugzilla (not bugzilla) on PyPI
           - python3-backoff # Bugzilla class
           #- python3-flask-restx # needs Fedora 32
+          - dnf-utils
+          - python3-pip
+          - make
+          # provides dnf builddep, required for packit dep installation from spec file
+          - dnf-plugins-core
         state: present
+    - name: Install dependencies for pakcit
```
jpopelka, line 33:
```suggestion
    - name: Install dependencies for packit
```

files/docker/Dockerfile
csomh: Would it make sense to add a short README in files/docker to explain the usage? I had to read the [docs](https://docs.docker.com/docker-hub/builds/advanced/#custom-build-phase-hooks) twice to understand how will this work :joy_cat:

files/install-deps-worker.yaml
```diff
@@ -22,11 +24,66 @@
...
+        - name: Install packit RPM runtime dependencies
+          shell: dnf install $(rpmspec -q --requires packit.spec | grep -v 'packit') -y
+          args:
+            warn: no
+          become: true
+        - name: Download packit setup.cfg
+          get_url:
+            url: https://raw.githubusercontent.com/packit-service/packit/master/setup.cfg
+            dest: ./packit_setup.cfg
+        - name: Install depndencies provided by setupcfg2rpm
```
jpopelka, line 57:
```suggestion
        - name: Install dependencies provided by setupcfg2rpm
```

files/install-deps-worker.yaml
```diff
@@ -22,11 +24,66 @@
...
+          shell: dnf install $(./setupcfg2rpm.py packit_setup.cfg)
+          become: true
+    - name: Install dependencies from setup.cfg as rpm packages for ogr
+      block:
+        # setupcfg2rpm.py was downloaded in previous block
+        - name: Download ogr setup.cfg
+          get_url:
+            url: https://raw.githubusercontent.com/packit-service/ogr/master/setup.cfg
+            dest: ./ogr_setup.cfg
+        - name: Install dependencies provided by setupcfg2rpm
```
jpopelka, line 67:
```suggestion
        - name: Install ogr dependencies provided by setupcfg2rpm
```

files/install-deps-worker.yaml
```diff
@@ -22,11 +24,66 @@
...
+            warn: no
+          become: true
     - name: Install pip deps
       pip:
         name:
           - git+https://github.com/packit-service/sandcastle.git
           - sentry-sdk
           - flask-restx
         executable: pip3
+    # to avoid unwanted pip packages installations, because of not installed rpms
```
jpopelka, line 79:
```suggestion
    # --no-deps: to avoid unwanted pip packages installations, because of not installed rpms
```

files/install-deps-worker.yaml
```diff
@@ -22,11 +24,66 @@
...
+    - name: Install pip deps with --no-deps
```
jpopelka, line 80:
```suggestion
    - name: pip install packit & ogr with --no-deps
```

files/docker/Dockerfile
sakalosj: sure - will update docs

sakalosj: documented in deployment as there is all related docu and added README.md with link to docu (https://github.com/packit-service/deployment/pull/108)

files/docker/Dockerfile
sakalosj: fixed

files/docker/README.md
```diff
@@ -0,0 +1 @@
+For details about builds configuration please check [deployment documentation](https://github.com/packit-service/deployment#image-build-process)
```
csomh, line 1:
```suggestion
For details about build configuration please check [deployment documentation](https://github.com/packit-service/deployment#image-build-process)
```

files/install-deps-worker.yaml
```diff
@@ -22,11 +30,67 @@
...
+            warn: no
+          become: true
     - name: Install pip deps
       pip:
         name:
           - git+https://github.com/packit-service/sandcastle.git
           - sentry-sdk
           - flask-restx
         executable: pip3
+    # to avoid unwanted pip packages installations, because of not installed rpm's
```
jpopelka, line 86:
```suggestion
    # --no-deps: to fail instead of installing from PyPI when we forget to add some dependency to packit.spec
```

PR #727, opened 2020-07-08
Allow to specify redis password

jpopelka:
This is needed if the Redis (broker) uses password authentication.
https://stackoverflow.com/questions/7537905/redis-set-a-password-for-redis


csomh:
Please tell in the commit message why this is required. It's not readily obvious. Thanks! :smiley:

PR #736, opened 2020-07-14
Use the last release tag as the commit_sha in IssueCommentEvents

csomh:
Currently getting the package configuration from IssueCommentEvents
fails as there is commit_sha is set to None, so there is no ref to be
checked out.

Packit commands in issues work with the last release from the repo, so
it's safe to return 'tag_name' as the 'commit_sha'.

### DIFF
--- a/packit_service/service/events.py
+++ b/packit_service/service/events.py
@@ -690,7 +690,6 @@ class IssueCommentEvent(AddIssueDbTrigger, AbstractGithubEvent):
         self.user_login = user_login
         self.comment = comment
         self.identifier = str(issue_id)
-        self.commit_sha = None
 
     @property
     def tag_name(self):
@@ -698,6 +697,10 @@ class IssueCommentEvent(AddIssueDbTrigger, AbstractGithubEvent):
             releases = self.project.get_releases()
             self._tag_name = releases[0].tag_name if releases else ""
         return self._tag_name
+
+    @property
+    def commit_sha(self):
+        return self.tag_name
 
     def get_dict(self, default_dict: Optional[Dict] = None) -> dict:
         result = super().get_dict()
--- a/tests/unit/test_events.py
+++ b/tests/unit/test_events.py
@@ -465,13 +465,16 @@ class TestEvents:
         assert event_object.project.full_repo_name == "packit-service/packit"
         assert not event_object.base_project
 
+        flexmock(event_object.project).should_receive("get_releases").and_return(
+            [flexmock(tag_name="0.5.0"), flexmock(tag_name="0.4.1")]
+        )
         flexmock(PackageConfigGetter).should_receive(
             "get_package_config_from_repo"
         ).with_args(
             base_project=event_object.base_project,
             project=event_object.project,
             pr_id=None,
-            reference=None,
+            reference="0.5.0",
             fail_when_missing=False,
             spec_file_path=None,
         ).and_return(


csomh:
Need help with writing/updating tests for this :smiley:

packit_service/worker/handlers/github_handlers.py
```diff
@@ -612,6 +599,20 @@ def run(self) -> TaskResults:
             self.project.issue_comment(self.db_trigger.issue_id, msg)
             return TaskResults(success=False, details={"msg": "Propose update failed"})
 
+        local_project = LocalProject(
+            git_project=self.project,
+            working_dir=self.service_config.command_handler_work_dir,
+            ref=self.data.tag_name,
```
TomasTomecek, line 605: in tests, I'd make sure that tag_name is properly parsed from the webhook and set in the Event

csomh: Yeah, actually the failing test above needed just that to be fixed.

lbarcziova:
I may be wrong, but getting a package config should happen before running the actual handler, so I think the original issue will still be there :thinking:

csomh:
> I may be wrong, but getting a package config should happen before running the actual handler, so I think the original issue will still be there

Sorry, I don't follow :confused: 

I've linked the Sentry issue which I'm trying to fix with this change.

csomh:
@lbarcziova I've updated this PR so that commit_sha returns the tag_name in IssueCommentEvents. This way 'event.package_config' will have the correct reference to get the package configuration from. PTAL.

I'm not sure, if the previous change (explicitly specifying the ref when creating LocalProject in the task) is still required... What do you think?

lbarcziova:
> I'm not sure, if the previous change (explicitly specifying the ref when creating LocalProject in the task) is still required... What do you think?

I am also not sure about this, can we try it on stg like this?

packit_service/service/events.py
```diff
@@ -699,6 +698,10 @@ def tag_name(self):
             self._tag_name = releases[0].tag_name if releases else ""
         return self._tag_name
 
+    @property
+    def commit_sha(self):
+        return self.tag_name
```
lachmanfrantisek, line 703: I know that I am late to the party, but this should really be a `commit_sha`. (We have `identifier` for the event name.)

(It should work but can be misleading and we can have troubles later: e.g. tags needs to be pulled, unlike the commits.)

lachmanfrantisek:
```python
    @property
        if not self._tag_name:
            releases = self.project.get_releases()
            self._tag_name = releases[0].git_tag.commit_sha if releases else ""
        return self._tag_name
```

Or some optimisation to not need to search for releases for twice.

csomh: Yeah, I am aware that this was a gross workaround but the way inheritance is set up in these events is really inconsistent and scary and don't have the bandwidth to fix it now.

If you look around the inheritance tree you'll see commit_sha, tag_name, git_ref, base_branch and all kind of weird stuff which probably could be unified. Unfortunately it's not clear based on what these attributes were assigned to a given layer of the inheritance tree, and it's also nowhere documented what their intended usage is.

The ideal solution would be to have a `git_ref` at the base and have it set to sha or branch or tag.

But on the long run I would like to see the `Event` classes to be gone (the way we do celery tasks now allows this), so from my point of view, there is no big hurry to fix this.

lachmanfrantisek: I understand your confusion, but this makes it much worse since the name and value do not match. And the solution is really easy as you can see from my proposal.

There are three common attributes for all events, each used in a different way:
- `commit_sha` - commit hash
- `git_ref` - git checkout-able reference
- `identifier` - can be shown to the user

One of the reasons it's like that is because commit of the PR cannot be easily checkout-ed without additional work.

Your points make sense but are not relevant to this. My point is to match attribute naming, which is not so hard.

csomh: Filed #744 to fix this.

lachmanfrantisek: Thanks

PR #740, opened 2020-07-16
Reporting build status while no permission

shreyaspapi:
In Gitlab while working with upstream-project we don't have access to the merge request commit coming from fork repos. To set commit status we need access to the repository. 

So, if the reporting fails we can report the state using the MR comment and request access to the forked repository ([docs](https://docs.gitlab.com/ee/api/access_requests.html)). 

- [x] Adding the functionality to report build status using MR comment.
- [x] Fixed enqueue error #741 
- [x] Requesting for project access (with right permissions).

### DIFF
--- a/files/docker/Dockerfile.tests
+++ b/files/docker/Dockerfile.tests
@@ -1,6 +1,6 @@
 # For running tests locally, see check_in_container target in Makefile
 
-FROM docker.io/usercont/packit-service-worker:dev
+FROM docker.io/usercont/packit-service-worker:stg
 
 ARG SOURCE_BRANCH
 RUN  if [[ -z $SOURCE_BRANCH ]]; then \
--- a/packit_service/service/api/webhooks.py
+++ b/packit_service/service/api/webhooks.py
@@ -219,12 +219,12 @@ class GitlabWebhook(Resource):
             namespace=namespace, repo_name=repo_name, project_url=http_url
         )
 
-        if not project_authentication_issue.issue_created:
+        if not project_authentication_issue:
             token = jwt.encode(
                 {"namespace": namespace, "repo_name": repo_name},
                 config.gitlab_token_secret,
                 algorithm="HS256",
-            )
+            ).decode("utf-8")
 
             project = config.get_project(url=http_url)
             packit_user = project.service.user.get_username()
@@ -232,12 +232,12 @@ class GitlabWebhook(Resource):
             project.create_issue(
                 title="Packit-Service Authentication",
                 body=f"To configure Packit-Service with `{repo_name}` you need to\n"
-                f"configure the webhook settings. Head to {http_url}/hooks and add\n"
+                f"configure the webhook settings. Head to {project.get_web_url()}/hooks and add\n"
                 f"the Secret Token `{token}` to authenticate requests coming to Packit.\n\n"
                 f"Packit needs rights to set commit status to merge requests, Please grant\n"
                 f"[{packit_user}](https://gitlab.com/{packit_user}) `admin` permissions\n"
                 f"on the {namespace}/{repo_name} project. You can add the rights by clicking\n"
-                f"[here]({http_url}/-/project_members).",
+                f"[here]({project.get_web_url()}/-/project_members).",
                 private=True,
             )
 
@@ -313,6 +313,8 @@ class GitlabWebhook(Resource):
         interesting_events = {
             "Note Hook",
             "Merge Request Hook",
+            "Push Hook",
+            "Issue Hook",
         }
         event_type = request.headers.get("X-Gitlab-Event")
         _interested = event_type in interesting_events
--- a/packit_service/worker/build/build_helper.py
+++ b/packit_service/worker/build/build_helper.py
@@ -27,6 +27,8 @@ from typing import Union, List, Optional, Tuple, Set
 from kubernetes.client.rest import ApiException
 
 from ogr.abstract import GitProject, CommitStatus
+from ogr.services.gitlab import GitlabProject
+from ogr.exceptions import GitlabAPIException
 from packit.api import PackitAPI
 from packit.config import JobType, JobConfig
 from packit.config.package_config import PackageConfig
@@ -79,6 +81,10 @@ class BaseBuildJobHelper:
         self._srpm_path: Optional[Path] = None
         self._job_tests: Optional[JobConfig] = None
         self._job_build: Optional[JobConfig] = None
+        self._base_project: Optional[GitProject] = None
+        self._pr_id: Optional[int] = None
+        self._is_reporting_allowed: Optional[bool] = None
+        self._is_gitlab_instance: Optional[bool] = None
 
     @property
     def local_project(self) -> LocalProject:
@@ -90,6 +96,47 @@ class BaseBuildJobHelper:
                 pr_id=self.metadata.pr_id,
             )
         return self._local_project
+
+    @property
+    def is_gitlab_instance(self) -> bool:
+        if self._is_gitlab_instance is None:
+            self._is_gitlab_instance = isinstance(self.project, GitlabProject)
+
+        return self._is_gitlab_instance
+
+    @property
+    def pr_id(self) -> Optional[int]:
+        if self._pr_id is None:
+            self._pr_id = self.metadata.pr_id
+        return self._pr_id
+
+    @property
+    def is_reporting_allowed(self) -> bool:
+        username = self.project.service.user.get_username()
+        if self._is_reporting_allowed is None:
+            self._is_reporting_allowed = self.base_project.can_merge_pr(username)
+        return self._is_reporting_allowed
+
+    @property
+    def base_project(self) -> GitProject:
+        """
+        Getting the source project info from PR,
+        In case of build events we loose the source info.
+        """
+        if self._base_project is None:
+            if self.pr_id:
+                self._base_project = self.project.get_pr(
+                    pr_id=self.pr_id
+                ).source_project
+            else:
+                self._base_project = self.project
+        return self._base_project
+
+    def request_project_access(self) -> None:
+        try:
+            self.base_project.request_access()
+        except GitlabAPIException:
+            logger.info("Access already requested")
 
     @property
     def api(self) -> PackitAPI:
@@ -208,7 +255,11 @@ class BaseBuildJobHelper:
     def status_reporter(self) -> StatusReporter:
         if not self._status_reporter:
             self._status_reporter = StatusReporter(
-                self.project, self.metadata.commit_sha, self.metadata.pr_id
+                project=self.base_project
+                if isinstance(self.project, GitlabProject) and self.is_reporting_allowed
+                else self.project,
+                commit_sha=self.metadata.commit_sha,
+                pr_id=self.metadata.pr_id,
             )
         return self._status_reporter
 
@@ -377,12 +428,28 @@ class BaseBuildJobHelper:
         The status reporting should be done through this method
         so we can extend it in subclasses easily.
         """
-        self.status_reporter.report(
-            description=description,
-            state=state,
-            url=url,
-            check_names=check_names,
-        )
+        if not self.is_gitlab_instance or self.is_reporting_allowed:
+            self.status_reporter.report(
+                description=description,
+                state=state,
+                url=url,
+                check_names=check_names,
+            )
+            return
+
+        # "Packit-User" does not have access to project.
+
+        self.request_project_access()
+        final_commit_states = [
+            CommitStatus.error,
+            CommitStatus.success,
+            CommitStatus.failure,
+        ]
+
+        # We are only commenting final states to avoid multiple comments for a build
+        # Ignoring all other states eg. pending, running
+        if state in final_commit_states and check_names is not None:
+            self.status_reporter.report_status_by_comment(state, url, check_names)
 
     def report_status_to_all(
         self, description: str, state: CommitStatus, url: str = ""
--- a/packit_service/worker/reporting.py
+++ b/packit_service/worker/reporting.py
@@ -22,6 +22,7 @@
 import hashlib
 import logging
 from typing import Optional, Union
+import gitlab
 
 from ogr.abstract import GitProject, CommitStatus
 from ogr.services.pagure import PagureProject
@@ -87,6 +88,24 @@ class StatusReporter:
                 uid=hashlib.md5(check_name.encode()).hexdigest(),
             )
 
+    def report_status_by_comment(
+        self,
+        state: CommitStatus,
+        url: str,
+        check_names: Union[str, list, None],
+    ):
+        """
+        Reporting build status with MR comment if no permission to the fork project
+        """
+
+        if isinstance(check_names, str):
+            check_names = [check_names]
+        comment_msg = [
+            f"| [{check}]({url}) | {state.name.upper()} |" for check in check_names
+        ]
+
+        self.project.pr_comment(self.pr_id, "\n\n".join(comment_msg))
+
     def set_status(
         self,
         state: CommitStatus,
@@ -99,9 +118,16 @@ class StatusReporter:
             url = "https://wiki.centos.org/Manuals/ReleaseNotes/CentOSStream"
 
         logger.debug(f"Setting status for check '{check_name}': {description}")
-        self.project.set_commit_status(
-            self.commit_sha, state, url, description, check_name, trim=True
-        )
+
+        try:
+            self.project.set_commit_status(
+                self.commit_sha, state, url, description, check_name, trim=True
+            )
+        except gitlab.exceptions.GitlabCreateError:
+            # Ignoring Gitlab 'enqueue' error
+            # https://github.com/packit-service/packit-service/issues/741
+            pass
+
         # Also set the status of the pull-request for forges which don't do
         # this automatically based on the flags on the last commit in the PR.
         self.__set_pull_request_status(check_name, description, url, state)
--- a/packit_service/service/events.py
+++ b/packit_service/service/events.py
@@ -509,6 +509,13 @@ class MergeRequestGitlabEvent(AddPullRequestDbTrigger, AbstractGitlabEvent):
         result["action"] = result["action"].value
         return result
 
+    def get_base_project(self) -> GitProject:
+        fork = self.project.service.get_project(
+            namespace=self.source_repo_namespace,
+            repo=self.source_repo_name,
+        )
+        return fork
+
 
 class PullRequestGithubEvent(AddPullRequestDbTrigger, AbstractGithubEvent):
     def __init__(
@@ -584,6 +591,13 @@ class MergeRequestCommentGitlabEvent(AddPullRequestDbTrigger, AbstractGitlabEven
         result = super().get_dict()
         result["action"] = result["action"].value
         return result
+
+    def get_base_project(self) -> GitProject:
+        fork = self.project.service.get_project(
+            namespace=self.source_repo_namespace,
+            repo=self.source_repo_name,
+        )
+        return fork
 
 
 class PullRequestCommentGithubEvent(AddPullRequestDbTrigger, AbstractGithubEvent):
--- a/tests/unit/test_copr_build.py
+++ b/tests/unit/test_copr_build.py
@@ -48,7 +48,10 @@ from packit_service.service.events import (
     MergeRequestGitlabEvent,
 )
 from packit_service.worker.build import copr_build
-from packit_service.worker.build.copr_build import CoprBuildJobHelper
+from packit_service.worker.build.copr_build import (
+    CoprBuildJobHelper,
+    BaseBuildJobHelper,
+)
 from packit_service.worker.parser import Parser
 from packit_service.worker.reporting import StatusReporter
 from tests.spellbook import DATA_DIR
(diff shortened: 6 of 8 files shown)


packit_service/worker/build/build_helper.py
```diff
@@ -373,9 +389,28 @@ def _report(
...
         so we can extend it in subclasses easily.
         """
-        self.status_reporter.report(
-            description=description, state=state, url=url, check_names=check_names,
-        )
+        try:
+            self.status_reporter.report(
+                description=description, state=state, url=url, check_names=check_names,
+            )
+        except Exception as e:
```
lachmanfrantisek, line 396: Catching such general exception can be dangerous -- there are a lot of other problems we can get.

packit_service/worker/build/build_helper.py
```diff
@@ -373,9 +389,28 @@ def _report(
...
+            logger.info("Unable to report status ", e)
+            self.report_status_by_comment(
+                int(self.metadata.identifier), state, url, check_names
```
lachmanfrantisek, line 399: - Please, use key-word arguments for better readability.
- `identifier` can be non-numeric for non-merge-request events.

packit_service/worker/build/build_helper.py
```diff
@@ -373,9 +389,28 @@ def _report(
...
+            )
+
+    # Reporting build status if no permission to the fork project
```
lachmanfrantisek, line 402: Use docstring for that information. (The `"""something"""` string bellow the function/class header.)

packit_service/worker/build/build_helper.py
shreyaspapi: I could only find identifier as common between all events.

lachmanfrantisek: My point was about `int(self.metadata.identifier)` that we can get non-numeric value from `self.metadata.identifier`. What will cause an exception...

packit_service/worker/build/build_helper.py
```diff
@@ -91,6 +95,35 @@ def local_project(self) -> LocalProject:
...
+        In case of build events we loose the source info.
+        """
+        if self._base_project is None:
+            self._base_project = self.project.get_pr(pr_id=self.pr_id).source_project
+        return self._base_project
+
+    def request_project_access(self) -> None:
+        try:
+            self.base_project.request_access()
+        except Exception:
```
lachmanfrantisek, line 124: I would use the most specific exception possible and ideally check also content. We can get more exception here. (E.g. network/authentication problems...).

packit_service/worker/build/build_helper.py
```diff
@@ -207,9 +240,14 @@ def job_tests(self) -> Optional[JobConfig]:
...
-                self.project, self.metadata.commit_sha, self.metadata.pr_id
-            )
+            if isinstance(self.project, GitlabProject):
+                self._status_reporter = StatusReporter(
+                    self.base_project, self.metadata.commit_sha, self.metadata.pr_id
+                )
+            else:
+                self._status_reporter = StatusReporter(
+                    self.project, self.metadata.commit_sha, self.metadata.pr_id
+                )
```
lachmanfrantisek, line 250: Maybe this to show the only difference more clearly?

```suggestion
            self._status_reporter = StatusReporter(
                project=self.base_project if isinstance(self.project, GitlabProject) else self.project,
                commit_sha=self.metadata.commit_sha,
                pr_id=self.metadata.pr_id
            )
```

packit_service/worker/build/build_helper.py
```diff
@@ -373,9 +411,59 @@ def _report(
         The status reporting should be done through this method
         so we can extend it in subclasses easily.
         """
-        self.status_reporter.report(
-            description=description, state=state, url=url, check_names=check_names,
-        )
+        if (
+            isinstance(self.project, GitlabProject)
+            and not self.check_project_permission
```
lachmanfrantisek, line 416: The name `check_project_permission` looks more like a method. With properties, I would use some `is_*` form of the name. E.g. `is_reporting_allowed`, or `is_fork_status_allowed`,...

packit_service/worker/build/build_helper.py
```diff
@@ -373,9 +411,59 @@ def _report(
...
+    ) -> None:
+        """
+        Reporting build status with MR comment if no permission to the fork project
+        """
+        comment_msg = f"[{check_names}]({url}): {state.name.upper()}"
+
+        all_pr_comments = self.project.get_pr(pr_id=self.pr_id)._get_all_comments()[
+            ::-1
+        ]
+
```
lachmanfrantisek, line 437: Don't we have some argument to do this with the public method?

packit_service/worker/build/build_helper.py
```diff
@@ -373,9 +411,59 @@ def _report(
...
+        if isinstance(check_names, list):
+            return
```
lachmanfrantisek, line 439: Can you please document this a bit?

packit_service/worker/build/build_helper.py
shreyaspapi: Yes added more specific exceptions.

packit_service/worker/build/build_helper.py
shreyaspapi: Done!

packit_service/worker/build/build_helper.py
shreyaspapi: Yup, is_reporting_allowed is better name here. Changed it!

packit_service/worker/build/build_helper.py
shreyaspapi: Thanks! Found it.

packit_service/worker/build/build_helper.py
shreyaspapi: Found the method get_comments(), changed the comment reporting a bit. Also added few comments.

packit_service/worker/reporting.py
```diff
@@ -95,9 +96,16 @@ def set_status(
...
-        )
+
+        try:
+            self.project.set_commit_status(
+                self.commit_sha, state, url, description, check_name, trim=True
+            )
+        except gitlab.exceptions.GitlabHttpError:
+            # Ignoring Gitlab 'enqueue' error
+            # https://github.com/packit-service/packit-service/issues/741
+            pass
```
lachmanfrantisek, line 107: What is done when we get that exception? Nothing is changed on GitLab side?

What can we do instead to show that info to the user? What about having a unique status for each step of our workflow? e.g. `packit-service/copr-build-srpm`, `packit-service/copr-build-fedora-31-rpm` (We do that in the past..;)

packit_service/worker/build/build_helper.py
```diff
@@ -91,6 +96,39 @@ def local_project(self) -> LocalProject:
...
+        username = self.project.service.user.get_username()
+        if self._is_reporting_allowed is None:
+            self._is_reporting_allowed = self.base_project.can_merge_pr(username)
+        return self._is_reporting_allowed
+
+    @property
+    def base_project(self) -> GitProject:
+        """
+        Getting the source project info from PR,
+        In case of build events we loose the source info.
```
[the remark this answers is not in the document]
shreyaspapi: > * If you mean results events, we can get that info from the database.

Yeah, I meant the result event. Okay, I will try to use the info from the database.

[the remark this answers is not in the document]
shreyaspapi: We don't store base_project info in database.

packit_service/worker/build/build_helper.py
```diff
@@ -91,6 +96,35 @@ def local_project(self) -> LocalProject:
...
+        """
+        if self._base_project is None:
+            self._base_project = self.project.get_pr(pr_id=self.pr_id).source_project
```
lachmanfrantisek, line 119: This will not work for non-pr cases. We should return `base_project` in that case.

shreyaspapi: The best I could think of was by cathing the value error here which will come from pr_id, in case of non-pr events.

packit_service/worker/build/build_helper.py
```diff
@@ -91,6 +96,41 @@ def local_project(self) -> LocalProject:
...
+        In case of build events we loose the source info.
+        """
+        if self._base_project is None:
+            try:
+                self._base_project = self.project.get_pr(
+                    pr_id=self.pr_id
+                ).source_project
+            # pr_id will result in value error for non-pr events
+            except (ValueError, gitlab.exceptions.GitlabGetError):
+                return self.project
```
lachmanfrantisek, line 125: What about this?

```suggestion
            if self.pr_id:
                self._base_project = self.project.get_pr(
                    pr_id=self.pr_id
                ).source_project
            else:
                self._base_project = self.project
```

packit_service/worker/build/build_helper.py
```diff
@@ -91,6 +96,43 @@ def local_project(self) -> LocalProject:
             )
         return self._local_project
 
+    @property
+    def pr_id(self) -> int:
+        if self._pr_id is None:
+            try:
+                self._pr_id = int(self.metadata.event_dict["identifier"])
```
lachmanfrantisek, line 103: @lbarcziova do we have `pr_id` in the `event_dict`? (So we don't need to do this `try-except`.) Thanks

lbarcziova: it depends on the content of the original Event, but in EventData (self.metadata) we are trying to get both [pr_id](https://github.com/packit-service/packit-service/blob/398d0d20060ea3af92d21d9dab9b3adef599f365/packit_service/service/events.py#L186)  and [identifier](https://github.com/packit-service/packit-service/blob/398d0d20060ea3af92d21d9dab9b3adef599f365/packit_service/service/events.py#L188), so this could be definitely simplified

lachmanfrantisek: @lbarcziova Thanks.

packit_service/worker/build/build_helper.py
```diff
@@ -373,9 +417,30 @@ def _report(
...
+
+    def report_status_by_comment(
+        self, state: CommitStatus, url: str, check_names: Union[str, list, None],
+    ) -> None:
+        """
+        Reporting build status with MR comment if no permission to the fork project
+        """
+        build_states = [CommitStatus.error, CommitStatus.success, CommitStatus.failure]
+
+        if isinstance(check_names, str):
```
lachmanfrantisek, line 437: What about the `check_names is None` scenario. (At least from the type annotation, it looks like we can get that.)

packit_service/worker/build/build_helper.py
```diff
@@ -373,9 +417,30 @@ def _report(
...
+            check_names = [check_names]
+
+        for check in check_names:
+            comment_msg = f"[{check}]({url}): {state.name.upper()}"
+            if state in build_states:
+                self.project.pr_comment(self.pr_id, comment_msg)
```
lachmanfrantisek, line 443: Can we merge those to one comment to not be so spammy?

packit_service/worker/build/build_helper.py
shreyaspapi: Ah, I have changed the type annotation.

shreyaspapi: Solved it here - [R438-R439](https://github.com/packit/packit-service/pull/740/files#diff-599fa650469aeda8bea187fa74ed9502R438-R439)

packit_service/worker/build/build_helper.py
shreyaspapi: Done!

shreyaspapi:
Thanks for the review! @lachmanfrantisek 
I have addressed all the comments here.

>What can we do instead to show that info to the user? What about having a unique status for each step of our workflow? e.g. >packit-service/copr-build-srpm, packit-service/copr-build-fedora-31-rpm (We do that in the past..;)

The only thing which is left here is to send unique build status for each step of the workflow, I will have different for it!

packit_service/worker/build/build_helper.py
lachmanfrantisek: If I read that correctly, we don't want to process status that is final (`error`/`success`/`failure`).

- What will be done for those "final" states?
- Is it possible to move this logic (= `if state not in ...` condition) out of this method? Is it too complicated?
- Can we add a comment and/or update docstring to say why we are doing this?

packit_service/worker/build/build_helper.py
```diff
@@ -375,9 +418,34 @@ def _report(
...
-            description=description, state=state, url=url, check_names=check_names,
-        )
+        if isinstance(self.project, GitlabProject) and not self.is_reporting_allowed:
+            # "Packit-User" does not have access to project.
+            self.request_project_access()
+            self.report_status_by_comment(state, url, check_names)
+        else:
+            self.status_reporter.report(
+                description=description, state=state, url=url, check_names=check_names,
+            )
```
lachmanfrantisek, line 428: What about moving the `report_status_by_comment` to `status_reporter` class?

lachmanfrantisek (commented):
Looks good!

I have only a few code-style suggestions.

packit_service/worker/build/build_helper.py
shreyaspapi: Yes, I have moved `report_status_by_comment ` to `StatusReporter` class.

packit_service/worker/build/build_helper.py
shreyaspapi: @lachmanfrantisek 
We only want to report status by comment if it's a final state (error / success / failure).

```
if state not in build_states or check_names is None:
    return
```
The above condition was still valid, as we are not doing anything if `state` has any value other then (error/success/failure).

Anyways I have updated and moved that part as suggested. 

I think this might be easier to read -
```
if state in build_states and check_names is not None:
    self.status_reporter.report_status_by_comment(state, url, check_names)
```

lachmanfrantisek: Ah, yes. My bad...;)

Thanks!

packit_service/service/api/webhooks.py
```diff
@@ -219,25 +219,25 @@ def create_confidential_issue_with_token(self):
...
+            ).decode("utf-8")
 
             project = config.get_project(url=http_url)
             packit_user = project.service.user.get_username()
 
             project.create_issue(
                 title="Packit-Service Authentication",
                 body=f"To configure Packit-Service with `{repo_name}` you need to\n"
-                f"configure the webhook settings. Head to {http_url}/hooks and add\n"
+                f"configure the webhook settings. Head to {http_url[:-4]}/hooks and add\n"
```
lachmanfrantisek, line 235: You can use [project.get_web_url method](https://github.com/packit/ogr/blob/176add79eeb7d71e765550da76c9cdc8aced9e92/ogr/services/gitlab/project.py#L554) instead for better readability:

```suggestion
                f"configure the webhook settings. Head to {project.get_web_url()}/hooks and add\n"
```

packit_service/service/api/webhooks.py
```diff
@@ -219,25 +219,25 @@ def create_confidential_issue_with_token(self):
...
                 f"the Secret Token `{token}` to authenticate requests coming to Packit.\n\n"
                 f"Packit needs rights to set commit status to merge requests, Please grant\n"
                 f"[{packit_user}](https://gitlab.com/{packit_user}) `admin` permissions\n"
                 f"on the {namespace}/{repo_name} project. You can add the rights by clicking\n"
-                f"[here]({http_url}/-/project_members).",
+                f"[here]({http_url[:-4]}/-/project_members).",
```
lachmanfrantisek, line 240:
```suggestion
                f"[here]({project.get_web_url()}/-/project_members).",
```

packit_service/worker/build/build_helper.py
```diff
@@ -91,6 +97,47 @@ def local_project(self) -> LocalProject:
             )
         return self._local_project
 
+    @property
+    def is_gitlab_instance(self) -> bool:
+        if self._is_gitlab_instance is None:
+            self._is_gitlab_instance = isinstance(self.project, GitlabProject)
+
+        return self._is_gitlab_instance
```
lachmanfrantisek, line 105: I am not sure if this helps too much. I am leaving that to your judgement if it is clearer now. (=You don't need to change that if you want..;)

packit_service/worker/build/build_helper.py
```diff
@@ -91,6 +97,47 @@ def local_project(self) -> LocalProject:
...
+                ).source_project
+            else:
+                self._base_project = self.project
+        return self._base_project
+
+    def request_project_access(self) -> None:
+        try:
+            self.base_project.request_access()
+        except GitlabAPIException:
+            logger.info("Access already requested")
```
lachmanfrantisek, line 139: Is there something in the exception instance or exception string representation what can we use to filter only those that mean `Access already requested`?

packit_service/worker/build/build_helper.py
```diff
@@ -377,12 +428,27 @@ def _report(
...
+            build_states = [
+                CommitStatus.error,
+                CommitStatus.success,
+                CommitStatus.failure,
+            ]
+
+            # We are only commenting final states to avoid multiple comments for a build
+            # Ignoring all other states eg. pending, running
+            if state in build_states and check_names is not None:
+                self.status_reporter.report_status_by_comment(state, url, check_names)
```
lachmanfrantisek, line 444:
```suggestion
            final_commit_states = [
                CommitStatus.error,
                CommitStatus.success,
                CommitStatus.failure,
            ]

            # We are only commenting final states to avoid multiple comments for a build
            # Ignoring all other states eg. pending, running
            if state in final_commit_states and check_names is not None:
                self.status_reporter.report_status_by_comment(state, url, check_names)
```

packit_service/worker/reporting.py
```diff
@@ -87,6 +88,25 @@ def __set_pull_request_status(
...
+        """
+        Reporting build status with MR comment if no permission to the fork project
+        """
+
+        if isinstance(check_names, str):
+            check_names = [check_names]
+
+        comment_msg = [
+            f"[{check}]({url}): {state.name.upper()}" for check in check_names
+        ]
```
lachmanfrantisek, line 106: We can use a table here (not need to be done here). The table looks better but only in case we don't need to provide any logs inside.

packit_service/worker/build/build_helper.py
```diff
@@ -377,12 +428,27 @@ def _report(
         The status reporting should be done through this method
         so we can extend it in subclasses easily.
         """
-        self.status_reporter.report(
-            description=description,
-            state=state,
-            url=url,
-            check_names=check_names,
-        )
+        if self.is_gitlab_instance and not self.is_reporting_allowed:
```
lachmanfrantisek, line 431: What about switching the order of those two `if` branches to have the shorter one in the beginning?

So, instead of this:
```python
if something:
    # long
else:
    # short
```

use this:

```python
if not something:
    # short
    return

# long
```

packit_service/worker/reporting.py
lachmanfrantisek: Do we want to solve this here or make a separate PR for this. We have multiple possibilities for that so it's worth a separate discussion.

lachmanfrantisek (commented):
Nice job!

I have a few notes but mostly styling nits or suggestions for the future. No worry about the number of comments..;)

packit_service/worker/build/build_helper.py
shreyaspapi: 👍

packit_service/worker/reporting.py
shreyaspapi: Yup we don't, I have changed it!

files/docker/Dockerfile.tests
```diff
@@ -1,6 +1,6 @@
 # For running tests locally, see check_in_container target in Makefile
 
-FROM docker.io/usercont/packit-service-worker:dev
+FROM docker.io/usercont/packit-service-worker:stg
```
lachmanfrantisek, line 3: Can you revert that, please?

packit_service/worker/reporting.py
```diff
@@ -87,6 +88,24 @@ def __set_pull_request_status(
...
+    ):
+        """
+        Reporting build status with MR comment if no permission to the fork project
+        """
+
+        if isinstance(check_names, str):
+            check_names = [check_names]
+        comment_msg = [
+            f"| [{check}]({url}) | {state.name.upper()} |" for check in check_names
+        ]
```
lachmanfrantisek, line 105: We are probably missing the header of the table. It should look like this:

```
| some text | in header |
| ------------- | ------------ |
| one | line |
| another | line |
```
| some text | in header |
| ------------- | ------------ |
| one | line |
| another | line |

or empty header:

```
| | |
| ------------- | ------------ |
| one | line |
| another | line |
```
| | |
| ------------- | ------------ |
| one | line |
| another | line |

Without that it looks like this:
```
| one | line |
| another | line |
```
| one | line |
| another | line |

packit_service/worker/reporting.py
```diff
@@ -87,6 +88,26 @@ def __set_pull_request_status(
...
+
+        header = ["| Type | Status |", "| ---- | ----- |"]
+
+        if isinstance(check_names, str):
+            check_names = [check_names]
+        comment_msg = [
+            f"| [{check}]({url}) | {state.name.upper()} |" for check in check_names
+        ]
+
+        self.project.pr_comment(self.pr_id, "\n\n".join(header + comment_msg))
```
lachmanfrantisek, line 109: I think we need only one new line here:
```suggestion
        self.project.pr_comment(self.pr_id, "\n".join(header + comment_msg))
```

packit_service/worker/reporting.py
shreyaspapi: I will make a different pr for this.

lachmanfrantisek: Thanks!

lachmanfrantisek (approved):
Good job! Thank you!

shreyaspapi:
@lachmanfrantisek problems with tests! We need to wait.

TomasTomecek:
Wow, merged \o/ good job Shreyas!

\o/

PR #742, opened 2020-07-16
Enable backend_cleanup Celery task to delete old task results

lbarcziova:
Needs packit-service/deployment#107

### DIFF
--- a/packit_service/celerizer.py
+++ b/packit_service/celerizer.py
@@ -19,7 +19,7 @@
 # LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
-
+from datetime import timedelta
 from os import getenv
 
 from celery import Celery
@@ -47,6 +47,11 @@ class Celerizer:
 
             # http://docs.celeryproject.org/en/latest/reference/celery.html#celery.Celery
             self._celery_app = Celery(backend=postgres_url, broker=redis_url)
+
+            days = int(getenv("CELERY_RESULT_EXPIRES", "30"))
+            # https://docs.celeryproject.org/en/latest/userguide/configuration.html#result-expires
+            self._celery_app.conf.result_expires = timedelta(days=days)
+
         return self._celery_app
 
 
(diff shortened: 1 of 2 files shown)


lbarcziova:
> so I'd suggest calling the variable $CELERY_COMMAND

Good suggestion, I will change it here and in second PR as well.

> Just to be sure: We don't need to configure the beat by setting the app.conf.beat_schedule? We just need to set the app.conf.result_expires and that will take care of the scheduling?

It should be run automatically (https://docs.celeryproject.org/en/latest/userguide/configuration.html#result-expires), when I was trying it locally with this changes, it showed:


```
beat                | [2020-07-16 10:23:56,857: DEBUG/MainProcess] Current schedule:
beat                | <ScheduleEntry: celery.backend_cleanup celery.backend_cleanup() <crontab: 0 4 * * * (m/h/d/dM/MY)>
```

packit_service/celerizer.py
```diff
@@ -47,6 +47,9 @@ def celery_app(self):
 
             # http://docs.celeryproject.org/en/latest/reference/celery.html#celery.Celery
             self._celery_app = Celery(backend=postgres_url, broker=redis_url)
+            # https://docs.celeryproject.org/en/latest/userguide/configuration.html#result-expires
+            self._celery_app.conf.result_expires = timedelta(days=30)
```
jpopelka, line 51: Can we also have it configurable via an environment variable?

lbarcziova:
I think the change in deployment repo has to be deployed at first since now there is missing CELERY_COMMAND env var in worker

jpopelka:
Aha, OK, then I'd say let's default the `CELERY_COMMAND` to "worker".
I don't remember why we were always explicitly setting it in the code where you took it from and I don't think we have to do it that way.

jpopelka:
To do that in bash, you can use [this syntax](https://github.com/packit-service/deployment/blob/master/cron-jobs/import-images/import-images.sh#L8).

lbarcziova:
@jpopelka thanks for helping, changed

packit_service/celerizer.py
```diff
@@ -47,6 +47,11 @@ def celery_app(self):
 
             # http://docs.celeryproject.org/en/latest/reference/celery.html#celery.Celery
             self._celery_app = Celery(backend=postgres_url, broker=redis_url)
+
+            days = int(getenv("CELERY_RESULTS_EXPIRATION", "30"))
```
jpopelka, line 51: I'm wondering whether we should rather call it CELERY_RESULT_EXPIRES, which from docs is also the old name for result_expires.
I know I'm going into details, so I'll leave this on you.

PR #746, opened 2020-07-19
Add /api/project/.../branches endpoint

IceWreck:


packit_service/models.py
```diff
@@ -366,6 +384,16 @@ def get_by_id(cls, id_: int) -> Optional["GitBranchModel"]:
         with get_sa_session() as session:
             return session.query(GitBranchModel).filter_by(id=id_).first()
 
+    def get_copr_builds(self):
+        return JobTriggerModel.get_or_create(
```
csomh, line 388: These will create branch build and branch test run entries in the DB, even if the branch builds are not configured for a given repo. Is this okay? Shouldn't we use a plain `get` here instead?

packit_service/service/api/projects.py
```diff
@@ -152,3 +152,46 @@ def get(self, forge, namespace, repo_name):
...
+
+@ns.route("/<forge>/<namespace>/<repo_name>/branches")
+@ns.param("forge", "Git Forge")
+@ns.param("namespace", "Namespace")
+@ns.param("repo_name", "Repo Name")
+class ProjectBranches(Resource):
+    @ns.response(HTTPStatus.OK, "OK, project branches handled by Packit Service follow")
+    def get(self, forge, namespace, repo_name):
+        """Project branches"""
+        branches_list = GitProjectModel.get_project_branches(
```
csomh, line 165: You could call this only `branches`, to make things simpler.

tests_requre/service/test_api.py
```diff
@@ -240,3 +240,20 @@ def test_get_projects_releases(client, clean_before_and_after, release_model):
     assert len(response_dict) == 1
     assert response_dict[0]["tag_name"] == SampleValues.tag_name
     assert response_dict[0]["commit_hash"] == SampleValues.commit_sha
+
+
+def test_get_projects_branches(
+    client, clean_before_and_after, a_koji_build_for_branch_push
+):
+    """Test Get Project's Releases"""
+    response = client.get(
```
csomh, line 249: Is a recording needed for this?

IceWreck: I want to blame autocomplete.
This should be _a_copr_build_for_branch_push_ but I guess tests still pass because SampleValues.branch is same for both  _a_copr_build_for_branch_push_ and  _a_koji_build_for_branch_push_

packit_service/models.py
IceWreck: Okay.

IceWreck:
@csomh I've used/changed to get instead of get_or_create whereever I made the methods but left it as it is in case of PullRequestModel.get_copr_builds() which is also used by jobs.py

PR #747, opened 2020-07-23
Fix build hook

sakalosj:
Starts build in project root.

### DIFF
--- a/files/docker/hooks/build
+++ b/files/docker/hooks/build
@@ -1,3 +1,8 @@
 #!/bin/bash
 
+# setting cwd to build context location defined in Dcker Hub build configuration -> build rules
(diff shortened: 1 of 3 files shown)


files/docker/hooks/build
```diff
@@ -1,3 +1,7 @@
 #!/bin/bash
 
+# required because this script is executed in Dockerfile location instead of provided
+# build context location, which is root for $DOCKERFILE_PATH
```
jpopelka, line 4: I've read it a couple of times and I'm still struggling with understanding the context.
Wouldn't it help if we moved this script in the "Dockerfile location", i.e. to `files/docker/`?

sakalosj:
recheck

PR #748, opened 2020-07-23
Revert JobTriggerModel.get(...)

IceWreck:
Introduced in #746 based on my incorrect understanding of JobTriggerModel.

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -225,7 +225,7 @@ class GitProjectModel(Base):
             branches = (
                 session.query(GitBranchModel)
                 .filter_by(project_id=project.id)
-                .order_by(desc(GitBranchModel.id))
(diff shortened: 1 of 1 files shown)


packit_service/models.py
```diff
@@ -225,7 +225,7 @@ def get_project_branches(
             branches = (
                 session.query(GitBranchModel)
                 .filter_by(project_id=project.id)
-                .order_by(desc(GitBranchModel.id))
+                .order_by(GitBranchModel.id)
```
csomh, line 228: Was this intentional?

IceWreck: Yes because master branch is more likely to have builds (ex. packit-master) than other branches. And it almost always has a lower id than the other branches. So I thought lets just show them in ascending order so that master is at the top.

PR #749, opened 2020-07-23
Try using a newer Fedora image for Zuul

csomh:


csomh:
Oh, yeah, this will not work, as most probably the issue is that >f30 images have cgroupsv2 configured which is not supported by docker.

jpopelka:
#381

PR #751, opened 2020-07-24
Use Fedora 31 as a base for the workers

csomh:
This is to make the tests work again. On the long run we should figure
out why the worker pods failed to start when the image was based on
Fedora 32 and why they had this issue only in the test environment but
not on the stage deployment.


sakalosj:
:+1:

PR #754, opened 2020-07-26
Misc API Changes

IceWreck:


packit_service/service/api/copr_builds.py
```diff
@@ -127,10 +123,7 @@ def get(self, id):
...
 
-            build = make_response(dumps(build_dict))
-            build.headers["Content-Type"] = "application/json"
-            build.headers["Access-Control-Allow-Origin"] = "*"
-            return build if build else ("", HTTPStatus.NO_CONTENT)
+            return response_maker(build_dict)
 
         else:
-            return "", HTTPStatus.NO_CONTENT
+            return response_maker([])
```
csomh, line 126: Hm.... wouldn't "not found" be a better response here? Why was "no content" used in the previous implementation?

packit_service/service/api/utils.py
```diff
@@ -0,0 +1,13 @@
+from flask import make_response
+from http import HTTPStatus
+from json import dumps
+
+
+def response_maker(result, content_range=None, status=HTTPStatus.OK):
+    """response_maker is a wrapper around flask's make_response"""
+    resp = make_response(dumps(result), status)
+    if content_range:
```
csomh, line 9: Doing like this it might make future updates cumbersome, as it would force us to update this method whenever we want to add a new optional header.

How about `response_maker` adding only the common headers and let the individual API endpoints to complete that with anything other they need and explicitly set the status code?

packit_service/service/api/projects.py
```diff
@@ -49,6 +49,29 @@ def get(self):
...
+@ns.param("forge", "Git Forge")
+@ns.param("namespace", "Namespace")
+@ns.param("repo_name", "Repo Name")
+class ProjectInfo(Resource):
+    @ns.response(HTTPStatus.OK, "Project details follow")
+    def get(self, forge, namespace, repo_name):
+        """Project Details"""
+        project = GitProjectModel.get_project(forge, namespace, repo_name)
+        if not project:
+            return response_maker([])
```
csomh, line 62: Shouldn't this be "not found" instead?

packit_service/service/api/utils.py
IceWreck: This returns a response so instead of doing
`return response_maker(params.....)` 
we can add additional headers by doing

```
response = response_maker(params.....)
response.headers["something"] = "blah blah"
return response
```

csomh: yes, exactly, and this is why I'm saying to remove the `content_range` argument and rather add it when it's needed :)

packit_service/service/api/copr_builds.py
IceWreck: > Why was "no content" used in the previous implementation?
Yes, previous implementation was wrong, no content should not be used here.

Also, from what Ive read its a good practice to have the same json structure at the same endpoint.
So an empty list here makes more sense compared to an error message.  

Similar case here https://softwareengineering.stackexchange.com/a/358245

packit_service/service/api/utils.py
IceWreck: Oh yeah 🤦

packit_service/models.py
```diff
@@ -176,6 +176,33 @@ def get_projects(cls, first: int, last: int) -> Iterable["GitProjectModel"]:
...
+        """Return projects of given forge and namespace"""
+        with get_sa_session() as session:
+            projects = (
+                session.query(GitProjectModel).filter_by(namespace=namespace).all()
+            )
+            matched_projects = []
+            for project in projects:
+                forge_domain = urlparse(project.project_url).netloc
+                if forge == forge_domain:
+                    matched_projects.append(project)
```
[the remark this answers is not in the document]
csomh: Are you sure this would be the right usage for that function? What would be the additional benefit? It's only the domain name that's needed here, `parse_git_repo` returns many other things which would not be used here.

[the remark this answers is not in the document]
lachmanfrantisek: `urlparse` needs to parse more things as well.

Benefits?

- Reusing our own code. Having the same way to get this info across the code-base. (We already use it in `packit`/`packit-service`.)
- I was confused about the `netloc` name.
- Dogfooding.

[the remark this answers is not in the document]
csomh: > * Reusing our own code. Having the same way to get this info across the code-base. (We already use it in `packit`/`packit-service`.)

That code would do the same thing as this code. Again: what's the benefit?

> * I've been confused about the `netloc` name. =>  `hostname` looks more clear to me

And you were rightly so: we should probably use `hostname` here, instead of `netloc`. Nice catch! :rocket: 

> * Dogfooding.

I would say that calling a function to call a library function is and extra step, which should be removed :slightly_smiling_face:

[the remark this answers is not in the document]
lachmanfrantisek: > That code would do the same thing as this code. Again: what's the benefit?

It's nearly the same, so we can choose both. I already wrote three benefits. We have multiple occurrences of the URL parsing -- some use more and some less of that function/class. My suggestion was to use the same function for all to be consistent.

That was meant as a suggestion of alternative approach. I don't care what we choose. My main problem was with the property name.

packit_service/service/api/copr_builds.py
```diff
@@ -113,24 +109,17 @@ def get(self, id):
...
```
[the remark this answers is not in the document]
IceWreck: Sure 👍

packit_service/models.py
```diff
@@ -176,6 +176,33 @@ def get_projects(cls, first: int, last: int) -> Iterable["GitProjectModel"]:
...
+    def get_project(
+        cls, forge: str, namespace: str, repo_name: str
+    ) -> Optional["GitProjectModel"]:
+        """Return one project which matches said criteria"""
+        with get_sa_session() as session:
+            project = cls.__choose_project(
+                session=session, forge=forge, namespace=namespace, repo_name=repo_name
+            )
+            if not project:
+                return None
```
csomh, line 203: `__choose_project` already returns `None`. Could this just `return project`—without the conditional?

packit_service/service/api/copr_builds.py
```diff
@@ -89,48 +85,40 @@ class InstallationItem(Resource):
...
-
-            build = make_response(dumps(build_dict))
-            build.headers["Content-Type"] = "application/json"
-            build.headers["Access-Control-Allow-Origin"] = "*"
-            return build if build else ("", HTTPStatus.NO_CONTENT)
-
-        else:
-            return "", HTTPStatus.NO_CONTENT
+        if not bool(builds_list.first()):
+            return response_maker({"error": "No info about build stored in DB"})
```
csomh, line 89: This should set "NOT_FOUND", right?

packit_service/service/api/projects.py
```diff
@@ -44,13 +43,59 @@ def get(self):
...
+@ns.param("forge", "Git Forge")
+@ns.param("namespace", "Namespace")
+@ns.param("repo_name", "Repo Name")
+class ProjectInfo(Resource):
+    @ns.response(HTTPStatus.OK, "Project details follow")
+    def get(self, forge, namespace, repo_name):
+        """Project Details"""
+        project = GitProjectModel.get_project(forge, namespace, repo_name)
+        if not project:
+            return response_maker({"error": "No info about project stored in DB"})
```
csomh, line 61: Again: should 404 be set?

csomh:
> Build failed (gate pipeline).
> 
> * [pre-commit ](https://softwarefactory-project.io/zuul/t/packit-service/build/e6ec7ce0252b4581be8f29433805d7a6) : FAILURE in 3m 25s

Please rebase :slightly_smiling_face:

PR #756, opened 2020-07-28
Fix Docker Hub CI test builds and update Makefile

sakalosj:
Because of Docker Hub CI test builds are triggered for all branches we need
set SOURCE_BRANCH to master in case a custom branch is used.

Makefile was updated because image builds require SOURCE_BRANCH 
build variable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,17 +9,18 @@ AP ?= ansible-playbook -vv -c local -i localhost, -e ansible_python_interpreter=
 PATH_TO_SECRETS ?= $(CURDIR)/secrets/
 COV_REPORT ?= term-missing
 COLOR ?= yes
+SOURCE_BRANCH ?= $(git branch --show-current)
 
 service: files/install-deps.yaml files/recipe.yaml
-	$(CONTAINER_ENGINE) build --rm -t $(SERVICE_IMAGE) -f files/docker/Dockerfile .
+	$(CONTAINER_ENGINE) build --rm -t $(SERVICE_IMAGE) -f files/docker/Dockerfile --build-arg SOURCE_BRANCH=$SOURCE_BRANCH .
 
 worker: files/install-deps-worker.yaml files/recipe-worker.yaml
-	$(CONTAINER_ENGINE) build --rm -t $(WORKER_IMAGE) -f files/docker/Dockerfile.worker .
+	$(CONTAINER_ENGINE) build --rm -t $(WORKER_IMAGE) -f files/docker/Dockerfile.worker --build-arg SOURCE_BRANCH=$SOURCE_BRANCH .
 
 # This is for cases when you want to deploy into production and don't want to wait for dockerhub
 # Make sure you have latest docker.io/usercont/packit:prod prior to running this
 worker-prod: files/install-deps-worker.yaml files/recipe-worker.yaml
-	$(CONTAINER_ENGINE) build --rm -t $(WORKER_IMAGE_PROD) -f files/docker/Dockerfile.worker .
+	$(CONTAINER_ENGINE) build --rm -t $(WORKER_IMAGE_PROD) -f files/docker/Dockerfile.worker --build-arg SOURCE_BRANCH=$SOURCE_BRANCH .
 
 worker-prod-push: worker-prod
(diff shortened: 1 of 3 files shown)


files/install-deps-worker.yaml
```diff
@@ -2,11 +2,23 @@
...
+    source_branch: "{{ lookup('env', 'SOURCE_BRANCH') }}"
   tasks:
+    - name: Check source_branch value
+      debug:
+        msg: source_branch is set to {{ source_branch }}
+    - name: Fail on empty source_branch variable
+      fail:
+        msg: Variable source_branch, which is set from env variable SOURCE_BRANCH is empty.
+      when: source_branch == ''
+    # Docker Hub CI test is build from non master/stable branch (eg. packit:fix_hook)
```
TomasTomecek, line 14:
```suggestion
    # Docker Hub CI image builds are performed from non contributor's branches (eg. packit:fix_hook), but we need it to be master or stable
```

csomh:
Sorry, but I don't remember, how is SOURCE_BRANCH used? Why is it needed and how does it's value influence the build?

sakalosj:
> Sorry, but I don't remember, how is SOURCE_BRANCH used? Why is it needed and how does it's value influence the build?

SOURCE_BRANCH contains a current branch. Based on it, a master/stable branch from other projects will be chosen. introduced in #717

Makefile
```diff
@@ -9,17 +9,18 @@ AP ?= ansible-playbook -vv -c local -i localhost, -e ansible_python_interpreter=
...
-	$(CONTAINER_ENGINE) build --rm -t $(SERVICE_IMAGE) -f files/docker/Dockerfile .
+	$(CONTAINER_ENGINE) build --rm -t $(SERVICE_IMAGE) -f files/docker/Dockerfile --build-arg SOURCE_BRANCH=$SOURCE_BRANCH .
 
 worker: files/install-deps-worker.yaml files/recipe-worker.yaml
-	$(CONTAINER_ENGINE) build --rm -t $(WORKER_IMAGE) -f files/docker/Dockerfile.worker .
+	$(CONTAINER_ENGINE) build --rm -t $(WORKER_IMAGE) -f files/docker/Dockerfile.worker --build-arg SOURCE_BRANCH=$SOURCE_BRANCH .
 
 # This is for cases when you want to deploy into production and don't want to wait for dockerhub
 # Make sure you have latest docker.io/usercont/packit:prod prior to running this
 worker-prod: files/install-deps-worker.yaml files/recipe-worker.yaml
```
csomh, line 22: Can this target be removed?

csomh:
> > Sorry, but I don't remember, how is SOURCE_BRANCH used? Why is it needed and how does it's value influence the build?
> 
> SOURCE_BRANCH contains a current branch. Based on it, a master/stable branch from other projects will be chosen. introduced in #717

A case insensitive grep would have helped me with finding this answer, but yeah, next time :)
Thanks!

sakalosj:
> > > Sorry, but I don't remember, how is SOURCE_BRANCH used? Why is it needed and how does it's value influence the build?
> > 
> > 
> > SOURCE_BRANCH contains a current branch. Based on it, a master/stable branch from other projects will be chosen. introduced in #717
> 
> A case insensitive grep would have helped me with finding this answer, but yeah, next time :)
> Thanks!

np ;)

sakalosj:
@TomasTomecek @csomh 

thanks for the review, but sorry I forgot that I was "testing" this and it requires to be moved to separate file and processed by both install-deps ansible playbooks

will implement re-request review

sakalosj:
> @TomasTomecek @csomh
> 
> thanks for the review, but sorry I forgot that I was "testing" this and it requires to be moved to separate file and processed by both install-deps ansible playbooks
> 
> will implement re-request review

pls ignore - will implement it in other PR which is focused on install-deps.yaml specifically #757

Makefile
sakalosj: I will not remove it in this PR because it is used in different places and probably will require some additional adjustments

raise an issue for this #759

files/install-deps-worker.yaml
sakalosj: forgot to include it now, but will do it another  commit

PR #757, opened 2020-07-28
Improve service dependency installation approach

sakalosj:
Worker dependency installation was improved in previous changes, here same
the approach will be applied for the service dependency installation.

Ansible configuration was split into multiple files.

Source code destination was unified - now it is /src in all containers.

### DIFF
--- a/files/tasks/install-ogr-deps.yaml
+++ b/files/tasks/install-ogr-deps.yaml
@@ -1,0 +1,16 @@
+# Install dependencies from setup.cfg as rpm packages for ogr
+- name: Download script setupcfg2rpm.py
+  get_url:
+    url: https://raw.githubusercontent.com/packit-service/deployment/master/scripts/setupcfg2rpm.py
+    dest: ./setupcfg2rpm.py
+    mode: "0744"
+- name: Download ogr setup.cfg
+  get_url:
+    # ogr has only master branch
+    url: https://raw.githubusercontent.com/packit-service/ogr/master/setup.cfg
+    dest: ./ogr_setup.cfg
+- name: Install ogr dependencies provided by setupcfg2rpm
+  shell: dnf install $(./setupcfg2rpm.py ogr_setup.cfg) -y
+  args:
+    warn: no
+  become: true
--- a/files/tasks/install-packit-deps.yaml
+++ b/files/tasks/install-packit-deps.yaml
@@ -1,0 +1,26 @@
+# Install dependencies for packit
+- name: Download script setupcfg2rpm.py
+  get_url:
+    url: https://raw.githubusercontent.com/packit-service/deployment/master/scripts/setupcfg2rpm.py
+    dest: ./setupcfg2rpm.py
+    mode: "0744"
+- name: Download packit spec file
+  get_url:
+    url: https://raw.githubusercontent.com/packit-service/packit/{{ source_branch }}/packit.spec
+    dest: ./packit.spec
+- name: Install packit RPM build dependencies from packit.spec
+  shell: dnf builddep packit.spec -y
+  args:
+    warn: no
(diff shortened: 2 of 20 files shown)


sakalosj:
recheck

files/tasks/install-ogr-deps.yaml
```diff
@@ -0,0 +1,28 @@
+# Install dependencies from setup.cfg as rpm packages for ogr
+- name: Check that the setupcfg2rpm.py exists
+  stat:
+    path: ./setupcfg2rpm.py
+  register: setupcf2rpm_stat_result
```
csomh, line 5: Maybe this was introduced in a prev PR, still: wouldn't be better to remove this check and just always download the script? When can it happen that the script is already present, and if so, how is it ensured that it's up to date?

files/tasks/install-packit-deps.yaml
```diff
@@ -0,0 +1,31 @@
+# Install dependencies for packit
+- name: Check that the setupcfg2rpm.py exists
+  stat:
+    path: ./setupcfg2rpm.py
+  register: setupcf2rpm_stat_result
```
csomh, line 5: The same as the prev comment.

files/tasks/install-ogr-deps.yaml
sakalosj: both tasks files/tasks/install-ogr-deps.yaml and files/tasks/install-packit-deps.yaml use this script. To avoid unnecessary downloads and related "delays" I decided to add this check. 
Because this is run during build time it will be always the most recent version.

On the other hand, if removed, we will avoid any potential future issues. So will remove it.

files/tasks/install-ogr-deps.yaml
```diff
@@ -0,0 +1,23 @@
...
+  shell: dnf install $(./setupcfg2rpm.py ogr_setup.cfg) -y
+  args:
+    warn: no
+  become: true
+- name: Install pip deps
+  pip:
+    name:
+      - git+https://github.com/packit-service/sandcastle.git
+      - sentry-sdk
+      - flask-restx
```
jpopelka, line 22: There are not ogr dependencies, are they? Then why they are here (in `install-ogr-deps.yaml`)?

sakalosj: looks like I copied more than I wanted - will remove it

files/tasks/install-packit-deps.yaml
sakalosj: fixed

files/tasks/install-ogr-deps.yaml
sakalosj: fixed

files/tasks/install-ogr-deps.yaml
sakalosj: fixed

IceWreck:
Hey @sakalosj I get 
```

TASK [Fail on empty source_branch variable] ************************************
task path: /src/files/tasks/process-source-branch.yaml:3
fatal: [localhost]: FAILED! => {
    "changed": false
}

MSG:

Variable source_branch, which is set from env variable SOURCE_BRANCH is empty.

PLAY RECAP *********************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=1    skipped=0    rescued=0    ignored=0   

ERROR: Service 'worker' failed to build: The command '/bin/sh -c dnf install -y ansible     && cd /src/     && ansible-playbook -vv -c local -i localhost, files/install-deps-worker.yaml     && dnf clean all' returned a non-zero code: 2
```
after building on rebase.
Is this error related to anything you did here? Do I have to change or set some config/envvar?

jpopelka:
I see it too, see #762

PR #761, opened 2020-07-30
Retry propose downstream when archive is not in PyPI yet

### LINKED ISSUE #469: Propose-downstream fails when project not released to PyPI yet.
See https://github.com/packit-service/ogr/issues/348#issuecomment-596520043 triggered by https://github.com/packit-service/ogr/pull/347#issuecomment-596503222

```
[DEBUG/ForkPoolWorker-1] All jobs finished!
WARNING  {'jobs': {'propose_downstream': {'success': False, 'details': {'msg': 'Propose update failed.', 'errors': {'master': "Failed to download file from URL https://files.pythonhosted.org/packages/source/o/ogr/ogr-0.11.0.tar.gz. Reason: 'Not Found'. ", 'f32': 'The distgit repository /tmp/packit-dist-gitpptwa97d is dirty.This is not supported.', 'f31': 'The distgit repository /tmp/packit-dist-gitpptwa97d is dirty.This is not supported.', 'f30': 'The distgit repository /tmp/packit-dist-gitpptwa97d is dirty.This is not supported.'}}}}, 'event': {'trigger': 'release', 'created_at': 1583757818, 'project_url': 'https://github.com/packit-service/ogr', 'repo_namespace': 'packit-service', 'repo_name': 'ogr', 'tag_name': '0.11.0'}}
ERROR    Propose update failed.
[ERROR/ForkPoolWorker-1] Propose update failed.
```

lbarcziova:
Fixes #469 
Implements [this](https://github.com/packit-service/packit-service/issues/469#issuecomment-597098103) workflow

### DIFF
--- a/packit_service/worker/handlers/github_handlers.py
+++ b/packit_service/worker/handlers/github_handlers.py
@@ -26,6 +26,8 @@ This file defines classes for job handlers specific for Github hooks
 import logging
 from typing import Optional, Callable, Set
 
+from celery.app.task import Task
+
 from ogr.abstract import GitProject, CommitStatus
 from packit.api import PackitAPI
 from packit.config import (
@@ -41,6 +43,8 @@ from packit_service import sentry_integration
 from packit_service.constants import (
     PERMISSIONS_ERROR_WRITE_OR_ADMIN,
     FAQ_URL_HOW_TO_RETRIGGER,
+    FILE_DOWNLOAD_FAILURE,
+    RETRY_LIMIT,
 )
 from packit_service.models import (
     InstallationModel,
@@ -141,6 +145,18 @@ class ProposeDownstreamHandler(JobHandler):
     triggers = [TheJobTriggerType.release]
     task_name = TaskName.propose_downstream
 
+    def __init__(
+        self,
+        package_config: PackageConfig,
+        job_config: JobConfig,
+        data: EventData,
+        task: Task,
+    ):
+        super().__init__(
+            package_config=package_config, job_config=job_config, data=data,
+        )
+        self.task = task
+
     def run(self) -> TaskResults:
         """
         Sync the upstream release to dist-git as a pull request.
@@ -162,6 +178,16 @@ class ProposeDownstreamHandler(JobHandler):
                     dist_git_branch=branch, version=self.data.tag_name
                 )
             except Exception as ex:
+                # the archive has not been uploaded to PyPI yet
+                if FILE_DOWNLOAD_FAILURE in str(ex):
+                    # retry for the archive to become available
+                    logger.info(f"We were not able to download the archive: {ex}")
+                    # when the task hits max_retries, it raises MaxRetriesExceededError
+                    # and the error handling code would be never executed
+                    retries = self.task.request.retries
+                    if retries < RETRY_LIMIT:
+                        logger.info(f"Retrying for the {retries + 1}. time...")
+                        self.task.retry(exc=ex, countdown=15 * 2 ** retries)
                 sentry_integration.send_to_sentry(ex)
                 errors[branch] = str(ex)
 
--- a/packit_service/constants.py
+++ b/packit_service/constants.py
@@ -22,6 +22,8 @@ MSG_RETRIGGER = (
     "into this pull request."
 )
 
+FILE_DOWNLOAD_FAILURE = "Failed to download file from URL"
+
 PERMISSIONS_ERROR_WRITE_OR_ADMIN = (
     "Only users with write or admin permissions to the repository "
     "can trigger Packit-as-a-Service"
@@ -34,6 +36,8 @@ COPR_API_FAIL_STATE = 2
 
 PG_COPR_BUILD_STATUS_FAILURE = "failure"
 PG_COPR_BUILD_STATUS_SUCCESS = "success"
+
+RETRY_LIMIT = 5
 
 WHITELIST_CONSTANTS = {
     "approved_automatically": "approved_automatically",
--- a/packit_service/worker/tasks.py
+++ b/packit_service/worker/tasks.py
@@ -22,6 +22,7 @@
 import logging
 from typing import Optional
 
+from packit_service.constants import RETRY_LIMIT
 from packit_service.celerizer import celery_app
 from packit_service.service.events import (
     CoprBuildEvent,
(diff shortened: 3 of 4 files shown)


packit_service/worker/handlers/github_handlers.py
```diff
@@ -154,13 +161,24 @@ def run(self) -> TaskResults:
...
+                try:
+                    self.api.sync_release(
+                        dist_git_branch=branch, version=self.data.tag_name
+                    )
+                    exception = None
+                except Exception as ex:
+                    exception = ex
+                    # the archive has not been uploaded to PyPI yet
+                    if FILE_DOWNLOAD_FAILURE in str(ex):
+                        sleep(10)
```
lachmanfrantisek, line 175: What about some constant for this? (And isn't it in seconds?)

packit_service/worker/handlers/github_handlers.py
```diff
@@ -179,11 +197,28 @@ def run(self) -> TaskResults:
...
-            self.project.create_issue(
+            issue = self.project.create_issue(
                 title=f"[packit] Propose update failed for release {self.data.tag_name}",
                 body=body_msg,
             )
 
+            if any(FILE_DOWNLOAD_FAILURE in error for error in errors.values()):
+                retry_message = (
+                    "Since your archive was not found in PyPI,"
+                    "we will try in 10 minutes again."
```
lachmanfrantisek, line 208: Here, we can use the constant as well.

lachmanfrantisek, line 208: It's not only about `PyPI`.

```suggestion
                    "Since your archive was not found, we can't do the update in Fedora."
                    "We will try in 10 minutes again."
```

packit_service/worker/handlers/github_handlers.py
lachmanfrantisek: OK, this is a different interval, but still worth using a constant or providing a comment why we are waiting here.

lbarcziova: comment added

packit_service/worker/handlers/github_handlers.py
lbarcziova: constant added

lachmanfrantisek:
I am only thinking if we can't merge those two loops of tries and try implementing some exponential interval. It would be useful also for other tasks (e.g. the result babysit one).

I found this on the celery level: https://stackoverflow.com/questions/9731435/retry-celery-tasks-with-exponential-back-off

Or some `retry` decorator?

(I still think/hope that this code can be avoided since it's a really common requirement.)

But ignore this if you need to spend untrivial time on this...;)

jpopelka:
> Or some `retry` decorator?

In case you need one, we already use [backoff](https://pypi.org/project/backoff/) in [worker/psbugzilla](https://github.com/packit-service/packit-service/blob/master/packit_service/worker/psbugzilla.py#L98)

packit_service/worker/handlers/github_handlers.py
```diff
@@ -179,11 +201,27 @@ def run(self) -> TaskResults:
...
                 body=body_msg,
             )
 
+            if any(FILE_DOWNLOAD_FAILURE in error for error in errors.values()):
+                retry_message = (
+                    f"Since your archive was not found, we can't do the update in Fedora. "
+                    f"We will try in {PROPOSE_DOWNSTREAM_RETRY_TIME} minutes again. "
+                    f"If you haven't uploaded your archive yet, do it now, please."
+                )
+                issue.comment(body=retry_message)
```
TomasTomecek, line 215: I'm thinking if this is not an overkill - would it make sense to only do a single comment if we run out of time?

My concern here is for p-s not to be too verbose.

packit_service/worker/handlers/github_handlers.py
```diff
@@ -154,13 +163,26 @@ def run(self) -> TaskResults:
...
-                errors[branch] = str(ex)
+            exception: Optional[Exception] = None
+            for _ in range(RETRY_LIMIT):
+                try:
+                    self.api.sync_release(
+                        dist_git_branch=branch, version=self.data.tag_name
+                    )
+                    exception = None
+                except Exception as ex:
+                    exception = ex
```
TomasTomecek, line 174: log the ex please

packit_service/worker/handlers/github_handlers.py
```diff
@@ -179,11 +201,27 @@ def run(self) -> TaskResults:
...
             )
 
-            self.project.create_issue(
+            issue = self.project.create_issue(
                 title=f"[packit] Propose update failed for release {self.data.tag_name}",
                 body=body_msg,
             )
 
+            if any(FILE_DOWNLOAD_FAILURE in error for error in errors.values()):
+                retry_message = (
```
TomasTomecek, line 210: please add some logging here as well

packit_service/worker/handlers/github_handlers.py
```diff
@@ -154,13 +163,26 @@ def run(self) -> TaskResults:
...
         ):
-            try:
-                self.api.sync_release(
-                    dist_git_branch=branch, version=self.data.tag_name
-                )
-            except Exception as ex:
-                sentry_integration.send_to_sentry(ex)
-                errors[branch] = str(ex)
+            exception: Optional[Exception] = None
+            for _ in range(RETRY_LIMIT):
```
TomasTomecek, line 167: if I understand this code correctly, it's trying to fetch archive for 50 seconds

I'd say it's better design to utilize celery here and instead of doing sleep(), use celery's retry

lbarcziova:
Reimplemented with just the celery retry, please check

packit_service/worker/handlers/github_handlers.py
```diff
@@ -141,6 +145,18 @@ class ProposeDownstreamHandler(JobHandler):
...
+        self,
+        package_config: PackageConfig,
+        job_config: JobConfig,
+        data: EventData,
+        task: Task,
+    ):
+        super().__init__(
+            package_config=package_config, job_config=job_config, data=data,
+        )
+        self.task = task
```
lachmanfrantisek, line 158: What about handling the exceptions/retries on the task level to not need to introduce a new attribute here?

lbarcziova:
@TomasTomecek thanks, added at least simple test case that retry is called

packit_service/worker/handlers/github_handlers.py
lbarcziova: This seems better to me as if the exception is called more then the RETRY_LIMIT, we want to do the handling like creating an issue and so and for that check of count we need again the celery Task

PR #762, opened 2020-07-30
[Makefile/docker-compose.yaml] set SOURCE_BRANCH

jpopelka:
This was (accidentally?) removed in daf86eed8

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ AP ?= ansible-playbook -vv -c local -i localhost, -e ansible_python_interpreter=
 PATH_TO_SECRETS ?= $(CURDIR)/secrets/
 COV_REPORT ?= term-missing
 COLOR ?= yes
-SOURCE_BRANCH ?=
+SOURCE_BRANCH ?= $(shell git branch --show-current)
 
 service: files/install-deps.yaml files/recipe.yaml
 	$(CONTAINER_ENGINE) build --rm -t $(SERVICE_IMAGE) -f files/docker/Dockerfile --build-arg SOURCE_BRANCH=$(SOURCE_BRANCH) .
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -58,6 +58,8 @@ services:
     build:
       context: .
       dockerfile: files/docker/Dockerfile.worker
+      args:
+        SOURCE_BRANCH: master
     image: docker.io/usercont/packit-service-worker:dev
     command: /usr/bin/run_worker.sh
     tty: true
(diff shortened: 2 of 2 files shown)


IceWreck:
Can you please add something like this to docker-compose as well
```
build:
      context: .
      dockerfile: files/docker/Dockerfile.worker
      args:
        SOURCE_BRANCH: master

```
In service and worker

csomh:
I'm also wondering weather it would make sense to have a default of `master` in the Dockerfiles.... it's weird when someone builds directly with docker and then finds out that there was a required build argument.

Makefile
```diff
@@ -9,7 +9,7 @@ AP ?= ansible-playbook -vv -c local -i localhost, -e ansible_python_interpreter=
 PATH_TO_SECRETS ?= $(CURDIR)/secrets/
 COV_REPORT ?= term-missing
 COLOR ?= yes
-SOURCE_BRANCH ?=
+SOURCE_BRANCH ?= $(shell git branch --show-current)
```
TomasTomecek, line 12: actually, now that I think about this, I second Hunor, shouldn't we hardcode "master" here?

will the build even work if it's something which is not stable nor master?

jpopelka: https://github.com/packit-service/packit-service/blob/master/files/tasks/process-source-branch.yaml#L9

Oh come on guys, this is a hotfix PR, not a new feature PR :-)
Let me merge this and leave this to @sakalosj :-)

jpopelka:
> I'm also wondering weather it would make sense to have a default of `master` in the Dockerfiles

I don't know why @sakalosj  explicitly [fails on SOURCE_BRANCH being empty](https://github.com/packit-service/packit-service/blob/master/files/tasks/process-source-branch.yaml#L6), so I'd rather let him answer this.
As I said, this is just a hotfix to unblock @IceWreck

sakalosj:
> > I'm also wondering weather it would make sense to have a default of `master` in the Dockerfiles
> 
> I don't know why @sakalosj explicitly [fails on SOURCE_BRANCH being empty](https://github.com/packit-service/packit-service/blob/master/files/tasks/process-source-branch.yaml#L6), so I'd rather let him answer this.
> As I said, this is just a hotfix to unblock @IceWreck

@jpopelka @IceWreck @csomh @TomasTomecek 
I decided to not default to any value is to avoid situations when because of some error or misconfiguration we can end up deploying master also to prod environment. From my perspective, this behavior is oriented toward automation - we should never achieve state when SOURCE_BRANCH is not defined. 

My suggestion is to implement a check if SOURCE_BRANCH is defined into Dockerfile so it fails fast if not defined and implement this into Dockerfile too, but I am open to any solution.

Issue #368, opened 2020-01-27
Schemathesis


lachmanfrantisek:
I've tried to use [kiwicom/schemathesis](https://github.com/kiwicom/schemathesis) to test our api endpoint.

- It uses the `https://prod.packit.dev/api/swagger.json` definition of our API and uses [hypothesis](https://hypothesis.works/) to create various inputs to tests the API.
- We can try that also to write tests for our API. WDYT?
- btw, it gives me a following output:

```
$ schemathesis run https://prod.packit.dev/api/swagger.json
======================================================================================================= Schemathesis test session starts ======================================================================================================
platform Linux -- Python 3.7.6, schemathesis-0.23.5, hypothesis-5.3.1, hypothesis_jsonschema-0.11.0, jsonschema-3.2.0
rootdir: /home/flachman
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/home/flachman/.hypothesis/examples')
Schema location: https://prod.packit.dev/api/swagger.json
Base URL: https://prod.packit.dev
Specification version: Swagger 2.0
Workers: 1
collected endpoints: 12

GET /api/copr-builds .                                                                                                                                                                                                                  [  8%]
GET /api/copr-builds/{id} .                                                                                                                                                                                                             [ 16%]
GET /api/healthz .                                                                                                                                                                                                                      [ 25%]
HEAD /api/healthz .                                                                                                                                                                                                                     [ 33%]
GET /api/installations E                                                                                                                                                                                                                [ 41%]
GET /api/installations/{id} .                                                                                                                                                                                                           [ 50%]
GET /api/tasks E                                                                                                                                                                                                                        [ 58%]
GET /api/tasks/{id} .                                                                                                                                                                                                                   [ 66%]
POST /api/testing-farm/results .                                                                                                                                                                                                        [ 75%]
POST /api/webhooks/github .                                                                                                                                                                                                             [ 83%]
GET /api/whitelist .                                                                                                                                                                                                                    [ 91%]
GET /api/whitelist/{login} .                                                                                                                                                                                                            [100%]

============================================================================================================== HYPOTHESIS OUTPUT ==============================================================================================================
Unreliable test timings! On an initial run, this test took 2338.67ms, which exceeded the deadline of 500.00ms, but on a subsequent run it took 52.12 ms, which did not. If you expect this sort of variability in your test timings, consider turning deadlines off for this test by setting deadline=None.
[...]

PR #766, opened 2020-08-04
Use integer response code values

IceWreck:
Thanks to @Stranger6667   for reporting it in https://github.com/packit/packit-service/issues/368#issuecomment-668556357

Basically we were sending and documenting the response status object instead of the integer code as [required by RestX](https://flask-restx.readthedocs.io/en/latest/swagger.html?highlight=response#documenting-with-the-api-response-decorator) and the return value to a view [should also have](https://flask.palletsprojects.com/en/1.1.x/api/#flask.make_response) an integer code as status.

https://docs.python.org/3/library/http.html


IceWreck:
Done

IceWreck:
what happened to zuul?

/recheck

PR #767, opened 2020-08-05
Dockerfiles will fail if build arg is missing

sakalosj:
Dockerfiles were improved to fail fast if build arg SOURCE_BRANCH is not
provided.

### DIFF
--- a/files/docker/Dockerfile
+++ b/files/docker/Dockerfile
@@ -3,6 +3,11 @@
 FROM fedora:31
 
 ARG SOURCE_BRANCH
+RUN  if [[ -z $SOURCE_BRANCH ]]; then \
+echo -e "\nMissing SOURCE_BRANCH build argument! Please add \
+\"--build-arg SOURCE_BRANCH=<val>\" to the build command to specify it!\n\
+This is the branch used when installing other Packit projects (e.g. ogr, packit).\n" && exit 1;\
+fi
 
 ENV LANG=en_US.UTF-8 \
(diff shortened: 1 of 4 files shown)


files/docker/Dockerfile
```diff
@@ -3,6 +3,11 @@
 FROM fedora:31
 
 ARG SOURCE_BRANCH
+RUN  if [[ -z $SOURCE_BRANCH ]]; then \
+echo -e "\nSOURCE_BRANCH build argument have to be defined using \
+\"--build-arg key=val\" option.\nIt is branch which should be used for installing internal \
+projects (e.g. ogr, packit).\n" && exit 1;\
+fi
```
csomh, line 10:
```suggestion
echo -e "\nMissing SOURCE_BRANCH build argument! Please add \
\"--build-arg SOURCE_BRANCH=<val>\" to the build command to specify it!\n\
This is the branch used when installing other Packit projects (e.g. ogr, packit).\n" && exit 1;\
fi
```

(use simple sentences, avoid complicated  grammatical structures, _present simple_ is your friend :slightly_smiling_face: )

sakalosj: fixed, thanks for suggestion

csomh:
recheck

sakalosj:
recheck

sakalosj:
> It's a pretty short script, so it's not such a big deal, but it's _really_ hard to read for me.
> 
> Approving nevertheless. Not a big deal.

I understand, but cannot be indented, because it will affect the output error message and I didn't want to move it script to avoid image build.

Thanks for approval

PR #768, opened 2020-08-05
Link SRPMBuildModel to JobTriggerModel

IceWreck:
- [x] Link SRPMBuildModel to JobTriggerModel (schema migration)
- [x] Copr trigger_ids from CoprBuildModel to SRPMBuildModel
- [x] Update tests

Rn, theres no way of accessing which project an SRPM build belongs to. Theyre only linked to CoprBuildModel. I've linked them to JobTriggerModel so that we get their GitProject and other things.

Note that I have zero idea of what Im doing so there might be obvious flaws here but this seems to be working. I was able to successfully complete a copr build from a pull request on my fork of packit/hello-world and then access the git project using something like
`session.query(SRPMBuildModel).order_by(desc(SRPMBuildModel.id)).all()[0].get_project()`


packit_service/models.py
```diff
@@ -920,15 +921,23 @@ class SRPMBuildModel(Base):
...
+        job_trigger = JobTriggerModel.get_or_create(
+            type=trigger_model.job_trigger_model_type, trigger_id=trigger_model.id,
+        )
         with get_sa_session() as session:
             srpm_build = cls()
             srpm_build.logs = logs
             srpm_build.success = success
+            srpm_build.job_trigger = job_trigger
             session.add(srpm_build)
             return srpm_build
```
IceWreck, line 943: Should this be converted to get_or_create style or is only create fine ?
If so, how and why?

alembic/versions/726cb0f70c6d_link_srpmbuildmodel_to_jobtriggermodel.py
```diff
@@ -0,0 +1,34 @@
...
+revision = "726cb0f70c6d"
+down_revision = "adbdc1c21d7e"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "srpm_builds", sa.Column("job_trigger_id", sa.Integer(), nullable=True)
```
csomh, line 22: Could some of the current values be populated with the corresponding `job_trigger_id` from the `copr_builds` table?

IceWreck: Yes, I can make a script to copy those over but someone will have to run it manually

csomh: Wait, cannot this be done as part of the migration? I'm no expert in this, just asking :thinking:

IceWreck: I dont know. This migration was autogenerated but can be manually tweaked, not sure if were allowed to do any random python code here.

I was thinking of something like:
```
builds = CoprBuildModel.get_merged_chroots()
for build in builds:
    project_details = build.project
    srpm_build = build.srpm_build
    <create a trigger, but now that i think of it, i have no idea how to do it this way either>

```

csomh: @lbarcziova any advice how to do this? IIRC you did something like this in the past.

lbarcziova: Do you mean something like [this](https://github.com/packit/packit-service/blob/362302f08d54a0de3fee028ae6a01881569bb5f6/alembic/versions/c6250555a36c_.py#L282) or [this] (https://github.com/packit/packit-service/blob/362302f08d54a0de3fee028ae6a01881569bb5f6/alembic/versions/d90948124e46_add_tables_for_triggers_koji_and_tests.py#L277)?

packit_service/models.py
csomh: @TomasTomecek ping :smiley:

alembic/versions/726cb0f70c6d_link_srpmbuildmodel_to_jobtriggermodel.py
csomh: Yes, that's what we are looking for. Do you have some place in the docs we could learn more about this? I wasn't able to find anything.

packit_service/models.py
TomasTomecek: I am used to django's API, hence the get_or_create methods; this project is the first time I work with flask/SA more.

If it's needed somewhere in the code, feel free to implement get_or_create, if not, we can have `create` only, that's fine. Since we don't update them, afaik, create should be perfectly fine.

alembic/versions/726cb0f70c6d_link_srpmbuildmodel_to_jobtriggermodel.py
lbarcziova: When I was implementing it, these were helpful to read: 
https://stackoverflow.com/questions/24612395/how-do-i-execute-inserts-and-updates-in-an-alembic-upgrade-script
http://ominian.com/2019/07/11/data-migration-with-sqlalchemy-and-alembic/

lachmanfrantisek: I agree here with Laura, you can easily do that in the migration script like we do in those two files. Just don't forget to define models also in the migration script so we preserve the state of the model which corresponds to the state the migration is done. (Those models can have only a subset of attributes -- only those you need.)

IceWreck: Thanks @lbarcziova @lachmanfrantisek

lachmanfrantisek:
> Rn, theres no way of accessing which project an SRPM build belongs to. They're only linked to CoprBuildModel. I've linked them to JobTriggerModel so that we get their GitProject and other things.

Before merging this, I would like to be sure we really need that.
- There is a way to get that info:
  - One direction: We can get all the `CoprBuildModel`/`KojiBuildModel` models for that trigger and get all the `SRPMBuildModel` models of those. 
  - Reverse direction:  `SRPMBuildModel` has `copr_builds`/`koji_builds` and those have one trigger.
- What is the problem with that? Time? High CPU/memory usage?
  - Is it a real problem or some premature optimisation?
  - What about trying the manual way and optimise later if needed?
- I am really not a database expert but I was taught that we should not have more ways how to get the data from database..;)

IceWreck:
> Before merging this, I would like to be sure we really need that.
> 
>     * There is a way to get that info:
>       
>       * One direction: We can get all the `CoprBuildModel`/`KojiBuildModel` models for that trigger and get all the `SRPMBuildModel` models of those.
>       * Reverse direction:  `SRPMBuildModel` has `copr_builds`/`koji_builds` and those have one trigger.
> 
>     * What is the problem with that? Time? High CPU/memory usage?
>       
>       * Is it a real problem or some premature optimisation?
>       * What about trying the manual way and optimise later if needed?
> 
>     * I am really not a database expert but I was taught that we should not have more ways how to get the data from database..;)

This is okay for successful SRPM builds but when an SRPM build fails, there is no copr/koji build so those srpm builds have no project associated to them

lachmanfrantisek:
> This is okay for successful SRPM builds but when an SRPM build fails, there is no copr/koji build so those srpm builds have no project associated to them

OK, that's the reason I was missing. Thanks for the explanation.

An alternative solution can be to create the build instance also each time.

csomh:
> An alternative solution can be to create the build instance also each time.

I think that storing builds in the DB which didn't even happen would be strange.

IceWreck:
Ive done schema and data migrations in different alembic scripts because it was easier for development. Will combine into one if you all want to.

IceWreck:
@lachmanfrantisek I copied over the models instead of importing from models.py like you said. However, I also had to import others that I did not require like KojiBuildsModel and TFTTestRunModel because theyre related to JobTrigger which I do require. Is that okay?

lachmanfrantisek:
> @lachmanfrantisek I copied over the models instead of importing from models.py like you said. However, I also had to copy others that I did not require like KojiBuildsModel and TFTTestRunModel because theyre related to JobTrigger which I do require. Is that okay?

You can remove also unnecessary relations if you want to avoid this. (But having them there should not harm anything.)

PR #769, opened 2020-08-06
Adding project token model and gitlab authentication

shreyaspapi:
Handing webhook authentication coming from Gitlab instance.

Currently, we are manually creating tokens for GitLab projects and adding them in gitlab_webhook_tokens.

Workflow for the users -
- Project Maintainer / Admin will add packit webhook in project settings -> webhooks without any token with permissions of all events.
- Project Maintainer / Admin will add Packit-User for that instance as an admin of the repository.
- Packit-Service (user) will automatically create a confidential issue to the repository with a project-specific unique token.
- Maintainer / Admin to use packit-service will have to update the webhook with the token found in confidential issue (created by packit-user).

- [x] Add tests


shreyaspapi:
recheck

lachmanfrantisek:
recheck

packit_service/config.py
```diff
@@ -91,8 +92,12 @@ def __init__(
         self.server_name: str = ""
 
         # Makeshift for now to authenticate webhooks coming from gitlab instances
+        # Old way of authenticating
         self.gitlab_webhook_tokens: Set[str] = set(gitlab_webhook_tokens or [])
 
+        # Gitlab token secret to decode JWT tokens
+        self.gitlab_token_secret: str = gitlab_token_secret
```
lachmanfrantisek, line 99: What is the benefit here? Can we avoid having just another value in the config?

packit_service/service/api/webhooks.py
```diff
@@ -206,31 +208,86 @@ def post(self):
...
+        )
+
+        if not project_token:
+            token = jwt.encode(
+                {"namespace": namespace, "repo_name": repo_name},
+                "secret",
+                algorithm="HS256",
+            )
+
+            _ = ProjectTokenModel.create(
```
lachmanfrantisek, line 228:
```suggestion
            ProjectTokenModel.create(
```

packit_service/service/api/webhooks.py
```diff
@@ -206,31 +208,86 @@ def post(self):
...
+                namespace=namespace,
+                repo_name=repo_name,
+                project_url=http_url,
+            )
+
+            project = config.get_project(url=self.http_url)
+            logger.info("CONFIDENTIAL ISSUE CREATED HERE")
+            project.create_issue(
+                title="Packit-Service",
+                body=f"Set token {token} in webhook settings.",
```
lachmanfrantisek, line 239: - Can we guess the link?
- Is this really clear? GitLab settings are a bit complicated and are divided into a lot of pages.
- What is the name of the variable sent with the webhook?
- I would use `  for token value or other markdown formating for better readability and being able to copy-paste it easily.

packit_service/config.py
lachmanfrantisek: Is it really used?

shreyaspapi: We are using `gitlab_token_secret` to encode the json payload which in this case its {namespace: "the-namespace", repo_name:"the-repo", project_url: "https://theurl.com"}. eg - https://pyjwt.readthedocs.io/en/latest/#example-usage

With the same secret then we can decode the token and directly check for data from the event (repo_name, Namespace)

shreyaspapi: I can use webhook_secret from the config But wanted to keep things different here.

packit_service/service/api/webhooks.py
shreyaspapi: > * Can we guess the link?

Yes, making users click the link is good idea.

> * Is this really clear? GitLab settings are a bit complicated and are divided into a lot of pages.

How about,

Title- Packit-Service Authentication

Body -

To configure Packit-Service with `repo_name` you need to configure the webhook settings.
Head to https://urltowebhook.com and add the secret token `secret_token` to authenticate
requests coming to Packit.

Grant [packit_user](https://userlink.com) `admin` permissions on the {owner}/{project} project.
You can add Packit user by clicking [here](link to the add_users settings) 
  manually to match the Packit configuration.

OR

Body-

Please following these steps to configure Packit correctly

- Packit needs to validate requests coming from `repo_name`, Head to https://urltowebhook.com 
and add the secret_token as `secret_token`, to webhook settings.
- Grant [packit_user](https://userlink.com) `admin` permissions on the {owner}/{project} project.
- You can add Packit user by clicking [here](link to the add_users settings) 
  manually to match the Packit configuration.


> * What is the name of the variable sent with the webhook?

👍

lachmanfrantisek: Nice.

I like the first version of the text. It's more friendly.



> > * What is the name of the variable sent with the webhook?
> 
> it's secret_token.

Isn't UPPER_CASE more common?

packit_service/config.py
lachmanfrantisek: My point is if this info is useful to us and worth specifying another token.

Wouldn't random tokens work as well?

lachmanfrantisek: > I can use webhook_secret from the config But wanted to keep things different from Github here.

I agree. I would not reuse a different token for another usage.

packit_service/service/api/webhooks.py
shreyaspapi: Yup! Changed the body of the issue.

packit_service/service/api/webhooks.py
```diff
@@ -206,31 +208,94 @@ def post(self):
...
+        http_url = project_data["git_http_url"]
+
+        project_token = ProjectTokenModel.get_project(
+            namespace=namespace, repo_name=repo_name, project_url=http_url
+        )
+
+        if not project_token:
+            token = jwt.encode(
+                {"namespace": namespace, "repo_name": repo_name},
+                "secret",
```
lachmanfrantisek, line 224: Don't we want to use the config secret here?

packit_service/service/api/webhooks.py
```diff
@@ -206,31 +208,94 @@ def post(self):
...
+                token=token,
+                namespace=namespace,
+                repo_name=repo_name,
+                project_url=http_url,
+            )
+
+            project = config.get_project(url=http_url)
+            packit_user = project.service.user.get_username()
+
+            logger.info("CONFIDENTIAL ISSUE CREATED HERE")
```
lachmanfrantisek, line 238: Do we need the UPPER_CASE here? It's nothing serious. It's a part of the regular workflow.

packit_service/service/api/webhooks.py
```diff
@@ -206,31 +208,94 @@ def post(self):
...
         """
         https://docs.gitlab.com/ee/user/project/integrations/webhooks.html#secret-token
         """
         if "X-Gitlab-Token" not in request.headers:
             if config.validate_webhooks:
                 msg = "X-Gitlab-Token not in request.headers"
-                logger.warning(msg)
-                raise ValidationFailed(msg)
+                logger.info(msg)
+                self.create_issue_token()
```
lachmanfrantisek, line 259: The naming is a bit misleading. Do we create token or issue here?


```suggestion
                self.create_confidential_issue_with_token()
```

packit_service/service/api/webhooks.py
```diff
@@ -206,31 +208,94 @@ def post(self):
...
+                return
             else:
                 # don't validate signatures when testing locally
                 logger.debug("Ain't validating token.")
                 return
 
         token = request.headers["X-Gitlab-Token"]
```
lachmanfrantisek, line 266: I am a bit confused why we have `SECRET_TOKEN` in the issue text and `X-Gitlab-Token` here.

packit_service/service/api/webhooks.py
```diff
@@ -206,31 +208,94 @@ def post(self):
...
+        gitlab_token_secret = config.gitlab_token_secret
+        if not gitlab_token_secret:
+            msg = "'gitlab_token_secret' not specified in the config."
+            logger.error(msg)
+            raise ValidationFailed(msg)
+
+        msg_failed = "Payload token {} failed".format
+
+        try:
+            token_decoded = jwt.decode(token, "secret", algorithms=["HS256"])
```
lachmanfrantisek, line 281: Don't we want to use configured secret also here?

packit_service/service/api/webhooks.py
shreyaspapi: Gitlab webhook settings have 2 fields - URL, Secret Token. The token entered in the field is sent as X-Gitlab-Token in the header. It will be easier for the admin/maintainer while reading `SECRET TOKEN` what they are dealing with.

Screenshot 2020-08-18 at 6 25 03 PM

packit_service/service/api/webhooks.py
shreyaspapi: Yes my mistake, Changing it.

packit_service/service/api/webhooks.py
shreyaspapi: 👍

packit_service/service/api/webhooks.py
shreyaspapi: Changed it.

packit_service/service/api/webhooks.py
lachmanfrantisek: Aha. Now I get it. So we should use `Secret Token` in the issue then... Sorry.

packit_service/service/api/webhooks.py
lachmanfrantisek: OK, I misunderstood that completely. We should use `Secret token` or `Secret token field` since it's the name of the settings field, not some key-value pair user needs to specify...

shreyaspapi: Ah okay!

shreyaspapi: Done!

packit_service/service/api/webhooks.py
```diff
@@ -206,31 +208,96 @@ def post(self):
...
+
+            logger.info("Confidential issue created.")
+
+            project.create_issue(
+                title="Packit-Service Authentication",
+                body=f"To configure Packit-Service with `{repo_name}` you need to\n"
+                f"configure the webhook settings. Head to {http_url}/hooks and add\n"
+                f"the Secret Token `{token}` to authenticate requests coming to Packit.\n\n"
+                f"Grant [{packit_user}](https://gitlab.com/{packit_user}) `admin` permissions\n"
+                f"on the {namespace}/{repo_name} project. You can add Packit user by clicking\n"
```
lachmanfrantisek, line 246: Would be nice to place some reason why we need such high permissions.

packit_service/service/api/webhooks.py
```diff
@@ -206,31 +208,96 @@ def post(self):
...
+                f"[here]({http_url}/-/project_members) manually to match the Packit configuration.",
```
lachmanfrantisek, line 247: What does it mean to `match the Packit configuration`? Cannot we remove that part?

```suggestion
                f"on the {namespace}/{repo_name} project. You can add Packit user"
                f"[here]({http_url}/-/project_members).",
```

packit_service/service/api/webhooks.py
```diff
@@ -206,31 +208,96 @@ def post(self):
...
+
+        if (token_namespace, token_repo_name) == (namespace, repo_name):
+            logger.debug("Payload signature is OK.")
+        else:
+            logger.warning(msg_failed("validation"))
+            raise ValidationFailed(msg_failed("validation"))
 
     @staticmethod
-    def interested():
+    def interested(self):
```
lachmanfrantisek, line 300: Why this change?

shreyaspapi: Yeah! I should make it a static method.

shreyaspapi: That should not have been there.

packit_service/service/api/webhooks.py
shreyaspapi: I have added - 

```
Packit needs rights to set commit status to merge requests, Please grant\n"
f"[{packit_user}](https://gitlab.com/{packit_user}) `admin` permissions\n"
f"on the {namespace}/{repo_name} project.
```

packit_service/service/api/webhooks.py
```diff
@@ -206,28 +208,94 @@ def post(self):
...
+                token=token,
+                namespace=namespace,
+                repo_name=repo_name,
+                project_url=http_url,
+            )
+
+            project = config.get_project(url=http_url)
+            packit_user = project.service.user.get_username()
+
+            logger.info("Confidential issue created.")
```
lachmanfrantisek, line 238:
```suggestion
            logger.info("Confidential issue will be created.")
```

or

```suggestion
            logger.info("Creating confidential issue.")
```

packit_service/service/api/webhooks.py
```diff
@@ -206,28 +208,94 @@ def post(self):
...
+            return
+
+        gitlab_token_secret = config.gitlab_token_secret
+        if not gitlab_token_secret:
+            msg = "'gitlab_token_secret' not specified in the config."
+            logger.error(msg)
+            raise ValidationFailed(msg)
 
-        logger.debug("Payload token is OK.")
+        msg_failed = "Payload token {} failed".format
```
lachmanfrantisek, line 279: This looks handy but a bit confusing -- I would use a regular string in those two places so one does not need to search and think about that `msg_failed` function.

You can then use a meaningful sentence -- e.g. `Signature of the payload token is not valid.` and `Payload token does not match the project."

packit_service/service/api/webhooks.py
```diff
@@ -206,28 +208,94 @@ def post(self):
...
+                config.gitlab_token_secret,
+                algorithm="HS256",
+            )
+
+            ProjectTokenModel.create(
+                token=token,
+                namespace=namespace,
+                repo_name=repo_name,
+                project_url=http_url,
+            )
```
lachmanfrantisek, line 233: Wait, do we need to save it in the database?

- Looks like you are not reading it from the database when we get it.
- With the configured secret token, we can regenerate the token if needed.

That's really nice!

lachmanfrantisek: One problem with that is that we can potentially create a lot of issues. That can be easily fixed later.

packit_service/service/api/webhooks.py
```diff
@@ -206,28 +208,94 @@ def post(self):
...
+
+            project.create_issue(
+                title="Packit-Service Authentication",
+                body=f"To configure Packit-Service with `{repo_name}` you need to\n"
+                f"configure the webhook settings. Head to {http_url}/hooks and add\n"
+                f"the Secret Token `{token}` to authenticate requests coming to Packit.\n\n"
+                f"Packit needs rights to set commit status to merge requests, Please grant\n"
+                f"[{packit_user}](https://gitlab.com/{packit_user}) `admin` permissions\n"
+                f"on the {namespace}/{repo_name} project. You can add the rights by clicking\n"
+                f"[here]({http_url}/-/project_members).",
```
lachmanfrantisek, line 248: This is probably unnecessary here -- since we are creating confidential issues we already have that permission..;)

(We might want to create a regular issue if we don't have that permission.)

lachmanfrantisek (commented):
I found out that it can be much easier -- good job with picking `jwt`. We can avoid saving tokens in the database with that..;)

shreyaspapi:
> I found out that it can be much easier -- good job with picking `jwt`. We can avoid saving tokens in the database with that..;)

We can't avoid saving tokens in the database as we can't keep a track that we have created a confidential issue or not.

packit_service/service/api/webhooks.py
```diff
@@ -206,28 +208,94 @@ def post(self):
...
+        project_data = request.json["project"]
+
+        namespace, repo_name = project_data["path_with_namespace"].split("/")
+        http_url = project_data["git_http_url"]
+
+        project_token = ProjectTokenModel.get_project(
+            namespace=namespace, repo_name=repo_name, project_url=http_url
+        )
+
+        if not project_token:
```
shreyaspapi, line 221: Here I am checking if the project token in present in the database or not. If present we can assume we have created an issue.

packit_service/service/api/webhooks.py
shreyaspapi: Anyone with or without permission to the repository can create confidential issues. User which don't have permission cannot see other confidential issues created.

packit_service/service/api/webhooks.py
shreyaspapi: Yes `Signature of the payload token is not valid.` this sounds better.

packit_service/service/api/webhooks.py
shreyaspapi: Do we really want to do this? While testing, I did exactly this and it created 7 issues in few seconds. On each event, it created an issue.

shreyaspapi: We might skip the database and check if the repo issues if we have created one or not (But this has a bit delay and will have issues with the race conditions).

lachmanfrantisek: > Do we really want to do this? While testing, I did exactly this and it created 7 issues in few seconds. On each event, it will create an issue.

7 issues? How come we got 7 messages at once?

packit_service/service/api/webhooks.py
lachmanfrantisek: Aha, ok. My bad. That does not make sense..;)

Thanks for the explanation.

packit_service/service/api/webhooks.py
lachmanfrantisek: > We might skip the database and check with the repo issues if we have created one or not (But this has a bit delay and will have issues with the race conditions).

We can at least try. The problem is that we can't list confidential issues if we don't have admin permissions.

shreyaspapi: Comment events, the issue created by the Packit itself counted as an event. 😅

lachmanfrantisek:
> We can't avoid saving tokens in the database as we can't keep a track that we have created a confidential issue or not.

We can avoid that. What you actually want to save in the database is the boolean (=if we have already created an issue) and not the token itself.

What about removing the model for now and fix the duplicate issues later?

packit_service/service/api/webhooks.py
shreyaspapi: > What about removing the model for now and fix the duplicate issues later?

Yes okay!

shreyaspapi:
> * If you really want, we can add a different model collecting only projects that has token generated and issue created. (Not the token itself.)

Yes, I think this will be a better idea to go with.

packit_service/service/api/webhooks.py
lachmanfrantisek: Yes, I got it. My point is that saving tokens is now only an implementation detail how to get rid of the duplicates and we really don't care about the token value.

packit_service/service/api/webhooks.py
shreyaspapi: Made them in different variables.

packit_service/models.py
```diff
@@ -1119,6 +1120,49 @@ def get_range(cls, first: int, last: int) -> Optional[Iterable["TFTTestRunModel"
             ]
 
 
+class ProjectTokenModel(Base):
```
lachmanfrantisek, line 1123: The name does not match the current content. We don't save tokens any more.

lachmanfrantisek (commented):
Looks really good now :hooray:

Just one name leftover and I would maybe like to see a test case for each scenario:
- no permissions, no token
- no permissions, issue with token created
- permissions, no token 
- permissions, issue with token created

packit_service/models.py
shreyaspapi: How about `ProjectAuthenticationIssueModel`

lachmanfrantisek: Sounds good.

lachmanfrantisek: Don't forget to rename table as well...

packit_service/service/api/webhooks.py
```diff
@@ -206,28 +208,95 @@ def post(self):
...
+        project_data = request.json["project"]
+        namespace, repo_name = project_data["path_with_namespace"].split("/")
+        token_namespace = token_decoded["namespace"]
+        token_repo_name = token_decoded["repo_name"]
+
+        if (token_namespace, token_repo_name) == (namespace, repo_name):
+            logger.debug("Payload signature is OK.")
+        else:
+            logger.warning(msg_failed_validation)
+            raise ValidationFailed(msg_failed_validation)
```
lachmanfrantisek, line 299: If I can suggest another nit-pick change..;)

```suggestion
        try:
            token_decoded = jwt.decode(
                token, config.gitlab_token_secret, algorithms=["HS256"]
            )
        except jwt.exceptions.InvalidSignatureError:
            msg_failed_signature = "Payload token does not match the project."
            logger.warning(msg_failed_signature)
            raise ValidationFailed(msg_failed_signature)

        project_data = request.json["project"]
        namespace, repo_name = project_data["path_with_namespace"].split("/")
        token_namespace = token_decoded["namespace"]
        token_repo_name = token_decoded["repo_name"]

        if (token_namespace, token_repo_name) == (namespace, repo_name):
            logger.debug("Payload signature is OK.")
        else:
            msg_failed_validation = "Signature of the payload token is not valid."
            logger.warning(msg_failed_validation)
            raise ValidationFailed(msg_failed_validation)
```

lachmanfrantisek (approved):
Nicely done!

:tada: Thanks for your hard work on this! :tada:

PR #773, opened 2020-08-11
Use stg parameter in PackitAPI

lbarcziova:
Needs packit/packit#922

### DIFF
--- a/packit_service/config.py
+++ b/packit_service/config.py
@@ -153,6 +153,9 @@ class ServiceConfig(Config):
             f"server_name='{self.server_name}')"
         )
 
+    def use_stage(self) -> bool:
+        return self.deployment != Deployment.prod
+
     @classmethod
     def get_from_dict(cls, raw_dict: dict) -> "ServiceConfig":
         # required to avoid circular imports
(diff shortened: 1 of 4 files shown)


packit_service/config.py
```diff
@@ -112,6 +112,9 @@ def hide(token: str) -> str:
             f"server_name='{self.server_name}')"
         )
 
+    def use_stage(self) -> bool:
+        return self.deployment != Deployment.prod
```
TomasTomecek, line 116: why not `self.deployment == Deployment.stg`? :)

jpopelka: In general: we compare to prod to be sure that we use prod configuration only on prod.
So when you create another 'deployment', like I did when trying the CentOS new OCP4 cluster, Laura's `use_stage()` would return `True`, while yours would return `False` (and prod configuration would be used).

So maybe we need to twist the logic and have `prod` attribute & `use_prod()` instead of `stage` & `use_stage()`?

lbarcziova:
Will be merged after *.stg.fedoraproject.org will work (https://fedoraproject.org/wiki/Infrastructure/2020-post-datacenter-move-known-issues#staging)

lachmanfrantisek:
Stale-bot, give us some more time to finish this...

jpopelka (approved):

Issue #755, opened 2020-07-27
only one job executed for 2 build definitations for the same trigger


TomasTomecek:
```yaml
---                                                                                                                                                                          
specfile_path: hello.spec                                                                                                                                                    
synced_files:                                                                                                                                                                
  - hello.spec                                                                                                                                                               
upstream_package_name: hello                                                                                                                                                 
downstream_package_name: hello                                                                                                                                               
jobs:                                                                                                                                                                        
- job: copr_build                                                                                                                                                            
  trigger: pull_request                                                                                                                                                      
  metadata:                                                                                                                                                                  
    targets:                                                                                                                                                                 
    - fedora-rawhide-x86_64                                                                                                                                                  
  actions:                                                                                                                                                                   
    post-upstream-clone: "ls /*"                                                                                                                                             
- job: copr_build                                                                                                                                                            
  trigger: pull_request                                                                                                                                                      
  metadata:                                                                                                                                                                  
    targets:                                                                                                                                                                 
    - fedora-32-x86_64                                                                                                                                                       
  actions:                                                                                                                                                                   
    post-upstream-clone: 'bash -c "ls /*"'   
```

And I only got:
```
packit-stg/rpm-build-fedora-rawhide-x86_64 — SRPM build failed, check the logs for details. 
```
no F32

PR #787, opened 2020-08-20
Use job config from handler as a first choice

lachmanfrantisek:
- Use job_config from handler as a first choice.
- Fix the copr-build tests.
- Fixes: https://github.com/packit/packit-service/issues/755


TomasTomecek (approved):
interesting

PR #792, opened 2020-08-28
Make PullRequestLabelPagureEvent JSON serializable

csomh:
The "labels" attribute of the class was defined as a set, but sets are
not JSON serializable. Change it to be a list instead and add a test to
check that the class is serializable.

### DIFF
--- a/packit_service/service/events.py
+++ b/packit_service/service/events.py
@@ -27,7 +27,7 @@ import copy
 import enum
 import logging
 from datetime import datetime, timezone
-from typing import Optional, List, Union, Dict, Set
+from typing import Optional, List, Union, Dict
 
 from ogr.abstract import GitProject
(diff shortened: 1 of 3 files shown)


jpopelka (approved):
Yes, this or have the `PullRequestLabelPagureEvent.labels` as a `list` instead of the `set` so that we don't convert back and forth (https://github.com/packit/packit-service/blob/master/packit_service/worker/parser.py#L1007)

csomh:
> Yes, this or have the `PullRequestLabelPagureEvent.labels` as a `list` instead of the `set` so that we don't convert back and forth (https://github.com/packit/packit-service/blob/master/packit_service/worker/parser.py#L1007)

@jpopelka yeah, this is a good idea and I did just that. Please take another look! :)

jpopelka (approved):
Thanks

PR #798, opened 2020-09-07
Better Testing-farm response for cluster problems

lachmanfrantisek:
- Response to testing-farm errors.
- Link centos-ci issue on known problem.

### DIFF
--- a/packit_service/worker/handlers/testing_farm_handlers.py
+++ b/packit_service/worker/handlers/testing_farm_handlers.py
@@ -106,14 +106,22 @@ class TestingFarmResultsHandler(JobHandler):
         if self.result == TestingFarmResult.passed:
             status = CommitStatus.success
             passed = True
-
+        elif self.result == TestingFarmResult.error:
+            status = CommitStatus.error
+            passed = False
         else:
             status = CommitStatus.failure
             passed = False
 
+        github_status_url = self.log_url
         if len(self.tests) == 1 and self.tests[0].name == "/install/copr-build":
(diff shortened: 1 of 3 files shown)


lachmanfrantisek:
recheck

packit_service/worker/handlers/testing_farm_handlers.py
```diff
@@ -106,14 +106,22 @@ def run(self) -> TaskResults:
...
 
+        github_status_url = self.log_url
         if len(self.tests) == 1 and self.tests[0].name == "/install/copr-build":
             logger.debug("No-fmf scenario discovered.")
             short_msg = "Installation passed" if passed else "Installation failed"
+        elif self.message.startswith(
+            "Command '['git', 'clone'"
+        ) and self.message.endswith("failed with exit code 128"):
+            short_msg = "Problem with Testing-Farm cluster"
+            github_status_url = "https://pagure.io/centos-infra/issue/85"
```
csomh, line 124: Would it make sense to make these (the identification of the issue, the message and url to be used) configurable, instead of hard-coding it?

This way we could turn these on and off from configuration - which might be needed as not all `git clone` issues of the future will have to do with this particular infra problam.

lachmanfrantisek: I agree that it will be nice but since this is only a short-term solution, I would stick with the current way.

- We will switch to a different TF API. => This code will be completely rewritten.
- I don't have much time for this and if we don't merge this soon, it will not bring any benefit.

csomh (approved):

PR #800, opened 2020-09-08
Support AWS SQS & RDS

jpopelka:
1. We use AWS SQS if both `AWS_ACCESS_KEY_ID` & `AWS_SECRET_ACCESS_KEY` are set.
It they are not, fall back to redis.
By default (if `QUEUE_NAME_PREFIX` is not set) we use `packit-$DEPLOYMENT-` queue prefix,
i.e. we have `packit-stg-celery` and `packit-prod-celery` queues.

2. Also allow to specify PostgreSQL host to be able to use other than local instance (e.g. AWS RDS PostgreSQL).

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,16 +4,16 @@
 
 repos:
   - repo: https://github.com/psf/black
-    rev: stable
+    rev: 20.8b1
     hooks:
       - id: black
   - repo: https://github.com/prettier/prettier
-    rev: 2.0.5
+    rev: 2.1.1
     hooks:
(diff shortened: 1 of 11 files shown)


jpopelka:
>   packit-service-tests-requre : RETRY_LIMIT

Will try again after we update deployment playbooks in https://github.com/packit/deployment/pull/131

jpopelka:
recheck

lachmanfrantisek (approved):
Looks good to me, but you might want to have another review by someone more openshift experienced.

Thanks for the pre-commit update!

PR #801, opened 2020-09-09
Turn PR labels into a set in the handler

csomh:
PR #792, made PullRequestLabelPagureEvents JSON serializable, but this
also means that the handler receives these events as a list, not as a
set.

Due to this the intersection used to find out whether the service should
react on the labels applied fails with an AttributeError ('list' object
has no attribute 'intersection'). To fix this, explicitly construct a
set in the handler's `__init__()`.

### DIFF
--- a/packit_service/worker/handlers/pagure_handlers.py
+++ b/packit_service/worker/handlers/pagure_handlers.py
@@ -21,7 +21,7 @@
 # SOFTWARE.
 
 import logging
-from typing import Optional, Set
+from typing import Optional, List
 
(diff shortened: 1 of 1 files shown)


csomh:
This was found [on stg by Sentry](https://sentry.io/organizations/red-hat-0p/issues/1884226335/events/bb0c4473bd804069adcf68f7e23f6f90/?project=1767823).

lachmanfrantisek (approved):

jpopelka (approved):

PR #804, opened 2020-09-10
[deployment.yaml] Set 'with_pushgateway: false' in dev.yml

jpopelka:
`with_redis_commander` & `with_flower` are now `false` by default


lbarcziova (approved):
Thanks!

PR #805, opened 2020-09-11
[models.get_pg_url] POSTGRES_SERVICE_HOST -> POSTGRESQL_HOST

jpopelka:
`POSTGRES_SERVICE_HOST` is set by k8s/openshift and points to local 'postgres'
while `POSTGRESQL_HOST` is set by us and points to
either AWS RDS instance url or also local 'postgres'

(I initially had this change in my previous #800 but then removed it when debugging failing tests in openshift)


jpopelka:
recheck

jpopelka:
merging myself :-)

PR #816, opened 2020-09-17
Add header to status table

lachmanfrantisek:
- Add missing markdown header to the status table.
- Example of the current state: https://gitlab.com/packit-service/src/drpm/-/merge_requests/2#note_414152514

### DIFF
--- a/packit_service/worker/reporting.py
+++ b/packit_service/worker/reporting.py
@@ -100,11 +100,13 @@ class StatusReporter:
 
         if isinstance(check_names, str):
             check_names = [check_names]
-        comment_msg = [
-            f"| [{check}]({url}) | {state.name.upper()} |" for check in check_names
-        ]
 
-        self.project.pr_comment(self.pr_id, "\n\n".join(comment_msg))
+        comment_table_rows = [
+            "| Job | Result |",
+            "| ------------- | ------------ |",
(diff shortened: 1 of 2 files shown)


packit_service/worker/reporting.py
```diff
@@ -100,11 +100,12 @@ def report_status_by_comment(
 
         if isinstance(check_names, str):
             check_names = [check_names]
-        comment_msg = [
+
+        comment_table_rows = ["| | |", "| ------------- | ------------ |"] + [
```
mfocko, line 104: The empty header looks a bit weird on GitLab, I cannot seem to find table like that on GitHub in hello-world, what about adding comment like:
```suggestion
        comment_table_rows = ["| Job | Result |", "| ------------- | ------------ |"] + [
```

mfocko (approved):
Looks nice :)

packit_service/worker/reporting.py
lachmanfrantisek: > I cannot seem to find table like that on GitHub in hello-world

Because we don't use comments on GitHub for that. (It's used when we can't set status on MR.)

lachmanfrantisek: I've added the text header and fix the tests. PTAL

PR #822, opened 2020-09-18
WIP: [files/install-deps*.yaml] base image is now F32 

jpopelka:
Depends-on: packit/deployment#141
Fixes: packit/packit#965

/usr/bin/alembic is at the moment broken symlink

### DIFF
--- a/files/docker/Dockerfile
+++ b/files/docker/Dockerfile
@@ -1,6 +1,6 @@
 # Image for the web service (httpd), for celery worker see files/docker/Dockerfile.worker
 
-FROM docker.io/usercont/base
+FROM fedora:32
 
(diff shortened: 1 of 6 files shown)


jpopelka:
I'm seeing a problem with `docker-compose up worker` (even if I put `command: sleep 1000` instead of `command: /usr/bin/run_worker.sh` and then run the `run_worker.sh` in the container it works OK)
but in local `oc cluster up` everything works OK.

sakalosj:
@jpopelka 
please check https://github.com/packit/deployment/pull/141#issuecomment-695987918

jpopelka:
hmmm, yeah, looks like the damn readiness probe is to blame here

PR #824, opened 2020-09-21
Pin celery to 4.4.7

jpopelka:
Also dnf install `python3-pycurl`.
This is to work-around failing reverse-dep tests in packit because I'm getting out of ideas why `reverse-dep-packit-service-tests` job [can't build pycurl](https://softwarefactory-project.io/logs/64/964/a7b8552475cbf4650465be7d6e260b276654bb28/check/reverse-dep-packit-service-tests/1ccb6d9/job-output.txt).
celery 4.4.7 requires pycurl-7.43.0.5 which is the same version as python3-pycurl on Fedora 32 (in Zuul)


lachmanfrantisek (approved):
Let's try that..;)

PR #826, opened 2020-09-21
Base images on F32

jpopelka:
Since we [removed worker probes](https://github.com/packit/deployment/pull/142) let's try whether it's feasible to base images on F32 or [Zuul would still object](https://github.com/packit/deployment/pull/141#issuecomment-695987918).

### DIFF
--- a/files/docker/Dockerfile.tests
+++ b/files/docker/Dockerfile.tests
@@ -1,6 +1,6 @@
 # For running tests locally, see check_in_container target in Makefile
 
-FROM docker.io/usercont/packit-service-worker:stg
+FROM docker.io/usercont/packit-service-worker:dev
 
 ARG SOURCE_BRANCH
 RUN  if [[ -z $SOURCE_BRANCH ]]; then \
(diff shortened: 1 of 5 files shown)


files/docker/Dockerfile
```diff
@@ -1,6 +1,6 @@
 # Image for the web service (httpd), for celery worker see files/docker/Dockerfile.worker
 
-FROM docker.io/usercont/base
+FROM fedora:32
```
sakalosj, line 3: you should update the [base image]( https://github.com/packit/deployment/blob/master/containers/Dockerfile.base)
it already contains ansible

sakalosj (changes requested):

files/docker/Dockerfile
jpopelka: I know, this is the 'Temporary commit, to be removed' :-)

sakalosj: ok sorry - got the point :)

sakalosj (approved):

jpopelka:
recheck

jpopelka:
It's green so once a [new base image](https://github.com/packit/deployment/pull/141) is built, I'll remove the last commit and merge (if someone approves this)

sakalosj (approved):

jpopelka:
recheck

PR #827, opened 2020-09-22
Push metrics about created copr-builds

lbarcziova:


packit_service/worker/monitoring.py
```diff
@@ -0,0 +1,37 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class Pushgateway:
+    def __init__(self):
+        self.pushgateway_address = (
+            "http://pushgateway" if os.getenv("DEPLOYMENT") in ("stg", "prod") else None
+        )
+        # so that workers don't overwrite their metrics,
```
csomh, line 14:
```suggestion
        # so that workers don't overwrite each other's metrics,
```

packit_service/worker/monitoring.py
```diff
@@ -0,0 +1,37 @@
...
+
+from prometheus_client import CollectorRegistry, Counter, push_to_gateway
+
+logger = logging.getLogger(__name__)
+
+
+class Pushgateway:
+    def __init__(self):
+        self.pushgateway_address = (
+            "http://pushgateway" if os.getenv("DEPLOYMENT") in ("stg", "prod") else None
```
csomh, line 12: I know this is going to be more complicated, but could we make this address to be configurable, defaulting to "http://pushgateway"?

With the current approach, I would need to change the code if I want to have this functionality on a local deployment, or I would need to name my deployment "stg" or "prod" which can be confusing.

packit_service/worker/monitoring.py
```diff
@@ -0,0 +1,37 @@
...
+        )
+        # so that workers don't overwrite their metrics,
+        # the job name corresponds to worker name (e.g. packit-worker-0)
+        self.worker_name = os.getenv("HOSTNAME")
+        self.registry = CollectorRegistry()
+
+        # metrics
+        self.copr_builds_created = Counter(
+            "copr_builds_created_count",
+            "Count of created copr-builds",
```
csomh, line 22: I think it's not required to indicate the type of the metric in the name, as that one is going to be reported in the endpoint.

And `_total` and `_created` seems to be automatically added by Prometheus. (Saying this based on how `github_webhook_calls` shows up in https://prod.packit.dev/metrics .)

```suggestion
        self.copr_builds = Counter(
            "copr_builds",
            "Number of Copr builds created",
```

lbarcziova:
recheck

csomh (approved):
Thank you! LGTM :rocket:

PR #828, opened 2020-09-22
Don't raise exception when REDIS_SERVICE_HOST not set

jpopelka:
Just default to 'redis', as we did before b0080eed454 so we don't have to set it in all the tests.


lachmanfrantisek (approved):
Thanks!

Issue #789, opened 2020-08-25
Wrong copr repo in testing farm for branch builds


lachmanfrantisek:
https://console-testing-farm.apps.ci.centos.org/pipeline/cd908ddc-a96e-4525-82a8-869d4fd216f7

```
Artifacts: https://artifacts-testing-farm.apps.ci.centos.org/cd908ddc-a96e-4525-82a8-869d4fd216f7/
Console:
[15:09:57] [+] [cruncher] Getting FMF metadata from git repository 'https://github.com/packit/packit.git' ref '8e151e88ed844a88d8908d37f3df3f3c5493136d' [15:10:33] [+] [cruncher] Downloading image 'https://kojipkgs.fedoraproject.org/compose/cloud/latest-Fedora-Cloud-31/compose/Cloud/x86_64/images/Fedora-Cloud-Base-31-20200824.0.x86_64.qcow2'
:
:
[15:13:08] [+] [cruncher] posting result message to URLs:[ "https://scheduler-testing-farm.apps.ci.centos.org/v0/result", "https://prod.packit.dev/api/testing-farm/results"]
[15:13:09] [+] [cruncher] [/ci/test/build/full] [finish] Destroying test machine
[15:13:09] [E] Pipeline reported an exception in module 'cruncher': Command 'cd ~/git-source; dnf -y copr enable packit/packit-packit-master' failed with return code '1'
[15:13:09] [E] Exiting with status -1
```

AC:
- [ ] check that we send correct owner+project for branch builds (also when custom settings are configured)
- [ ] correct possible problems
- [ ] create tests to check that

thrix:
Thanks, let me know if we need to fix something

lachmanfrantisek:
Some action to trigger sync2jira.

lachmanfrantisek:
To make some conflusion, the problem was that the test was triggered when it should not be triggered. The linked issue fixed that job parsing and mapping to handlers and job-configs.

PR #837, opened 2020-10-01
no hardcoded mapping

lachmanfrantisek:
- Do not hardcode mapping from event to job.
- Fixes: #789


lachmanfrantisek:
recheck

lachmanfrantisek:
recheck

TomasTomecek (approved):
I yield to tests and stg

lachmanfrantisek:
I've added a `TODO` comment for mapping from issue comments that needs to be solved properly in the future. (For now, we have `issue -> propose-update`, but we need to make it general.)

lachmanfrantisek:
@jscotka I've skipped the test so we can use zuul for merging.

PR #843, opened 2020-10-08
already enable requre test for fedpkg

jscotka:
test already passed in zuul.
Solution was clean, just copy all files via glob instead of just `yaml` file and regenerate test_data

```
tests_requre/openshift_integration/test_fedpkg.py::FedPkg::test_fedpkg_clone PASSED [ 74%]
```


lachmanfrantisek (approved):
Thanks a lot!!!

Issue #793, opened 2020-08-28
Unexpected username and e-mail in a changelog entry generated by propose_downstream job


nforro:
See for example [here](https://src.fedoraproject.org/rpms/rebase-helper/blob/master/f/rebase-helper.spec#_104).

`rebase-helper <rebase-helper@localhost.local>` are fallback values used by rebase-helper in case it's not possible to obtain the proper values from git configuration (i.e. `git config --get user.name` and `git config --get user.email` fail).

jpopelka:
hmm, the image looks OK to me:
```
$ docker run -ti --rm usercont/packit-service-worker:prod git config -l
user.name=Packit Service
user.email=user-cont-team+packit-service@redhat.com
```

TomasTomecek:
...but sandcastle doesn't have any

```
$ docker run -ti --rm usercont/sandcastle:prod git config --get user.name
$ echo $?
1
```

jpopelka:
I thought we use sandcastle only to build SRPM.

lachmanfrantisek:
> ...but sandcastle doesn't have any

Can we move this issue to sandcastle then?

TomasTomecek:
sandcastle is being used in multiple places and it seems bumpspec is executed in sandcastle

this definitely needs some investigation

anyway, @nforro this is great for you, your project is getting the credit :))

lachmanfrantisek:
ping sync2jira

lbarcziova:
Some more investigation:

- updating of the changelog [is not happening](https://github.com/packit/packit/blob/b531c63eac475c4909d51aa3c4cf636c78ef6c77/packit/specfile.py#L63) in sandcastle

- the commits in which the changelog entries have these default values have the names in `Signed off` section set correctly, but not the email (should be `user-cont-team+packit-service@redhat.com`):
` Signed-off-by: Packit Service <packit@packit-worker-2.packit-worker.packit-prod.svc.cluster.local> `
` Signed-off-by: Packit Service <packit@packit-worker-1.packit-worker.packit-prod.svc.cluster.local>` (haven't found record with worker-0)

-  happened in that time in other projects too  - https://src.fedoraproject.org/rpms/python-ogr/blob/f32/f/python-ogr.spec#_58, https://src.fedoraproject.org/rpms/packit/blob/f32/f/packit.spec#_84

TomasTomecek:
what does rpmdev-packager print?
```
$ rpmdev-packager                                                                                                                                
[...]

PR #849, opened 2020-10-15
Test the git config values in worker /src

lbarcziova:
This should test whether the gitconfig values in worker are set correctly (#793), the other tests in openshift run in the pod using the packit-service-tests image where the pytest is installed, but we want to run these checks in worker, where the pytest is not installed.

### DIFF
--- a/files/check-inside-openshift.yaml
+++ b/files/check-inside-openshift.yaml
@@ -127,3 +127,19 @@
     # rsync works only on a running pod
     - name: get generated responses from the pod
       command: oc rsync pod/get-requre-data:{{ remote_data_dir }} {{ local_data_dir }}
+
+    - name: get the git config name in worker /src
+      command: oc exec packit-worker-0 -- bash -c "cd /src && git config --global user.name"
+      register: git_name
+
+    - name: get the git config email in worker /src
+      command: oc exec packit-worker-0 -- bash -c "cd /src && git config --global user.email"
+      register: git_email
+
+    - name: check the git config values
+      assert:
+        that:
+          - git_name.stdout == "Packit Service"
+          - git_email.stdout == "user-cont-team+packit-service@redhat.com"
+        msg:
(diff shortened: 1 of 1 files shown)


lbarcziova:
```Incorrect git config values: zuul , zuul@softwarefactory-project.io'``` :confused:

csomh:
Isn't the problem here, that the worker image is built on a Zuul node, and during building [the `.git` directory is copied to the image](https://github.com/packit/packit-service/blob/f064ca1ea0eefd43a41172a60d08be4307af7649/files/docker/Dockerfile.worker#L28)? Maybe Zuul will set these values in `.git/config` when cloning the repository?

Would it make sense to explicitly configure git when building the image?

lbarcziova:
> Isn't the problem here, that the worker image is built on a Zuul node, and during building the .git directory is copied to the image? Maybe Zuul will set these values in .git/config when cloning the repository?

Yes, Zuul probably sets their values when cloning the repo during building in tests, therefore it is not possible to test.

> Would it make sense to explicitly configure git when building the image?

 How do you mean this? We already [copy the gitconfig file into ~/.gitconfig](https://github.com/packit/packit-service/blob/5de67b82d56b5dffe86a2f084673e4c707804997/files/recipe-worker.yaml#L15)

TomasTomecek:
Would it make sense to use `--global` or `--system`, since we don't really care about `--local`. Also, you could invoke the command outside of a git repo to get the global values.

csomh:
Or another idea: maybe use `--file` and explicitly specify which file to use for Git configuration. The user (and so the home directory) which builds the image might not be the same that runs it.

lbarcziova:
> Would it make sense to use --global or --system, since we don't really care about --local. Also, you could invoke the command outside of a git repo to get the global values.

I just wanted to test exactly what happens when [getting values for changelog](https://github.com/packit/packit-service/issues/793#issuecomment-708330663), but we can try the test with --global.

TomasTomecek (approved):
LGTM

I wonder if we should change this in rebase-helper

PR #852, opened 2020-10-19
Set error status when running non-scratch production build

### LINKED ISSUE #848: Document that we can't run production (non-scratch) builds from upstream
It looks (#842) like we can't do non-scratch builds and we need to check with releng if and how they can allow that. (See https://pagure.io/releng/issue/9801 for more info.)

Until we need more, we need to:
- [ ] document the requirement for scratch=true in the docs
- [ ] put a link with a link to the pagure issue above when scratch=false is used.

lbarcziova:
Fixes #848

### DIFF
--- a/packit_service/worker/build/koji_build.py
+++ b/packit_service/worker/build/koji_build.py
@@ -31,7 +31,7 @@ from packit.exceptions import PackitCommandFailedError
 
 from packit_service import sentry_integration
 from packit_service.config import ServiceConfig
-from packit_service.constants import MSG_RETRIGGER
+from packit_service.constants import MSG_RETRIGGER, KOJI_PRODUCTION_BUILDS_ISSUE
 from packit_service.models import KojiBuildModel
 from packit_service.service.urls import (
     get_srpm_log_url_from_flask,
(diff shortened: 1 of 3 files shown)


packit_service/worker/build/koji_build.py
```diff
@@ -113,6 +113,17 @@ def supported_koji_targets(self):
         return self._supported_koji_targets
 
     def run_koji_build(self) -> TaskResults:
+        if not self.is_scratch:
+            msg = (
+                "It is not possible to run non-scratch production builds from upstream."
```
lachmanfrantisek, line 118: I would try to make it shorter. The space is really limited on Github so I would try to put most important info to the beginning...

```suggestion
                "Non-scratch builds not possible from upstream."
```

lachmanfrantisek (commented):
Nice!!

Just one suggestion about status text, otherwise LGTM

packit_service/worker/build/koji_build.py
lbarcziova: right, done

lachmanfrantisek (approved):
:+1:

PR #853, opened 2020-10-21
Update pre-commit configuration for prettier

csomh:
In order to solve this installation issue:

https://github.com/prettier/prettier/issues/9459

It seems that prettier is moving the pre-commit parts in a separate
repository.


mfocko (approved):
pre-commit is passing, LGTM :)

PR #861, opened 2020-11-09
Fix auto aliases issue

sakalosj:
`get_valid_build_targets` will be used instead of `get_build_targets`
Tests were updated


tests/conftest.py
```diff
@@ -199,3 +200,12 @@ def gitlab_mr_webhook():
...
     return Parser.parse_mr_event(gitlab_mr_webhook)
+
+
+@pytest.fixture
+def clean_packit_function_cache():
+    functions = [
+        packit.config.aliases.get_aliases,
+        packit.copr_helper.CoprHelper.get_available_chroots,
+    ]
+    [f.cache_clear() for f in functions]
```
TomasTomecek, line 211: could you please document what this does, never heard of it

TomasTomecek (approved):
one docs request, otherwise LGTM

tests/conftest.py
sakalosj: @TomasTomecek 
documented and a bit reworked, its just clearing lru_cache in functions which were decorated using `@functools.lru_cache`
This decorator adds "method" (or callable attribute) to function `.cache_clear` and this is just invoking it

PR #863, opened 2020-11-11
Don't fall back to flask_restplus 

jpopelka:
We are now [Fedora-32 based](https://github.com/packit/deployment/blob/master/containers/Dockerfile.base#L4), which [ships flask_restx](https://src.fedoraproject.org/rpms/python-flask-restx).


lachmanfrantisek (approved):
:+1:

PR #864, opened 2020-11-12
Add field source to GitProjectModel

sakalosj:
Agreed on updating GitProjectModel:

* field: instance_url
* content: git service url (str)


packit_service/models.py
```diff
@@ -131,6 +132,11 @@ class GitProjectModel(Base):
     # Example: https://github.com/packit/hello-world.git
     https_url = Column(String)
     project_url = Column(String)
+    source = Column(String, nullable=False)
```
lachmanfrantisek, line 135: What about calling it `instance_url` like in other places across the code and in OGR?

The thing that is (probably) not consistent is if we add protocol or not.

alembic/versions/e35bb9ce7313_added_source_field_to_gitproject.py
```diff
@@ -0,0 +1,36 @@
...
+depends_on = None
+
+
+def upgrade():
+    bind = op.get_bind()
+    session = orm.Session(bind=bind)
+    op.add_column("git_projects", sa.Column("source", sa.String(), nullable=True))
+    for git_project in session.query(GitProjectModel):
+        git_project.source = urlparse(git_project.project_url).netloc
+    session.commit()
```
lachmanfrantisek, line 29: Nice!

lachmanfrantisek (commented):
Nice! Do you want to work on the usage in another PR?

(e.g. simplifying/getting rid of `__choose_project`)

packit_service/models.py
TomasTomecek: agreed with the rename, "source" can literally mean anything

alembic/versions/e35bb9ce7313_added_instance_url_field_to_gitproject.py
```diff
@@ -0,0 +1,36 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    bind = op.get_bind()
+    session = orm.Session(bind=bind)
+    op.add_column("git_projects", sa.Column("instance_url", sa.String(), nullable=True))
+    for git_project in session.query(GitProjectModel):
+        git_project.instance_url = urlparse(git_project.project_url).netloc
```
lachmanfrantisek, line 28: I am only wondering if `.hostname` (without protocol) will not be more useful. What do you think?

```suggestion
        git_project.instance_url = urlparse(git_project.project_url).hostname
```

sakalosj: fixed

packit_service/models.py
sakalosj: fixed

TomasTomecek (approved):
LGTM

though I'm no longer familiar with p-s' codebase so I'd suggest to account for other reviews as well

packit_service/models.py
```diff
@@ -131,6 +132,11 @@ class GitProjectModel(Base):
     # Example: https://github.com/packit/hello-world.git
     https_url = Column(String)
     project_url = Column(String)
+    instance_url = Column(String, nullable=False)
+
+    @orm.reconstructor
+    def __init__(self):
+        self.instance_url = urlparse(self.project_url).netloc
```
lachmanfrantisek, line 139: Sorry, forgot to mention also this one:

```suggestion
        self.instance_url = urlparse(self.project_url).hostname
```

sakalosj: np, was bit tired later yesterday, fixed now

sakalosj:
> Nice! Do you want to work on the usage in another PR?
> 
> (e.g. simplifying/getting rid of `__choose_project`)

Yes, I can check it, but not sure if will have enough time for it.

PR #875, opened 2020-11-20
New Testing Farm API

jpopelka:


packit_service/worker/testing_farm.py
```diff
@@ -93,6 +75,49 @@ def _trigger_payload(self, pipeline_id: str, chroot: str) -> dict:
...
+                    ],
+                }
+            ],
+            "notification": {"webhook": f"{self.api_url}/testing-farm/results"},
+        }
+
+    def get_compose_arch(self, chroot) -> Tuple[str, str]:
+        # fedora-33-x86_64 -> Fedora-33, x86_64
+        compose, arch = chroot.rsplit("-", 1)
+        compose = compose.title()
```
lachmanfrantisek, line 110: We might want to use some logic from aliases...  There are some weird names for that this will not work:
https://github.com/packit/packit/blob/c571c810bf633477cd470bb2652a4461cab8c1c0/tests/unit/test_config_aliases.py#L79

lachmanfrantisek (commented):
Looking good so far!

packit_service/worker/testing_farm.py
jpopelka: Well, I can't tell how would a compose (which the new Testing Farm API uses) for this `opensuse-leap-15.0` be named, because currently [there are fedora composes only](https://api.dev.testing-farm.io/v0.1/composes), but I'd expect `Opensuse-Leap-15.0`, which is what my code does:

```python
compose, arch = "opensuse-leap-15.0-x86_64".rsplit("-", 1)
print(compose.title())
'Opensuse-Leap-15.0'
```

or maybe I just don't understand what you're suggesting?

lachmanfrantisek: My bad. I've missed the `r` in `rsplit`....

psss:
Hi, guys! Thanks much for working on this. Do you have an estimation when the testing farm integration might be working? The old cruncher is getting more and more unreliable. It's getting almost unusable with all the different failures. And it is essential for us to get the full `tmt` support for test execution. Thanks for any update.

jpopelka:
The last missing piece is processing notifications that testing farm sends to our webhook URL.
@thrix does the new testing farm already send them? (I haven't tried recently)

thrix:
@jpopelka I have the code, I am deploying it to the api instance today, will write here once it is ready for testing

thrix:
sorry guys, found a bunch of issues during deployment and I was not able to fix them until my PTO :( Seems this will slip until January as I am already on PTO from this week :( :-1:

jpopelka:
At the moment, the links in the checks would lead to the request URL (i.e. e.g. [here](https://api.dev.testing-farm.io/v0.1/requests/ab18144c-777b-4b54-aaea-a8f32f1c3d59)).
Converting that to something understandable to end-users is to be done in a follow-up PR as we want this PR to be deployed ASAP.

packit_service/worker/parser.py
```diff
@@ -722,52 +727,77 @@ def parse_distgit_event(event) -> Optional[DistGitEvent]:
...
+            return None
+
+        request_id: str = event["request_id"]
+        logger.info(f"Testing farm notification event. Request ID: {request_id}")
+
+        # Testing Farm sends only request/pipeline id in a notification.
+        # We need to get more details ourselves.
+        # It'd be much better to do this in TestingFarmResultsHandler.run(),
+        # but all the code along the way to get there expects we already know the details.
+        event = TestingFarmJobHelper.get_request_details(request_id)
```
lachmanfrantisek, line 762: What is the downside of calculating the missing attributes here? That we don't want to do it in the service?

I think we do some calculating also for other events.

But we can optimize later.

lachmanfrantisek (approved):
I can't find any significant problem. I would merge it and test on stage. We can't break anything because it does not work currently at all.

packit_service/constants.py
```diff
@@ -14,9 +14,7 @@
 
 CONFIG_FILE_NAME = "packit-service.yaml"
 
-TESTING_FARM_TRIGGER_URL = (
-    "https://scheduler-testing-farm.apps.ci.centos.org/v0/trigger"
-)
+TESTING_FARM_API_URL = "https://api.dev.testing-farm.io/v0.1/"
```
TomasTomecek, line 17: "dev" - doesn't sound like a prod deployment, I wonder if there will be two or more

packit_service/worker/parser.py
```diff
@@ -722,52 +727,77 @@ def parse_distgit_event(event) -> Optional[DistGitEvent]:
...
-        repo_namespace: str = nested_get(event, "artifact", "repo-namespace")
-        ref: str = nested_get(event, "artifact", "git-ref")
-        https_url: str = nested_get(event, "artifact", "git-url")
-        commit_sha: str = nested_get(event, "artifact", "commit-sha")
-        tests: List[TestResult] = [
+    def _parse_tf_result_xunit(xunit: Optional[str]) -> List[TestResult]:
+        """ Parse event["result"]["xunit"] to get tests results """
+        if not xunit:
+            return []
+        xunit_dict = xmltodict.parse(xunit)
```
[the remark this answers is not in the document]
TomasTomecek: I'd say that a more proper reaction would be

TomasTomecek (approved):
+1 on trying this on stg

packit_service/constants.py
jpopelka: Yes, @thrix said that it will run here for about half a year and then move to `api.testing-farm.io` when it's time. He'll let us know when prod API is ready.

packit_service/worker/parser.py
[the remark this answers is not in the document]
jpopelka: yes, xunit, which is xml, see for example ["result"]["xunit"] [here](https://api.dev.testing-farm.io/v0.1/requests/ab18144c-777b-4b54-aaea-a8f32f1c3d59)

packit_service/worker/testing_farm.py
```diff
@@ -117,18 +144,7 @@ def run_testing_farm_on_all(self):
...
-            state=CommitStatus.success,
-            description="Testing farm is temporarily disabled.",
-            chroot=chroot,
-            url="https://github.com/packit/packit-service/issues/803",
-        )
-
-        # do not send the request to the testing farm API
-        return TaskResults(success=True, details={})
-
+    def run_testing_farm(self, build_id: int, chroot: str) -> TaskResults:
```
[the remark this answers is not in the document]
jpopelka: I see. 80 lines is IMHO not that huge and I don't see any clear pattern how to split it so I'll leave it as it is for now and might think about it later - this might not be the last TF-related PR.

packit_service/worker/parser.py
jpopelka: Not sure I understand the question.

I don't see how we could calculate anything else from just a request_id, which is all we get in a notification from TF.

Example: When something changes in the TF request, TF sends a notification that contains only request id (for example `ab18144c-777b-4b54-aaea-a8f32f1c3d59`) and we have to get https://api.dev.testing-farm.io/v0.1/requests/ab18144c-777b-4b54-aaea-a8f32f1c3d59 to know all the other attributes.

There are 3 places when we can do this (get request details from TF):
- in service (httpd): we don't want to block service when something's wrong with TF and we want to have some retry mechanism
- in task.run_testing_farm_results_handler: that's what I wanted initially, but it didn't go well, hence the comment
- here in task.steve_jobs.process_message

packit_service/worker/parser.py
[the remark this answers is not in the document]
csomh: Hm, do we need to parse this? This seems to me as a detailed breakdown of all the testcases that were executed. Do we provide this information to our users or we just set the flags based on the overall status?

(maybe I'm missing something, so please gently point me in the right direction ;) )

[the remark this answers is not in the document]
jpopelka: Imagine your tests failed and you need to know why, but all you get when clicking on the check link is [this](https://api.dev.testing-farm.io/v0.1/requests/ab18144c-777b-4b54-aaea-a8f32f1c3d59).
How user friendly would you consider it?

On the other hand, Tomas said that @thrix said ;-) that he will add this xunit -> HTML conversion on their side so we probably actually don't need to do this.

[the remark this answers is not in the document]
TomasTomecek: we at least need to set the commit status to be passed/failed right? not sure if this info is present only in the xunit or also in some other metadata

would be awesome if TFT had some pretty view for the test results

packit_service/constants.py
TomasTomecek: thanks for info

packit_service/worker/parser.py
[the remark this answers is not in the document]
jpopelka: status is in ["result"]["overall"], we don't have to parse the xunit for that

packit_service/worker/parser.py
lachmanfrantisek: Ok, thanks for the info. I now understand that more.

Last question:
- Do we have the info about test-run already in the database? (Added when we trigger the test?)

packit_service/worker/parser.py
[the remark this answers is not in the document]
jpopelka: summary of IRC discussion:
jpopelka: we parse the xunit, but don't use the parsed data anywhere atm. It's because I thought I would need to create an HTML from of it at some point, but then Tomas said Miro will do that on TF side.
csomh: what about using http://artifacts.dev.testing-farm.io/ab18144c-777b-4b54-aaea-a8f32f1c3d59/work-fullQMe0r3
as it has all the logs but is pretty complicated
jpopelka: it's IMHO even more complicated than finding the links in the xunit. I had to go back to the xunit to see how to go through the directory structure to find the logs.

packit_service/worker/parser.py
jpopelka: Hopefully, I answer your question by saying that yes, we [create a TFTTestRunModel at the end of TestingFarmJobHelper.run_testing_farm()](https://github.com/packit/packit-service/pull/875/files#diff-bdd081387e0aa002e9e49b42836616251db4ab0822b9d6d47dfbdc5c0c0e6a46R218)

lachmanfrantisek: So, what about getting the missing information from the database?

If we know the pipeline-id, we have a relation to the project and trigger model.

jpopelka: Aha, I see now. I've been always weak in these databazy thingy.
Look, I don't feel like doing it right now. I'd rather merge this ASAP and do this change later - I've added a TODO comment there.

jpopelka:
recheck

PR #877, opened 2020-11-23
Build result page - provide more info

### LINKED ISSUE #865: Improve the build result page
We have a single view for logs and a "results page" for a build, let's split those into two:

The results page contains:
* Link to copr
* Link to copr logs
* Link to srpm logs
* status
* date/time
* how to install locally instructions
* link to the dashboard

Make sure all the information is correct for branch and release builds.

We got this feedback from Carl George.

lbarcziova:
- build status
- submitted time of SRPM
- installation instructions
- link to our website and dashboard

Snímka obrazovky z 2020-11-25 13-12-22

Fixes #865

### DIFF
--- a/packit_service/service/templates/build_info.html
+++ b/packit_service/service/templates/build_info.html
@@ -6,20 +6,40 @@
 <h1>
   Builds for {{ project.namespace }}/{{ project.repo_name }}: {{ title_identifier }}
 </h1>
+<p>Status: {{build.status}}</p>
 <p>
   <a
     href="{{ url_for('builds.get_srpm_build_logs_by_id', id_=build.srpm_build_id) }}"
   >
     SRPM build
   </a>
+  started {{srpm_submitted_time}}
 </p>
 {% if build.build_id %}
 <p>
   <a href="{{ build.web_url }}">{{ build_description }}</a>
   {% if build.build_logs_url %} &gt; Build for
-  <a href="{{ build.build_logs_url }}">{{ build.target }}</a> started at {{
+  <a href="{{ build.build_logs_url }}">{{ build.target }}</a> started {{
   build_submitted_time }}, status: {{ build.status }} {% endif %}
 </p>
 <!-- prettier-ignore -->
 {% endif %}
+{% if build.status == "success" and owner and project_name%}
+
+<p>You can install the built RPMs by following these steps:</p>
+<ul>
+  <li>
+    <code>sudo yum install -y dnf-plugins-core</code> on RHEL 8 or CentOS Stream
+  </li>
+  <li><code>sudo dnf install -y dnf-plugins-core</code> on Fedora</li>
+  <li><code>dnf copr enable {{owner}}/{{project_name}}</code></li>
+  <li>And now you can install the packages.</li>
+</ul>
+{% if is_pr %}
+<p>Please note that the RPMs should be used only in a testing environment.</p>
+{% endif %} {% endif %}
+<p>
+  For more info see our <a href="https://packit.dev">website</a> or
+  <a href="https://dashboard.packit.dev">dashboard</a>.
+</p>
 {% endblock %}
--- a/packit_service/utils.py
+++ b/packit_service/utils.py
@@ -20,6 +20,8 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 import logging
+from datetime import datetime
+from typing import Optional
 
 from packit.config import JobConfig, PackageConfig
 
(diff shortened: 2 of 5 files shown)


packit_service/constants.py
```diff
@@ -46,6 +46,15 @@
     "approved_manually": "approved_manually",
 }
 
+INSTALLATION_INSTRUCTIONS = (
+    "You can install the built RPMs by following these steps:\n\n"
+    "* `sudo yum install -y dnf-plugins-core` on RHEL 8\n"
+    "* `sudo dnf install -y dnf-plugins-core` on Fedora\n"
+    "* `dnf copr enable {owner}/{project_name}`\n"
+    "* And now you can install the packages.\n"
+    "\nPlease note that the RPMs should be used only in a testing environment."
```
lachmanfrantisek, line 55: Is this line valid also for the branch/release builds? Is it possible to add this not only for PRs?

lachmanfrantisek (commented):
I like that, thanks!

packit_service/constants.py
lbarcziova: You mean the line `dnf copr enable {owner}/{project_name}`? Since the `owner` and `project_name` are for the copr project, we have it in db for branches and releases as well. Or am I missing something?

lachmanfrantisek: No, I mean the line 55 -- `"\nPlease note that the RPMs should be used only in a testing environment."`.

That it does not need to be testing version.

lbarcziova: Oh, I see. I will change that

packit_service/service/templates/build_info.html
```diff
@@ -6,12 +6,14 @@
...
   Builds for {{ project.namespace }}/{{ project.repo_name }}: {{ title_identifier }}
 </h1>
+<p>Status: {{build.status}}</p>
 <p>
   <a
     href="{{ url_for('builds.get_srpm_build_logs_by_id', id_=build.srpm_build_id) }}"
   >
     SRPM build
   </a>
+  started at {{srpm_submitted_time}}
```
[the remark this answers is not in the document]
lbarcziova: I added a function for humanizing it

packit_service/service/templates/build_info.html
```diff
@@ -6,20 +6,40 @@
...
   build_submitted_time }}, status: {{ build.status }} {% endif %}
 </p>
 <!-- prettier-ignore -->
 {% endif %}
+{% if build.status == "success" and owner and project_name%}
+
+<p>You can install the built RPMs by following these steps:</p>
+<ul>
+  <li>
+    <code>sudo yum install -y dnf-plugins-core</code> on RHEL 8 or CentOS Stream
```
TomasTomecek, line 32: nice, I wasn't aware of the code HTML tag

tests_requre/service/test_views.py
```diff
@@ -1,17 +1,11 @@
 from flask import url_for
```
TomasTomecek, line 1: would be nice to add some cases here, not just drop them

TomasTomecek (commented):
looking great! I'd love to see some integration tests added, not just dropped

packit_service/utils.py
```diff
@@ -65,3 +67,41 @@ def load_job_config(job_config: JobConfig):
...
 def dump_job_config(job_config: JobConfig):
     return JobConfigSchema().dump_config(job_config) if job_config else None
+
+
+def pretty_time(time: datetime) -> Optional[str]:
+    """
+    Get a pretty string like 'an hour ago', 'yesterday', '3 months ago',
+    'just now'.
+    https://stackoverflow.com/questions/1551382/user-friendly-time-format-in-python
+    """
```
lachmanfrantisek, line 77: Nice improvement!

I am only wondering if we can't use some library for that. After (very quick) search, [arrow](https://arrow.readthedocs.io/en/latest/) should be packaged for Fedora.

packit_service/service/templates/build_info.html
```diff
@@ -6,20 +6,40 @@
...
+  </li>
+  <li><code>sudo dnf install -y dnf-plugins-core</code> on Fedora</li>
+  <li><code>dnf copr enable {{owner}}/{{project_name}}</code></li>
+  <li>And now you can install the packages.</li>
+</ul>
+{% if is_pr %}
+<p>Please note that the RPMs should be used only in a testing environment.</p>
+{% endif %} {% endif %}
```
lachmanfrantisek, line 40: Thanks!

packit_service/utils.py
TomasTomecek: I would argue if it's worth replacing a single function with a library

lachmanfrantisek: That's a question, but
- this looks too manual and too reinvent-the-wheel to me,
- it's already in Fedora what I am using as some quality certificate,
- we might find also other features useful (timezone support, locales, easier shifts, ...).

I am wondering if flask does not have something like this already there?

TomasTomecek: yes, those are good points, my intention is that when we add a new dependency, our images grow and we start relaying on the new library and all of its dependencies - in this case I'd say it's "cheaper" to go with the function for now and use a library once we realize we need more - which we probably will in the dashboard when we replace the build result page

luckily, this is p-s, we're not packaging it as an rpm, so we can do w/e we want

tests_requre/service/test_views.py
lbarcziova: done

packit_service/utils.py
lbarcziova: As I read the pros and cons, I would go for now with this simple function and not install the library for this functionality. If we will extend the functionality in the future, we can do it then.

lachmanfrantisek (approved):

TomasTomecek (commented):
Thanks for adding the tests and making the page look great!

PR #888, opened 2020-11-25
Don't skip build for tests when build with different trigger defined

### LINKED ISSUE #872: Build not run when test configured for trigger and build for different trigger
Example PR: https://github.com/psss/fmf/pull/88

As @lbarcziova has proposed, we should check that we don't skip build in that case.

lbarcziova:
Fixes #872


lachmanfrantisek (approved):
:tada:

PR #891, opened 2020-11-26
Add disclaimer regarding removal of :champagne: comment

### LINKED ISSUE #881: Include warning about our plans to disable the "Congratulations" comments in this comment
Update the content of comment about successfully created Copr build:
- [ ] add a warning about our plan to disable these comments in 2 weeks
- [ ] add a note that users can reach out to us if they would really like to receive these comments

mfocko:
Closes #881

### DIFF
--- a/packit_service/worker/handlers/fedmsg_handlers.py
+++ b/packit_service/worker/handlers/fedmsg_handlers.py
@@ -271,6 +271,11 @@ class CoprBuildEndHandler(AbstractCoprBuildReportHandler):
         ):
             msg = (
                 f"Congratulations! One of the builds has completed. :champagne:\n\n"
+                ":warning: Please note that our current plans include removal of these "
+                "comments in the near future (at least 2 weeks after including this "
+                "disclaimer), if you have serious concerns regarding their removal "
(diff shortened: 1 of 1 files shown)


jpopelka (approved):

lachmanfrantisek (commented):
Nice, what about linking this or the other issue?

mfocko:
> Nice, what about linking this or the other issue?

I thought about this, but since we "smack" the comment on every PR, I can see the issue being spammed just by use referencing it. If it's not a big problem, could be added. :)

lachmanfrantisek:
> I thought about this, but since we "smack" the comment on every PR, I can see the issue being spammed just by use referencing it. If it's not a big problem, could be added. :)

Good point. Let's go without that.

lachmanfrantisek (approved):

mfocko:
> > I thought about this, but since we "smack" the comment on every PR, I can see the issue being spammed just by use referencing it. If it's not a big problem, could be added. :)
> 
> Good point. Let's go without that.

We could probably pin the issue for removal of it?

PR #893, opened 2020-11-30
Fix fonts in build status page

mfocko:

### DIFF
--- a/packit_service/service/static/style.css
+++ b/packit_service/service/static/style.css
@@ -2,7 +2,11 @@
   font-family: sans-serif;
 }
 
-pre {
+pre,
+code,
+kbd,
+var,
+samp {
   font-family: monospace;
 }
 


packit_service/service/static/style.css
```diff
@@ -2,7 +2,9 @@
   font-family: sans-serif;
 }
 
-pre {
+pre,
+code,
+kbd {
```
csomh, line 7: Wow, HTML really turned fancy :smile:  

How about adding `samp` and `var`, too? Just in case :)

mfocko: I'd be thumbs up for Latin Modern for `var` :smirk: :joy: gonna add them

packit_service/service/static/style.css
```diff
@@ -2,7 +2,11 @@
   font-family: sans-serif;
 }
 
-pre {
+pre,
+code,
+kbd,
+var,
+samp {
```
TomasTomecek, line 9: what are these? :D

TomasTomecek (approved):

packit_service/service/static/style.css
TomasTomecek: answered myself: https://stackoverflow.com/questions/32284477/what-are-the-differences-between-kbd-samp-and-code-in-html

csomh (approved):

PR #904, opened 2020-12-11
Ignore push events that delete branch

mfocko:

### DIFF
--- a/tests/unit/test_steve.py
+++ b/tests/unit/test_steve.py
@@ -23,7 +23,7 @@
 """
 Let's test that Steve's as awesome as we think he is.
 """
-from json import dumps
+from json import dumps, load
 
 import pytest
 from celery.canvas import Signature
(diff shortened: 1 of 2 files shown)


tests/unit/test_steve.py
```diff
@@ -108,3 +108,20 @@ def test_process_message(event, private, enabled_private_namespaces, success):
     )
     assert "propose_downstream" in next(iter(results["job"]))
     assert first_dict_value(results["job"])["success"]
+
+
+@pytest.fixture()
+def github_push():
+    with open(DATA_DIR / "webhooks" / "github" / "push.json") as outfile:
```
lachmanfrantisek, line 115: We already have an example. Nice!

lachmanfrantisek (approved):
Looks good to me. Thank you!

tests/unit/test_steve.py
mfocko: that's also a problem :D it's copy-paste, would be nice to do a PR and reorganize all those fixtures

lachmanfrantisek: Would be really nice!

PR #906, opened 2020-12-16
Support Minishift in check-inside-openshift

jpopelka:
and don't rebuild images everytime

### DIFF
--- a/files/deployment.yaml
+++ b/files/deployment.yaml
@@ -49,12 +49,20 @@
       register: kubeconfig_token
       become: "{{ as_root }}"
 
+    - name: get minishift ip
+      command: minishift ip
+      register: minishift_ip
+      ignore_errors: True
+
+    - set_fact:
+        cluster_ip: "{{ minishift_ip.stdout if not minishift_ip.failed else '127.0.0.1' }}"
+
     - name: Create dev.yml
       copy:
         content: |
           # See https://github.com/packit/deployment/blob/master/vars/dev_template.yml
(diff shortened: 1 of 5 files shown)


files/deployment.yaml
```diff
@@ -49,12 +49,17 @@
...
+      register: oc_server
+      become: "{{ as_root }}"
+
     - name: Create dev.yml
       copy:
         content: |
           # See https://github.com/packit/deployment/blob/master/vars/dev_template.yml
           project: myproject
-          host: https://127.0.0.1:8443
+          host: {{ oc_server.stdout }}
```
jpopelka, line 62: Uh, thinking about it more, this might be a disaster if you happen to forget to spin up a local cluster before running `make check_in_container` - in which case you might still be logged to a prod/stg cluster.

csomh: Indeed it would :slightly_smiling_face: 

Would it be possible to try to use the `minishift` command first (`minishift openshift` should have some way to get this value) and if that does not work for some reason, fall back to localhost?

jpopelka: Yup, `minishift ip` is the way. I changed the code, the current version doesn't work, but I'll fix it once I have my proper internet connection back.

jpopelka: changed, PTAL

jpopelka: solved

csomh (approved):
Looks good. Thanks!

PR #907, opened 2020-12-18
Remove `GitProjectModel.__choose_project`

sakalosj:
Because `instace_url` field was introduced previously,
`GitProjectModel.__choose_project` is not required anymore.

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -138,24 +138,6 @@ class GitProjectModel(Base):
         self.instance_url = urlparse(self.project_url).hostname
 
     @classmethod
-    def __choose_project(
-        cls, session, forge, namespace, repo_name
-    ) -> "GitProjectModel":
-        """Returns a project (GitProjectModel) given session, forge, namespace and repo_name"""
-        projects = (
-            session.query(GitProjectModel)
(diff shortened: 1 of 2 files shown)


packit_service/models.py
```diff
@@ -213,33 +197,33 @@ def get_project_prs(
...
                 session.query(PullRequestModel)
-                .filter_by(project_id=project.id)
+                .join(GitProjectModel)
+                .filter(
+                    PullRequestModel.project_id == GitProjectModel.id,
+                    GitProjectModel.instance_url == forge,
+                    GitProjectModel.namespace == namespace,
+                    GitProjectModel.repo_name == repo_name,
+                )
                 .order_by(desc(PullRequestModel.pr_id))[first:last]
```
lachmanfrantisek, line 209: What is the difference between the filtering `PullRequestModel` by `project_id` and this join?

sakalosj: In the previous implementation, you required to query a project first.
Here you are getting a list of PR's direct without the need for GitProjectModel instance.

lachmanfrantisek: OK, just curious...

lachmanfrantisek (approved):

PR #910, opened 2021-01-04
don't install python3-kubernetes from Fedora

TomasTomecek:
we can't install kube from Fedora because API changes b/w major versions

### DIFF
--- a/files/install-deps-worker.yaml
+++ b/files/install-deps-worker.yaml
@@ -12,7 +12,10 @@
           - nss_wrapper # openshift anyuid passwd madness
           - redis # redis-cli for debugging
           - origin-clients # for sandcastle
-          - python3-kubernetes # for sandcastle
+          # we can't install kube from Fedora because kubernetes client API
+          # changes b/w major versions and we want to pin to the specific
+          # version to be sure it works well - it's 12.0.1 for now
+          # - python3-kubernetes  # for sandcastle
           - python3-fedora # to access FAS
           - python3-requests
           - python3-alembic


files/install-deps-worker.yaml
```diff
@@ -12,7 +12,7 @@
           - nss_wrapper # openshift anyuid passwd madness
           - redis # redis-cli for debugging
           - origin-clients # for sandcastle
-          - python3-kubernetes # for sandcastle
+          # we can't install kube from Fedora because API changes b/w major versions
```
jpopelka, line 15: Can you keep the `- python3-kubernetes # for sandcastle` there, just comment it out?
Otherwise, it looks like (to me) that this ("we can't install kube...") comment is meant for the below `python3-fedora`.

Also, we don't install kube from Fedora because we need a pinned version which might and might not be in fedora.
(The reason why we use pinned version in sandcastle is "because API changes b/w major versions")

TomasTomecek: good point, I just wanted to be sure mainly it builds fine with the initial commit, the comment should be more clear now

TomasTomecek:
okay, we have correct kube now:
```
2021-01-04 09:24:48.533007 | test-node | Collecting kubernetes<=12
2021-01-04 09:24:48.533032 | test-node |   Downloading https://files.pythonhosted.org/packages/3b/4e/6a16c118f72dd7b6334b2a6662d2a0693cf31f8b76dfe0fbcdaebb91d33b/kubernetes-12.0.0-py3-none-any.whl (1.7MB)
```

jpopelka (approved):

PR #911, opened 2021-01-04
[install-deps-worker] install sandcastle@source_branch instead of master

jpopelka:
This installs sandcastle@stable instead of sandcastle@master in `worker:prod`.

Hunor says he thinks it's been intentional but doesn't remember more details. @sakalosj do you remember?

### DIFF
--- a/files/install-deps-worker.yaml
+++ b/files/install-deps-worker.yaml
@@ -43,7 +43,7 @@
     - name: Install pip deps
       pip:
         name:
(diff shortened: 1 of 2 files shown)


TomasTomecek (approved):
LGTM, we really want this

jpopelka:
> Hunor says he thinks it's been intentional but doesn't remember more details.

The source_branch change is older than the sandcastle stable branch so I guess we just forgot to do this change when adding the sandcastle stable branch.

PR #913, opened 2021-01-04
Rename /packit propose-update to propose-downstream

jpopelka:
Related to packit/packit#756

- [x] have `/packit propose-update` alias to `/packit propose-downstream`

### DIFF
--- a/packit_service/worker/handlers/github_handlers.py
+++ b/packit_service/worker/handlers/github_handlers.py
@@ -28,7 +28,6 @@ import logging
 from typing import Optional
 
 from celery.app.task import Task
-
 from ogr.abstract import CommitStatus, GitProject
 from packit.api import PackitAPI
 from packit.config import (
@@ -39,6 +38,7 @@ from packit.config.aliases import get_branches
 from packit.config.package_config import PackageConfig
 from packit.exceptions import PackitException
 from packit.local_project import LocalProject
+
 from packit_service import sentry_integration
 from packit_service.constants import (
     FAQ_URL_HOW_TO_RETRIGGER,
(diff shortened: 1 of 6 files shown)


packit_service/worker/handlers/github_handlers.py
```diff
@@ -144,7 +144,7 @@ def run(self) -> TaskResults:
 
 
 @configured_as(job_type=JobType.propose_downstream)
-@run_for_comment(command="propose-update")
+@run_for_comment(command="propose-downstream")
```
lachmanfrantisek, line 147: We can have both if we really want some middle step.

jpopelka: oh, nice, that's what I've been looking for! thanks

packit_service/worker/handlers/github_handlers.py
```diff
@@ -204,12 +205,15 @@ def run(self) -> TaskResults:
...
+            msg_retrigger = MSG_RETRIGGER.format(
+                job="update", command="propose-downstream", place="issue"
+            )
             body_msg = (
                 f"Packit failed on creating pull-requests in dist-git:\n\n"
                 f"| dist-git branch | error |\n"
                 f"| --------------- | ----- |\n"
                 f"{branch_errors}\n\n"
-                f"{MSG_RETRIGGER.format(job='update', command='propose-update', place='issue')}\n"
+                f"{msg_retrigger}\n"
```
TomasTomecek, line 216: I'm just wondering if we want to place a comment here if a user writes propose-update instead of propose-downstream

TomasTomecek (approved):
lgtm, one UX question

packit_service/worker/handlers/github_handlers.py
jpopelka: I don't think so. The message already says to use `propose-downstream` so the user should realize the change anyway. Thanks for review.

PR #919, opened 2021-01-10
Fix ignoring of push that deletes branch

mfocko:
Instead of parsing delete by push, explicitly log it's delete by push and do not parse the event

### DIFF
--- a/packit_service/worker/parser.py
+++ b/packit_service/worker/parser.py
@@ -321,6 +321,11 @@ class Parser:
 
         if not (raw_ref and commits and before and pusher):
             return None
+        elif event.get("after").startswith("0000000"):
+            logger.info(
+                f"GitLab push event on '{raw_ref}' by {pusher} to delete branch"
+            )
+            return None
 
         number_of_commits = event.get("total_commits_count")
 
@@ -382,6 +387,11 @@ class Parser:
         )
 
         if not (raw_ref and head_commit and before and pusher):
+            return None
+        elif event.get("deleted"):
+            logger.info(
+                f"GitHub push event on '{raw_ref}' by {pusher} to delete branch"
+            )
             return None
 
         number_of_commits = event.get("size")
--- a/packit_service/worker/parser.py
+++ b/packit_service/worker/parser.py
@@ -102,83 +102,30 @@ class Parser:
             logger.warning("No event to process!")
             return None
 
-        response: Optional[
-            Union[
-                PullRequestGithubEvent,
-                InstallationEvent,
-                ReleaseEvent,
-                DistGitEvent,
-                TestingFarmResultsEvent,
-                PullRequestCommentGithubEvent,
-                IssueCommentEvent,
-                KojiBuildEvent,
-                AbstractCoprBuildEvent,
-                PushGitHubEvent,
-                MergeRequestGitlabEvent,
-                MergeRequestCommentGitlabEvent,
-                IssueCommentGitlabEvent,
-                PushGitlabEvent,
-            ]
-        ] = Parser.parse_pr_event(event)
-        if response:
-            return response
+        for response in map(
+            lambda parser: parser(event),
+            (
+                Parser.parse_pr_event,
+                Parser.parse_pull_request_comment_event,
+                Parser.parse_issue_comment_event,
+                Parser.parse_release_event,
+                Parser.parse_push_event,
+                Parser.parse_installation_event,
+                Parser.parse_distgit_event,
+                Parser.parse_testing_farm_results_event,
+                Parser.parse_copr_event,
+                Parser.parse_mr_event,
+                Parser.parse_koji_event,
+                Parser.parse_merge_request_comment_event,
+                Parser.parse_gitlab_issue_comment_event,
+                Parser.parse_gitlab_push_event,
+            ),
+        ):
+            if response:
+                return response
 
-        response = Parser.parse_pull_request_comment_event(event)
-        if response:
-            return response
-
-        response = Parser.parse_issue_comment_event(event)
-        if response:
-            return response
-
-        response = Parser.parse_release_event(event)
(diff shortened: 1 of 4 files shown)


packit_service/service/events.py
```diff
@@ -405,6 +405,9 @@ def get_package_config(self) -> Optional[PackageConfig]:
             f"\tpr_id: {self.pr_id}"
         )
 
+        if self.commit_sha is not None and self.commit_sha.startswith("0000000"):
+            return None
```
lachmanfrantisek, line 409: Some comment would be nice to express that `00000...` means branch deletion.

lachmanfrantisek (approved):
Nice and easy, thanks!

tests/unit/test_events.py
```diff
@@ -518,12 +518,16 @@ def test_parse_gitlab_issue_comment(self, gitlab_issue_comment):
         assert event_object.package_config
 
     def test_parse_github_push(self, github_push):
```
TomasTomecek, line 520: new test case like this would be lovely

packit_service/service/events.py
csomh: Could the `deleted` key of the event checked instead, and if it's `True`, simply not create anything in the Celery Queue?

This would be a more straightforward approach, than aborting while getting the package config.

lachmanfrantisek: This PR solves a slightly different issue:

- The skipping of push that deletes branch is already implemented. (#904 )
- In my refactor, I started adding package-config to the result object which results in errors for this situation. (This pr solves that problem.)

But we can sure do the filtering of events in a different way.

csomh: Hm, I see. From what I can see #904 didn't solve the original issue (#739) as there were Sentry events for this problem during the weekend (I've linked them to #739).

My point is: I see this filtering happening at this level of the architecture a little bit too late. It's not a "natural" place to do it. I would not expect such filtering to happen neither in `process_jobs()` nor in `get_package_config()`.

Relying on a special value ("00000") to tell that this is branch deletion is also not the best possible solution. There is an explicit field for this in the event payload. We should rather rely on that.

Creating a task for the service to process, when it's obvious at the time of receiving the event payload that this is not going to be needed is problematic, and leads to issues similar to the one this PR tries to solve.

lachmanfrantisek: > Hm, I see. From what I can see #904 didn't solve the original issue (#739) as there were Sentry events for this problem during the weekend (I've linked them to #739).

It did solve it, but I broke it again by adding package-config to result..;)  (It causes loading config from unknown commit again.)

> My point is: ....

I don't argue. I only wanted to point out that we are speaking about something slightly different.

And, yes, we can implement your proposal. Here or as a follow-up.

mfocko:
@lachmanfrantisek @csomh I've changed the way it's handled to **not parse** the push event and log it
@TomasTomecek the test case for parsing push has been changed to use different json; test case for checking if it's ignored was already included

packit_service/worker/parser.py
```diff
@@ -102,83 +102,30 @@ def parse_event(
...
+                Parser.parse_merge_request_comment_event,
+                Parser.parse_gitlab_issue_comment_event,
+                Parser.parse_gitlab_push_event,
+            ),
+        ):
+            if response:
+                return response
+
+        logger.debug("We don't process this event.")
+        return None
```
lachmanfrantisek, line 128: :+1:

lachmanfrantisek (approved):
Thanks!

packit_service/worker/parser.py
```diff
@@ -383,6 +335,11 @@ def parse_push_event(event) -> Optional[PushGitHubEvent]:
 
         if not (raw_ref and head_commit and before and pusher):
             return None
+        elif event.get("deleted"):
```
csomh, line 338: Why not do this even before starting any processing of the event in [`api/webooks`](https://github.com/packit/packit-service/blob/33af9662ba6bb6df94a3fe2a4805b9c746b3e0fa/packit_service/service/api/webhooks.py#L82)? I'm aware that this might be outside the scope of this PR.

lachmanfrantisek: > I'm aware that this might be outside the scope of this PR.
 
Agree. This is relevant to the filtering of other events as well.

mfocko: I have looked into it, we basically just do validation there, so I didn't want to include this kind of a logic that is specific to push event. It should be discussed further imo.

csomh: IIRC we already had several discussions on this topic on arch meetings.

The conclusion there was that it's okay to do some filtering on the API level, if that filtering can be done solely based on the information available in the event payload or HTTP headers. This would not require any considerable extra time spent in the API as inspecting a few values in dictionaries is cheap and there wouldn't be any external resources involved (no additional API calls, lengthy disk reads or such).

Yes, an approach like this would cause having two places where events can be filtered out, but it would also save us quite a few extra CPU cycles we currently use just to encode and then decode the whole payload. Furthermore, workers would need to deal only with events where there is actually some extra work to be done (events which are actionable), like fetching a `.packit.yaml` at the minimum.

And I'm pretty sure such an approach would also help with preventing mistakes like this PR tries to fix :slightly_smiling_face:

lachmanfrantisek: Yes, I recall we want to solve this in general. Not only for this specifically. (But when we discussed this last time, we mentioned that this will require changes in a lot of test cases.)

I would merge this and create an issue/card for the move of filtering to the API, what do you think?

csomh: Sure :+1:

TomasTomecek (approved):
thanks for the explanation, nicely done!

mfocko:
recheck

PR #924, opened 2021-01-13
De-master-ize code

jpopelka:
Similar to https://github.com/packit/packit/pull/1074, i.e.
allow upstream/downstream default branch other than master.


lachmanfrantisek (approved):

TomasTomecek (commented):
looking good, can we get some test case added or updated to make sure this works?

jpopelka:
> can we get some test case added or updated to make sure this works?

sure, tests were kinda work-in-progress. Should be updated & green now.
I have ACK from Franta, so adding `mergeit`.

Issue #803, opened 2020-09-09
The tests are temporarily disabled


lachmanfrantisek:
### :warning: Testing-farm support is disabled. We are moving to a new test runner. :warning: 

The cluster with the old testing-farm runner has died, it will never work again and we are finalizing the support for a new one.
* [x] We have prepared (#875) implementation of the new API of the testing-farm.
* [x] Miro V. has been working hard on the new API. You can expect full tmt support so be patient, please.
* [x] Jirka is finalizing our part and making sure the connection works smoothly (#875).

We are sorry about that but give us some time, please.

More information about that can be found in #802 or in packit/packit#1078.

You will now get `Testing farm is temporarily disabled.` result until the new testing-farm is ready. We set the result status to `success` to not mark your PRs as failed. The neutral state is not possible with our current procedure how we set the status (#760).

![Screenshot from 2021-01-25 09-58-13](https://user-images.githubusercontent.com/20214043/105683432-ebbc9280-5ef3-11eb-954c-10bcbb96c32c.png)

psss:
> Miro V. is working hard on the new API but it's not ready yet. 

I guess that part has been done, right? @thrix?

jpopelka:
Yes, I'm working on implementing it on our side.

lachmanfrantisek:
@psss I've updated the description to be fair..;)

psss:
Thanks for the update :)

psss:
This seems to be already in production, right?

jpopelka:
yes, it's been in production since this week

Issue #760, opened 2020-07-29
Make check runs conclusion "neutral" for permission problems


StykMartin:
## Current state
Check runs for all chroots are marked as "failed" when the user is not part of organization.

## Proposed state
Check runs for all chroots are marked as "neutral" therefore it will not mark the whole PR as "failed".

I expect that some users would like to see "failed" so they don't forget to run CI. So I suggest making this part of `packit` configuration per project.

For example, LGTM is using neutral conclusion.
![image](https://user-images.githubusercontent.com/188918/88800075-d9ada180-d1a7-11ea-8f66-d29f5ae479fd.png)              ![image](https://user-images.githubusercontent.com/188918/88800091-e0d4af80-d1a7-11ea-8293-c6cc5771fbf9.png)

Added by @lachmanfrantisek:

The use-case above is not relevant anymore since we've changed the permission workflow but let's use `neutral` status for all the permission problems.

- [x] implement the support for check runs in OGR (packit/ogr#592)
- [x] ask for the `Check-runs` permission for the Github apps.
- [ ] find all the places where we check the permissions and use check runs with a neutral state there.

TomasTomecek:
+1, good suggestion and should be trivial to implement on our side

anyone wants to picks this up on Friday?

lachmanfrantisek:
The API endpoint we are using for setting commit status [does not provide](https://docs.github.com/en/rest/reference/repos#statuses) `neutral` state, only `error`, `failure`, `pending`, or `success`.


We will need to switch to [checks API](https://docs.github.com/en/rest/reference/checks) to be able to use it.

StykMartin:
👀

TomasTomecek:
Given the fact we are not using the checks API and commit statuses do not support the neutral status, I believe we can close this.

lachmanfrantisek:
But if we use check API in OGR, we can. (And it should be already implemented in the pygithub.) Not a high-prio, but I would still like to see it.

mfocko:
Does it still matter? We require just the namespace to be allowed now.

lachmanfrantisek:
But we still sometimes want to have some neutral state (e.g. warning during testing-farm problems).

lachmanfrantisek:
I'll change the title and add some TODO to the description to cover all the permission errors because this specific permission problem can't occur nowadays.

lachmanfrantisek:
Ping the sync to Jira!

lachmanfrantisek:
[...]

PR #929, opened 2021-01-21
Don't submit the TFT tests, inform users about it

lbarcziova:
Related to packit/packit#1078

### DIFF
--- a/packit_service/worker/testing_farm.py
+++ b/packit_service/worker/testing_farm.py
@@ -118,6 +118,17 @@ class TestingFarmJobHelper(CoprBuildJobHelper):
         )
 
     def run_testing_farm(self, chroot: str) -> TaskResults:
+        # inform users about temporarily disabled testing farm
+        self.report_status_to_test_for_chroot(
+            state=CommitStatus.success,
+            description="Testing farm is temporarily disabled.",
+            chroot=chroot,
+            url="https://github.com/packit/packit-service/issues/803",
+        )
+
+        # do not send the request to the testing farm API
+        return TaskResults(success=True, details={})
+
         if chroot not in self.tests_targets:
             # Leaving here just to be sure that we will discover this situation if it occurs.
             # Currently not possible to trigger this situation.
--- a/tests/integration/test_listen_to_fedmsg.py
+++ b/tests/integration/test_listen_to_fedmsg.py
@@ -346,6 +346,7 @@ def test_copr_build_end_release(copr_build_end, pc_build_release, copr_build_rel
     )
 
 
+@pytest.mark.skip(reason="Testing farm currently disabled")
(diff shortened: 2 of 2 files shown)


jpopelka (approved):
that was quick, thanks

packit_service/worker/testing_farm.py
```diff
@@ -118,6 +118,17 @@ def run_testing_farm_on_all(self):
         )
 
     def run_testing_farm(self, chroot: str) -> TaskResults:
+        # inform users about temporarily disabled testing farm
+        self.report_status_to_test_for_chroot(
+            state=CommitStatus.success,
```
lachmanfrantisek, line 123: I hope that `success` will not be misleading to users. But I agree that only success will not break the whole PR status.

packit_service/worker/testing_farm.py
```diff
@@ -118,6 +118,17 @@ def run_testing_farm_on_all(self):
...
+            description="Testing farm is temporarily disabled.",
+            chroot=chroot,
+            url="https://github.com/packit/packit/issues/1078",
```
lachmanfrantisek, line 126: What about using this one (https://github.com/packit/packit-service/issues/803) and update the top comment?

lachmanfrantisek (approved):
Thanks!

I am glad we can have it before this week's deployment. Please, check it on stage and let me know when I can move the stable branches...

packit_service/worker/testing_farm.py
lachmanfrantisek: My point is to have a very clear message about what is a problem and what needs to be done to make it work.
(And we can edit the message in the meantime.)

I am not sure if the conversation in the issue 1078 clearly describes the problem.

packit_service/worker/testing_farm.py
TomasTomecek: I thought there is also neutral but it's not:
https://github.com/packit/packit-service/issues/760
https://docs.github.com/en/rest/reference/repos#create-a-commit-status

yes, either set error, success or don't set it at all

lachmanfrantisek: Yes, we can't use neutral for now. (Would be nice to mention in the linked issue.)

> don't set it at all
This might require more changes. (And we don't have place to let user know.)

TomasTomecek: > > don't set it at all
> This might require more changes. (And we don't have place to let user know.)

agreed, I'd say all three are "bad" and not setting at all is probably the worst solution

lbarcziova: yes, I also wanted to set neutral but that's not possible, so I will leave there `success`

packit_service/worker/testing_farm.py
lbarcziova: yes, probably #803 is more clear, changing the link to it

PR #931, opened 2021-01-27
Autoretry the Celery handler tasks on any error

### LINKED ISSUE #873: Retry the celery task on error
Celery docs: https://docs.celeryproject.org/en/stable/userguide/tasks.html#automatic-retry-for-known-exceptions

@dhodovsk

lbarcziova:
Fixes #873 
Retry the handler task on any error, maximally 2 times, with exponential backoff (3s, 6s) (these parameters can be configured via env vars), and by default, the jitter is set to True.

### DIFF
--- a/packit_service/constants.py
+++ b/packit_service/constants.py
@@ -36,9 +36,9 @@ COPR_API_FAIL_STATE = 2
 PG_COPR_BUILD_STATUS_FAILURE = "failure"
 PG_COPR_BUILD_STATUS_SUCCESS = "success"
 
-RETRY_LIMIT = 2
+DEFAULT_RETRY_LIMIT = 2
 # retry in 3s, 6s
-RETRY_BACKOFF = 3
+DEFAULT_RETRY_BACKOFF = 3
 
 WHITELIST_CONSTANTS = {
     "approved_automatically": "approved_automatically",
--- a/packit_service/constants.py
+++ b/packit_service/constants.py
@@ -39,8 +39,6 @@ PG_COPR_BUILD_STATUS_SUCCESS = "success"
 RETRY_LIMIT = 2
 # retry in 3s, 6s
 RETRY_BACKOFF = 3
-# retry on any error
-EXCEPTIONS = (Exception,)
 
 WHITELIST_CONSTANTS = {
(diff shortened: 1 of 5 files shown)


jpopelka:
>  it is possible from version 4.4.

maybe it's time to update [our base image](https://github.com/packit/deployment/blob/master/containers/Dockerfile.base#L4) to fedora-33 which [has 5.0.5](https://src.fedoraproject.org/rpms/python-celery)

jpopelka (approved):

packit_service/constants.py
```diff
@@ -38,7 +38,11 @@
 PG_COPR_BUILD_STATUS_FAILURE = "failure"
 PG_COPR_BUILD_STATUS_SUCCESS = "success"
 
-RETRY_LIMIT = 5
+RETRY_LIMIT = 2
+# retry in 3s, 6s
+RETRY_BACKOFF = 3
```
lachmanfrantisek, line 43: Would it be hard to have it configurable?
- So we can have a different setup for prod/stage.
- Edits possible on the deployment level. => Quick change without code/image change and we can increase the numbers by steps.

lachmanfrantisek (commented):
Thanks!

packit_service/constants.py
TomasTomecek: having it in the packit service config (or env vars) would be great

lbarcziova: no, that's a good point, will make it configurable

lbarcziova:
```
TASK [Getting packit-worker pod]
....
"containerStatuses": [
{
"state": {
    "terminated": {
```

need to debug this

lbarcziova:
PTAL, I made the retry parameters configurable via env vars, since when defining the base Celery task we don't have the config available. The tests should pass after the base image will be rebuilt (packit/deployment#173), since the class-based task should be available from Celery 5.

jpopelka (approved):

lachmanfrantisek (approved):

TomasTomecek (commented):
wow, this is pretty impressive, I love how the implementation is simple and clear

PR #935, opened 2021-01-27
Move setting of SRPM URL to avoid temporary one

### LINKED ISSUE #934: Link to SRPM resolves to 404
COPR sets `source_project['url']` to a temporary address (probably, where it's uploaded by us) instead of the one where it can be accessed afterwards.

mfocko:
Since COPR sets just a temporary URL for SRPM when triggering the build,
move setting of the URL to `check_copr_build` where it will be set after
the build is finished.

Fixes #934

### DIFF
--- a/packit_service/worker/build/copr_build.py
+++ b/packit_service/worker/build/copr_build.py
@@ -371,7 +371,6 @@ class CoprBuildJobHelper(BaseBuildJobHelper):
             build = self.api.copr_helper.copr_client.build_proxy.create_from_file(
                 ownername=owner, projectname=self.job_project, path=self.srpm_path
             )
-            self.srpm_model.set_url(build.source_package.get("url"))
         except CoprRequestException as ex:
             if "You don't have permissions to build in this copr." in str(
                 ex
--- a/packit_service/worker/build/babysit.py
+++ b/packit_service/worker/build/babysit.py
@@ -85,6 +85,8 @@ def check_copr_build(build_id: int) -> bool:
             ),  # this seems to be the SRPM name
             timestamp=chroot_build.ended_on,
         )
+        if build.srpm_build.url is None:
+            build.srpm_build.set_url(build_copr.source_package.get("url"))
 
         job_configs = get_config_for_handler_kls(
             handler_kls=CoprBuildEndHandler,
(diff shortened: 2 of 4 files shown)


packit_service/worker/build/copr_build.py
```diff
@@ -371,7 +371,6 @@ def run_build(
             build = self.api.copr_helper.copr_client.build_proxy.create_from_file(
                 ownername=owner, projectname=self.job_project, path=self.srpm_path
             )
-            self.srpm_model.set_url(build.source_package.get("url"))
```
lachmanfrantisek, line 374: We can update it multiple times. => Do we need to remove this?

lachmanfrantisek (commented):
What about processing that in the build-start/build-end handler (as well)?

packit_service/worker/build/copr_build.py
mfocko: it was 404 even during the build

lachmanfrantisek: OK, I thought it can't harm anything but since we don't show anything for `None`, it's probably better to not set the URL here.

mfocko:
> What about processing that in the build-start/build-end handler (as well)?

I'm not really sure about that, both `AbstractCoprBuildEvent` and `CoprBuildEndHandler` are using precomputed values (not having COPR build object at all); maybe if we kept COPR build object in `CoprBuildJobHelper`? and for objects that were not created by running a build have it as a property?

lachmanfrantisek:
> I'm not really sure about that, both `AbstractCoprBuildEvent` and `CoprBuildEndHandler` are using precomputed values (not having COPR build object at all); maybe if we kept COPR build object in `CoprBuildJobHelper`? and for objects that were not created by running a build have it as a property (get the build object from COPR by ID)?


Sure, do it where you think it's the best place. I haven't thought about that so much.

mfocko:
@lachmanfrantisek added check to set the URL only if unset
---
regarding moving it to `CoprBuildEndHandler`, I need to think about it; proposing to fix it as a bug for now and refactor it later

lachmanfrantisek (approved):
Thanks for the quick fix!

PR #936, opened 2021-01-28
Refactor permission system + rename whitelist to allowlist

### LINKED ISSUE #920: Don't require `/packit build` by user with write permissions
Since Packit creates separate projects for each PR IIUC, it is not necessary
to - at least from Copr POV - require maintainers to trigger the build.  We
can simply start them.

### LINKED ISSUE #937: Rename whitelist to allowlist
Rename whitelist to allowlist (or something else we agree on)

- [ ] rename class(es)
- [ ] update logging
- [ ] update docs
- [ ] database (rename table)

mfocko:
# TODO

- [x] what about Koji?
   _resolution_: keeping as is
- [ ] ~~GitLab namespace (they are multi-part, we probably want to support allowlisting `redhat`, but accepting all nested namespaces, e.g. `redhat/rpms`)~~
  - [ ] safest would be iteratively stripping subnamespace from right and checking against DB
- [ ] ~~we need to differentiate between users with permissions on GitLab and GitHub~~
- [x] Switch on allowlisting for GitLab (turned off right now; all events are processed)
   _resolution_: turned on for MRs, push and issue comments
- [x] add migration script for allowlist model
   _resolution_: done, not pushed, may be merged together with enum for gitlab/github

**Differentiating GitLab/GitHub allowed users (with related not done TODOs) will be done in follow-up PR**

# Changes

- permissions on pull requests:
  - before:
    `write access > ( Packit admins OR allowlisted namespace of repository OR user that triggered event )`
  - after:
    `Packit admins > ( allowlisted namespace AND ( author of pull request OR write access ) )`
- issue comments:
  - before: `allowlisted user OR namespace`
  - after: `allowlisted namespace AND write permissions of user`
- checking of administrator is moved to allowlist
- checking of write permissions of user on repository is moved to allowlist
- renamed whitelist to allowlist

---
Fixes #920
Fixes #937


lachmanfrantisek (approved):

TomasTomecek:
yes, please let's leave koji as it is now and do this only for the CI PR builds

TomasTomecek (approved):
nice, LGTM!

interesting that no tests need an update :thinking:

mfocko:
> nice, LGTM!
> 
> interesting that no tests need an update 🤔

removed code is unreachable, it gets denied even sooner 🙃😪

mfocko:
@TomasTomecek @lachmanfrantisek please check, there is also description of allow/deny changes in last commit

lachmanfrantisek:
> GitLab namespace

BTW, do we differentiate between git-forges? What about GitLab namespace, that matches the name of the GitHub one that is whitelisted...

packit_service/constants.py
```diff
@@ -1,9 +1,10 @@
 from enum import Enum
 
-DOCS_URL = "https://packit.dev/packit-as-a-service/"
-FAQ_URL = f"{DOCS_URL}#faq"
+DOCS_URL = "https://packit.dev/docs"
+FAQ_URL = f"{DOCS_URL}/faq"
 FAQ_URL_HOW_TO_RETRIGGER = (
-    f"{DOCS_URL}#how-to-re-trigger-packit-service-actions-in-your-pull-request"
+    f"{DOCS_URL}/packit-as-a-service/"
+    "#how-to-re-trigger-packit-service-actions-in-your-pull-request"
```
lachmanfrantisek, line 7: Thanks for fixing that!

packit_service/worker/jobs.py
```diff
@@ -284,11 +261,10 @@ def process_jobs(self, event: Event) -> List[TaskResults]:
...
             # check whitelist approval for every job to be able to track down which jobs
             # failed because of missing whitelist approval
-            whitelist = Whitelist()
-            user_login = getattr(event, "user_login", None)
-            if user_login and user_login in self.service_config.admins:
-                logger.info(f"{user_login} is admin, you shall pass.")
-            elif not whitelist.check_and_report(
+            whitelist = (
+                Whitelist()
+            )  # TODO: Why do we create separate instance for each iteration?
```
lachmanfrantisek, line 266: It's maybe because of some refactorings. I don't think we need to create multiple instances.

I am thinking if this can't be moved away (=before) the for loop? WDYT?

mfocko:
> > GitLab namespace
> 
> BTW, do we differentiate between git-forges? What about GitLab namespace, that matches the name of the GitHub one that is whitelisted...

there's no difference between user and namespace either

packit_service/worker/whitelist.py
```diff
@@ -191,95 +301,45 @@ def check_and_report(
...
+        # Administrators
+        user_login = getattr(event, "user_login", None)
+        if user_login and user_login in service_config.admins:
+            logger.info(f"{user_login} is admin, you shall pass.")
+            return True
 
-        # whitelist checks dont apply to CentOS (Pagure, Gitlab)
-        if isinstance(
-            event,
+        for callback, related_events in (
```
lachmanfrantisek, line 310: What about some mapping, this is becoming too magical to me.

lachmanfrantisek (commented):
Thanks for making it more straightforward!

packit_service/worker/jobs.py
mfocko: I don't see anything shared, but there was/is a comment for that, so I'm just asking :D

packit_service/worker/whitelist.py
mfocko: it is a mapping, without map :D i'll add it

mfocko:
also I could probably switch the naming of whitelist since I'm probably going to touch the database because of github vs gitlab

packit_service/worker/whitelist.py
lachmanfrantisek: My point was more about readability. Just that it's really clear what is going on here.

packit_service/worker/allowlist.py
```diff
@@ -0,0 +1,346 @@
...
+        :return:
+        """
+        account = AllowlistModel.get_account(account_name)
+        if account:
+            db_status = account.status
+            s = AllowlistStatus(db_status)
+            return (
+                s == AllowlistStatus.approved_automatically
+                or s == AllowlistStatus.approved_manually
+            )
```
lachmanfrantisek, line 139:
```suggestion
            return AllowlistStatus(db_status) in (
                AllowlistStatus.approved_automatically,
                AllowlistStatus.approved_manually
            )
```

packit_service/worker/allowlist.py
```diff
@@ -0,0 +1,346 @@
...
+        :return: list of accounts waiting for approval
+        """
+        return [
+            account.account_name
+            for account in AllowlistModel.get_accounts_by_status(
+                AllowlistStatus.waiting.value
+            )
+        ]
+
+    def unchecked_event(
```
lachmanfrantisek, line 175: What about some prefix for those methods? Something like `validate_unchecked_event`, `validate_release_push_event`,...

packit_service/worker/allowlist.py
```diff
@@ -0,0 +1,346 @@
...
+
+        # Administrators
+        user_login = getattr(event, "user_login", None)
+        if user_login and user_login in service_config.admins:
+            logger.info(f"{user_login} is admin, you shall pass.")
+            return True
+
+        for related_events, callback in CALLBACKS.items():
+            if isinstance(event, related_events):
+                return callback(event, project, service_config, job_configs)
```
lachmanfrantisek, line 342: Personally, I like that approach, but tbh, I am still not sure about the readability of this.

But let's use this and improve later if someone complains...;)

packit_service/worker/jobs.py
lachmanfrantisek: I am now not sure.

We want to have a result for each handler, but if there are multiple handlers, we will run validation multiple times and potentially add multiple comments. Am I right?

lachmanfrantisek (commented):
Sorry for another group of comments...

mfocko:
> Sorry for another group of comments...

definitely won't be last one till finish :D

packit_service/worker/allowlist.py
mfocko: thx for bringing it up, i noticed it, must have carried on with renaming and then forgot :)

packit_service/worker/jobs.py
mfocko: I'll move it out of the for-loop
it doesn't matter really, cause first failure collects `processing_results` for each `job_config` and returns them

packit_service/service/api/allowlist.py
```diff
@@ -0,0 +1,32 @@
...
+
+from packit_service.models import AllowlistModel
+
+logger = getLogger("packit_service")
+
+ns = Namespace("allowlist", description="Allowlisted FAS accounts")
+
+
+@ns.route("")
+class Allowlist(Resource):
```
lachmanfrantisek, line 17: Just to be sure. Do we use it somewhere?

mfocko: I'm not sure tbh, internally we don't need to use API I guess, permitting/denying happens also happens in the cluster. Informative value maybe?

lachmanfrantisek: We use API from the dashboard but I am not sure if it is used there. Can you please check that?

lachmanfrantisek: Approving, please merge once this is checked.

lachmanfrantisek (approved):
Thanks! That was a huge amount of work. :rocket:

packit_service/service/api/allowlist.py
mfocko: haven't found any reference from dashboard

PR #942, opened 2021-01-29
TF hack: Temporarily load fmf from base/main

jpopelka:
Prior to this commit, we were loading the FMF from `self.metadata.commit_sha` of the base project,
which did not work, because that commit is not yet merged into the base project.

More in the code. If you know how to make this a proper way, please let me know, otherwise please ack so that submitting TF works at least like this and I'll keep looking for a proper solution.

### DIFF
--- a/packit_service/worker/testing_farm.py
+++ b/packit_service/worker/testing_farm.py
@@ -66,12 +66,21 @@ class TestingFarmJobHelper(CoprBuildJobHelper):
 
         """
         compose, arch = self.get_compose_arch(chroot)
+        # FIXME FIXME FIXME !!!
+        # This tells TF to load fmf metadata from main/master of the base project.
+        # What we need is to load it from self.metadata.commit_sha of the FORKED project.
+        # This is ugly hack, but I need to make this work at least like this ASAP
+        # and find proper way later. Sorry.
+        url = self.metadata.project_url
+        ref = self.project.default_branch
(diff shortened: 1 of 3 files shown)


packit_service/worker/testing_farm.py
```diff
@@ -66,12 +66,21 @@ def _payload(self, build_id: int, chroot: str) -> dict:
...
         compose, arch = self.get_compose_arch(chroot)
+        # FIXME FIXME FIXME !!!
+        # This tells TF to load fmf metadata from main/master of the base project.
+        # What we need is to load it from self.metadata.commit_sha of the FORKED project.
+        # This is ugly hack, but I need to make this work at least like this ASAP
+        # and find proper way later. Sorry.
+        url = self.metadata.project_url
+        ref = self.project.default_branch
+        # url = ? url of the fork from which the PR has been created ?
+        # ref = self.metadata.commit_sha
```
lachmanfrantisek, line 77: Yes, the fork seems like a way to go. I think we already have that in `self.metadata`.

lachmanfrantisek (approved):

packit_service/worker/testing_farm.py
jpopelka: I was of course already looking into `self.metadata` but  give it another try if you say so.
Merging this for now. Thanks

PR #947, opened 2021-02-01
TF: store original/base project url in TFTTestRunModel

jpopelka:
#946 has not worked. Trying to get a parent of forked project results in https://sentry.io/organizations/red-hat-0p/issues/2191846769

```
OgrException
Couldn't retrieve token from Tokman: (400) {"error": "Failed to retrieve a token: App is not installed on <forked repo>"}
```

So we need to store the original/base project URL in a db when submitting tests to TF and get it from there when processing TF notification.


lachmanfrantisek (approved):
Let's try this..

PR #957, opened 2021-02-05
TF: run TF's install test when no fmf configured

jpopelka:


packit_service/worker/handlers/testing_farm_handlers.py
```diff
@@ -102,6 +102,14 @@ def run(self) -> TaskResults:
             status = CommitStatus.failure
             summary = self.summary or "Tests failed ..."
 
+        if len(self.tests) >= 0 and self.tests[0].name == "/packit/install-and-verify":
```
mfocko, line 105: are you sure about `>= 0` and then `self.tests[0]`? didn't you mean to do
```suggestion
        if len(self.tests) > 0 and self.tests[0].name == "/packit/install-and-verify":
```

mfocko (approved):
LGTM, apart from that one remark

packit_service/worker/handlers/testing_farm_handlers.py
jpopelka: yes, thanks!

PR #960, opened 2021-02-08
[parse TF results] return original commit_sha as ref in case of install test

jpopelka:
Fixes https://sentry.io/organizations/red-hat-0p/issues/1809076278

### DIFF
--- a/packit_service/constants.py
+++ b/packit_service/constants.py
@@ -16,6 +16,7 @@ SANDCASTLE_PVC = "SANDCASTLE_PVC"
 CONFIG_FILE_NAME = "packit-service.yaml"
 
 TESTING_FARM_API_URL = "https://api.dev.testing-farm.io/v0.1/"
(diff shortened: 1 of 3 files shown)


lachmanfrantisek (approved):
Looks good. Thanks!

packit_service/constants.py
```diff
@@ -16,6 +16,7 @@
 CONFIG_FILE_NAME = "packit-service.yaml"
 
 TESTING_FARM_API_URL = "https://api.dev.testing-farm.io/v0.1/"
+TESTING_FARM_INSTALL_TEST_URL = "https://gitlab.com/testing-farm/tests"
```
csomh, line 19: 'install' is a verb, so can wrongly indicate this is some kind of action flag. Maybe `TESTING_FARM_INSTALLABILITY_TEST` would be better?

csomh (approved):

PR #971, opened 2021-02-12
Support testing in toolbox

mfocko:
- Add Makefile target for
  - Setting up from inside toolbox
  - Testing in toolbox (checks if exists + sets it up if necessary)
- Script that checks if toolbox exists, otherwise sets it up
- Playbook with necesary targets

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,5 +90,16 @@ check-inside-openshift: service worker test_image
 check-inside-openshift-zuul: test_image
 	ANSIBLE_STDOUT_CALLBACK=debug $(AP) files/check-inside-openshift.yaml
 
+setup-inside-toolbox:
+	@if [[ ! -e /run/.toolboxenv ]]; then \
+		echo "Not running in a toolbox!"; \
+		exit 1; \
+	fi
+	dnf install -y ansible
+	SOURCE_BRANCH=$(SOURCE_BRANCH) ANSIBLE_STDOUT_CALLBACK=debug $(AP) files/setup-toolbox.yaml
+
+check-inside-toolbox:
+	bash files/test_in_toolbox.sh packit_service
+
 requre-purge-files:
 	pre-commit run requre-purge --all-files --verbose --hook-stage manual
--- a/files/test_in_toolbox.sh
+++ b/files/test_in_toolbox.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+set -eu
+
+TOOLBOX_NAME=$1
+
+(toolbox list | grep -q $TOOLBOX_NAME) || {
+	toolbox create -c $TOOLBOX_NAME
+	toolbox run -c $TOOLBOX_NAME sudo make setup-toolbox
+}
+
+toolbox run -c $TOOLBOX_NAME make check
--- a/files/setup-toolbox.yaml
+++ b/files/setup-toolbox.yaml
@@ -1,0 +1,4 @@
+---
+- include: install-deps.yaml
+- include: install-deps-worker.yaml
+- include: recipe-tests.yaml


Makefile
```diff
@@ -90,5 +90,12 @@ check-inside-openshift: service worker test_image
 check-inside-openshift-zuul: test_image
 	ANSIBLE_STDOUT_CALLBACK=debug $(AP) files/check-inside-openshift.yaml
 
+setup-toolbox:
+	dnf install -y ansible
```
lachmanfrantisek, line 94: I am not sure if it is necessary. We don't do that elsewhere.

lachmanfrantisek (commented):
Nice!

Makefile
mfocko: fedora-toolbox image doesn't have ansible preinstalled

files/test_in_toolbox.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
```
csomh, line 2: `set -eu` to fail on the first error and when using undefined variables.

csomh (commented):
I'm not sure I understand what the intended utility for these targets is... Why not just use one of the existing images to create a toolbox? How is `check-inside-toolbox` different from `check_in_container`? (besides using `toolbox` vs `podman`/`docker`)

mfocko:
> I'm not sure I understand what the intended utility for these targets is... Why not just use one of the existing images to create a toolbox? How is `check-inside-toolbox` different from `check_in_container`? (besides using `toolbox` vs `podman`/`docker`)

@csomh It seems to me it's more swift compared to running in container where I've encountered rebuilds of image even throughout the day
(also it was more straightforward to set up local mirror for toolbox rather than `usercont/base` that managed to keep both images from registry and local and continued to use local ones -__-)

lachmanfrantisek:
> Why not just use one of the existing images to create a toolbox?

I don't think this will be easy. The toolbox images do not have easy build instructions.

> How is check-inside-toolbox different from check_in_container?

For users that don't use toolbox, there is probably only a small benefit (maybe easier access to the test environment).

Otherwise, it's more a question of "Why use toolbox at all?." For me, it's really handy, especially because I am using Silveblue and have all the development tools (GUI included) inside toolbox.

Makefile
lachmanfrantisek: Aha, this is not run on the host? Then I would suggest some better name.

csomh:
> I don't think this will be easy. The toolbox images do not have easy build instructions.

Strange, I thought something like `toolbox create --container packit-service --image usercont/packit-service:prod` is enough. Trying it out seems to be working and it seems to me that it's the right environment. What am I missing?

mfocko:
> > I don't think this will be easy. The toolbox images do not have easy build instructions.
> 
> Strange, I thought something like `toolbox create --container packit-service --image usercont/packit-service:prod` is enough. Trying it out seems to be working and it seems to me that it's the right environment. What am I missing?

pytest and flexmock from `files/recipe-tests.yaml` (you probably have them installed in `$HOME` through pip? iirc there were some problems with pytest from pip instead of RPM, maybe they're resolved)

Makefile
mfocko: what name do you suggest? :)

I could also move ansible installation into the playbook and add check for `/run/.envrc` to make sure we're running in container / toolbox

lachmanfrantisek: `setup-inside-toolbox`/`install-dependecies-into-toolbox`/`dependencies-inside-toolbox`/..

I now understand how your name makes sense but was confused originally if we are setting-up the toolbox or making setup inside toolbox.

jpopelka (approved):

PR #977, opened 2021-02-16
Speed up CI by using pre-built tests image

jpopelka:
We now have [tests image in Quay.io](https://quay.io/repository/packit/packit-service-tests) which is re-built on a push to this repo.
When we run tests, we don't rebuild the image (as we used to), but use the `quay.io/packit/packit-service-tests` and mount sources into its `/src`.

Pros:
- [packit-service-tests](https://github.com/packit/packit-service/blob/main/.zuul.yaml#L14) run about 2min instead of 8min
- [packit-service-tests-requre](https://github.com/packit/packit-service/blob/main/.zuul.yaml#L20) are a bit faster, around 17min instead of 18min - the most time takes spawning an Openshift cluster and building `worker:dev` & `service:dev`. It's actually 2 tests (1. tests with pre-recorded responses, 2. deployment test), but we run them together because spawning an Openshift cluster is such an expensive operation. If we split it, we could use pre-built `worker:dev` & `service:dev` for the requre test, which would speed up it a lot. But we need to build them for the deployment test anyway, which would again take (with spawning a cluster for it) about 18min so all tests together would take the same time as now.

Cons:
- If a PR adds a new dependency to [worker](https://github.com/packit/packit-service/blob/main/files/install-deps-worker.yaml) or [service](https://github.com/packit/packit-service/blob/main/files/install-deps.yaml) that dependency won't be in the test image during the tests run and the tests will probably fail. Work-around is to `make test_image` and `podman push`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-BASE_IMAGE ?= docker.io/usercont/base
+BASE_IMAGE ?= quay.io/packit/base
 SERVICE_IMAGE ?= docker.io/usercont/packit-service:dev
 WORKER_IMAGE ?= docker.io/usercont/packit-worker:dev
 WORKER_IMAGE_PROD ?= docker.io/usercont/packit-worker:prod
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,8 @@ check:
 test_image: files/install-deps-worker.yaml files/install-deps.yaml files/recipe-tests.yaml
 	$(CONTAINER_ENGINE) build --rm -t $(TEST_IMAGE) -f files/docker/Dockerfile.tests --build-arg SOURCE_BRANCH=$(SOURCE_BRANCH) .
 
-check_in_container: test_image
+check_in_container:
+	$(CONTAINER_ENGINE) pull $(TEST_IMAGE)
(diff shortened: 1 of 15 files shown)


Makefile
```diff
@@ -1,8 +1,8 @@
-BASE_IMAGE ?= docker.io/usercont/base
+BASE_IMAGE ?= quay.io/packit/base
 SERVICE_IMAGE ?= docker.io/usercont/packit-service:dev
 WORKER_IMAGE ?= docker.io/usercont/packit-worker:dev
 WORKER_IMAGE_PROD ?= docker.io/usercont/packit-worker:prod
-TEST_IMAGE ?= packit-service-tests
+TEST_IMAGE ?= quay.io/packit/packit-service-tests
```
mfocko, line 5: isn't this the reason for failing tests? Since it tries to pull the image from the quay instead of using the local build?

jpopelka: Thanks Maťo, no I don't think so.
But the ultimate goal here is to pull the image instead of building it.
It's still heavy WIP though.

TomasTomecek (approved):
:O

PR #980, opened 2021-02-16
Add description to MR comments on GitLab

### LINKED ISSUE #978: Add a status message to the GitLab comments
When we don't have permission to create status, we put a comment instead:

Screenshot from 2021-02-16 14-07-23

The table in the comment does not contain a message/description of the status.

mfocko:
- Add description of the status to the comment
- Also mention that we asked for access to the project

Fixes #978

### DIFF
--- a/packit_service/worker/build/build_helper.py
+++ b/packit_service/worker/build/build_helper.py
@@ -434,7 +434,16 @@ class BaseBuildJobHelper:
         # We are only commenting final states to avoid multiple comments for a build
         # Ignoring all other states eg. pending, running
         if state in final_commit_states and check_names is not None:
-            self.status_reporter.report_status_by_comment(state, url, check_names)
+            self.status_reporter.report_status_by_comment(
+                state,
+                url,
+                check_names,
+                (
+                    f"{description}\n\n---\n*Packit-User does not have access to the "
+                    "source project. We have requested access to be able to set "
+                    "commit statuses / pipelines instead of the comments.*"
+                ),
+            )
 
     def report_status_to_all(
         self, description: str, state: CommitStatus, url: str = ""
--- a/packit_service/worker/reporting.py
+++ b/packit_service/worker/reporting.py
@@ -115,6 +115,7 @@ class StatusReporter:
         state: CommitStatus,
         url: str,
         check_names: Union[str, list, None],
+        description: str,
     ):
         """
         Reporting build status with MR comment if no permission to the fork project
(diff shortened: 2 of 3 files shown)


packit_service/worker/build/build_helper.py
```diff
@@ -434,7 +434,15 @@ def _report(
...
         # Ignoring all other states eg. pending, running
         if state in final_commit_states and check_names is not None:
-            self.status_reporter.report_status_by_comment(state, url, check_names)
+            self.status_reporter.report_status_by_comment(
+                state,
+                url,
+                check_names,
+                (
+                    f"{description}\n---\n"
+                    "Packit-User does not have access to project. We have requested access."
```
lachmanfrantisek, line 443: What about using italics and mention the difference?
```suggestion
                    "*Packit-User does not have access to the source project. We have requested access to be able to use merge-request status instead of the comments.*"
```

lachmanfrantisek (commented):
Thanks!

packit_service/worker/build/build_helper.py
mfocko: :+1:

lachmanfrantisek (approved):

TomasTomecek (approved):
Very nice!

mfocko:
added one more newline before `---`, since it interprets it with the text on the line above as header

packit_service/worker/reporting.py
```diff
@@ -128,7 +129,8 @@ def report_status_by_comment(
             "| ------------- | ------------ |",
         ] + [f"| [{check}]({url}) | {state.name.upper()} |" for check in check_names]
 
-        self.comment("\n".join(comment_table_rows))
+        table = "\n".join(comment_table_rows)
+        self.comment(table + f"\n### Description\n\n{description}")
```
lachmanfrantisek, line 133: @mfocko I am looking at it now and thinking if we are not doing this too complicated:

* If we want to match status formatting, I would add the description to the table.
    * => Differentiate between `description` and `footer`.
* As a second option, we can change the behaviour a bit:
  * Use table only when chroots are relevant, unfortunately, we don't process all chroots together and I'm not sure how easy is to check this condition.
  * In case of permission errors, it does not make sense to print some chroots at all.

WDYT?

mfocko: I'm looking at it and there's nothing to match tbh
- GitLab
  ```
  IF reporting is allowed THEN
    StatusReporter.report()
    // that tries commit status, if fails, falls back to commit comment
  ELSE
    report by comment (if merge request is present then MR comment else commit comment)
  FI
   ```
- GitHub
  ```
  StatusReporter.report()
  // that tries commit status, if fails, falls back to commit comment
  ```

I'll try to go through the commits if there is any reason we check reporting and don't go by default (I guess the reason would be requesting access which is something that `StatusReporter` doesn't take care of... but we probably could check if it's needed to request permissions and then just call that `report` method

Ad 2nd question, it's weird, because we can get `str` or `List[str]` of check names, but only one description, since there is no problem with that on GitHub, I guess that's something GitLab specific.

mfocko: AI:
- [ ] check the stage as is now
- [ ] clean it up (*aargh*)

mfocko: current state of stage: https://gitlab.com/packit-service/src/chrony/-/merge_requests/1#note_510963266

Issue #967, opened 2021-02-10
parsing TFT results in a traceback


TomasTomecek:
```
Traceback (most recent call last):                                                                                                                                                            
  File "/usr/lib/python3.9/site-packages/celery/app/trace.py", line 405, in trace_task                                                                                                        
    R = retval = fun(*args, **kwargs)                                                                                                                                                         
  File "/usr/lib/python3.9/site-packages/celery/app/trace.py", line 697, in __protected_call__                                                                                                
    return self.run(*args, **kwargs)                                                                                                                                                          
  File "/usr/local/lib/python3.9/site-packages/sentry_sdk/integrations/celery.py", line 197, in _inner                                                                                        
    reraise(*exc_info)                                                                                                                                                                          File "/usr/local/lib/python3.9/site-packages/sentry_sdk/_compat.py", line 54, in reraise                                                                                                        raise value                                                                                                                                                                                 File "/usr/local/lib/python3.9/site-packages/sentry_sdk/integrations/celery.py", line 192, in _inner                                                                                            return f(*args, **kwargs)                                                                                                                                                                 
  File "/src/packit_service/worker/tasks.py", line 98, in process_message                                                                                                                     
    return SteveJobs().process_message(event=event, topic=topic, source=source)                                                                                                               
  File "/src/packit_service/worker/jobs.py", line 332, in process_message                                                                                                                     
    event_object = Parser.parse_event(event)                                                                                                                                                  
  File "/src/packit_service/worker/parser.py", line 113, in parse_event                                                                                                                       
    for response in map(                                                                                                                                                                      
  File "/src/packit_service/worker/parser.py", line 114, in <lambda>                                                                                                                          
    lambda parser: parser(event),                                                                                                                                                             
  File "/src/packit_service/worker/parser.py", line 778, in parse_testing_farm_results_event                                                                                                  
    tests: List[TestResult] = Parser._parse_tf_result_xunit(                                                                                                                                  
  File "/src/packit_service/worker/parser.py", line 738, in _parse_tf_result_xunit                                                                                                            
    return [                                                                                                                                                                                  
  File "/src/packit_service/worker/parser.py", line 740, in <listcomp>                                                                                                                        
    name=testcase["@name"],                                                                                                                                                                   
TypeError: string indices must be integers   
```

not sure if this is known

jpopelka:
Trying to reproduce it in our worker:prod image, but with no luck.
```python
[...]

PR #988, opened 2021-02-19
_parse_tf_result_xunit(): silence TypeError

jpopelka:
Related to #967

I've been not able to reproduce the issue, trying to parse the same xunit string in `worker:prod` container.

Given that we don't need the parsed data in most cases, let's just continue instead of raising the exception, which would prevent us from updating the check in GUI.


TomasTomecek (approved):
good-enough for now

PR #992, opened 2021-02-23
Github actions to build & push images (worker, service, tests)

jpopelka:
We need to specify `SOURCE_BRANCH` build arg, which AFAIK can't be done in Quay.io build trigger so we have to build it in Github.


csomh:
Would it be possible to get rid of some of the duplication by using context variables, for example? Also, is there some kind of "matrix" functionality in GitHub workflows? Maybe that could be used to get away with a single workflow file.

jpopelka:
> Would it be possible to get rid of some of the duplication by using context variables, for example?

Yes, it is possible to merge them but at a price of much more code.
I decided to go the simpler, but duplicated code way.
We can always merge later (once I finally read something more about actions and know the best practices).

> Also, is there some kind of "matrix" functionality

no idea what's that

csomh:
> We can always merge later (once I finally read something more about actions and know the best practices).
> 

I would like to see this work done in this PR, in order to avoid merging something which adds more to the pile of technical dept.

csomh:
At least let's have a look if [using a build matrix](https://docs.github.com/en/actions/learn-github-actions/managing-complex-workflows#using-a-build-matrix) could help to remove some of the duplication.

.github/workflows/rebuild-and-push-images.yml
```diff
@@ -0,0 +1,62 @@
...
+          - dockerfile: files/docker/Dockerfile.tests
+            image: packit-service-tests
+
+    steps:
+      - uses: actions/checkout@v2
+
+      # https://stackoverflow.com/a/58035262/2768738
+      - name: Extract branch name
+        shell: bash
+        run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
```
csomh, line 29:
```suggestion
        run: |
            branch=${GITHUB_REF#refs/heads/}
            tag=""
            case $branch in:
                main)
                    tag=stg
                    ;;
                stable)
                    tag=prod
                    ;;
            esac
            echo "::set-output name=branch::$branch"
            echo "::set-output name=tag::$tag"            
```

This should set `steps.extract_branch.outputs.tag`  and the next two steps could be removed.

jpopelka: Thanks, exactly what I've been looking for!
Works OK in my fork :rocket: 
Final review, please :-)

csomh (approved):
Thanks a lot for all the improvements! :rocket:

PR #993, opened 2021-02-23
Switch to ogr's get_latest_release

mfocko:

### DIFF
--- a/packit_service/service/events.py
+++ b/packit_service/service/events.py
@@ -710,8 +710,9 @@ class IssueCommentGitlabEvent(AddIssueDbTrigger, AbstractGitlabEvent):
     @property
     def tag_name(self):
         if not self._tag_name:
-            releases = self.project.get_releases()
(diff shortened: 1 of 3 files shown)


packit_service/service/events.py
```diff
@@ -710,8 +710,9 @@ def __init__(
     @property
     def tag_name(self):
         if not self._tag_name:
-            releases = self.project.get_releases()
-            self._tag_name = releases[0].tag_name if releases else ""
+            self._tag_name = ""
+            if latest_release := self.project.get_latest_release():
```
lachmanfrantisek, line 714: I hope we don't break anything with this..;)

lachmanfrantisek (approved):

jpopelka (approved):

packit_service/service/events.py
mfocko: base is F33, so we should be fine :)

PR #994, opened 2021-02-24
Deny processing GitLab webhooks without token

mfocko:

### DIFF
--- a/packit_service/service/api/webhooks.py
+++ b/packit_service/service/api/webhooks.py
@@ -221,7 +221,7 @@ class GitlabWebhook(Resource):
                 msg = "X-Gitlab-Token not in request.headers"
                 logger.info(msg)
                 self.create_confidential_issue_with_token()
-                return
+                raise ValidationFailed(msg)
             else:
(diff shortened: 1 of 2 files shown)


mfocko:
issue with token should be created only once, so I don't expect possibility of spamming; also I haven't noticed new issue when testing

lachmanfrantisek (commented):
Would be nice to have some test-case but LGTM otherwise. Thanks, good catch!

jpopelka (approved):

mfocko:
> Would be nice to have some test-case but LGTM otherwise. Thanks, good catch!

extended, we had tests for valid, invalid secrets/tokens in headers, but not for missing

lachmanfrantisek (approved):
Thanks!

mfocko:
recheck

mfocko:
recheck

TomasTomecek (approved):
nice

jpopelka (approved):
rebase please, CI should be fixed

PR #995, opened 2021-02-24
Do not log failed GitLab status as error + refactor later

mfocko:
TODO:
- [ ] refactor
   **currently**: if there are no permissions we ask for them and report via comment (PR or commit comment)
   I've noticed TF bypasses it and posts commit comments... with regards to refactoring I suggest to just post commit comments as a fallback; on the other hand, there is also an option to use `.comment` that prefers PR comment to commit comment

Since the status is more tied to the commit itself rather than PR (and commit comments are shown in PRs), I suggest commenting only on commits.
   

### DIFF
--- a/packit_service/worker/build/build_helper.py
+++ b/packit_service/worker/build/build_helper.py
@@ -410,37 +410,30 @@ class BaseBuildJobHelper:
         The status reporting should be done through this method
         so we can extend it in subclasses easily.
         """
-        if not self.is_gitlab_instance or self.is_reporting_allowed:
-            self.status_reporter.report(
-                description=description,
-                state=state,
-                url=url,
-                check_names=check_names,
+        if self.is_gitlab_instance and not self.is_reporting_allowed:
+            self.request_project_access()
+            description = (
+                f"{description}\n\n---\n*Packit-User does not have access to the "
+                "source project. We have requested access to be able to set "
+                "commit statuses / pipelines instead of the comments.*"
             )
-            return
 
-        # "Packit-User" does not have access to project.
-
-        self.request_project_access()
-        final_commit_states = [
-            CommitStatus.error,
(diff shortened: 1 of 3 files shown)


lachmanfrantisek (approved):

jpopelka (approved):

mfocko:
@lachmanfrantisek would it make sense to keep only commit comments for statuses and once the build result page is running, just keep one comment like so (reason for datetime of last update: so we don't spam, but still there would be some notion of change):

---

For results of the build see \<url goes here\>

*\<in case there are no permissions to set commit status, comment goes here\>*

Latest update: \<datetime of latest change of status\>

TomasTomecek:
@mfocko I wanted to check how kernel does it but can't see their Stream repos :/ as I remember, they are using pipelines to report statuses

one MR comment sounds like a good solution to me, since the fork permission mechanism is so unfriendly

lachmanfrantisek:
@mfocko To conclude this (and to check I get that right..;):

* If we have permissions to place statuses,  put a commit status for the process updates.
* If we don't have permissions to use statuses, use a commit comment.

But when we need to put a comment with more info -- config problems, Copr project settings,... Which one do we want to use, MR comment? (Feels better to me.)

mfocko:
> But when we need to put a comment with more info -- config problems, Copr project settings,... Which one do we want to use, MR comment? (Feels better to me.)

@lachmanfrantisek there is `.comment` method that prefers PR/MR comment to commit comment, iirc it's the one used for config, COPR settings

mfocko:
> Build failed (gate pipeline).
> 
>     * [pre-commit ](https://softwarefactory-project.io/zuul/t/packit-service/build/None) : NODE_FAILURE in 0s

*me screams and runs* not this again -_-

PR #996, opened 2021-02-24
[build targets] use targets both from build job and test job

lbarcziova:
When we have configuration like this:
```yaml
jobs:
- job: copr_build
  trigger: pull_request
  metadata:
    targets:
    - fedora-stable-x86_64

- job: tests
  trigger: pull_request
  metadata:
    targets:
    - fedora-rawhide-x86_64
 ```
 build for targets both from build job and test job (currently the build would not run for `fedora-rawhide-x86_64`).
 
 The tests were testing only mocked functionality, so there are some changes. They now test the `configured_build_targets` and `configured_test_targets` instead of the `build_targets` and `test_targets` which call `get_valid_build_targets` which uses Copr and Bodhi, so now we check the targets used as input for `get_valid_build_targets`.
 
 Related to #976

### DIFF
--- a/packit_service/worker/build/build_helper.py
+++ b/packit_service/worker/build/build_helper.py
@@ -156,31 +156,28 @@ class BaseBuildJobHelper:
         """
         Return the targets to build.
 
-        1. If the job is not defined, use the test_targets.
-        2. If the job is defined, but not the targets, use "fedora-stable" alias otherwise.
+        1. Use targets defined for build job and targets defined for test job.
+        2. Use "fedora-stable" alias if neither defined.
         """
-        if (
-            (not self.job_build or not self.job_build.metadata.targets)
-            and self.job_tests
-            and self.job_tests.metadata.targets
-        ):
-            return self.configured_tests_targets
+        targets = set()
+        if self.job_build:
+            targets.update(self.job_build.metadata.targets)
 
-        if self.job_build and self.job_build.metadata.targets:
-            return self.job_build.metadata.targets
+        if self.job_tests:
+            targets.update(self.job_tests.metadata.targets)
 
-        return {"fedora-stable"}
+        return targets or {"fedora-stable"}
 
     @property
     def configured_tests_targets(self) -> Set[str]:
         """
-        Return the list of chroots used in the testing farm.
-        Has to be a sub-set of the `build_chroots`.
+        Return the targets used in the testing farm.
+        Has to be a sub-set of the `configured_build_targets`.
 
-        Return an empty list if there is no job configured.
+        Return an empty set if there is no test job configured.
 
         If not defined:
-        1. use the build_chroots if the job si configured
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by lbarcziova** [comment]
recheck

File: packit_service/worker/build/build_helper.py [hunk 1]
```diff
@@ -156,31 +156,28 @@ def configured_build_targets(self) -> Set[str]:
...
-        ):
-            return self.configured_tests_targets
+        targets = set()
+        if self.job_build:
+            targets.update(self.job_build.metadata.targets)
 
-        if self.job_build and self.job_build.metadata.targets:
-            return self.job_build.metadata.targets
+        if self.job_tests:
+            targets.update(self.job_tests.metadata.targets)
```
mfocko, line 167: I think this could be problem / ambiguous with the following configuration:

```yaml
jobs:
- job: copr_build
  trigger: pull_request

- job: tests
  trigger: pull_request
  metadata:
    targets:
    - fedora-rawhide-x86_64
```

Based on the previous description I'd expect COPR build for `fedora-stable` and tests for `fedora-rawhide`. (Based on code we built `fedora-rawhide` and ran tests for it.)
Based on the current description it should be just `fedora-rawhide`. (Code conforms to the description).

Is this intended change though?

I can imagine use case:
> I want COPR builds for `stable` and run tests for `rawhide`, so I would just *abuse* defaults for `copr_build` and expect the service to build the `rawhide` for TF too.

mfocko (commented):
otherwise LGTM :)

packit_service/worker/build/build_helper.py
mfocko: Looking at the documentation my question might be answered. We "require" targets to be set :D

lbarcziova: But this is not really a change. 
Previous code from `configured_build_targets`:
```python
        if (
            (not self.job_build or not self.job_build.metadata.targets)
            and self.job_tests
            and self.job_tests.metadata.targets
        ):
            return self.configured_tests_targets
 ```
 
 On the other hand, it is true it can be misleading and we can change it now, if we decide or document it properly.

lachmanfrantisek (approved):
Thanks, looks good to me. And it's also simplified -- nice!

lbarcziova:
recheck

TomasTomecek (approved):

jpopelka:
recheck

jpopelka:
rebase please, CI should be fixed

TomasTomecek:
Awesome, another UX bug fixed!

lbarcziova:
CI apparently doesn't like this PR :persevere:

mfocko:
> CI apparently doesn't like this PR :persevere:

it's just telling you to start the extended weekend sooner :upside_down_face:

PR #1001, opened 2021-02-25
Allowlist for GitLab

mfocko:
TODO:
- [x] DB migration to prefix current entries
- [ ] automatically add GitLab namespaces to waiting list
  - [x] thinking about it: **fix the bug when adding from GitHub installations**
- [x] Agree on the implementation
- [x] Decide if we should support `AllowlistStatus.denied`
- [x] Fix tests and adjust to new behaviour
- [x] Implement decided interface
- [X] Check format of events to prevent ambiguity, e.g. `https://github.com/mfocko/dotfiles.git.git` (ofc GitLab allows it)
   - [x] Shall we save it in the format `github.com/packit/packit` or `github.com/packit/packit.git`? I prefer `.git`, since on GitLab you could get your repository approved and then create subnamespace O.o
   - URLs in hooks don't contain `.git` suffix
- [x] UX of allowlisting new accounts (especially if we add hostname and support separate repositories)

### DIFF
--- a/alembic/versions/800abbbb23c9_allow_multiple_forges_in_allowlist.py
+++ b/alembic/versions/800abbbb23c9_allow_multiple_forges_in_allowlist.py
@@ -1,0 +1,105 @@
+"""Allow multiple forges in allowlist
+
+Revision ID: 800abbbb23c9
+Revises: a5c06aa9ef30
+Create Date: 2021-03-25 10:43:00.679552
+
+"""
+import enum
+from typing import TYPE_CHECKING, Dict
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy import Column, Integer, String, Enum, orm
+
+# from sqlalchemy.dialects import postgresql
+from sqlalchemy.ext.declarative import declarative_base
+
+
+# revision identifiers, used by Alembic.
+revision = "800abbbb23c9"
+down_revision = "a5c06aa9ef30"
+branch_labels = None
+depends_on = None
+
+
+# https://github.com/python/mypy/issues/2477#issuecomment-313984522 ^_^
+if TYPE_CHECKING:
+    Base = object
+else:
+    Base = declarative_base()
+
+ALLOWLIST_CONSTANTS = {
+    "approved_automatically": "approved_automatically",
+    "waiting": "waiting",
+    "approved_manually": "approved_manually",
+    "denied": "denied",
+}
+
+
+class AllowlistStatus(str, enum.Enum):
+    approved_automatically = ALLOWLIST_CONSTANTS["approved_automatically"]
+    waiting = ALLOWLIST_CONSTANTS["waiting"]
+    approved_manually = ALLOWLIST_CONSTANTS["approved_manually"]
+    denied = ALLOWLIST_CONSTANTS["denied"]
+
+
+class AllowlistModel(Base):
+    __tablename__ = "allowlist"
+    id = Column(Integer, primary_key=True)
+    namespace = Column(String, index=True)  # renamed from account_name
+    status = Column(Enum(AllowlistStatus))
+    fas_account = Column(String)
+
+    def to_dict(self) -> Dict[str, str]:
+        return {
+            "namespace": self.namespace,
+            "status": self.status,
+            "fas_account": self.fas_account,
+        }
+
+    def __repr__(self):
+        return (
+            f'<AllowlistModel(namespace="{self.namespace}", '
+            f'status="{self.status}", '
+            f'fas_account="{self.fas_account}")>'
+        )
+
+
+def upgrade():
+    op.add_column("allowlist", sa.Column("fas_account", sa.String(), nullable=True))
+
+    # rename account_name to namespace
+    op.execute("ALTER TABLE allowlist RENAME COLUMN account_name TO namespace")
+
+    # replaces creating new index and dropping old one
+    op.execute("ALTER INDEX ix_allowlist_account_name RENAME TO ix_allowlist_namespace")
+
+    # update all of the entries
+    bind = op.get_bind()
+    session = orm.Session(bind=bind)
+
+    for entry in session.query(AllowlistModel).all():
+        if "/" not in entry.namespace:
+            entry.namespace = f"github.com/{entry.namespace}"
+
+    session.commit()
+
+
+def downgrade():
+    bind = op.get_bind()
+    session = orm.Session(bind=bind)
+
+    for entry in session.query(AllowlistModel).all():
+        if entry.namespace.startswith("github.com/"):
+            _, entry.namespace = entry.namespace.split("/", 1)
+    session.commit()
+
+    # drop additional information
+    op.drop_column("allowlist", "fas_account")
+
+    # rename back to account_name
+    op.execute("ALTER TABLE allowlist RENAME COLUMN namespace TO account_name")
+
+    # fix index; will it work with just rename? O.o
+    op.execute("ALTER INDEX ix_allowlist_namespace RENAME TO ix_allowlist_account_name")
--- a/tests/unit/test_allowlist.py
+++ b/tests/unit/test_allowlist.py
@@ -1,7 +1,7 @@
 # Copyright Contributors to the Packit project.
 # SPDX-License-Identifier: MIT
 
-from typing import List, Tuple, TYPE_CHECKING
+from typing import Tuple, Iterable
 
 import pytest
 from copr.v3 import Client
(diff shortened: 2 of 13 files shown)


TomasTomecek (commented):
I'm not sure what exactly is your thought-process but I wouldn't personally do it with regexes.

TomasTomecek:
I'd personally do it like this:
```
class AllowlistModel(Base):
    __tablename__ = "allowlist"
    id = Column(Integer, primary_key=True)

    # this should be the account name of the person who installed packit in github - not sure what would the gitlab equiv would be
    account_name = Column(String, index=True)

    # ideally we should also store the fas account name of the requester
    fas_account = Column(String)

    # string or enum
    forge = {github.com, gitlab.com}

    # root namespace, let's just approve everything for users or namespaces - I don't see what the benefit would be for being able to block subnamespaces
    namespace = Column(String)
    status = Column(Enum(AllowlistStatus))
```

and translate the existing data in the schema migration

is this clear, does it make sense?

lachmanfrantisek:
> since on GitLab you could get your repository approved and then create subnamespace O.o

If they want to be evil, they can do evil things in the already-approved repo as well.

mfocko:
based on Tomas' suggestion I propose:
```py
class AllowlistModel(Base):
    __tablename__ = "allowlist"
    id = Column(Integer, primary_key=True)
    namespace = Column(String, index=True) # previously account_name

    parent_id = Column(Integer, ForeignKey("allowlist.id"), nullable=True)
    parent = relationship("AllowlistModel")

    # parent = relationship("AllowlistModel", back_populates="subnamespaces")
    # subnamespaces = relationship("AllowlistModel", back_populates="parent")
 
    status = Column(Enum(AllowlistStatus))
    # ideally we should also store the fas account name of the requester
    fas_account = Column(String)
```

I would probably approach it as Trie(s)
- it would allow us to store github/gitlab only once (with implicit `AllowlistStatus.Deny`)
- we can differentiate forges, e.g. `github.com/some_namespace` is allowed and we get triggered on `gitlab.com/some_namespace/some_repo.git`, we will find `some_namespace` entry, but parent is `github.com` => denied
- FAS username is fine (related to plans for implementation of automatic approval based on `packager` group and/or CLA)

TomasTomecek:
I'd say the parent mechanism is overkill but won't stop you if you think it's a good idea :) since you'll be writing the code :D 

On the other hand I agree this will give us complete flexibility and explicit permissions for every single repo in a namespace - I just wanna make sure:
1. we'll have code to allow all repos in a namespace (and all subnamespaces)
2. p-s will work the same way for current users - meaning, they will not notice we have changed in how allowlist works

mfocko:
> I'd say the parent mechanism is overkill but won't stop you if you think it's a good idea :) since you'll be writing the code :D

"A bit" :D also Franta has suggested we should hit around 3 levels at most, so majority of entries will be namespaces like are now, they would just keep one link to "root namespace" = github/gitlab (even allowing self-hosted instances of gitlab nicely)

> 1. we'll have code to allow all repos in a namespace (and all subnamespaces)

That's the intention... Let's say we allowlist `gitlab.com/packit-service/src`,
- we get request for `gitlab.com/packit-service/src/glibc.git`
  1. can't find `glibc.git`, let's go one level up
  2. found `src` (allowed), let's check rest of the path, done
- we get request for `github.com/src/something.git`
  1. can't find `something.git`, level up
  2. found `src` (allowed), let's check rest of the path, from database i get `packit-service`, but I have `github.com` => denied

it mostly hits UX of the allowlist script we use; will have to adjust it too

> 2. p-s will work the same way for current users - meaning, they will not notice we have changed in how allowlist works

yeah, apart from allowlisting namespaces of some bots and developers yeah :D (that's relict of allowing namespace OR user; could be probably cleaned up against projects from which we received webhooks)

TomasTomecek:
@mfocko got it now, thanks for the detailed explanation, makes complete sense to me, very nice design!

mfocko:
updated the tests, will probably create fake DB in fixture, since it looks awful right now; there are added comments also, the format is `(queried_namespace, returned_db_entries)`, in that order (as described [here](https://github.com/packit/packit-service/pull/1001#issuecomment-786076251))

tests/unit/test_allowlist.py
```diff
@@ -46,39 +46,301 @@ def allowlist():
...
+                                status=AllowlistStatus.denied,
+                                parent_id=8,
+                                parent=flexmock(
+                                    id=8,
+                                    namespace="gitlab.com",
+                                    status=AllowlistStatus.denied,
+                                    parent_id=None,
+                                ),
+                            ),
+                        )
```
mfocko, line 238: missing github.com with implicit deny...

tests/unit/test_allowlist.py
```diff
@@ -46,39 +46,301 @@ def allowlist():
...
+                            parent=flexmock(
+                                id=1,
+                                account_name="github.com",
+                                status=AllowlistStatus.denied,
+                                parent_id=None,
+                            ),
+                        )
+                    ),
+                )
+            ),
```
mfocko, line 135: missing gitlab.com

TomasTomecek:
@mfocko would you have time to finish the work this week?

mfocko:
> @mfocko would you have time to finish the work this week?

I hope so, majority is done, so card can be closed imo
we're talking about:
- database migration
- script for approving that we use internally
- handling of `waiting` since we can handle the repositories themselves

mfocko:
Please go by commits when reviewing. In case I rebased and squashed something I didn't want, **YELL**. :D

alembic/versions/800abbbb23c9_allow_multiple_forges_in_allowlist.py
```diff
@@ -0,0 +1,106 @@
...
+
+    session.commit()
+
+
+def downgrade():
+    bind = op.get_bind()
+    session = orm.Session(bind=bind)
+
+    for entry in session.query(AllowlistModel).all():
+        if entry.namespace.startswith("github.com/"):
```
mfocko, line 95: this way we can keep the gitlab accounts when downgrading, they are unusable with the previous implementation... could be probably a better idea to check for domain when upgrading too :thinking:

alembic/versions/800abbbb23c9_allow_multiple_forges_in_allowlist.py
```diff
@@ -0,0 +1,106 @@
...
+        )
+
+
+def upgrade():
+    op.add_column("allowlist", sa.Column("fas_account", sa.String(), nullable=True))
+
+    # rename account_name to namespace
+    op.execute("ALTER TABLE allowlist RENAME COLUMN account_name TO namespace")
+
+    # fix index; will it work with just rename? O.o
```
TomasTomecek, line 75: you can try locally with stg data :)

alembic/versions/800abbbb23c9_allow_multiple_forges_in_allowlist.py
```diff
@@ -0,0 +1,106 @@
...
+    op.execute("ALTER INDEX ix_allowlist_account_name RENAME TO ix_allowlist_namespace")
+    # op.create_index(op.f('ix_allowlist_namespace'), 'allowlist', ['namespace'], unique=False)
+    # op.drop_index('ix_allowlist_account_name', table_name='allowlist')
+
+    # update all of the entries
+    bind = op.get_bind()
+    session = orm.Session(bind=bind)
+
+    for entry in session.query(AllowlistModel).all():
+        entry.namespace = f"github.com/{entry.namespace}"
```
TomasTomecek, line 85: so one of the first things we'll need to do will be to approve gitlab.com/packit-service after we deploy this to stg

tests/unit/test_allowlist.py
```diff
@@ -46,39 +46,151 @@ def allowlist():
...
+            id=8,
+            namespace="gitlab.com/packit",
+            status=AllowlistStatus.approved_manually.value,
+        ),
+        "github.com/packit": flexmock(
+            id=10,
+            namespace="github.com/packit",
+            status=AllowlistStatus.denied.value,
+        ),
+        "gitlab.com/packit-service/src": flexmock(
```
TomasTomecek, line 88: could we get another layer for gitlab here? e.g. `gitlab.com/packit-service/src/systemd`

tests/unit/test_allowlist.py
```diff
@@ -46,39 +46,151 @@ def allowlist():
...
+            ("github.com/konipas", "github.com"),
+            False,
+        ),
+        (  # Krasomila is not present at all, checks parent
+            "github.com/krasomila",
+            ("github.com/krasomila", "github.com"),
+            False,
+        ),
+        (  # gitlab.com/packit-service/src is approved
+            "gitlab.com/packit-service/src/glibc.git",
```
TomasTomecek, line 163: oh, you're checking it here

tests/unit/test_allowlist.py
TomasTomecek: ...so what should happen here actually for gitlab.com?

TomasTomecek (commented):
Awesome! Really love that you added way-more tests and updated docs and the script as well.

alembic/versions/800abbbb23c9_allow_multiple_forges_in_allowlist.py
TomasTomecek: afaik for gitlab it's being used only in gitlab.com/packit-service (and soon gitlab.com/redhat/centos-stream/src)

alembic/versions/800abbbb23c9_allow_multiple_forges_in_allowlist.py
mfocko: already tried with prod data :D seems to be fine... we probably also could set the FAS accounts based on `automatically_approved`, but it's just matched login, which would go against the already open issue for actually verifying through the login to FAS.

I'm still thinking about adding the option to keep `gitlab.com/` prefixes when downgrading, just in case

tests/unit/test_allowlist.py
mfocko: > could we get another layer for gitlab here? e.g. `gitlab.com/packit-service/src/systemd`

test? sure

> ...so what should happen here actually for gitlab.com?

nothing, it just looks it up as parent namespace that can be approved; this could be potentially helpful with self-hosted instances with restricted access (e.g. for testing), it would allow allowlisting whole instance

tests/unit/test_allowlist.py
mfocko: oh, I interpreted it as another subnamespace in the comment above

tests/unit/test_allowlist.py
TomasTomecek: yup, as a test case

thanks for explanation

tests/unit/test_allowlist.py
TomasTomecek: I just wanted to be sure the behaviour is correct above

alembic/versions/800abbbb23c9_allow_multiple_forges_in_allowlist.py
lachmanfrantisek, line 75: If this is tested, let's remove that comment...;)

alembic/versions/800abbbb23c9_allow_multiple_forges_in_allowlist.py
```diff
@@ -0,0 +1,106 @@
...
+def upgrade():
+    op.add_column("allowlist", sa.Column("fas_account", sa.String(), nullable=True))
+
+    # rename account_name to namespace
+    op.execute("ALTER TABLE allowlist RENAME COLUMN account_name TO namespace")
+
+    # fix index; will it work with just rename? O.o
+    op.execute("ALTER INDEX ix_allowlist_account_name RENAME TO ix_allowlist_namespace")
+    # op.create_index(op.f('ix_allowlist_namespace'), 'allowlist', ['namespace'], unique=False)
+    # op.drop_index('ix_allowlist_account_name', table_name='allowlist')
```
lachmanfrantisek, line 78: Why is this here?

lachmanfrantisek (commented):
Thanks for all the work!

Please, update the migration script (only comments?) so it looks like it will really work..;)

alembic/versions/800abbbb23c9_allow_multiple_forges_in_allowlist.py
mfocko: alembic initially generated `create + drop` for the index, i checked after the migration (with the `rename` that is uncommented):
```
packit=# \d allowlist 
                                    Table "public.allowlist"
   Column    |       Type        | Collation | Nullable |                Default                
-------------+-------------------+-----------+----------+---------------------------------------
 id          | integer           |           | not null | nextval('allowlist_id_seq'::regclass)
 namespace   | character varying |           |          | 
 status      | allowliststatus   |           |          | 
 fas_account | character varying |           |          | 
Indexes:
    "allowlist_pkey" PRIMARY KEY, btree (id)
    "ix_allowlist_namespace" btree (namespace)

packit=# \d ix_allowlist_namespace 
     Index "public.ix_allowlist_namespace"
    Column    |       Type        | Definition 
--------------+-------------------+------------
 account_name | character varying | namespace
btree, for table "public.allowlist"
```

so it is OK even if it keeps the old column name in index, right? cause it registers `ix_*` as index from the table

lachmanfrantisek: Aha, so we can remove those lines or put a comment that the rename replaces create&drop.

lachmanfrantisek (approved):
:rocket:

PR #1019, opened 2021-03-12
Increase delay in retrying failed propose-downstream

jpopelka:
Before this, we were retrying in 15s and then in 30s.

That's too little because the action for uploading the file to PyPI [took over 1 minute](https://github.com/packit/requre/actions/runs/645692636).

With this commit, we'll retry in 1m and then again in 2m.

Related to https://github.com/packit/requre/issues/167


lbarcziova (approved):
LGTM, thanks!

PR #1020, opened 2021-03-15
Fix handling of separately processed events

lbarcziova:
We need to assign handlers for `InstallationEvent` and `PullRequestLabelPagureEvent` directly, but we were doing:
```python
if isinstance(event, InstallationEvent):
```
 where the `event` was the original dictionary and not the parsed object, therefore this was always False. Therefore the installations were not handled properly and issues in notifications repo were not created.


jpopelka (approved):
good catch!

TomasTomecek (approved):

Issue #968, opened 2021-02-10
Reference test repository in the test job metadata


mrc0mmand:
Hello!

Would it be possible to use the test job metadata to reference a repository which contains the whole FMF tree with tests to run? Something along these lines:

```yaml
    jobs:
    - job: tests
      trigger: pull_request
      metadata:
        targets:
        - fedora-all
        repo_url: "https://github.com/johnsmith/my-fabulous-tests"
```

I know that currently you can reference a test repository [using FMF](https://packit.dev/docs/testing-farm/#systemd-tests), but even for this you need to have an initialized FMF tree in the repo plus the respective FMF file. Being able to reference the test repo directly in `packit.yml` would be way nicer, imo.

I'm not sure if this RFE should be filed for Packit or TMT/TFT, so apologies if I chose the *wrong* project :-)

csomh:
Hey @mrc0mmand :wave: 

This is definately the _right_ project to file this issue and this is a good proposal. Let's see what the others say, but I'm all for implementing this... and shouldn't be that difficult either.

jpopelka:
I'd call it `fmf_url` and optional `fmf_ref`, but yes, makes sense.
```yaml
    jobs:
    - job: tests
      trigger: pull_request
      metadata:
        targets:
          - fedora-all
        fmf_url: https://github.com/johnsmith/my-fabulous-tests
        fmf_ref: main  # optional, defaults to repo's default branch
```

It should be as easy as putting this [into TF request payload instead of the source repo url/ref](https://github.com/packit/packit-service/blob/d12836284f5e34f54dc6de20279175493b5f5d31/packit_service/worker/testing_farm.py#L77).

mrc0mmand:
Looking at a possible implementation (so I could start playing with TFT in systemd) - wouldn't be necessary to extend the metadata related classes in Packit as well? Like [JobMetadataSchema](https://github.com/packit/packit/blob/ff61fe51420148c9a41d9342e74656cc4c179293/packit/schema.py#L183) and [JobMetadataConfig](https://github.com/packit/packit/blob/0e4f301c0703ac5a40e797c73f8d6b1dfa2d1372/packit/config/job_config.py#L54).

mfocko:
Closed by packit/packit#1155, in case you feel the point of this issue was not fulfilled, please reopen

mrc0mmand:
I suspect this needs https://github.com/packit/packit-service/pull/1021 as well, so we can use it with PaaS?

[...]

PR #1021, opened 2021-03-15
Allow to reference a test repo in test job metadata

mmuzila:
This change allows to specify distinct repository for tests.

Fixes: #968

### DIFF
--- a/packit_service/worker/testing_farm.py
+++ b/packit_service/worker/testing_farm.py
@@ -56,6 +56,20 @@ class TestingFarmJobHelper(CoprBuildJobHelper):
                 self._tft_api_url += "/"
         return self._tft_api_url
 
+    @property
+    def fmf_url(self):
+        return (
+            self.job_config.metadata.fmf_url
+            or self.project.get_pr(self.metadata.pr_id).source_project.get_web_url()
+        )
+
+    @property
+    def fmf_ref(self):
+        if self.job_config.metadata.fmf_url:
+            return self.job_config.metadata.fmf_ref
+
+        return self.metadata.commit_sha
+
     def _payload(self, build_id: int, chroot: str) -> dict:
         """
         Testing Farm API: https://testing-farm.gitlab.io/api/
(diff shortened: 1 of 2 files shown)


packit_service/worker/testing_farm.py
```diff
@@ -56,6 +56,20 @@ def tft_api_url(self) -> str:
...
+    @property
+    def fmf_url(self):
+        return (
+            self.job_config.metadata.fmf_url
+            or self.project.get_pr(self.metadata.pr_id).source_project.get_web_url()
+        )
+
+    @property
+    def fmf_ref(self):
+        if self.job_config.metadata.fmf_url:
```
jpopelka, line 68: Why do we check for `fmf_url` and not `fmf_ref`?

jpopelka (commented):
LGTM with one question.

Would be nice to finish packit/packit#1155 so that tests here can pass.

mmuzila:
> LGTM with one question.
> 
> Would be nice to finish [packit/packit#1155](https://github.com/packit/packit/pull/1155) so that tests here can pass.

I want to merge it after release of  [packit/packit#1155](https://github.com/packit/packit/pull/1155). I probably should use WIP prefix for this PR.

packit_service/worker/testing_farm.py
mmuzila: There is no need to check for `fmf_ref`. If it isn't set, `None` is returned and it shouldn't be a problem.

I check for `fmf_url` because I want to use either (project repo + project ref) or (remote repo + remote ref). I don't want to mix them. 

This way if `fmf_ref` is set, but `fmf_url` isn't, tests are taken from project repository and current branch/ref (`fmf_ref` is ignored).

If those parameters were independent it would lead to problems. Eg. if `fmf_url` was set, `fmf_ref` wasn't set and project ref was used instead, testing farm would receive the remote repo url (`fmf_url`) with invalid ref (project repo commit-sha)

I agree that this may be a bit confusing.

packit_service/worker/testing_farm.py
```diff
@@ -68,16 +82,12 @@ def _payload(self, build_id: int, chroot: str) -> dict:
         """
         distro, arch = self.chroot2distro_arch(chroot)
         compose = self.distro2compose(distro)
-        pr = self.project.get_pr(self.metadata.pr_id)
-        # url of the source/fork from which the PR has been created
-        from_url = pr.source_project.get_web_url()
+        fmf = {"url": self.fmf_url, "ref": self.fmf_ref}
```
jpopelka, line 85: Per [TF docs](https://testing-farm.gitlab.io/api/#operation/requestsPost), the "ref" is optional so if `fmf_ref` is not set (and `fmf_url` is set), we should not send it at all, rather than sending `{"ref": None}`.

TomasTomecek (approved):
LGTM, though I'd wait for Jirka's approval as well

TomasTomecek:
recheck

jpopelka (approved):

TomasTomecek:
recheck

jpopelka:
recheck

jpopelka:
>   * [packit-service-tests] : FAILURE in 2m 23s

It's still failing because we are now (since #977) using pre-built tests image instead of building it each time.
The pros are speed (over 2min instead of over 8min), the cons are that if you need to rebuild it because something has changed in packit since last push to this repo, you have to do it yourself in [Actions](https://github.com/packit/packit-service/actions), which is what I did just now.

PR #1023, opened 2021-03-16
Make syncing specfile from downstream work

lbarcziova:


packit_service/config.py
```diff
@@ -34,6 +34,35 @@ class Deployment(enum.Enum):
     prod = "prod"
 
 
+class ProjectToSync(NamedTuple):
+    forge: str
```
TomasTomecek, line 38: a brief docstring would be very nice here - took me a bit to parse this :)

packit_service/worker/handlers/fedmsg_handlers.py
```diff
@@ -124,33 +122,24 @@ def __init__(
...
-                details={
-                    "msg": "URL in specfile is not set. "
-                    "We don't know where the upstream project lives."
-                },
-            )
-
-        n, r = get_namespace_and_repo_name(self.job_config.upstream_project_url)
-        up = self.project.service.get_project(repo=r, namespace=n)
-        self.local_project = LocalProject(
-            git_project=up, working_dir=self.service_config.command_handler_work_dir
```
TomasTomecek, line 144: wow, this look so much simpler now!

packit_service/worker/handlers/fedmsg_handlers.py
```diff
@@ -124,33 +122,24 @@ def __init__(
...
+        # TODO: check if rev is HEAD on {branch}, warn then?
 
-        self.api = PackitAPI(self.service_config, self.job_config, self.local_project)
         self.api.sync_from_downstream(
-            # rev is a commit
-            # we use branch on purpose so we get the latest thing
-            # TODO: check if rev is HEAD on {branch}, warn then?
-            dist_git_branch=self.branch,
-            upstream_branch=None,  # repo's default branch
+            dist_git_branch=self.dg_branch, upstream_branch=self.branch
```
TomasTomecek, line 142: it would be awesome if we could store the result of this action in database.

lbarcziova:
`packit-service-tests : FAILURE in 2m 28s` needs the packit change to be merged

TomasTomecek (approved):
I would give this a go with packit, ogr and requre on stg.

packit_service/worker/handlers/fedmsg_handlers.py
lbarcziova: you mean having some table where we would reference the trigger (dg project with the revision), the upstream project, timestamp and so? should I implement it here or later when we implement storing of the projects to sync in db instead of the service config (and discuss the design of this table on arch? )

lbarcziova:
recheck

packit_service/worker/handlers/fedmsg_handlers.py
TomasTomecek: yup, that's exactly what I had in mind, so that we have "an audit trail" of what p-s did and can have a dashboard of those actions

I think that would blow this PR up a lot and we are not doing that now anyway, so let's implement it in the follow-up work.

+1 discussing on the architecture

PR #1032, opened 2021-03-30
Add fedpkg-stage to worker deps

lbarcziova:
After merging packit/packit-service#773 and other related PRs, we will use `fedpkg-stage` instead of `fedpkg` in staging.
I added it as a dependency here since it will be used only in packit-service for now, not packit itself. WDYT, should I rather move this to packit?


jpopelka (approved):
I'm fine with having it here.

PR #1040, opened 2021-04-01
COPR builds API bug fixes and enhancements

### LINKED ISSUE #982: /api/copr-builds shows same packit_id_per_chroot for all chroots
Hey @lachmanfrantisek , all entries in `packit_id_per_chroot` have the same id.  
https://stg.packit.dev/api/copr-builds?page=1&per_page=10

This is because https://github.com/packit/packit-service/blob/c922f17561d8270fc64dd181ddab16e391555cc1/packit_service/service/api/copr_builds.py#L52 
uses new_id which is just one id out of the ids of the combined models. new_id was created because we needed a unique identifier after merging 3-4 rows into one. 

https://github.com/packit/packit-service/blob/c922f17561d8270fc64dd181ddab16e391555cc1/packit_service/models.py#L682
Here is the complete query which merges three CoprBuildModel rows into one row

```python
    @classmethod
    def get_merged_chroots(
        cls, first: int, last: int
    ) -> Optional[Iterable["CoprBuildModel"]]:
        """Returns a list of unique build ids with merged status, chroots
        Details:
        https://github.com/packit/packit-service/pull/674#discussion_r439819852
        """
            builds = (
                session.query(
                    # We need something to order our merged builds by,
                    # so set new_id to be min(ids of to-be-merged rows)
                    func.min(CoprBuildModel.id).label("new_id"),
                    # Select identical element(s)
                    CoprBuildModel.build_id,
                    # Merge chroots and statuses from different rows into one
                    func.array_agg(psql_array([CoprBuildModel.target])).label("target"),
                    func.array_agg(psql_array([CoprBuildModel.status])).label("status"),
                )
                .group_by(CoprBuildModel.build_id)  # Group by identical element(s)
                .order_by(desc("new_id"))[first:last]
            )

            return builds

```

To get individual packit ids of all chroots, do it analogous to build status like done here
https://github.com/packit/packit-service/blob/c922f17561d8270fc64dd181ddab16e391555cc1/packit_service/models.py#L687
and
https://github.com/packit/packit-service/blob/c922f17561d8270fc64dd181ddab16e391555cc1/packit_service/service/api/copr_builds.py#L51

Just don't label it new_id (call it `id` or `packit_id` ) because thats already being used to identify merged rows.

In #972 you did say

> I do not want to spend much time on the values that require a connection between multiple models because I am finishing the refactoring of the scheme in another PR (#954). I will update/enhance those values once the refactor is merged.

So I'm not sure if your were talking about this so maybe you are already aware of this but I'll make this issue just in case. 

And fixing this is not needed for the results page anyways but this is required while linking the jobs page to the results page

mfocko:
1. Add missing `packit_id` to the list of builds
2. Do not discard `packit_id` of separate chroots

---
Related to packit/dashboard#86
Fixes #982

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -785,6 +785,9 @@ class CoprBuildModel(ProjectAndTriggersConnector, Base):
                     # Merge chroots and statuses from different rows into one
                     func.array_agg(psql_array([CoprBuildModel.target])).label("target"),
                     func.array_agg(psql_array([CoprBuildModel.status])).label("status"),
+                    func.array_agg(psql_array([CoprBuildModel.id])).label(
+                        "packit_id_per_chroot"
+                    ),
                 )
                 .group_by(CoprBuildModel.build_id)  # Group by identical element(s)
                 .order_by(desc("new_id"))[first:last]
--- a/packit_service/service/api/copr_builds.py
+++ b/packit_service/service/api/copr_builds.py
@@ -32,6 +32,7 @@ class CoprBuildsList(Resource):
             build_info = CoprBuildModel.get_by_build_id(build.build_id, None)
             project_info = build_info.get_project()
             build_dict = {
+                "packit_id": build.id,
                 "project": build_info.project_name,
                 "build_id": build.build_id,
                 "status_per_chroot": {},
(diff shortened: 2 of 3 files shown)


lachmanfrantisek (approved):
Some test case would be nice..;)

mfocko:
> Some test case would be nice..;)

Lemme try, but it looks like requre tests O.o

TomasTomecek:
> > Some test case would be nice..;)
> 
> Lemme try, but it looks like requre tests O.o

I'd say you shouldn't use requre itself in the tests: just put some data in the database, retrieve them via API and make sure they are the one you expected :) I believe there should be some tests like this already

PR #1043, opened 2021-04-06
[DistGitCommitEvent] Remove the trigger object in get_dict()

lbarcziova:
As we do in some other tasks, remove the trigger object so that the dict is serializable and task can be sent (we have the db ID of the trigger in the dict).

[Sentry error](https://sentry.io/organizations/red-hat-0p/issues/2310730457/?project=1767823&query=is%3Aunresolved): `Object of type GitBranchModel is not JSON serializable`


jpopelka (approved):
thanks!

PR #1046, opened 2021-04-07
Status before building SRPM

### LINKED ISSUE #1038: "accepted" / "stay tuned" status before SRPM build starts
When asking  @martinpitt and @marusak about their expectations they both suggested it would be nice to have some "accepted" / "stay tuned" status, which would appear "immediately" and be replaced with "Building SRPM" later.

Martin:
In one case, creating the packit statuses in a GitHub PR took unusually long (> 5 mins), which made me nervous and ping Tomas, as usually the statuses get created rather quickly. I think that kind of "I heard you" feedback is quite important.

Matej:
I agree with Martin that packit should be reasonably quick to post status to Github - just to let us know it was recognized and it will be dealt with. This should be measured in "a couple of seconds". I would assume there is some queue where you keep tasks that will be run. If that is the case, then I would say there should be some limit on how much time a task spends in a queue waiting for a free worker. While posting status should take at most a couple of seconds, this should take a maximum of "a couple of minutes" and then the task should be running.

lbarcziova:
- move all checks to pre-check methods of CoprBuildHandler/KojiBuildHandler (so that we don't create handler tasks that won't really run the builds)
- do the precheck in `process_message` task (for this refactor the handlers argument processing), only if pre-check succeeds, set status about accepted task (in case of CoprBuildHandler/KojiBuildHandler) and then create Celery handler task

Fixes #1038

### DIFF
--- a/packit_service/constants.py
+++ b/packit_service/constants.py
@@ -30,6 +30,8 @@ PERMISSIONS_ERROR_WRITE_OR_ADMIN = (
     "can trigger Packit-as-a-Service"
 )
 
+TASK_ACCEPTED = "The task was accepted :+1:"
+
 COPR_SUCC_STATE = "succeeded"
 COPR_FAILED_STATE = "failed"
 COPR_API_SUCC_STATE = 1
--- a/packit_service/worker/jobs.py
+++ b/packit_service/worker/jobs.py
@@ -29,11 +29,14 @@ from typing import List, Set, Type, Union
 
 from celery import group
 
+from ogr.abstract import CommitStatus
 from packit.config import JobConfig, PackageConfig
+from packit_service.constants import TASK_ACCEPTED
 from packit_service.config import ServiceConfig
 from packit_service.log_versions import log_job_versions
 from packit_service.service.events import (
     Event,
+    EventData,
     InstallationEvent,
     IssueCommentEvent,
     IssueCommentGitlabEvent,
@@ -42,10 +45,13 @@ from packit_service.service.events import (
     PullRequestCommentPagureEvent,
     PullRequestLabelPagureEvent,
 )
+from packit_service.worker.build import CoprBuildJobHelper, KojiBuildJobHelper
 from packit_service.worker.handlers import (
+    CoprBuildHandler,
     CoprBuildEndHandler,
     CoprBuildStartHandler,
     GithubAppInstallationHandler,
+    KojiBuildHandler,
     TestingFarmResultsHandler,
 )
 from packit_service.worker.handlers.abstract import (
@@ -251,7 +257,7 @@ class SteveJobs:
             return []
 
         allowlist = Allowlist()
-        job_configs = []
+        processing_results: List[TaskResults] = []
 
         for handler_kls in handler_classes:
             # TODO: merge to to get_handlers_for_event so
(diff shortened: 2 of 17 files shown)


jpopelka (commented):
That was quick, looks OK to me, thanks!

lbarcziova:
rechceck

lbarcziova:
recheck

packit_service/constants.py
```diff
@@ -30,6 +30,8 @@
     "can trigger Packit-as-a-Service"
 )
 
+TASK_ACCEPTED = "The task was accepted."
```
TomasTomecek, line 33:
```suggestion
TASK_ACCEPTED = "The task was accepted 👍"
```
:D am I too crazy?

TomasTomecek (commented):
I can see that you had to switch argument of many methods from EventData to dict - was the reason celery? My point here is that passing dicts is really hard to debug and that it would be nice to have classes with attributes there instead.

packit_service/worker/jobs.py
```diff
@@ -282,24 +288,52 @@ def process_jobs(self, event: Event) -> List[TaskResults]:
...
+                    )
+
+                    job_helper.report_status_to_all(
+                        description=TASK_ACCEPTED,
+                        state=CommitStatus.pending,
+                        url="",
+                    )
+                signatures.append(
+                    handler_kls.get_signature(event=event, job=job_config)
+                )
```
TomasTomecek, line 325: I'm wondering if we could create a celery task chain here: https://docs.celeryproject.org/en/stable/userguide/canvas.html#chains so we can reduce the size of this method

No need to do any changes here right now, just a food for thought.

lbarcziova:
> I can see that you had to switch argument of many methods from EventData to dict - was the reason celery? My point here is that passing dicts is really hard to debug and that it would be nice to have classes with attributes there instead.

I just moved the parsing of the dict to EventData (we won't store only the dict) so that we can create Handler object and call pre-check on the Handler object here in the processing task https://github.com/packit/packit-service/blob/0c8fc450dcb1579cc82f19b04ee09272663f3c2d/packit_service/worker/jobs.py#L295-L302 and don't create useless task. This is not explicitly needed for creating the `accepted` status, but if we do pre-check in the handler tasks, it can happen that e.g. we create `Task was accepted` status, create new Celery handler task and then in handler task we find out that pre-check returns False because the user doesn't have the permissions and we change the status to inform that we won't run the build. 
Currently, we would only inform about the needed permissions and not create the `Task was accepted` status nor the new Celery task.
Should I not include this pre-check change here and really only create the `Task was accepted` status?

TomasTomecek:
> > I can see that you had to switch argument of many methods from EventData to dict - was the reason celery? My point here is that passing dicts is really hard to debug and that it would be nice to have classes with attributes there instead.
> 
> I just moved the parsing of the dict to EventData (we won't store only the dict) so that we can create Handler object and call pre-check on the Handler object here in the processing task
> 
> https://github.com/packit/packit-service/blob/0c8fc450dcb1579cc82f19b04ee09272663f3c2d/packit_service/worker/jobs.py#L295-L302
> and don't create useless task. This is not explicitly needed for creating the `accepted` status, but if we do pre-check in the handler tasks, it can happen that e.g. we create `Task was accepted` status, create new Celery handler task and then in handler task we find out that pre-check returns False because the user doesn't have the permissions and we change the status to inform that we won't run the build.
> Currently, we would only inform about the needed permissions and not create the `Task was accepted` status nor the new Celery task.
> Should I not include this pre-check change here and really only create the `Task was accepted` status?

Thanks for the explanation, it makes sense how you implemented it - I'd leave it as it is.

I'd say that long-term we should figure out a better way to do pre-checks so that we'd move logic from this method there.

TomasTomecek (approved):
let's see this on stg :)

PR #1056, opened 2021-04-19
Dashboard links

mfocko:
TODO
- [x] What are we going to do, if there is no dashboard URL set up?
  Right now it resolves to garbage.
- [x] `test_views.py` that requires local deployment for tests, for now is failing. I do not expect to have dashboard running with it, also it would prolong the time to set up the openshift tests. I suggest dropping it.
- [x] Testing farm

To be merged after:
- [x] Updating configs (already open)
- [x] Updating deployment (already open packit/deployment#200)


jpopelka (approved):
LGTM

> What are we going to do, if there is no dashboard URL set up?

You mean set up in the config, right? I'd consider it a must (to be set up).

TomasTomecek:
> TODO
> 
>     * [ ]  What are we going to do, if there is no dashboard URL set up?
>       Right now it resolves to garbage.

At this point, dashboard should be always up and running: if it's not, it's an outage and we should follow the outage process which is being defined right now: https://github.com/packit/research/pull/87

>     * [ ]  `test_views.py` that requires local deployment for tests, for now is failing. I do not expect to have dashboard running with it, also it would prolong the time to set up the openshift tests. I suggest dropping it.

+1, let's just make sure the URL is correct, no need to check the content

packit_service/service/app.py
```diff
@@ -48,17 +48,19 @@ def get_flask_application():
...
-    s = ServiceConfig.get_service_config()
+    service_config = ServiceConfig.get_service_config()
     # https://flask.palletsprojects.com/en/1.1.x/config/#SERVER_NAME
     # also needs to contain port if it's not 443
-    app.config["SERVER_NAME"] = s.server_name
+    app.config["SERVER_NAME"] = service_config.server_name
     app.config["PREFERRED_URL_SCHEME"] = "https"
     if getenv("DEPLOYMENT") in ("dev", "stg"):
         app.config["DEBUG"] = True
+    app.config["DASHBOARD"] = service_config.dashboard_url
```
TomasTomecek, line 58: nit: I'd name it DASHBOARD_URL :)

TomasTomecek (approved):
wow, very nice!!

packit_service/service/app.py
mfocko: thx for noticing, I can remove it, it is loaded in `service.urls` as global constant :D this one is no longer needed

packit_service/service/urls.py
```diff
@@ -20,49 +20,47 @@
...
+    Creates a dashboard URL that can be used for setting commit status or redirect.
+
+    Args:
+        build (str): Type of build, e.g. `srpm`, `copr` or `koji`.
+        id_ (int): Packit ID of the build.
+
+    Returns:
+        URL to the results of `id_` build of kind `build`.
+    """
+    return f"{DASHBOARD_URL}/results/{build}-builds/{id_}"
```
lachmanfrantisek, line 39: What about `build_type`?

```suggestion
def _get_dashboard_url(build_type: str, id_: int) -> str:
    """
    Creates a dashboard URL that can be used for setting commit status or redirect.

    Args:
        build_type (str): Type of build, e.g. `srpm`, `copr` or `koji`.
        id_ (int): Packit ID of the build.

    Returns:
        URL to the results of `id_` build of kind `build`.
    """
    return f"{DASHBOARD_URL}/results/{build_type}-builds/{id_}"
```

packit_service/service/urls.py
```diff
@@ -20,49 +20,47 @@
...
 
-from flask import url_for
+from packit_service.config import ServiceConfig
 
-from packit_service.service.app import packit_as_a_service as application
+DASHBOARD_URL = ServiceConfig().get_service_config().dashboard_url
 
 
-def get_srpm_log_url_from_flask(
+def _get_dashboard_url(build: str, id_: int) -> str:
```
lachmanfrantisek, line 28: What about something like `_get_dashboard_result_url` or something like that?

(I expect DASHBOARD_URL as a return value of a function with this name..;)

packit_service/service/views.py
```diff
@@ -23,91 +23,38 @@
...
-        srpm_build=srpm_build,
-        build_submitted_time=pretty_time(build.build_submitted_time),
-        srpm_submitted_time=pretty_time(srpm_build.build_submitted_time)
-        if srpm_build
-        else None,
-        owner=build.owner if hasattr(build, "owner") else None,  # type: ignore
-        project_name=build.project_name if hasattr(build, "project_name") else None,  # type: ignore
-        is_pr=trigger.job_trigger_model_type == JobTriggerModelType.pull_request,
-    )
+    return redirect(get_srpm_build_info_url(id_), code=308)
```
lachmanfrantisek, line 40: So easy? Nice!

lachmanfrantisek (approved):
Well done!

Just two nit picks..;)

packit_service/service/urls.py
mfocko: Yeah, I looked into testing farm, not sure if keep it, or generalize for TF too

lachmanfrantisek: Good point. Since it's mostly the same thing, I would generalize for TF as well.

mfocko:
should be done from my side, if there are no remarks, I would rebase some of the commits and redeploy stage after merge

PR #1059, opened 2021-04-21
Set up 2 separate Celery queues for long and short running tasks

lbarcziova:


files/run_worker.sh
```diff
@@ -16,6 +16,12 @@ else
   LOGLEVEL="DEBUG"
 fi
 
+if [[ ${HOSTNAME} == "packit-worker-1" ]]; then
+  QUEUES="short-running"
+else
+  QUEUES="short-running,long-running"
+fi
```
lachmanfrantisek, line 23: What about stage deployment with only one worker?

lachmanfrantisek: Aha, we are indexing from `0`..;)

lachmanfrantisek (approved):

files/run_worker.sh
lachmanfrantisek: Maybe not necessary but maybe, we can have an env. variable with the number of short-running-queue workers so we can edit the proportion more easily. But this will be more relevant once we have a different image for different types of workers and that can be done on the openshift level and not via env. var.

jpopelka (approved):

packit_service/celerizer.py
```diff
@@ -59,6 +59,8 @@ def celery_app(self):
             days = int(getenv("CELERY_RESULT_EXPIRES", "30"))
             # https://docs.celeryproject.org/en/latest/userguide/configuration.html#result-expires
             self._celery_app.conf.result_expires = timedelta(days=days)
+            # https://docs.celeryproject.org/en/latest/userguide/configuration.html#std-setting-task_default_queue
+            self._celery_app.conf.task_default_queue = "short-running"
```
TomasTomecek, line 63: small nit to make the string a constant

TomasTomecek (commented):
I'd say it's essential to test this on stg and am wondering how to do that

files/run_worker.sh
lbarcziova: not so nice solution could be to have env vars e.g. `SHORT_RUNNING`=x, `LONG_RUNNING`=y (so currently we would have `SHORT_RUNNING`=1, `LONG_RUNNING`=0), `packit-worker-0` would always serve both queues and then `x` workers beginning from `packit-worker-1` would serve the short running queue and `y` workers beginning from `packit-worker-{x+1}` would serve only long-running queue (in this way we would be able to set `QUEUES` in the `run_worker` script depending on the `HOSTNAME`)...is this something that you had in mind? we would need to make sure that `SHORT_RUNNING`+`LONG_RUNNING`+1 = number of workers, or schedule the number of workers according to this calculation

lachmanfrantisek: @lbarcziova Yes, that's what I had in mind.

But, I am not sure if this is needed for now:

* If we want to differentiate images for different types of workers (what I think we've discussed), we can leave that work for later. (Scaling will be done on openshift level, not worker code level.)
* If we don't have such a change in mind,  having `SHORT_RUNNING`/`LONG_RUNNING` env. variables will be nice.

lbarcziova: I implemented it in the way described in the comment, PTAL

files/run_worker.sh
```diff
@@ -16,6 +16,22 @@ else
...
+
+WORKER_NUMBER="$(cut -d'-' -f3 <<< ${HOSTNAME})"
+
+if [ "$WORKER_NUMBER" -ge 1 ] && [ "$WORKER_NUMBER" -lt $((1 + SHORT_RUNNING)) ]; then
+  QUEUES="short-running"
+elif [ "$WORKER_NUMBER" -ge $((1 + SHORT_RUNNING)) ] && [ "$WORKER_NUMBER" -lt $((1 + SHORT_RUNNING + LONG_RUNNING)) ]; then
+  QUEUES="long-running"
+else
+  QUEUES="short-running,long-running"
+fi
```
lachmanfrantisek, line 33: Nice! I hope you've tested those conditionals..;)

lachmanfrantisek (approved):

files/run_worker.sh
jpopelka: Could you add a comment translating the condition into human language? Thanks

csomh: I see this env var as configuration so why not just define `QUEUES` in the [deployment](https://github.com/packit/deployment/blob/900fbd85a2529e3c8d569bf25f1e821d033d82c7/openshift/packit-worker.yml.j2#L75) instead of calculating it here?

You could do this by splitting the current `worker_replicas` var in the deployment playbook into `workers_long_tasks` and `workers_all_tasks` (or some better names if you have an idea :smiley:), with defaults `workers_long_tasks: 0` and `workers_all_tasks: 1`, and decide based on this whether to run [a second task to deploy](https://github.com/packit/deployment/blob/900fbd85a2529e3c8d569bf25f1e821d033d82c7/playbooks/deploy.yml#L172) a set of workers with the value of `QUEUES` set only to `long-running`.

The end result in prod would be that there would be two sets of stateful sets deployed for workers each with the specified number of replicas.

lbarcziova: I was thinking about this option, but if we would have one set of workers for both queues and then one set of workers for short-running tasks and another one for long-running tasks, we would have 3 sets of stateful sets and the only difference between them would be that one variable...it made better sense to me to have just one stateful set and not divide it, since it will create duplication and basically all would do the same thing. On the other hand, this would make things more straightforward, so I am really not sure about it.
What do others think?

TomasTomecek: I'd say both solutions have pros and cons: we should discuss this tomorrow on the architecture. I'm wondering if we could have the mapping definition b/w workers and queues in a file as a configuration so we wouldn't need to rebuild the image if we wanted to change the setup and at the same time, have only a single stateful set.

TomasTomecek (approved):

csomh (approved):
Thanks!

PR #328, opened 2020-01-13
Report invalid or missing packit config

dhodovsk:
Requires https://github.com/packit-service/packit/pull/664
Fixes https://github.com/packit-service/packit-service/issues/222

PR #1064, opened 2021-04-22
Deprecate packit options and create issue if needed

mfocko:
As we agreed on deprecating:
- `current_version_command`
- `create_tarball_command`

Refactor creating of the issues on the projects in case of package config
problems.

Use issue filtering based on authenticated user.

Create issues in case deprecated options are being used.

### DIFF
--- a/packit_service/config.py
+++ b/packit_service/config.py
@@ -8,7 +8,7 @@ from typing import List, Optional, Set, Union, NamedTuple
 
 from yaml import safe_load
 
-from ogr.abstract import GitProject
+from ogr.abstract import GitProject, Issue
 from packit.config import (
     Config,
     PackageConfig,
@@ -225,6 +225,23 @@ class ServiceConfig(Config):
 
 class PackageConfigGetter:
     @staticmethod
+    def create_issue_if_needed(
+        project: GitProject, message: str, details: str = ""
+    ) -> Optional[Issue]:
+        issues = project.get_issue_list(author=project.service.user.get_username())
+
+        if "[packit] Invalid config" in (x.title for x in issues):
+            return None
+
+        if details:
+            details = f": {details}"
+
+        # TODO: store in DB
+        return project.create_issue(
+            title=f"[packit] Invalid config{details}", body=message
+        )
+
+    @staticmethod
     def get_package_config_from_repo(
         project: GitProject,
         reference: Optional[str] = None,
@@ -232,7 +249,7 @@ class PackageConfigGetter:
         pr_id: int = None,
         fail_when_missing: bool = True,
         spec_file_path: Optional[str] = None,
-    ):
+    ) -> Optional[PackageConfig]:
         """
         Get the package config and catch the invalid config scenario and possibly no-config scenario
         """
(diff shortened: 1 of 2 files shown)


packit_service/config.py
```diff
@@ -224,6 +224,23 @@ def get_project_to_sync(self, dg_repo_name, dg_branch) -> Optional[ProjectToSync
 
 
 class PackageConfigGetter:
+    @staticmethod
+    def create_issue_if_needed(
+        project: GitProject, message: str, details: str = ""
+    ) -> Optional[Issue]:
+        issues = project.get_issue_list(author=project.service.user.get_username())
+
+        if "[packit] Invalid config" in (x.title for x in issues):
```
lachmanfrantisek, line 233: We don't have any filtering? (That would be handy.)

packit_service/config.py
```diff
@@ -252,27 +269,56 @@ def get_package_config_from_repo(
...
+                "https://packit.dev/packit-as-a-service/ or contact us - "
+                "[Packit team]"
+                "(https://github.com/orgs/packit/teams/the-packit-team)"
+            )
+
+            if pr_id:
+                project.pr_comment(pr_id, message)
+            elif created_issue := PackageConfigGetter.create_issue_if_needed(
+                project, message
+            ):
```
lachmanfrantisek, line 319: :snake: :rocket:

packit_service/config.py
```diff
@@ -224,6 +224,23 @@ def get_project_to_sync(self, dg_repo_name, dg_branch) -> Optional[ProjectToSync
...
+    ) -> Optional[Issue]:
+        issues = project.get_issue_list(author=project.service.user.get_username())
+
+        if "[packit] Invalid config" in (x.title for x in issues):
+            return None
+
+        if details:
+            details = f": {details}"
+
+        # TODO: store in DB
```
lachmanfrantisek, line 239: Do you want to create just a model, or something more sophisticated?

lachmanfrantisek (approved):
Well done! Looks good to me.

packit_service/config.py
mfocko: I am not sure about the benefits vs space complexity for this o.o
I guess @lbarcziova can provide better answer, since git blame blames her ;) I cannot seem to find any related issue, so I don't know what were the intentions :/

packit_service/config.py
mfocko: only author/assignee

packit_service/config.py
lbarcziova: If you look deeper, the TODO comment was added [here](https://github.com/packit/packit-service/commit/7361bf92e03f2d0a7ba62fc80025d337187d258b), so I am not the right person to answer..but I looked into that PR and this comment was probably added based on [this](https://github.com/packit/packit-service/pull/328#discussion_r366765240)

mfocko: thanks :)

packit_service/config.py
mfocko: thinking about it, `get_issue_list` fetches `open` issues by default, and if we filter it by `packit` user, it shouldn't be that big list 🤔

packit_service/config.py
lachmanfrantisek: OK, it looks like Tomáš has wanted to store it to avoid the issue searching. And as Maťo said, we are filtering it by `packit` user, it's not such a big problem.

PR #1066, opened 2021-04-23
Update/sync license boilerplates

jpopelka:
It's Friday ;)

### DIFF
--- a/tests/unit/test_parser.py
+++ b/tests/unit/test_parser.py
@@ -1,5 +1,6 @@
 # Copyright Contributors to the Packit project.
 # SPDX-License-Identifier: MIT
+
 from packit_service.models import TestingFarmResult
 from packit_service.service.events import TestResult
 from packit_service.worker.parser import Parser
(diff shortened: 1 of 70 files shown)


tests/unit/test_parser.py
```diff
@@ -1,5 +1,6 @@
 # Copyright Contributors to the Packit project.
 # SPDX-License-Identifier: MIT
+
```
mfocko, line 3: will we ever match on 0-2 empty lines after license? :D

mfocko (commented):
if you want to be consistent, there are no license headers in `tests_requre/.*/__init__.py`; not a blocker for me though :)

tests/unit/test_parser.py
jpopelka: thanks

> there are no license headers in tests_requre/.*/__init__.py

yeah, I couldn't decide whether to remove from all `__init__.py` or add to all, so I let it as it is :-)

> will we ever match on 0-2 empty lines after license? :D

:-) should be 1 line everywhere now

mfocko (approved):

PR #1068, opened 2021-04-25
Revert filtering of issues created by Packit

mfocko:
GitHub App does not have PyGithub instance, thus we cannot get username.
GitLab does not have `app/` prefix => cannot be guessed from deployment.

Related to: https://sentry.io/organizations/red-hat-0p/issues/2361275931/


mfocko:
recheck

mfocko:
recheck

lachmanfrantisek (approved):
Thanks!

PR #1072, opened 2021-04-28
Add run_ids to API and extend TF endpoint

mfocko:
- Add run IDs to endpoints which are related
- Add COPR build ID to testing farm endpoint


lachmanfrantisek (approved):
Nice! Thanks!

Will this work on prod data from the memory perspective?

mfocko:
> Will this work on prod data from the memory perspective?

It is done only for the requests with specific ID, so we do not return all run IDs for tables.

jpopelka (approved):

PR #1087, opened 2021-05-04
WIP: Create a Bugzilla bug for a new Merge Request (Gitlab)

jpopelka:
We already have the functionality, but only for Pagure. This makes it work on Gitlab.

There are some refactoring commits, so it's better to check them separately.


packit_service/worker/handlers/__init__.py
```diff
@@ -16,7 +16,7 @@
     FedmsgHandler,
     DistGitCommitHandler,
 )
-from packit_service.worker.handlers.github_handlers import (
+from packit_service.worker.handlers.forges_handlers import (
```
lachmanfrantisek, line 19: :tada: Thanks!

lachmanfrantisek: I am only thinking about the new name -- by now, handlers are forge-independent and can be grouped by the focus. E.g. I like that you have moved the `TestingFarmHandler` to other TF handlers.

(`forges_handlers` now does not make much sense)

packit_service/worker/handlers/forges_handlers.py
```diff
@@ -384,9 +384,9 @@ def pre_check(self) -> bool:
         return True
 
 
-@reacts_to(event=PullRequestLabelPagureEvent)
-class PagurePullRequestLabelHandler(JobHandler):
-    task_name = TaskName.pagure_pr_label
+@reacts_to(MergeRequestLabelGitlabEvent)
+class PullRequestLabelHandler(JobHandler):
+    task_name = TaskName.pr_label
```
lachmanfrantisek: Do we want to preserve the old functionality? (=Do we want to remove the old `reacts_to` decorator?)

lachmanfrantisek (commented):
Nice, thanks!

packit_service/worker/handlers/__init__.py
jpopelka: Ok, let me know if you figure out something better than `forges_handlers` :-)

packit_service/worker/handlers/forges_handlers.py
jpopelka: My thinking is that this create-bug-from-mr feature is tightly coupled with c8s source-git 2 dist-git workflow.
Since we've already moved c8s source-git repos from CentOS Pagure to Gitlab, there's in my POV, minimal chance that we will ever need this also for Pagure.
And I'm a fan of cleaning unused code at the moment I stumble upon it as one can always dig it from history.

lachmanfrantisek: Yes, I agree. I just wanted to be sure that it's what you want and that we can support multiple events.

packit_service/worker/handlers/__init__.py
lachmanfrantisek: (can be done later/elsewhere)

What about splitting the handlers by purpose -- `copr_build_handlers`/`koji_build_handlers`/`testing_farm_handlers`/...?

(We just need to be sure that the file with handler is always imported to trigger the decorators so we don't ignore them.)

lachmanfrantisek: Thanks!

lachmanfrantisek (approved):

lachmanfrantisek (approved):
Sorry for happily reviewing the label-version of this workflow. I've thought the same.

Thanks for all the work!

lachmanfrantisek:
Do you think we will ever need to react on labels? (I am just thinking about whether it makes sense to preserve the parsing you've implemented and is now not present?)

TomasTomecek:
> Do you think we will ever need to react on labels? (I am just thinking about whether it makes sense to preserve the parsing you've implemented and is now not present?)

I can easily imagine a future workflow where p-s, or probably "the-new-thing", will react to labels.

jpopelka:
> I am just thinking about whether it makes sense to preserve the parsing you've implemented and is now not present?

It's gone, sorry :-)
One can dig it out from [history](https://github.com/packit/packit-service/commit/d1613eae17a4ffaaf1ad9fc18c57e07f1865ed82) if we ever need it in the future.

Thanks for review, will merge this after we move stable branches today.

PR #1090, opened 2021-05-12
Gitlab: allow webhook token for whole group/namespace

jpopelka:
With Group Hooks (Gitlab EE only) we can setup a webhook for whole group/namespace.
In that case the same token would be sent with events from each repo, so the token can't be encoded with a repo name, only with the namespace.

When we receive/decode the token we now accept tokens without repo_name as well.

### DIFF
--- a/packit_service/service/api/webhooks.py
+++ b/packit_service/service/api/webhooks.py
@@ -177,8 +177,13 @@ class GitlabWebhook(Resource):
         )
 
         if not project_authentication_issue:
-            token = jwt.encode(
+            token_project = jwt.encode(
                 {"namespace": parsed_url.namespace, "repo_name": parsed_url.repo},
+                config.gitlab_token_secret,
+                algorithm="HS256",
+            ).decode("utf-8")
+            token_group = jwt.encode(
+                {"namespace": parsed_url.namespace},
                 config.gitlab_token_secret,
                 algorithm="HS256",
             ).decode("utf-8")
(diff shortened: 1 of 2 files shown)


packit_service/service/api/webhooks.py
```diff
@@ -230,7 +240,7 @@ def validate_token(self):
         token = request.headers["X-Gitlab-Token"]
 
         if token in config.gitlab_webhook_tokens:
-            logger.debug("Deprecation Warning: Old Gitlab tokens used.")
+            logger.debug("Deprecation Warning: Old/static Gitlab tokens used.")
```
lachmanfrantisek, line 243: Do we want to get rid of those "static" ones?

I treat them more like a different use-case, but:
* After this PR, I am not sure if we can find some use-case.
* Looks like having two ways is confusing to people.

lachmanfrantisek (approved):
So easy? Nice!

lachmanfrantisek (approved):
Some test would be nice.

*/edit: Wanted to use `Request changes` but hit the wrong button..;) /*

packit_service/service/api/webhooks.py
jpopelka: I'd keep it there for some more time until we prove that this new group webhook tokens work OK.

jpopelka:
> Wanted to use `Request changes` but hit the wrong button..;)

Please re-review then :-)

lachmanfrantisek:
> > 
> 
> Please re-review then :-)

Sorry, I've missed the notification. Sorry. (LGTM)

PR #1106, opened 2021-05-27
Move #packit to Libera.Chat

csomh:

### DIFF
--- a/packit_service/worker/build/build_helper.py
+++ b/packit_service/worker/build/build_helper.py
@@ -383,7 +383,7 @@ class BaseBuildJobHelper:
 
             srpm_logs += (
                 f"\nMessage: {exception}\nException: {exception!r}\n{self.msg_retrigger}"
-                "\nPlease join the freenode IRC channel #packit for the latest info.\n"
+                "\nPlease join #packit on irc.libera.chat if you need help with the error above.\n"
             )
 
         self._srpm_model, self.run_model = SRPMBuildModel.create_with_new_run(


packit_service/worker/build/build_helper.py
```diff
@@ -383,7 +383,7 @@ def _create_srpm(self):
 
             srpm_logs += (
                 f"\nMessage: {exception}\nException: {exception!r}\n{self.msg_retrigger}"
-                "\nPlease join the freenode IRC channel #packit for the latest info.\n"
+                "\nPlease join #packit on irc.libera.chat for the latest info.\n"
```
csomh, line 386: @lachmanfrantisek This message seems a little bit vague to me (ie: latest info on what?). What was the original intention here? I'm looking for a way to improve this.

TomasTomecek: I guess the intention was to get the latest information about the current status of the deployment: check /topic in the channel or latest messages. Mainly to know if there are any issues in the production env and the ETA to resolve them.

csomh: I've updated this to "Please join #packit on irc.libera.chat if you need help with the error above.", which covers the service-status aspect, but also more, and maybe it's less confusing. Let me know what you think about it!

TomasTomecek (approved):
Nice and clear!

packit_service/worker/build/build_helper.py
lachmanfrantisek: @csomh Sorry, I've missed the notification.

I'm worried I'm not the original author of that message (price for refactoring..;), but agree with your improvement.

PR #1114, opened 2021-06-07
Fix the _get_run_item method in models.py

lbarcziova:
We get the `RunModel` objects with `self.get_runs()` and so we can get the builds/test models directly from them (and remove None values in the end if the run doesn't have a corresponding object). Before, the query to get the corresponding model was not correct.


packit_service/models.py
```diff
@@ -129,9 +129,22 @@ def get_runs(self) -> List["RunModel"]:
...
+        if model == KojiBuildModel:
+            models = [run.koji_build for run in runs]
+
+        if model == SRPMBuildModel:
+            models = [run.srpm_build for run in runs]
+
+        if model == TFTTestRunModel:
+            models = [run.test_run for run in runs]
+
+        return [model for model in models if model is not None]
```
lachmanfrantisek, line 147: Can we use a different name, please? So we don't collide with the argument...

lachmanfrantisek, line 147: What about using `set` or `list(set(` to get rid of the duplicates?

packit_service/models.py
```diff
@@ -129,9 +129,22 @@ def get_runs(self) -> List["RunModel"]:
             return trigger_list[0].runs if trigger_list else []
 
     def _get_run_item(self, model):
-        run_ids = [run.id for run in self.get_runs()]
-        with get_sa_session() as session:
-            return session.query(model).filter(RunModel.id.in_(run_ids)).all()
```
lachmanfrantisek, line 134: It's some time I was playing with that. Can you please describe what actually was wrong? And ideally, create a test case for that?

packit_service/models.py
lachmanfrantisek: And looking at that, can you please add the type annotation of that function? (Probably my bad, but after some time, it took me a while to get what the `model` argument is about -- maybe, we can rename it to `model_type` or `model_class` to make it obvious.)

packit_service/models.py
lbarcziova: It is covered by the extension of the existing test..before the changes, after adding other copr build triggered by branch push to the db in the test, the `pr.get_copr_builds()` returned a list not only with the 1 correct copr build object, but also with this added copr build triggered by branch push, because all the run models were iterated when filtering (I assume therefore the API endpoint was dead, since this needed a lot of memory).

lachmanfrantisek: When looking at it now, I am really not sure what it does and how comes it works. (There is no single `RunModel.id` property for build/test model.)

lachmanfrantisek (approved):
Thanks for the explanation, LGTM

packit_service/models.py
lbarcziova: It worked in tests only because there was only one copr build object in the DB during the test, after adding more it didn't behave correctly.

PR #1126, opened 2021-06-17
Implement different mapping for internal TF

lachmanfrantisek:
The /composes endpoint is not valid for internal TF.
This will be improved in the upcoming version of TF,
but for now, we need to hardcode the mapping.

It works like this:
* `Fedora-3x` -> `Fedora-3x-Updated`
* `CentOS-x` ->  `CentOS-x-latest`
* `CentOS-Stream-8` -> `RHEL-8.5.0-Nightly`

@thrix @psss You might be interested in this.

The last bit is epel and here are the possibilities:

upstream:
* `epel-7` -> `centos-7` distro and  `CentOS-7` compose
* `epel-8` -> `centos-8` distro and  `CentOS-8` compose

for internal, we can do the same:
* `epel-7` -> `centos-7` distro a  `CentOS-7-latest` compose
* `epel-8` -> `centos-8` distro a  `CentOS-8-latest` compose

Or:
* `epel-7` -> ??? /We can use `centos-7 ` but that's not consistent /
* `epel-8` -> `rhel-8.5` distro a `RHEL-8.5.0-Nightly` compose

### DIFF
--- a/packit_service/worker/testing_farm.py
+++ b/packit_service/worker/testing_farm.py
@@ -195,18 +195,37 @@ class TestingFarmJobHelper(CoprBuildJobHelper):
         return distro, arch
 
     def distro2compose(self, distro: str) -> str:
-        """Create a compose string from distro, e.g. fedora-33 -> Fedora-33
-        https://api.dev.testing-farm.io/v0.1/composes"""
+        """
+        Create a compose string from distro, e.g. fedora-33 -> Fedora-33
+        https://api.dev.testing-farm.io/v0.1/composes
+
+        The internal TF has a different set and behaves differently:
+        * Fedora-3x -> Fedora-3x-Updated
+        * CentOS-x ->  CentOS-x-latest
+        * CentOS-Stream-8 -> RHEL-8.5.0-Nightly
+        """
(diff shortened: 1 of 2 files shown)


TomasTomecek:
Franto, very nice description!

I'd suggest thing mapping for EL:

| x |Internal|Public|
|---|---|---|
|el7| RHEL 7*   | CentOS Linux 7 |
|el8| RHEL 8.5 | CentOS Stream 8 |

*Since el7 is quite old at this point, I don't expect much usage here though we should really strive for RHEL 7 internally and nothing else.

lachmanfrantisek:
OK, I've updated the code to match the following schema:
    
upstream:
* `epel-7` -> distro=`centos-7` compose=`CentOS-7`
* `epel-8` -> distro=`centos-stream-8` compose=`CentOS-Stream-8`
    
internal:
* `epel-7` -> distro=`rhel-7` compose=`RHEL-7-LatestReleased`
* `epel-8` -> distro=`rhel-8` compose=`RHEL-8.5.0-Nightly`

TomasTomecek (approved):

jpopelka (commented):
:+1:

Issue #1036, opened 2021-03-31
Service sometimes does not react to /packit test


lachmanfrantisek:
@psss has encountered multiple times that the service did nothing after he put `/packit test` comment. (Adding another comment helped.)

Example PRs:
* https://github.com/psss/tmt/pull/676
* https://github.com/psss/fmf/pull/124

TODO:
* [ ] Try reproducing the same behaviour.
* [ ] Check the content of the database.
* [ ] Check the logs/sentry.
* [ ] Find and fix the problem.

lachmanfrantisek:
Note that there is only one build for PR 124 and it failed on SRPM (because of the quota).

lbarcziova:
> Note that there is only one build for PR 124 and it failed on SRPM (because of the quota).

That may explain the failure of the task (Sentry issue says that it wasn't successful). I haven't found relevant logs, but assuming from the info in issue, it probably failed because there is [no latest copr build](https://github.com/packit/packit-service/blob/92d9a6f80cc44216c8c3c11dfa968e78e46124e7/packit_service/worker/testing_farm.py#L214), in that case (when tests are triggered by comment), we do not run the needed build for tests (but I'm not sure whether this is the expected behaviour)

lachmanfrantisek:
> (when tests are triggered by comment), we do not run the needed build for tests (but I'm not sure whether this is the expected behaviour)

That's a good question. When the build is only some internal step to run the test (=only the test is defined), the build should be retriggered I would say.

I am not sure about the other variant (=test and build is defined). On one side, they will lose results from the previous build, on the other side, rerunning the build is the only thing that can help. (Otherwise, they need to push new changes.)

@psss What do you think about that?

psss:
> When the build is only some internal step to run the test (=only the test is defined), the build should be retriggered I would say.

Make sense to me.

lachmanfrantisek:
So to conlcude this:

- [ ] The task here is to make sure that we run the build for `/packit test` when there is no build for HEAD commit or there is no successful target for the last build of the HEAD commit.

@lbarcziova @psss Does it make sense?

psss:
Yeah, sounds good.

mmuzila:
> So to conlcude this:
> 
>     * [ ]  The task here is to make sure that we run the build for `/packit test` when there is no build for HEAD commit or there is no successful target for the last build of the HEAD commit.

[...]

PR #1132, opened 2021-06-22
Make TF to trigger Copr build if no suitable build is found

mmuzila:
Resolves: #1036

### DIFF
--- a/packit_service/worker/handlers/testing_farm.py
+++ b/packit_service/worker/handlers/testing_farm.py
@@ -7,6 +7,7 @@ This file defines classes for job handlers specific for Testing farm
 import logging
 from typing import Optional
 
+from celery import signature
 from ogr import GitlabService
 from ogr.abstract import CommitStatus
 from packit.config import JobConfig, JobType
@@ -36,8 +37,7 @@ from packit_service.worker.reporting import StatusReporter
 from packit_service.worker.result import TaskResults
 from packit_service.worker.testing_farm import TestingFarmJobHelper
 from packit_service.constants import PG_COPR_BUILD_STATUS_SUCCESS
-from packit_service.utils import dump_job_config
-from packit_service.worker.build import CoprBuildJobHelper
+from packit_service.utils import dump_job_config, dump_package_config
 
 logger = logging.getLogger(__name__)
 
@@ -119,21 +119,21 @@ class TestingFarmHandler(JobHandler):
             logger.info("No suitable copr-build found, run copr build.")
 
             result_details = {
-                "msg": "Build required, triggering the build",
+                "msg": "Build required, triggering copr build",
                 "event": self.data,
                 "package_config": self.package_config,
                 "job": self.job_config.type.value if self.job_config else None,
                 "job_config": dump_job_config(self.job_config),
             }
 
-            CoprBuildJobHelper(
-                self.service_config,
-                self.package_config,
-                self.project,
-                self.data,
-                self.db_trigger,
-                self.job_config,
-            ).run_copr_build()
+            signature(
+                TaskName.copr_build.value,
+                kwargs={
+                    "package_config": dump_package_config(self.package_config),
+                    "job_config": dump_job_config(self.job_config),
+                    "event": self.data.get_dict(),
+                },
+            ).apply_async()
 
             return TaskResults(success=True, details=result_details)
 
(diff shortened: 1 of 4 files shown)


packit_service/worker/handlers/testing_farm_handlers.py
```diff
@@ -102,6 +105,35 @@ def run(self) -> TaskResults:
...
+            }
+
+            CoprBuildJobHelper(
+                self.service_config,
+                self.package_config,
+                self.project,
+                self.data,
+                self.db_trigger,
+                self.job_config,
+            ).run_copr_build()
```
lachmanfrantisek, line 133: Just thinking out loud, what about creating a celery task? But on the other hand, this is a really simple solution,

packit_service/worker/handlers/testing_farm_handlers.py
```diff
@@ -102,6 +105,35 @@ def run(self) -> TaskResults:
...
+
+            return TaskResults(success=True, details=result_details)
```
lachmanfrantisek, line 135: This is tricky:
* `CoprBuildJobHelper.run_copr_build` returns `TaskResult` so we might want to reuse it (and maybe enhance to mention why this has been triggered).
* `CoprBuildJobHelper.run_copr_build` can fail, or in other words, it can return `TaskResult` with `success=False`.

lachmanfrantisek (commented):
Looking good so far!

mmuzila:
@lachmanfrantisek, I have problems with tests. I'm unable to make them ([1](https://github.com/packit/packit-service/blob/6ca80e34c6cad130fd4eadcc281844497be40611/tests/integration/test_listen_to_fedmsg.py#L349),[2](https://github.com/packit/packit-service/blob/6ca80e34c6cad130fd4eadcc281844497be40611/tests/integration/test_listen_to_fedmsg.py#L646),[3](https://github.com/packit/packit-service/blob/6ca80e34c6cad130fd4eadcc281844497be40611/tests/integration/test_listen_to_fedmsg.py#L532)) work. The source of problems is [this line](https://github.com/packit/packit-service/blob/6ca80e34c6cad130fd4eadcc281844497be40611/tests/integration/test_listen_to_fedmsg.py#L317) - status of copr build is not actually set to `success`, thus the new part of code (triggering new copr build) is executed even when it shouldn't be.

I tried to work around this (tried to set it to `success`; tried to redefine `copr_build_pr` in order to better resemble `CoprBuildModel`; ...), but with no success.

packit_service/worker/handlers/testing_farm_handlers.py
mmuzila: Based on our [previous discussion](https://github.com/packit/packit-service/issues/1036#issuecomment-853743715), I assumed that triggering a new copr build and then return `TaskResult` with `success=True` is ok.

But maybe it'd be better for TF to return `TaskResult` with `success=False` always when triggering new copr build.
After all, TF was not successfull.

I'm still not sure how does this work as a whole. Is TF triggered when copr build  finishes successfully? If yes, then after the new copr build is ready second instance of TF is triggered and it should find appropriate copr build, run tests and return proper result. If the copr build fails, second instance of TF is not triggered - having successful result from the first instance of TF  probably isn't what we want.

Does the above make at least a bit of sense, or I'm completely lost?

lachmanfrantisek: > Does the above make at least a bit of sense, or I'm completely lost?

No, not at all..;)

> s TF triggered when copr build finishes successfully? If yes, then after the new copr build is ready second instance of TF is triggered and it should find appropriate copr build, run tests and return proper result.

Yes, exactly.

> If the copr build fails, second instance of TF is not triggered - having successful result from the first instance of TF probably isn't what we want.

Here, we are speaking about two things -- task result (=status for service) and build result. (=status for user).
I previously thought that we will not submit the actual build, but add it to the queue -- so we would be able to return successful task result (and do nothing else).

Here, I would return the same result we get from `CoprBuildJobHelper.run_copr_build` and attach some note about the reason.

lachmanfrantisek:
@mmuzila I've managed to fix your problems with tests -- can I push the changes to this PR?

mmuzila:
> @mmuzila I've managed to fix your problems with tests -- can I push the changes to this PR?

Yes, you can, thank you.

mmuzila:
recheck

mmuzila:
recheck

jpopelka:
I meanwhile renamed `handlers/testing_farm_handlers.py` to just `testing_farm.py` in 4d96aff827
Hope it doesn't cause you too much pain, sorry.

TomasTomecek (approved):
LGTM, let's try this on stg :)

TomasTomecek:
regate

PR #54, opened 2019-07-18
Refactor jobs

rpitonak:
Tested locally. I will be glad for comments on design.

TODO

- [x] fix tests

PR #1144, opened 2021-06-30
Move events from service/ to worker/

jpopelka:
The events were initially created in #54 by Rado,
so I can't ask why they were added to service/ while parser was added to worker/.

But I don't see why they should be in service/ when they're not used anywhere there.


jpopelka:
@lachmanfrantisek @lbarcziova you probably know that code best, let me know what do you think, thanks.

jpopelka:
recheck

lbarcziova (approved):
+1, it should be in worker/, thanks!

lachmanfrantisek (commented):
:+1:

PR #1146, opened 2021-07-01
Switch from error to debug logging in allowlist

mfocko:
Switch from logging action of not approved user from error to debug as
it is not a vital issue from our side and thus doesn't need to be tracked
in Sentry.


lachmanfrantisek (approved):

jpopelka:
`error` -> `debug` is quite a big jump, I'd probably use `warning`/`info`, but never mind :-)

PR #1147, opened 2021-07-01
Log projects for internal TF

lachmanfrantisek:

### DIFF
--- a/packit_service/worker/testing_farm.py
+++ b/packit_service/worker/testing_farm.py
@@ -316,6 +316,12 @@ class TestingFarmJobHelper(CoprBuildJobHelper):
             and f"{self.project.service.hostname}/{self.project.full_repo_name}"
             not in self.service_config.enabled_projects_for_internal_tf
         ):
+            logger.debug(
+                f"Internal TF not enabled for"
+                f"'{self.project.service.hostname}/{self.project.full_repo_name}'\n"
+                f"enabled_projects_for_internal_tf"
+                f"={self.service_config.enabled_projects_for_internal_tf}"
+            )
             self.report_status_to_test_for_chroot(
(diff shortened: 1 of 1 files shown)


TomasTomecek (approved):

packit_service/worker/testing_farm.py
```diff
@@ -316,6 +316,12 @@ def run_testing_farm(self, build: "CoprBuildModel", chroot: str) -> TaskResults:
             and f"{self.project.service.hostname}/{self.project.full_repo_name}"
             not in self.service_config.enabled_projects_for_internal_tf
         ):
+            logger.debug(
+                f"Internal TF not enabled for"
+                f"'{self.project.service.hostname}/{self.project.full_repo_name}'\n"
+                f"enabled_projects_for_internal_tf"
+                f"={self.service_config.enabled_projects_for_internal_tf}"
```
mfocko, line 323: Do you want to log all of them every time? :thinking:
imo it might be an unnecessary noise, since we can „SSH“ into the worker

TomasTomecek: that's a valid concern although there is just 2 of them for now

lachmanfrantisek: I know I can check the config in the OpenShift UI or connect to the worker. But when adding new projects, I want to be sure what is in the config when we hit this. (e.g. When I checked the worker, the config had already been updated and I was wondering about what had gone wrong here.)

lachmanfrantisek: I can remove it if you want. Just wanted to add my motivation...

lachmanfrantisek:
Let's close this. I now don't think we need it.

PR #1149, opened 2021-07-02
[Dockerfiles] Don't COPY everything

jpopelka:
This was first added in 447f215fa without an explanation and I don't see why we'd need all those files.
I'm going to reuse the image in hardly and don't want them.

I tried the image locally via docker-compose and all looks fine.


TomasTomecek (approved):
+1, well done!

we now exactly know what we need; back in 2019, we didn't :)

PR #1156, opened 2021-07-13
Make main task name configurable

jpopelka:
To be hardly ;-) utilized, as discussed in
https://github.com/packit/hardly/pull/11#discussion_r668688299

The 'or' is used so that we use the default even when the env. var. is set to empty string.


jpopelka:
recheck

TomasTomecek (approved):

lachmanfrantisek (approved):
Thanks!

PR #1162, opened 2021-07-16
Fix typo when creating TF runs internally

mfocko:
Add missing default for submitted time for the TF runs.

Part of packit/dashboard#80


TomasTomecek (approved):
Doesn't look like a typo to me :D I'd say it's a plain old bug.

mfocko:
> Doesn't look like a typo to me :D I'd say it's a plain old bug.

The comment above it states usage of the default, so to me it seems like overlooked or maybe caused by rebase with conflicts (I lost `@staticmethod` once like that :D)

lachmanfrantisek (commented):
Probably some bad rebase. Thanks!

PR #1163, opened 2021-07-16
Install scripts for worker after deleting /src

mfocko:
Since the contents of ‹/src› directory is deleted to be reused for hardly,
install the scripts (mainly ‹allowlist.py›) in ‹/usr/bin› and add shebangs
to be able to run the scripts without specifying interpreter explicitly.


TomasTomecek (approved):
whoopsie

I wonder if we should introduce some kind of test for this

jpopelka (approved):

Issue #591, opened 2020-04-29
Flagging PRs in Pagure


csomh:
There are flags for commits and PRs. How do we make it clear which commit the PR flags are referring too?

This issue is to track the discussion started in https://pagure.io/pagure/issue/4835.

lachmanfrantisek:
The Pagure issue is being worked on. It looks like we can use commit-flag for flagging PRs like on GitHub. (So, we will be able to remove that pr-flag completely.)

TomasTomecek:
what work on our side is this issue tracking exactly?

lachmanfrantisek:
Isn't this about using commit flags instead of the PR flags?

So, remove the "hack" for Pagure from the service code:

https://github.com/packit/packit-service/blob/17e07017b640fe03efb5236c0dd51132727b8a6d/packit_service/worker/reporting.py#L94-L111

https://github.com/packit/packit-service/blob/17e07017b640fe03efb5236c0dd51132727b8a6d/packit_service/worker/reporting.py#L190-L192

lachmanfrantisek:
It would be nice to get rid of the pagure-specific tweaks.

lbarcziova:
Implemented in #1167

PR #1167, opened 2021-07-20
Use check runs for reporting in Github

lbarcziova:
- create BaseCommitStatus to unify statuses and checks
- create subclasses of StatusReporter for each forge
- StatusReporterGithubStatuses parent class for StatusReporterGithubChecks so that we can fallback to statuses (e.g. no permissions for checks)
- improved mapping of statuses which don't exist for some forge (e.g. Gitlab has running status, Github and Pagure not), therefore  some statuses changed from pending to running `e.g. Build in progress...` 
- neutral statuses for permission problems 
- get rid of flagging PRs in Pagure

Implements #760, #591

TODO:

- [x] agree on the check run output form (https://github.com/packit/hello-world/pull/111/checks?check_run_id=3114072495)
- [x] figure out what to do with re-run on failed check runs

### DIFF
--- a/packit_service/constants.py
+++ b/packit_service/constants.py
@@ -58,6 +58,13 @@ CELERY_DEFAULT_QUEUE_NAME = "short-running"
 
 CELERY_DEFAULT_MAIN_TASK_NAME = "task.steve_jobs.process_message"
 
+MSG_MORE_DETAILS = "You can find more details about the job [here]({url}).\n\n"
+
+MSG_RERUN_NOT_SUPPORTED = (
+    "*Rerunning tasks via `Re-run` button is currently not supported, "
+    "please don't click it.*"
+)
+
 
 class KojiBuildState(Enum):
     """
--- a/packit_service/worker/reporting.py
+++ b/packit_service/worker/reporting.py
@@ -1,18 +1,56 @@
 # Copyright Contributors to the Packit project.
 # SPDX-License-Identifier: MIT
 
-import hashlib
 import logging
-from typing import Optional, Union
+from enum import Enum
+from typing import Optional, Union, Dict
 
 import github
 import gitlab
 
 from ogr.abstract import CommitStatus, GitProject
+from ogr.services.github import GithubProject
+from ogr.services.github.check_run import (
+    create_github_check_run_output,
+    GithubCheckRunResult,
+    GithubCheckRunStatus,
+)
 from ogr.services.gitlab import GitlabProject
 from ogr.services.pagure import PagureProject
 
+from packit_service.constants import MSG_MORE_DETAILS, MSG_RERUN_NOT_SUPPORTED
+
 logger = logging.getLogger(__name__)
+
+
+class BaseCommitStatus(Enum):
+    failure = "failure"
+    neutral = "neutral"
+    success = "success"
+    pending = "pending"
+    running = "running"
+    error = "error"
+
+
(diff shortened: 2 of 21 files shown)


packit_service/constants.py
```diff
@@ -58,6 +58,13 @@
 
 CELERY_DEFAULT_MAIN_TASK_NAME = "task.steve_jobs.process_message"
 
+MSG_MORE_DETAILS = "You can find more details about the job [here]({url}).\n\n"
+
+MSG_RERUN_NOT_SUPPORTED = (
+    "**Rerunning tasks via `Re-run` button is currently not supported, "
+    "please don't click it.**"
+)
```
lachmanfrantisek, line 66: I would use italics instead and amplify other info instead.

packit_service/worker/reporting.py
```diff
@@ -30,6 +69,22 @@ def __init__(
         self.commit_sha: str = commit_sha
         self.pr_id: Optional[int] = pr_id
 
+    @classmethod
+    def get(
+        cls, project: GitProject, commit_sha: str, pr_id: Optional[int] = None
+    ) -> "StatusReporter":
```
lachmanfrantisek, line 75: Maybe some more specific name?

```suggestion
    def get_instance(
        cls, project: GitProject, commit_sha: str, pr_id: Optional[int] = None
    ) -> "StatusReporter":
```

packit_service/worker/reporting.py
```diff
@@ -113,42 +184,92 @@ def report_status_by_comment(
...
+        self.project_with_commit.set_commit_status(
+            self.commit_sha, state_to_set, url, description, check_name, trim=True
         )
+        # Also set the status of the pull-request for forges which don't do
+        # this automatically based on the flags on the last commit in the PR.
+        self.__set_pull_request_status(check_name, description, url, state_to_set)
+
+    def __set_pull_request_status(
+        self, check_name: str, description: str, url: str, state: CommitStatus
+    ):
```
lachmanfrantisek, line 232: Do you want to get rid of the pagure-specific hacks here or leave it as a follow-up?

(context: https://github.com/packit/packit-service/issues/591#issuecomment-804120367)

packit_service/constants.py
```diff
@@ -58,6 +58,13 @@
 
 CELERY_DEFAULT_MAIN_TASK_NAME = "task.steve_jobs.process_message"
 
+MSG_MORE_DETAILS = "You can find more details about the job [here]({url}).\n\n"
```
lachmanfrantisek, line 61: I am thinking about how we can make the URL more visible.

Maybe put it on a table?

| | |
| --- | --- |
| job details | https://dashboard.packit.dev/results/copr-builds/5000 |

Or some icon/button/image with a link?

But that's really a small thing we can easily improve later.

lachmanfrantisek (commented):
Well done!

Just a few really small notes.

packit_service/worker/reporting.py
lbarcziova: removed

packit_service/constants.py
lbarcziova: I don't like the table without a header so much but will try to find some better and more visible way. I will leave it simple as it is for now.

lachmanfrantisek (approved):
:rocket:

TomasTomecek (approved):
very nice!

PR #1172, opened 2021-07-26
Runs API

mfocko:
TODO:
- [x] add endpoint for merged run
- [x] add endpoint for run itself
- [x] update SRPM, COPR, Koji, TF endpoints to provide run ID
  - [x] decide how to handle multiple COPR builds and TF runs from one SRPM
- [x] tests
- [x] get rid of the code duplication of runs query :disappointed: 

---

*release notes will go here*

### DIFF
--- a/packit_service/service/api/runs.py
+++ b/packit_service/service/api/runs.py
@@ -1,0 +1,126 @@
+# Copyright Contributors to the Packit project.
+# SPDX-License-Identifier: MIT
+
+from http import HTTPStatus
+from logging import getLogger
+
+from flask_restx import Namespace, Resource
+
+from packit_service.models import (
+    CoprBuildModel,
+    KojiBuildModel,
+    RunModel,
+    SRPMBuildModel,
+    TFTTestRunModel,
+    optional_time,
+)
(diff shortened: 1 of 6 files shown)


packit_service/service/api/runs.py
```diff
@@ -0,0 +1,126 @@
...
+    SRPMBuildModel,
+    TFTTestRunModel,
+    optional_time,
+)
+from packit_service.service.api.parsers import indices, pagination_arguments
+from packit_service.service.api.utils import get_project_info_from_build, response_maker
+
+logger = getLogger("packit_service")
+
+ns = Namespace("runs", description="Pipelines")
```
lachmanfrantisek, line 22: Is this some convention that we name it `ns` each time and need to rename it when importing?..;)

lachmanfrantisek (approved):
Looks good to me!

packit_service/service/api/runs.py
mfocko: Probably because it's shorter in decorators afterwards? Otherwise have no clue

PR #1173, opened 2021-07-26
Add PR template

mfocko:
Add PR template to simplify adding release notes for blog post to the PRs
and ensure unified format across our projects and PRs.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,20 @@
+TODO:
+
+- [ ] ‹fill in›
+
+<!-- notes for reviewers -->
+
+<!-- Links to other issues or pull requests,
+     for cross-repository links use: ‹namespace›/‹repository›#‹ID of issue›
+       (‹namespace›/‹repository›!‹ID of PR› respectively)
+-->
+
+Fixes
+
+Related to
+
+Merge before/after
+
+---
+
+<!-- release notes for blog follow -->


csomh (approved):
:parachute:

mfocko:
> One question, though: How should I craft a PR if I don't want to be included in the blog post?

I could add template without it (going to test on my own repo before merging, cause I can't find in docs how it works with default template, etc.), not sure how are we going to parse/scrape those yet

mfocko:
@csomh apparently if we had more than one PR template, it can be used only via URL that explicitly specifies which template to use (aaargh…)

So I suggest something like this:

```md

<!-- release notes that are included in the blogpost -->

---

<!--
TODO:
- [ ] ‹fill in›
-->

<!-- notes for reviewers -->

<!-- Links to other issues or pull requests,
     for cross-repository links use: ‹namespace›/‹repository›#‹ID of issue›
	(‹namespace›/‹repository›!‹ID of PR› respectively)
Fixes
Closes
Related to
Merge before/after
-->

```
^ also this is directly from opening PR, which leads me to suggestion to swap the release notes to the end, since GitHub prefills commit message if PR has only one commit and it doesn't look like it can be prevented

also it could help parsing if we did split by `--- <!-- release notes follow -->` and took the rest :man_shrugging:

csomh:
So much to discuss :)

I don't have a strong opinion about the order: the part that goes into the blog-post could be first or second. On the long run I think we will just write the commit message this way, so whatever gets pre-filled by GitHub is already good. If one has multiple commits, the merge-request message needs to be crafted by hand anyway. So in this case the order won't matter much.

The algorithm to pick parts to include in blog post could be: look for merge commits that have `---` and pick whatever comes before/after that.

mfocko:
> So much to discuss :)
> 
> I don't have a strong opinion about the order: the part that goes into the blog-post could be first or second. On the long run I think we will just write the commit message this way, so whatever gets pre-filled by GitHub is already good. If one has multiple commits, the merge-request message needs to be crafted by hand anyway. So in this case the order won't matter much.
> 
> The algorithm to pick parts to include in blog post could be: look for merge commits that have `---` and pick whatever comes before/after that.

I would probably discard pre-filled commit message, because Zuul adds PR description, that way we will end up with the same commit message twice in the history (once for the commit itself and second time for the merge commit).

mfocko:
Update:
- Zuul doesn't copy the PR description 1:1, so scraping commit messages doesn't appear to be an option
- GitHub API doesn't offer listing PRs by merge timestamp (even though you can do it on the frontend)
- maybe PR IDs from `move-stable` and then ogr?

@TomasTomecek @jpopelka any other remarks, or we're good to merge and distribute to other repositories? :)

For label I have been thinking about `has-release-notes` with description: _PR includes release notes for weekly blog post._

PR #1179, opened 2021-07-30
Improve admin's UX for allowlisting

mfocko:
1. Strips protocol, since it's not used.
2. Allows interactive input when no argument is given.
3. Allows allowlisting from ‹waiting› command.

### DIFF
--- a/files/scripts/allowlist.py
+++ b/files/scripts/allowlist.py
@@ -1,5 +1,8 @@
 #!/usr/bin/env python3
 
+from urllib.parse import urlparse
+from typing import Optional
+
 import click
 
 from packit_service.worker.allowlist import Allowlist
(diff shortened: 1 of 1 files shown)


files/scripts/allowlist.py
```diff
@@ -15,32 +18,137 @@
...
+
+    Returns:
+        Path from user input.
+    """
+    domain = click.prompt("Please input domain (e.g. github.com, gitlab.com)")
+    namespace_or_repo = click.prompt(
+        "(in case of specific repository, type the whole path suffixed with "
+        "‹.git›, e.g. for ‹ogr› repository in ‹packit› namespace type "
+        "‹packit/ogr.git›)\nPlease input namespace"
+    )
```
csomh, line 79: If this is broken up like this, then I think `PATH_HELP` should be rewritten to avoid confusion.

jpopelka (approved):

files/scripts/allowlist.py
mfocko: that function is used as fallback, you can do

    $ allowlist.py approve github.com # or
    $ allowlist.py approve # <- with this you get the prompt from that function

csomh: Okay, I see. No update needed then. :)

PR #1185, opened 2021-08-03
Measure indicators

lbarcziova:
- Implemented indicators listed [here](https://github.com/packit/research/blob/main/error-budgets/SLOs-SLIs.md#packit-slis) except of `Number of unfinished builds that are in progress for more than 12 hours`, `Number of unfinished test runs that are in progress for more than 12 hours` (these could be implemented e.g.with babysit task)
- removed original metrics for copr build, since we now have 3 instead which are more defined
- save the time when the `Task was accepted` status is set in the CoprBuildModel so that we can measure how long it takes from `accepted` to `finished` status
- for test runs we discussed with Tomas, what time we should measure, but since the test run usually includes also building in copr, we agreed on measuring the time from when the test run is submitted

---

TBD


packit_service/models.py
```diff
@@ -779,6 +779,7 @@ class CoprBuildModel(ProjectAndTriggersConnector, Base):
     web_url = Column(String)
     # url to copr build logs
     build_logs_url = Column(String)
+    task_accepted_time = Column(DateTime)
```
TomasTomecek, line 782: nit: please provide a comment what the value is meant to contain, I'm assuming it's something like

    # time when we receive events and create respective copr builds

packit_service/worker/jobs.py
```diff
@@ -190,6 +193,25 @@ def get_config_for_handler_kls(
...
+    pushgateway = Pushgateway()
+
+    task_accepted_time = datetime.now()
+    response_time = (task_accepted_time - event.created_at).total_seconds()
+    pushgateway.initial_status_time.observe(response_time)
+    if response_time > 15:
+        pushgateway.no_status_after_15_s.inc()
+
+    # set the time when the accepted status was set so that we can use it later for measurements
+    event.task_accepted_time = task_accepted_time
```
TomasTomecek, line 206: it took me some time to figure out where is the `event.task_accepted_time` being set and I find it a bit misleading it's "buried" here. I think it would be more transparent to define that value sooner, e.g. in `process_jobs()` or even during the event instance creation/parsing. What do you think?

TomasTomecek (commented):
Very nice work! just one main suggestion about the code clarity

packit_service/worker/jobs.py
lbarcziova: I implemented it here (directly after the `Task is accepted` status is set), because we discussed we want to measure the time to build in Copr from this point, when we set the statuses for the first time. So do we want to measure the copr build time from the point when the event happened? (event.created_at from Github event)

TomasTomecek: we sorted this out over chat and the existing implementation does exactly what it's meant to do, I just got confused.

packit_service/worker/handlers/copr.py
```diff
@@ -275,12 +276,22 @@ def run(self):
             logger.info(msg)
             return TaskResults(success=True, details={"msg": msg})
 
+        self.pushgateway.copr_builds_finished.inc()
+
+        # if the build is needed only for test, it doesn't have the task_accepted_time
```
mfocko, line 281: so we don't take in account Copr builds, right?

also related question
> How are we going to handle the length of test run if the Copr build needs to be done?

imo it will basically track the time of running the whole pipelline, which is intended? or not?

mfocko (commented):
lgtm

packit_service/worker/handlers/copr.py
lbarcziova: > so we don't take in account Copr builds, right?

when do you mean this?

> How are we going to handle the length of test run if the Copr build needs to be done?

we agreed with Tomas on measuring the time from when the test run is submitted so that we don't include the time for building in Copr here

TomasTomecek: it's definitely worth considering but from user PoV - if they are not interested in actual RPM builds let's not account for them in the build-specific metrics

+1 wrt tests

I'm really excited about finally starting to see these numbers and charts: once we have those, we can start tinkering with the actual measurements to get numbers and charts which have the biggest use for us

TomasTomecek (approved):

packit_service/worker/handlers/copr.py
mfocko: @lbarcziova sorry, I just realized I wrote like an idiot :D I was going back and forth between the commits and it merged for me…

1. OK, Copr build is ignored if needed for TF. Got it.
2. I just checked TF helper, it creates the entry **when actually** submitting tests to TF, so my question is irrelevant. :)

lbarcziova:
Thank you for the reviews, I won't merge this today, since we move the stable branches today and I would like to have this in staging for few days before getting to prod.

jpopelka (approved):

Issue #1182, opened 2021-08-03
Tests triggered by comment without an existing build end without an action


lbarcziova:
This was supposed to be fixed, but [the condition](https://github.com/packit/packit-service/blob/2fd5e3496ef0abe780ce76af947337471d10f618/packit_service/worker/handlers/testing_farm.py#L110) to check whether there is already a build doesn't cover the case when we run testing farm [on all targets](https://github.com/packit/packit-service/blob/2fd5e3496ef0abe780ce76af947337471d10f618/packit_service/worker/handlers/testing_farm.py#L103) (triggered by comment).

This was seen here https://github.com/psss/tmt/pull/853.

mfocko:
related to #1174

psss:
Not sure the issue has been fixed. A recent `/packit test` comment in https://github.com/psss/tmt/pull/853 ended up with a `Missing Copr build for some target, running a new Copr build.` message. Now it seems to be stuck in this phase.

lachmanfrantisek:
@lbarcziova Do you think it's already fixed by fixing the serialisation in #1205 ? Or, is there anything else we can do?

PR #1187, opened 2021-08-03
Set the accepted status also when reacting to /packit test

lbarcziova:
Currently, when we trigger tests by `/packit test`, we got directly the `Submitting tests` status. If there are more handler tasks in queue, that can take some time, so the `Task was accepted` status can be helpful to get feedback and also after fixing #1182, we will run the build if there is none and waiting to get to the worker which builds SRPMs can take even longer.

---
When packit-service reacts to `/packit test`, `Task was accepted` status is firstly set to unify the behaviour with reacting to triggering builds.


mfocko (approved):
lgtm :)

lbarcziova:
recheck

PR #1188, opened 2021-08-04
Fix running tests if there is no build or failed build

### LINKED ISSUE #1182: Tests triggered by comment without an existing build end without an action

lbarcziova:
This mainly changes the behaviour for tests triggered by `/packit test`
- If there is a missing build for some target, trigger a new task to build in Copr.
- If the latest Copr build has failed status for some target, set the Github check/status for that target so that user is informed about this and run tests for the other targets (currently, the tests are submitted also for failed builds and we got later failure from TF, [example](https://github.com/packit/hello-world/pull/426/checks?check_run_id=3238411179)). We can discuss this behaviour and it can be improved in the future e.g. to run the build.

Fixes #1182 

---
The behaviour of running tests triggered by `/packit test` comment was improved. If there is no existing Copr build when the tests are triggered, Packit service should now react and create a new build. Also when the last Copr build status is failed, tests are not submitted and users are informed about this.

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -889,19 +889,26 @@ class CoprBuildModel(ProjectAndTriggersConnector, Base):
             return query.first()
 
     @classmethod
-    def get_all_by_owner_and_project_and_target(
+    def get_all_by_owner_project_target_commit(
         cls,
         owner: str,
         project_name: str,
         target: str,
+        commit_sha: str,
     ) -> Optional[Iterable["CoprBuildModel"]]:
         """
         All owner/project_name builds sorted from latest to oldest
+        with the given target and commit_sha.
         """
         with get_sa_session() as session:
             query = (
                 session.query(CoprBuildModel)
-                .filter_by(owner=owner, project_name=project_name, target=target)
+                .filter_by(
+                    owner=owner,
+                    project_name=project_name,
+                    target=target,
+                    commit_sha=commit_sha,
+                )
                 .order_by(CoprBuildModel.build_id.desc())
             )
             return query.all()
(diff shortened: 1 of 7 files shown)


jpopelka (approved):

packit_service/models.py
```diff
@@ -889,19 +889,26 @@ def get_by_build_id(
             return query.first()
 
     @classmethod
-    def get_all_by_owner_and_project_and_target(
+    def get_all_by_owner_project_target_commit(
         cls,
         owner: str,
         project_name: str,
         target: str,
+        commit_sha: str,
```
mfocko, line 897: Shouldn't commit sha and project name (just to make sure, but I think at least GitHub is switched to new SHA after the collision was discovered) be enough?

mfocko (approved):

packit_service/models.py
lbarcziova: I'm not sure how do you mean it, which parameters do you suggest removing?

mfocko: Keeping only `target` and `commit_sha`, but it's just a thought; It is safer the way it is :)

PR #1189, opened 2021-08-05
Fix inconsistency of PR template

mfocko:
Update PR template to also include usage of release notes in changelogs.


jpopelka (approved):
well, we don't have changelog here, but :+1:

PR #1194, opened 2021-08-06
Add Run API endpoint for project

mfocko:
For the enhancements of project view of pipelines in dashboard.

TODO:

- [ ] figure out better solution that has smaller performance impact
- [ ] refactor, clean out; there are unused methods from trying different solutions

My main issue is with joining the tables to be able to filter by project. I haven't been able to do it properly, the way it is done now doesn't appear to be that problematic (unless you run it for non-existing project… it takes a while, cause it „crunches“ through the whole run table). **Suggestions are welcome.**

Merge before packit/dashboard#135

---

_TODO_

### DIFF
--- a/packit_service/service/api/runs.py
+++ b/packit_service/service/api/runs.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: MIT
 
 from http import HTTPStatus
+import itertools
 from logging import getLogger
 
 from flask_restx import Namespace, Resource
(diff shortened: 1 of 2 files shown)


mfocko:
as we discussed with @TomasTomecek we could probably factor out the git project from separate trigger models to `JobTriggerModel`, but it would result in tables with not very meaningful data, e.g. for triggering from git branch we would have table with just names of branches, which may be a bit weird

packit_service/service/api/runs.py
```diff
@@ -86,6 +87,35 @@ def get(self):
...
+                ),
+                first,
+                last,
+            )
+        )
+        resp = response_maker(
+            result,
+            status=HTTPStatus.PARTIAL_CONTENT.value,
+        )
+        resp.headers["Content-Range"] = f"runs {first + 1}-{last}/*"
```
TomasTomecek, line 115: interesting, didn't know about this header

PR #1198, opened 2021-08-10
Install python3-flexmock-0.10.4 in test image

jpopelka:
This is just a workaround and something better will need to be done once that build disappears from fedora repos.

I spent some time trying to figure out the cause, but gave up as I have better things to do.

The tests here will fail anyway, because the image is cached in the registry, i.e. we first need to merge this to build the new tests image.

---

N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,6 +23,9 @@ check:
 	find . -name "*.pyc" -exec rm {} \;
 	PYTHONPATH=$(CURDIR) PYTHONDONTWRITEBYTECODE=1 python3 -m pytest --color=$(COLOR) --verbose --showlocals --cov=packit_service --cov-report=$(COV_REPORT) $(TEST_TARGET)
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -23,6 +23,9 @@ check:
 	find . -name "*.pyc" -exec rm {} \;
 	PYTHONPATH=$(CURDIR) PYTHONDONTWRITEBYTECODE=1 python3 -m pytest --color=$(COLOR) --verbose --showlocals --cov=packit_service --cov-report=$(COV_REPORT) $(TEST_TARGET)
 
+# In most cases you don't need to build your test-image, the one in registry should be all you need.
+# But if you think you need your (because of some new dependency, for example),
+# remove the '--pull=always' in check-in-container target before running it.
```
mfocko, line 28: Thanks for the comment!

mfocko (approved):

Issue #1184, opened 2021-08-03
Do not run jobs that can touch the local repo from the short-running worker


lachmanfrantisek:
Benefit: We can make the `short-running` worker more lightweight and does not need to attach the repository-cache volume to it.

The task is:
* [ ] check the jobs attached to the `short-running` queue and move out the ones that touch the git repo or can run any action in the sandbox.
* [ ] document this distinction in the deployment or in the packit-service code.

Optionally: How busy is the current short-running worker? What about scaling it up and/or reducing the resource limits for it.

lbarcziova:
Just some notes (@lachmanfrantisek correct me please if I got something wrong):
- sandcastle is used for executing code in actions, actions should be used only for `propose-downstream` and creating SRPM, so the tasks: copr_build, koji_build, propose_downstream
- worker repository cache is used whenever we clone the repo (=using LocalProject), so the tasks which do some operations with local repo, these should be: copr_build, koji_build, propose_downstream, distgit_commit
- but from checking the worker logs, I found `Cloning repo ...  -> /sandcastle using repository cache at /repository-cache` also in copr_build_end task (even if it probably isn't really needed), the BaseBuildJobHelper has the local_project property so I need to check its usages
- maybe it would be safer to only remove the pvc for sandcastle which is used by short-running worker for the beginning?

lachmanfrantisek:
> * sandcastle is used for executing code in actions, actions should be used only for `propose-downstream` and creating SRPM, so the tasks: copr_build, koji_build, propose_downstream

:+1: I would say so

> * worker repository cache is used whenever we clone the repo (=using LocalProject), the tasks which I think really do some operations with local repo, should be: copr_build, koji_build, propose_downstream, distgit_commit

:+1:

> * but from checking the worker logs, I found `Cloning repo ...  -> /sandcastle using repository cache at /repository-cache` also in copr_build_end task (even if it probably isn't really needed) and also some other tasks, since the BaseBuildJobHelper (which is used in lot of tasks) has the `api` property and when this is called, the `local_project` [is passed and initialized ](https://github.com/packit/packit-service/blob/14e7a86977b8298df2583434e939ea1cde922d2d/packit_service/worker/build/build_helper.py#L134). It doesn't make sense to send all these tasks to long-running queue since most of them don't really need the cloned repo, this could be maybe somehow refactored so that the `local_project` is called really only if needed.

[...]

PR #1208, opened 2021-08-24
Use local project in BaseBuildJobHelper.api only if needed

lbarcziova:
Related to #1184 

This should prevent cloning of the repo when not needed. Another proposed solution was to use `offline` or `refresh` arguments of `LocalProject`, but since we use the local_project in `BaseBuildJobHelper`, we use this also when building SRPM and there we need to clone the repo. Let me know if you have some other suggestions.

---

N/A

### DIFF
--- a/packit_service/worker/build/build_helper.py
+++ b/packit_service/worker/build/build_helper.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: MIT
 
 import logging
+from functools import partial
 from io import StringIO
 from packit_service.service.urls import get_srpm_build_info_url
 from packit_service.worker.result import TaskResults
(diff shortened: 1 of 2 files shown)


packit_service/worker/build/build_helper.py
```diff
@@ -131,7 +134,8 @@ def api(self) -> PackitAPI:
             self._api = PackitAPI(
                 self.service_config,
                 self.job_config,
-                self.local_project,
+                # so that the local_project is evaluated only if needed
+                Proxy(partial(BaseBuildJobHelper.local_project.__get__, self)),  # type: ignore
```
lachmanfrantisek, line 138: It's really easy with the `Proxy` object, nice! Do we want to have this for other occurrences in packit's codebase as well?

lachmanfrantisek (approved):

packit_service/worker/build/build_helper.py
lbarcziova: that would  be nice, will look into it

PR #1209, opened 2021-08-25
Set sandcastle repository cache env var only when serving long-running queue

lbarcziova:
TODO:

- [x] in packit code don't create the sandcastle volume if this var is not set

Related to #1184 

---
N/A


lachmanfrantisek (approved):
Thanks! :+1:

Issue #642, opened 2020-06-01
[testing-farm] Select packages to be installed from the copr repo


lachmanfrantisek:
AC:
- [ ] Get the info about packages that were built.
- [ ] Send the list of packages with the testing-farm payload.
- [ ] Since we may use this info in service's API in dashboard (and maybe even in other places), we may need to store this in the DB

TomasTomecek:
Sounds good, is this also aligned with the new API and how tmt works?

lachmanfrantisek:
> Sounds good, is this also aligned with the new API and how tmt works?

- See [Miro's comment](https://github.com/packit-service/packit-service/issues/628#issuecomment-635854313) for the context.
- Yes, it will be in the new API.
- Miro wants to fix that also in the cruncher. (Old API.)
- Regarding the TMT in the new API, we can use STI, TMT or command(s) for running tests.
- You can see the API structure [here](https://testing-farm.gitlab.io/api/#/paths/~1requests/post).
   - Specifically, we are speaking about `environments` -> `artifacts` -> `packages`. ([relevant code](https://gitlab.com/testing-farm/api/-/blob/master/testing-farm.yaml#L326))

thrix:
@lachmanfrantisek until the new api is available, can you expose that information also in the current API, so I can pick it up in cruncher pls? List of RPMs which I should install is fine.

lachmanfrantisek:
> @lachmanfrantisek until the new api is available, can you expose that information also in the current API, so I can pick it up in cruncher pls? List of RPMs which I should install is fine.

I'll try to look at it next week.

lachmanfrantisek:
I've created [an issue](https://pagure.io/copr/copr/issue/1411) in copr-repo since I could not find that info in the API/client trying both v2 and v3.

lachmanfrantisek:
The issue on the Copr side should be solved -- now it's our turn.

thrix:
@lachmanfrantisek oh cool! let me know if that is available via API, so I can try to consume it with cruncher

lbarcziova:
[The issue](https://pagure.io/copr/copr/issue/1411) in copr-repo was reopened since the functionality we need was not provided.

lachmanfrantisek:
We are still blocked..;(

TomasTomecek:
The issue is still open.

TomasTomecek:
is the information we need available in copr API already?

lachmanfrantisek:
[...]

PR #1212, opened 2021-09-01
Send list of packages built in copr to testing farm

mmuzila:
Fixes: #642 

TODO:
- [x] squash the commits (keeping them now so that it's clear what is added now)


packit_service/worker/testing_farm.py
```diff
@@ -110,6 +110,21 @@ def _payload(self, build_id: int, chroot: str) -> dict:
         if self.fmf_ref:
             fmf["ref"] = self.fmf_ref
 
+        artifacts = [
+            {
+                "id": f"{build_id}:{chroot}",
+                "type": "fedora-copr-build",
+            }
+        ]
```
lbarcziova, line 118: maybe I just don't get it correctly, but if the `built_packages` are present, we want to send only info about them without this, or am I wrong?

mmuzila: I think that we need to send also info about the copr build, otherwise the tft wouldn't know where to look for specified packages.

mmuzila:
recheck

lbarcziova:
`packit-service-tests-requre : FAILURE in 23m 36s`

For the new field in DB you need to generate a migration script, it is documented a bit [here](https://github.com/packit/packit-service/blob/c93f851a4767b79df1a18bb81da05575c33d5488/CONTRIBUTING.md#database).

TomasTomecek (commented):
I think this approach makes sense. We could create a new table for those N-V-R values but I don't think it's worthy.

Seems that only the migration is left to be created.

lbarcziova:
I'm not sure about it, but from the docs for me it looks like we should send the packages to install as a part of the artifact we are already sending:

`packages | Array of strings (artifact.packages)  List of packages to install, if applicable to the artifact. By default all packages from the artifact are installed.` 

(I implemented this in the latest commit, but need to verify it)

lbarcziova:
From the docs to me it looks like we should send the packages to install as a part of the artifact we are already sending:

`packages | Array of strings (artifact.packages)  List of packages to install, if applicable to the artifact. By default all packages from the artifact are installed.` 

(I implemented this in the latest commit, but need to verify it)

packit_service/models.py
```diff
@@ -816,8 +816,15 @@ class CoprBuildModel(ProjectAndTriggersConnector, Base):
     # metadata is reserved to sqlalch
     data = Column(JSON)
 
+    built_packages = Column(JSON)
```
TomasTomecek, line 819: nit: can we have a comment here to describe the expected schema please?

TomasTomecek (approved):
pretty straightforward, thanks for finishing this!

PR #1215, opened 2021-09-08
Shorten path to get testing farm results

### LINKED ISSUE #1178: Make it shorter to get to the test results
When github checks are enabled, check [results page](https://github.com/psss/tmt/pull/795/checks?check_run_id=3182856758) contains a link to [packit page](https://dashboard.packit.dev/results/testing-farm/32270) which contains a link to actual test results. This is three clicks. Could you please make the path shorter? Thanks.

nikromen:
Fixes #1178 

---

Now you can find URL pointing directly to testing farm results on GitHub Checks page.

### DIFF
--- a/packit_service/constants.py
+++ b/packit_service/constants.py
@@ -66,6 +66,8 @@ MSG_RERUN_NOT_SUPPORTED = (
     "please don't click it.*"
 )
 
+MSG_TABLE_HEADER_WITH_DETAILS = "| Name/Job | URL |\n" "| --- | --- |\n"
+
 
 class KojiBuildState(Enum):
     """
--- a/packit_service/worker/reporting.py
+++ b/packit_service/worker/reporting.py
@@ -18,7 +18,10 @@ from ogr.services.github.check_run import (
 from ogr.services.gitlab import GitlabProject
 from ogr.services.pagure import PagureProject
 
-from packit_service.constants import MSG_MORE_DETAILS, MSG_RERUN_NOT_SUPPORTED
+from packit_service.constants import (
+    MSG_RERUN_NOT_SUPPORTED,
+    MSG_TABLE_HEADER_WITH_DETAILS,
+)
 
 logger = logging.getLogger(__name__)
 
@@ -112,6 +115,7 @@ class StatusReporter:
         description: str,
         check_name: str,
         url: str = "",
+        links_to_external_services: Optional[Dict[str, str]] = None,
     ):
         raise NotImplementedError()
 
@@ -120,15 +124,29 @@ class StatusReporter:
         state: BaseCommitStatus,
         description: str,
         url: str = "",
+        links_to_external_services: Optional[Dict[str, str]] = None,
         check_names: Union[str, list, None] = None,
     ) -> None:
         """
-        set commit check status
+        Set commit check status.
 
-        :param state: state accepted by github
-        :param description: the long text
-        :param url: url to point to (logs usually)
-        :param check_names: those in bold
+        Args:
+            state: State accepted by github.
+            description: The long text.
+            url: Url to point to (logs usually).
+
+                Defaults to empty string
+            links_to_external_services: Direct links to external services.
+                e.g. `{"Testing Farm": "url-to-testing-farm"}`
+
+                Defaults to None
+            check_names: Those in bold.
+
+                Defaults to None
+
+        Returns:
+            None
+
         """
 
         if not check_names:
@@ -140,7 +158,11 @@ class StatusReporter:
 
         for check in check_names:
             self.set_status(
-                state=state, description=description, check_name=check, url=url
+                state=state,
+                description=description,
+                check_name=check,
+                url=url,
+                links_to_external_services=links_to_external_services,
             )
 
     def _add_commit_comment_with_status(
@@ -209,6 +231,7 @@ class StatusReporterPagure(StatusReporter):
         description: str,
         check_name: str,
         url: str = "",
+        links_to_external_services: Optional[Dict[str, str]] = None,
     ):
         state_to_set = self.get_commit_status(state)
         logger.debug(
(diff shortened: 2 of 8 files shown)


nikromen:
Maybe if we want (in the future) to add another links to the GitHub Checks page (related to [this](https://github.com/packit/packit-service/issues/1178#issuecomment-890964478)) it will be useful to have it formatted e.g. on table like this:

| Name/Job | URL |
| --- | --- |
| Dashboard | link-to-dashboard |
| TF results | link-to-TF-results |
| COPR build | link-to-copr-build | 
| something else | link-to-something-else |

or something more clear rather than:

You can find more details about the job here.
You can view the testing farm results here.
and so on...

Also if we want to do it this way - should I replace the `tf_url` argument in `StatusReporter.set_status` and `StatusReporter.report` methods with dictionary which will map the "job" to the link or with some object which will contain these links? Because having all links as method parameters is nasty.

packit_service/worker/reporting.py
```diff
@@ -112,6 +116,7 @@ def set_status(
         description: str,
         check_name: str,
         url: str = "",
+        tf_url: str = "",
```
mfocko, line 119: There is a reason why you can't have `url` as `None` (Pagure requires `url`, but can be empty string), doesn't apply to `tf_url` though.

mfocko (commented):
LGTM

Though I would probably not focus so hard on the testing farm, since similar request may come for Copr or Koji too…

Cannot really think about something more appropriate… I would probably go with something like „direct link to external service“, but that sounds a bit «meh» to me

packit_service/constants.py
```diff
@@ -66,6 +66,8 @@
     "please don't click it.*"
 )
 
+MSG_TABLE_HEADER_WITH_DETAILS = "| Name/Job | URL |\n" "| --- | --- |\n"
```
mfocko, line 69: It looks bit weird, but I see where you are going with it.

packit_service/worker/reporting.py
```diff
@@ -120,15 +124,29 @@ def report(
...
+        Set commit check status.
+
+        Args:
+            state: State accepted by github.
+            description: The long text.
+            url: Url to point to (logs usually).
+
+                Defaults to empty string
+            links_to_external_services: Direct links to external services.
+                e.g. {"Testing Farm": "url-to-testing-farm"}
```
mfocko, line 140:
```suggestion
                e.g. `{"Testing Farm": "url-to-testing-farm"}`
```

Google style docstrings support markdown.

packit_service/worker/reporting.py
```diff
@@ -300,23 +325,46 @@ def set_status(
...
+        table_content = []
+        if url:
+            table_content.append(f"| Dashboard | {url} |\n")
+        if links_to_external_services is not None:
+            table_content += [
+                f"| {name} | {link} |\n"
+                for name, link in links_to_external_services.items()
+            ]
+
+        return "".join(table_content)
```
mfocko, line 341:
```suggestion
        return "\n".join(table_content)
```
Or you can do this and leave out the `\n`s before return

packit_service/worker/reporting.py
```diff
@@ -300,23 +325,46 @@ def set_status(
...
+        url: str, links_to_external_services: Optional[Dict[str, str]]
+    ) -> str:
+        table_content = []
+        if url:
+            table_content.append(f"| Dashboard | {url} |\n")
+        if links_to_external_services is not None:
+            table_content += [
+                f"| {name} | {link} |\n"
+                for name, link in links_to_external_services.items()
+            ]
```
mfocko, line 339:
```suggestion
            table_content += (
                f"| {name} | {link} |\n"
                for name, link in links_to_external_services.items()
            )
```

This is „just“ syntax thing. `[]` constructs list, `()` is a generator (iterator).

In this use-case it shouldn't matter :thinking: I wonder where's the threshold where the construction of list starts to take more time than the generator. (IIRC there were some benchmarks by Hunor on one of the PRs before, but not sure if it was really this or just related to this.)

[link to SO with benchmark, both time and space](https://stackoverflow.com/a/62709748)

_Conclusion_: Shouldn't make much of a difference for up to 5 links.

tests/unit/test_reporting.py
```diff
@@ -150,7 +156,11 @@ def test_set_status_gitlab(
             flexmock(),
             BaseCommitStatus.success,
             "We made it!",
-            MSG_MORE_DETAILS.format(url="https://api.packit.dev/build/111/logs"),
+            MSG_TABLE_HEADER_WITH_DETAILS
+            + create_table_content(
```
mfocko, line 160: Careful here, if the `create_table_content` is broken, the tests will pass. Ideally you could cover this by unit test «just for» the `create_table_content`

mfocko (approved):
LGTM, minor suggestions apart from the tests.

packit_service/worker/reporting.py
```diff
@@ -300,23 +325,46 @@ def set_status(
...
             f" {description}"
         )
-        summary = (MSG_MORE_DETAILS.format(url=url) if url else "") + (
-            MSG_RERUN_NOT_SUPPORTED
-            if state_to_set == GithubCheckRunResult.failure
-            else ""
+
+        table_content = self._create_table_content(url, links_to_external_services)
+        summary = (
+            (MSG_TABLE_HEADER_WITH_DETAILS if table_content else "")
```
lbarcziova, line 359: wouldn't it be better to move this to `_create_table_content` (and rename the method) so that this method would manage the whole table creation?

lbarcziova (approved):
LGTM, nice work!

lachmanfrantisek (approved):
Nicely done! :rocket:

(Can you please click on `Resolve conversation` on the discussions/suggestions you've already covered? This will allow us to see if all the comments were resolved and we can merge the pull request...)

PR #1220, opened 2021-09-22
check-in-container tweaks

csomh:
Tweak the _check-in-container_ target so that it pulls the test image only if it's missing and make `podman run` interactive when running the target locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@ PATH_TO_SECRETS ?= $(CURDIR)/secrets/
 COV_REPORT ?= term-missing
 COLOR ?= yes
 SOURCE_BRANCH ?= $(shell git branch --show-current)
+CONTAINER_RUN_INTERACTIVE ?= -it
 
 service: files/install-deps.yaml files/recipe.yaml
 	$(CONTAINER_ENGINE) pull $(BASE_IMAGE)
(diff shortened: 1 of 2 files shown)


TomasTomecek (approved):

Makefile
```diff
@@ -24,15 +26,14 @@ check:
...
-# remove the '--pull=always' in check-in-container target before running it.
 build-test-image: files/install-deps-worker.yaml files/install-deps.yaml files/recipe-tests.yaml
 	$(CONTAINER_ENGINE) build --rm -t $(TEST_IMAGE) -f files/docker/Dockerfile.tests --build-arg SOURCE_BRANCH=$(SOURCE_BRANCH) .
 
 check-in-container:
 	@# don't use -ti here in CI, TTY is not allocated in zuul
 	echo $(SOURCE_BRANCH)
-	$(CONTAINER_ENGINE) run --rm --pull=always \
+	$(CONTAINER_ENGINE) run --rm $(CONTAINER_RUN_INTERACTIVE) \
+		--pull="$(PULL_TEST_IMAGE)" \
```
jpopelka, line 36: Both, `always` and `missing` have pros and cons.
|   | pros | cons |
| --- | --- | --- |
| always | always up-to-date image | user has to tweak the Makefile if (s)he wants to run own image | 
| missing | no need to tweak Makefile in case user wants  to run own image | the image will become outdated after some time and one has to remember to pull it manually| 

For me, `always` still wins, but while writing this comment I see it's already been merged :-D

csomh: AFAIK, we agreed not so long ago to follow the `make build-test-image && make check-in-container` pattern in all of our repositories in order to have a consistent workflow so that we don't have to spend brainpower on figuring out how to test things based on which repo we are working on.

`--pull=always` broke this pattern, and personally, I found it annoying to remember to edit the Makefile every time I wanted to use a locally built image **and** remember not to commit this change.

In this regard defaulting to `--pull=missing` seems to be a good compromise. And if the `--pull=always` usecase is more appealing, it can be set through the `PULL_TEST_IMAGE` env var, without the need to touch the Makefile.

PR #1222, opened 2021-09-22
+1 on comment commands

### LINKED ISSUE #1081: Give :+1: on comment commands
When reacting to the comment command, give a :+1: reaction to the comment to provide some clarity that we:
* have processed the comment
* are working on it.

*This is a part of our UX effort to provide more information to users about the process of ongoing packit-service jobs.*

nikromen:
TODO:

- [x] add a test case which will work
- [x] modify tests to make them work :confounded:
- [x] implement `get_comment` method in ogr

Fixes #1081 

---

If you trigger packit tests with one of `/packit _` commands, our bot gives you a :+1: reaction to let you know that we are working on it.

### DIFF
--- a/packit_service/worker/events/github.py
+++ b/packit_service/worker/events/github.py
@@ -3,7 +3,7 @@
 
 from typing import Dict, Optional, Union, List, Set
 
-from ogr.abstract import GitProject
+from ogr.abstract import GitProject, PRComment, IssueComment
 
 from packit_service.models import (
     AllowlistStatus,
@@ -128,7 +128,9 @@ class PullRequestCommentGithubEvent(AddPullRequestDbTrigger, AbstractGithubEvent
         project_url: str,
         user_login: str,
         comment: str,
+        comment_id: int,
         commit_sha: Optional[str] = None,
+        comment_object: Optional[PRComment] = None,
     ):
         super().__init__(project_url=project_url, pr_id=pr_id)
         self.action = action
@@ -139,11 +141,13 @@ class PullRequestCommentGithubEvent(AddPullRequestDbTrigger, AbstractGithubEvent
         self.target_repo_name = target_repo_name
         self.user_login = user_login
         self.comment = comment
+        self.comment_id = comment_id
         self.identifier = str(pr_id)
         self.git_ref = None  # pr_id will be used for checkout
 
         # Lazy properties
         self._commit_sha = commit_sha
+        self._comment_object = comment_object
 
     @property
     def commit_sha(self) -> Optional[str]:  # type:ignore
@@ -156,10 +160,19 @@ class PullRequestCommentGithubEvent(AddPullRequestDbTrigger, AbstractGithubEvent
         result = super().get_dict()
         result["action"] = result["action"].value
         result["commit_sha"] = self.commit_sha
+        result.pop("_comment_object")
         return result
 
     def get_base_project(self) -> Optional[GitProject]:
         return None  # With Github app, we cannot work with fork repo
+
+    @property
+    def comment_object(self) -> Optional[PRComment]:
+        if not self._comment_object:
+            self._comment_object = self.project.get_pr(self.pr_id).get_comment(
+                self.comment_id
+            )
+        return self._comment_object
 
 
 class IssueCommentEvent(AddIssueDbTrigger, AbstractGithubEvent):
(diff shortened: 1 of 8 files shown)


packit_service/worker/jobs.py
```diff
@@ -93,6 +93,11 @@ def get_handlers_for_event(
         ),
     ):
         handlers_triggered_by_comment = get_handlers_for_comment(event.comment)
+
+        if handlers_triggered_by_comment and not isinstance(
+            event, PullRequestCommentPagureEvent
```
mfocko, line 98: Careful about issue comment on Pagure, not sure if we handle those. Better safe than sorry…

packit_service/worker/parser.py
```diff
@@ -383,8 +385,10 @@ def parse_issue_comment_event(event) -> Optional[IssueCommentEvent]:
         issue_id = nested_get(event, "issue", "number")
         action = event.get("action")
         comment = nested_get(event, "comment", "body")
-        if action != "created" or not issue_id or not comment:
+        comment_object = event.get("comment")
+        if action != "created" or not (issue_id and comment and comment_object):
             return None
+        comment_object = GithubIssueComment(comment_object)
```
mfocko, line 391: If you look in ogr, these constructors expect the comment object from the dependency (e.g. PyGithub or python-gitlab). Here you are given webhook = `comment_object` is some kind of a JSON, not the thing `GithubIssueComment` processes.

mfocko: Also thinking about it, it is likely that this kind of attributes (e.g. projects, pull request / issue or comments) are initialized via property once they are needed.

nikromen:
recheck

nikromen:
recheck

mfocko:
recheck

tests/integration/test_issue_comment.py
```diff
@@ -57,6 +57,11 @@ def mock_comment(request):
         .and_return(True)
     )
     flexmock(project_class).should_receive("issue_comment").and_return(None)
+    issue = flexmock()
+    flexmock(project_class).should_receive("get_issue").and_return(issue)
+    comment = flexmock()
+    flexmock(issue).should_receive("get_comment").and_return(comment)
+    flexmock(comment).should_receive("add_reaction").with_args("+1").once()
```
mfocko, line 64: Just out of interest, what happens here with both prod and stg deployed?

I hope it doesn't fail [here](https://github.com/packit/packit-service/pull/1222/files#diff-1d719ae684f53243de1efeadb4b79e8cd77b2408c27b563bd43e0c4b6530236aR100)

mfocko (approved):

tests/integration/test_issue_comment.py
mfocko: Now that I'm thinking about it, I hope GitHub apps can react :thinking: but i think i've seen some „voting“ mechanism implemented via bot, so it should be fine.

packit_service/worker/events/github.py
```diff
@@ -204,6 +222,14 @@ def get_dict(self, default_dict: Optional[Dict] = None) -> dict:
         result["issue_id"] = self.issue_id
         return result
 
+    @property
+    def comment_object(self) -> Optional[IssueComment]:
+        if not self._comment_object:
+            self._comment_object = self.project.get_issue(self.issue_id).get_comment(
```
lbarcziova, line 228: since at the end of processing of the event we send the event as a JSON to handler Celery task, the dictionary from `Event.get_dict()` needs to be JSON serializable (I think it is not really tested in other events, so I will probably create an issue for it), so there can be a problem with this attribute and I think we need to call `event_dict.pop("_comment_object")` in `get_dict()` method for these classes (please, check how it is done in some other Event subclasses (e.g.we are removing the `project` as well)

lbarcziova (commented):
Nice work! Just one note..

tests/integration/test_issue_comment.py
TomasTomecek: I guess we'll find out once this is deployed to stg :D

TomasTomecek: definitely a good point though, I hope we won't need to request a new perm for this

lbarcziova (approved):
:rocket:

PR #1225, opened 2021-09-23
Change mapping to centos for internal_tf

kokesak:
Using internal_tf for testing in convert2rhel we need to get centos distro instead of rhel one.


TomasTomecek (approved):
@lachmanfrantisek PTAL

TomasTomecek:
please update these two test cases and once @thrix gives us green here, we're good to merge:
```
FAILED tests/unit/test_testing_farm.py::test_chroot2distro_arch[epel-7-x86_64-rhel-7-x86_64-True]
FAILED tests/unit/test_testing_farm.py::test_chroot2distro_arch[epel-8-x86_64-rhel-8-x86_64-True]
```

thrix:
hmm, I am not sure about this, I would rather see epel-8 tested on centos-8 on public TF ? I see this a major benefit that users can test epel-8 builds against RHEL nightlies? Note that for secrets, testing farm now supports passing secret environment variables, or is there some other reason why public TF cannot be used? (where the mapping is like this?)

bocekm:
@thrix, we can't use public TF because for conversions we need access to internal RH network - to access testing/staging RHSM server and to access internal RHEL repo mirrors directly.
There's no CentOS Linux chroot available in Fedora Copr, that's why we're using epel-X chroots.

thrix:
@bocekm thanks for the explanation. Still, I believe sooner or later somebody will want to test epel-8 against RHEL. Could we make the switch optional via another setting? by default we map to rhel on internal-tf, optionally we can map to CentOS?

TomasTomecek:
...or we could build against centos-stream in Copr and map that to RHEL internally

another possibility is that you'd be able to explicitly specify the mapping in packit's config so you could choose what you want to map and where

I'm personally LGTM on this as a short-term solution and figure out something better for the long-term

bocekm:
The possibility to choose the mapping myself in the upstream repo packit config sounds like most appealing to me. But I'll be fine with any solution that will allow me to run tests on CentOS Linux and having access to the Red Hat internal network.

I symphatize with Tomas' POV that considering you don't have anybody now who'd need the current epel-X to rhel-X mapping, you can temporarily change it per this PR to unblock us and implement a long term solution after that.

PR #1227, opened 2021-09-24
WIP: Support Github check re-runs

### LINKED ISSUE #1204: Support re-runs via GitHub's Checks interface
We do not support that right now and need to write code to have it:

- accept those webhooks
- have Event class(es) for them
- Process correctly and submit jobs only to the specific subset of the targets
- Make sure it works for builds and tests

lbarcziova:
This is still WIP, I want to discuss possible changes in implementation before fixing the broken tests

TODO:
- [ ] agree on this implementation
- [ ] better check of app name (currently filtering both stg and prod)
- [ ] check whether all needed info are stored in the events
- [ ] test get_dict of new events -> JSON serializable
- [ ] new integration tests for all events
- [ ] fix broken tests

---

- parse the job and target from the check run name
- in the `external_id` which we set for each check run store the ID of JobTriggerModel from our db
- when parsing the event, from this ID get the original job trigger and needed info (e.g. pr_id for PR, tag_name for release, branch name for commit)
- have `targets_override` property which can be used for other events in future as well (rerunning of specific targets in comment)
- `targets_override` is included in the payload to the handler task and then used in EventData
- mapping of the event to handler via job we get from check run name
- existing handlers (Copr, Koji, TF) only use the `targets_override` from EventData if present and propagate them to builder classes
- I added pre-filtering directly in the service so that we don't process created and completed check runs in the worker, since for each check we create we will process this event in worker again

Fixes #1204 

---

TBD


packit_service/worker/reporting.py
```diff
@@ -61,19 +61,26 @@ def __init__(
...
         self.pr_id: Optional[int] = pr_id
 
     @classmethod
     def get_instance(
-        cls, project: GitProject, commit_sha: str, pr_id: Optional[int] = None
+        cls,
+        project: GitProject,
+        commit_sha: str,
+        trigger_id: int,
+        pr_id: Optional[int] = None,
```
mfocko, line 83: I saw some `trigger_id=None`, this may be better and wouldn't make you update code in `Bugzilla` for example:

```suggestion
        pr_id: Optional[int] = None,
        trigger_id: Optional[int] = None,
```

However I see that you are explicit about the arguments, so I leave the choice up to you… _It is_ easier to read with those arguments you have there :shrug:

mfocko (commented):
lgtm

packit_service/worker/allowlist.py
```diff
@@ -357,6 +364,9 @@ def check_and_report(
                 InstallationEvent,
                 KojiBuildEvent,
                 DistGitCommitEvent,
+                CheckRerunPullRequestEvent,
+                CheckRerunReleaseEvent,
+                CheckRerunCommitEvent,
             ): self._check_unchecked_event,
```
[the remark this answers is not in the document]
mfocko: Nope, doesn't work…

packit_service/worker/parser.py
```diff
@@ -601,6 +617,111 @@ def parse_pull_request_comment_event(
...
+
+        check_name = nested_get(event, "check_run", "name")
+        logger.info(f"Github check run {check_name} rerun event.")
+
+        deployment = ServiceConfig.get_service_config().deployment
+        app = nested_get(event, "check_run", "app", "slug")
+        if (deployment == Deployment.prod and app != "packit-as-a-service") or (
+            deployment == Deployment.stg and app != "packit-as-a-service-stg"
+        ):
+            logger.warning("Check run not created by this app.")
```
TomasTomecek, line 641: can we please log the name of the app? so we can debug potential issues

packit_service/worker/parser.py
```diff
@@ -601,6 +617,111 @@ def parse_pull_request_comment_event(
...
+                    # e.g. [testing-farm-]fedora-rawhide-x86_64
+                    check_name_target = check_name_job_info[
+                        (len(job_name) + 1) :  # noqa
+                    ]
+
+        if not (check_name_job and check_name_target):
+            logger.warning(
+                "We were not able to parse the job and target from the check run name."
+            )
+            return None
```
TomasTomecek, line 660: also log the variables please (even though they will be visible in sentry, but still, let's have them in logs as well)

TomasTomecek (commented):
impressive! :O

TomasTomecek (approved):
LGTM, looking fwd to stg deployment :rocket:

packit_service/worker/allowlist.py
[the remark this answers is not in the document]
mfocko: But i wonder if it supports inheritance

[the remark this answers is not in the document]
mfocko: It does :tada: you should be able to do just `CheckRerunEvent`

packit_service/worker/jobs.py
```diff
@@ -93,18 +100,31 @@ def get_handlers_for_event(
         ),
     ):
         handlers_triggered_by_comment = get_handlers_for_comment(event.comment)
-    else:
-        handlers_triggered_by_comment = None
+
+    if isinstance(
+        event,
+        (CheckRerunCommitEvent, CheckRerunPullRequestEvent, CheckRerunReleaseEvent),
+    ):
```
mfocko, line 107: same here

```suggestion
if isinstance(event, CheckRerunEvent):
```

packit_service/service/api/webhooks.py
```diff
@@ -88,6 +92,25 @@ def post(self):
 
         return "Webhook accepted. We thank you, Github.", HTTPStatus.ACCEPTED
 
+    @staticmethod
+    def interested():
+        """
+        Check whether we want to process this event.
+        :return: False if we are not interested in this kind of event
```
mfocko, line 99: google style comments :P

```suggestion
        Returns:
            `False` if we are not interested in this kind of event.
```

packit_service/worker/handlers/testing_farm.py
```diff
@@ -111,47 +111,60 @@ def run(self) -> TaskResults:
...
+                    state=BaseCommitStatus.pending,
+                    description="Missing Copr build for this target, running a new Copr build.",
+                    url="",
+                    chroot=missing_target,
+                )
 
             # monitor queued builds
-            for _ in range(len(targets)):
+            for _ in range(len(targets_without_build)):
                 self.pushgateway.copr_builds_queued.inc()
```
mfocko, line 151: It won't count it twice, right? :)

mfocko (approved):
LGTM

PR #1235, opened 2021-10-06
clean up after doing propose-downstream work

TomasTomecek:
the dist-git repositories linger on worker in /tmp


mfocko:
Also L149-L181 would be nice in a separate function something like:
```py
def sync_branch(self, branch, ...) -> Optional[str]: ...
```
which would optionally return the `str(ex)` that is the end of the `except` block right now…

mfocko (approved):
LGTM, thanks

PR #1238, opened 2021-10-07
Define custom buckets for Prometheus histograms

lbarcziova:
We could use these custom buckets for displaying our SLOs in Grafana ([source](https://prometheus.io/docs/practices/histograms/#apdex-score)).
For our current SLOs it would be enough to have (15, inf) for statuses and (43200 (12 h) , inf) for the other 2, should we have also the other buckets for the future?

---
N/A

### DIFF
--- a/packit_service/worker/monitoring.py
+++ b/packit_service/worker/monitoring.py
@@ -66,18 +66,37 @@ class Pushgateway:
             "initial_status_time",
             "Time it takes to set the initial status",
             registry=self.registry,
+            buckets=(5, 15, 30, float("inf")),
         )
 
         self.copr_build_finished_time = Histogram(
             "copr_build_finished_time",
             "Time it takes from setting accepted status for Copr build to finished",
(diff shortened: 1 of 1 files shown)


csomh:
More buckets would allow us to gain a better understanding of build/test/feedback times, although the resolution is probably a matter of personal taste and curiosity :smiley: 

For the feedback time I would have: 5s, 15s, 30s, Inf - I would remove the 10s you've proposed, as I don't really see it brings too much of an additional value. Having the 30s as a bucket before infinite, would help to have a better understanding of how much late the late feedbacks are.

In a similar fashion for build and test times: 30min, 1h, 3h, 6h, 12h, 24h, Inf.

Most importantly: thank you for taking the time and reading up on this, this is indeed the right way to do this!

csomh (approved):
Other than my previous comment, this LGTM.

Not being able to transform already registered metrics makes sens, and I don't think it's an issue.

PR #1248, opened 2021-10-18
Include built_packages in API response for getting a copr build

lbarcziova:
Related to packit/dashboard#118

---
N/A


lbarcziova:
Ok, I will leave it as it is now, thanks!

PR #1256, opened 2021-10-26
Enable running only tests without build phase

### LINKED ISSUE #1078: Enable running tests on repos which only contain FMF-based tests
the use case here is that a project ( https://github.com/RedHat-SP-Security/tests ) contains tests for multiple components and the team would love to run TF tests on every PR: the expectation is that RPMs would come directly from a compose or basically "the latest released versions". This workflow would serve as a way to add new tests and run them during the development.

The tasks for us are:
* allow empty specfile_path
* skip the build phase completely
* submit a new TF job which would installed selected components from the official repos and run FMF tests defined in the repo

lbarcziova:
TODO:
- [x] break the copr-build -> test dependency
- [ ]  verify correctness of the format of payload if skip_build=True
- [x] refactor tests

Fixes #1078

---

TBD

### DIFF
--- a/packit_service/constants.py
+++ b/packit_service/constants.py
@@ -67,6 +67,8 @@ MSG_MORE_DETAILS = "You can find more details about the job [here]({url}).\n\n"
 
 MSG_TABLE_HEADER_WITH_DETAILS = "| Name/Job | URL |\n" "| --- | --- |\n"
 
+REQUESTED_PULL_REQUEST_COMMENT = "/packit"
+
 
 class KojiBuildState(Enum):
     """
--- a/packit_service/worker/handlers/abstract.py
+++ b/packit_service/worker/handlers/abstract.py
@@ -11,7 +11,7 @@ from collections import defaultdict
 from datetime import datetime
 from os import getenv
 from pathlib import Path
-from typing import Dict, Optional, Set, Type
+from typing import Dict, Optional, Set, Type, List
 
 from celery import signature
 from celery.canvas import Signature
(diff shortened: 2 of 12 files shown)


TomasTomecek (approved):
let's try on stg :)

packit_service/worker/handlers/abstract.py
```diff
@@ -196,6 +197,26 @@ def add_topic(kls: Type["FedmsgHandler"]):
...
+        return []
+
+    comment_lines = comment_parts.split("\n")
+
+    for line in filter(None, map(str.strip, comment_lines)):
+        (packit_mark, *packit_command) = line.split(maxsplit=3)
+        # packit_command[0] has the first cmd and [1] has the second, if needed.
+
+        if packit_mark == REQUESTED_PULL_REQUEST_COMMENT:
+            if packit_command:
```
mfocko, line 214:
```suggestion
        if packit_mark == REQUESTED_PULL_REQUEST_COMMENT and packit_command:
```

packit_service/constants.py
```diff
@@ -67,6 +67,8 @@
 
 MSG_TABLE_HEADER_WITH_DETAILS = "| Name/Job | URL |\n" "| --- | --- |\n"
 
+REQUESTED_PULL_REQUEST_COMMENT = "/packit"
```
mfocko, line 70: Could we potentially bend this in future to be able to trigger only stage or only prod? :thinking:

lachmanfrantisek (approved):
Looks really good! :rocket:

lachmanfrantisek:
@lbarcziova thanks for working on this! :rocket: 

What about having a PR on hello-world with this usecase?

TomasTomecek:
We should also document this. Once in prod, we can even remove the spec file in RedHat-SP-Security/tests and finally, I'd love to announce this to QE (together with @thrix) since this is a big feature.

lbarcziova:
> What about having a PR on hello-world with this usecase?

Definitely, I had it in the plan.

I would like to properly test it at first (also existing functionality, since there were refactoring changes) and then if we see it works as expected, add documentation and announce it.

PR #1265, opened 2021-11-02
TF runs: fill submitted_time if missing

TomasTomecek:
...and don't traceback on the fact that submitted_time is not filled
since we use it to compute timedelta

If submitted_time is not filled in, we obtain it from TF's result,
"created" field and populate it in the TF Result handler.

Fixes
https://sentry.io/organizations/red-hat-0p/issues/2762340421/?environment=prod&project=1767823

---

N/A, internal change

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -58,7 +58,7 @@ repos:
     hooks:
       - id: check-rebase
         args:
-          - git://github.com/packit-service/packit-service.git
+          - https://github.com/packit/packit-service.git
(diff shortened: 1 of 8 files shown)


lachmanfrantisek (approved):

FrNecas (approved):
LGTM, thanks for fixing this after me :D

TomasTomecek:
> LGTM, thanks for fixing this after me :D

No worries, was glad to do some p-s coding after a long time

PR #1266, opened 2021-11-02
Fail when event is triggered by `\packit` command without packit config file in repo

### LINKED ISSUE #765: Give feedback if Packit command is in comment, but Packit is not configured
See [this PR](https://github.com/avocado-framework/avocado-vt/pull/2597) as an example: `/packit copr-build` was triggered, but there is no Packit configuration in the repository (though the application seems to be enabled on the namespace).

In these cases Packit should comment telling users the configuration is missing and pointing them to the relevant documentation on how to do the configuration.

nikromen:
If event is triggered by `\packit` command and there is no packit config file in repository, let user know that we require it.

TODO:

- [x] tests
- [ ] don't check every comment twice
- [x] mypy complains a lot - solve it
- [x] AbstractCommentEvent?

Also this implementation has a few issues...
- I don't like that I am checking every comment twice for handlers triggered by comment. I can think of two solutions but both of them are quite messy for me... any ideas?
    - when I call `get_handlers_triggered_by_comment` for first time, just store those job handlers to class attribute to remember them and I can use them later.
    - remember those handlers as function arguments
- I don't want to check if event is somewhat \*Comment\* type of event every time because it is like 10-lines if statement. But if it is separated in function, mypy don't know (and obviously can't know) that I am operating only with \*Comment\* type of event outside this function. 
What about implementing `AbstractCommentEvent` class? So I can check whether event is  \*Comment\* type much cleaner than checking it for every subtype of comment.
Now we want to check this in the code like 3 times but this can be really handy in the future because if we want to provide more feedback to users via comments and operate more with comments, we may want some easier way to check whether event is \*Comment\* type. 
And also I can store in this class `handlers_triggered_by_comment` so this will solve my problem with checking every comment twice.

Fixes #765 

---

N/A

### DIFF
--- a/packit_service/worker/events/event.py
+++ b/packit_service/worker/events/event.py
@@ -294,6 +294,7 @@ class AbstractForgeIndependentEvent(Event):
         super().__init__(created_at)
         self.project_url = project_url
         self._pr_id = pr_id
+        self.fail_when_config_file_missing = False
 
         # Lazy properties
         self._project: Optional[GitProject] = None
@@ -359,7 +360,7 @@ class AbstractForgeIndependentEvent(Event):
             project=self.project,
             reference=self.commit_sha,
             pr_id=self.pr_id,
-            fail_when_missing=False,
+            fail_when_missing=self.fail_when_config_file_missing,
             spec_file_path=spec_path,
         )
 
--- a/packit_service/worker/events/event.py
+++ b/packit_service/worker/events/event.py
@@ -9,7 +9,7 @@ from datetime import datetime, timezone
 from logging import getLogger
 from typing import Dict, Optional, Union, Set, List
 
-from ogr.abstract import GitProject
+from ogr.abstract import GitProject, Comment
 from ogr.services.pagure import PagureProject
 from packit.config import PackageConfig
 
@@ -376,3 +376,22 @@ class AbstractForgeIndependentEvent(Event):
         result.pop("_base_project")
         result.pop("_package_config")
         return result
+
+
+class AbstractCommentEvent(AbstractForgeIndependentEvent):
+    def __init__(
+        self,
+        project_url: str,
+        comment: str,
+        pr_id: Optional[int] = None,
+        comment_object: Optional[Comment] = None,
+    ) -> None:
+        super().__init__(project_url=project_url, pr_id=pr_id)
+        self.comment = comment
+
+        # Lazy properties
+        self._comment_object = comment_object
+
+    @property
+    def comment_object(self) -> Optional[Comment]:
+        raise NotImplementedError()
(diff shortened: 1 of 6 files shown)


packit_service/worker/jobs.py
```diff
@@ -58,6 +58,22 @@
...
+    if isinstance(
+        event,
+        (
+            PullRequestCommentGithubEvent,
+            PullRequestCommentPagureEvent,
+            IssueCommentEvent,
+            MergeRequestCommentGitlabEvent,
+            IssueCommentGitlabEvent,
+        ),
+    ):
```
lachmanfrantisek, line 71: > What about implementing AbstractCommentEvent class? So I can check whether event is *Comment* type much cleaner than checking it for every subtype of comment.

:+1: Looks good to me.

packit_service/worker/jobs.py
```diff
@@ -264,6 +270,10 @@ def process_jobs(self, event: Event) -> List[TaskResults]:
         """
         Create a Celery task for a job handler (if trigger matches) for every job defined in config.
         """
+        if get_handlers_triggered_by_comment(event):
+            # if event is triggered by `/packit` command, we require packit config file
+            # (mypy can't know that we get here only if `event` is `*Comment*` type)
+            event.fail_when_config_file_missing = True  # type: ignore
```
lachmanfrantisek, line 276: I would try to avoid this. Such things can be hard to debug. And when reading the code, it's not super clear why we are getting handlers (and don't use them).

TomasTomecek (approved):

packit_service/worker/events/pagure.py
```diff
@@ -58,7 +63,7 @@ def __init__(
         comment: str,
         commit_sha: str = "",
     ):
```
jpopelka, line 65: Why not adding  `comment_object: Optional[Comment] = None,` also here?

nikromen: the thing is that Pagure API doesn't support getting comment by its id so we can't get comment_object for pagure

jpopelka (approved):

packit_service/worker/events/event.py
```diff
@@ -359,7 +360,7 @@ def get_package_config(self) -> Optional[PackageConfig]:
             project=self.project,
             reference=self.commit_sha,
             pr_id=self.pr_id,
-            fail_when_missing=False,
+            fail_when_missing=self.fail_when_config_file_missing,
             spec_file_path=spec_path,
         )
```
lachmanfrantisek, line 365: What about adding an optional argument to this method instead of introducing a new attribute?

(Just an idea to make this more readable...)

PR #1267, opened 2021-11-02
Do not expect copr builds in DB to exist

FrNecas:
Some builds in production may be old and already deleted so checking
their status through the copr API may fail.

Fixes the unhandled exception in prod

Related to

Merge before/after

---
N/A

### DIFF
--- a/packit_service/worker/build/babysit.py
+++ b/packit_service/worker/build/babysit.py
@@ -6,6 +6,7 @@ import datetime
 import logging
 from typing import Iterable
 
+import copr.v3
 import requests
 from copr.v3 import Client as CoprClient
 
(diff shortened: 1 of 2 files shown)


jpopelka (approved):

packit_service/worker/build/babysit.py
```diff
@@ -142,7 +143,16 @@ def update_copr_builds(build_id: int, builds: Iterable["CoprBuildModel"]) -> boo
         bool: Whether the run was successful, False signals the need to retry.
     """
     copr_client = CoprClient.create_from_config_file()
-    build_copr = copr_client.build_proxy.get(build_id)
+    try:
+        build_copr = copr_client.build_proxy.get(build_id)
+    except copr.v3.CoprNoResultException:
+        logger.error(
```
TomasTomecek, line 149: please change this to info

logger.error lines end up in our sentry and this is definitely not an error which needs to be acted upon - as you say, builds are gone, nothing we can do about it

thanks!

TomasTomecek (changes requested):

packit_service/worker/build/babysit.py
FrNecas: Thanks for pointing this out, good to know :) I will push the fix once https://github.com/packit/packit-service/pull/1268 merges so that I don't have to rebase twice...

TomasTomecek (approved):
thanks Franto, very nice work!

PR #1268, opened 2021-11-02
Use read-only PR for mock testing

FrNecas:
As @mfocko described in https://github.com/packit/ogr/pull/652#discussion_r740103503 , the constructor for `PullRequest` was supposed to be changed (I will hopefully get around to refactoring it soon). Anyway, for the use-case of testing with flexmock, it seems that the read-only PR object should be the way to go so modifying in advance.

Fixes

Related to https://github.com/packit/ogr/pull/652

Merge before/after

---

N/A


mfocko (approved):
lgtm

TomasTomecek (approved):

Issue #1230, opened 2021-09-24
Enable users configure mapping between build and test targets


TomasTomecek:
The mapping between build and test targets is usually 1:1, e.g. fedora-34 -> fedora-34

But for the enterprise linux it's a bit more tricky since different distros have the same ABI (epel, centos, centos-stream, rhel, rocky, oraclelinux...). For these especially, we should have a way inside packit config to define build targets and map those to test targets, since we have 2 deployments of TF (public, internal) which offer different composes and evolve on the fly.

TODO:

- [x] propose packit config changes which would allow the 1-to-n mapping
- [x] implement 1-to-1 mapping in p-s
- [x] make sure it works properly for convert2rhel
- [ ] implement 1-to-n mapping in p-s

Context: https://github.com/packit/packit-service/pull/1225

mfocko:
I think this becomes „a bit“ more complex once we start taking into account public **and** internal Copr (as we have spike for internal Copr in backlog too) **and also** public and internal Testing Farm instance… Ideally I would imagine some mapping via to instances via alias or hostname at least…

TomasTomecek:
I like the structure - the 1:n mapping

would `distro_mapping` be a more descriptive name of the field? or `target[s]_mapping`

jpopelka:
> would `distro_mapping` be a more descriptive name of the field? or `target[s]_mapping`

Prior to `targets_distros` I had `from_targets_mapping` and `target2distros_mapping`, but I removed the `mapping`, because it just means "a dictionary". We also don't have `targets_list`, but only `targets`. But yeah, I can return it back so it's more obvious what it is.

Anyone else any suggestions?

lachmanfrantisek:
What something containing `build_to_test`?

jpopelka:
ok, then `build_targets_to_test_distros` if that's not too long :-)

mfocko:
I think we should be fine with `targets_distros` or `targets2distros` as long as we have it described in documentation. I would also consider an option like:

```yml
- job: tests
  trigger: pull_request
  metadata:
    targets:
    - epel-7-x86_64
      distros: [centos-7, rhel-7]
    - epel-8-x86_64
      distros: [centos-stream-8]
    - epel-8-aarch64
  use_internal_tf: True
```

But that might be more complicated to implement…

jpopelka:
> I think we should be fine with `targets_distros` or `targets2distros` as long as we have it described in documentation.

That's what I thought too.

> ```yaml
[...]

PR #1269, opened 2021-11-08
Take distro (to test build on) from user's config

jpopelka:
This is the first part of #1230
The 1-to-n mapping will be in a separate commit since it will be much more work (and nobody needs it at the moment).

Related to: #1230 

Merge after: packit/packit#1407

---

It is possible to specify in the packit config which distro to test a build on in Testing Farm.


jpopelka:
Tests are expected to fail now until packit/packit#1407 is merged.

packit_service/worker/testing_farm.py
```diff
@@ -207,15 +207,19 @@ def chroot2distro_arch(self, chroot: str) -> Tuple[str, str]:
...
         distro, arch = chroot.rsplit("-", 1)
 
-        epel_mapping = {
-            "epel-6": "centos-6",
-            "epel-7": "centos-7",
-        }
+        distros = self.job_config.metadata.targets_dict.get(chroot, {}).get("distros")
+        if distros:
+            # For now take just the first one
+            return distros[0], arch
```
lachmanfrantisek, line 213: What about [assignment expression](https://www.python.org/dev/peps/pep-0572/)?

```suggestion
        if distros := self.job_config.metadata.targets_dict.get(chroot, {}).get("distros")
            # For now take just the first one
            return distros[0], arch
```

lachmanfrantisek (approved):
Looks good!

packit_service/worker/testing_farm.py
jpopelka: Ah, actually black reformatted it into
```python
        if distros := self.job_config.metadata.targets_dict.get(chroot, {}).get(
            "distros"
        ):
```
in which case I think the old
```python
        distros = self.job_config.metadata.targets_dict.get(chroot, {}).get("distros")
        if distros:
```
is a bit nicer and easier to "parse". So I'd probably leave it as it is.

jpopelka:
recheck

PR #1272, opened 2021-11-09
Fix checking the status of Testing Farm pipelines

csomh:
Packit Service implementation mixes the pipeline state and result values
into a single enum, called TestingFarmResult. This lead to a
misunderstanding when determining the state of pipelines in the
baby-sitting task.

Fix this by first parsing the response received from Testing Farm, and
then looking at the value of 'result' to tell the state.

### DIFF
--- a/packit_service/worker/build/babysit.py
+++ b/packit_service/worker/build/babysit.py
@@ -58,13 +58,8 @@ def check_pending_testing_farm_runs() -> None:
             )
             run.set_status(TestingFarmResult.error)
             continue
-        details = response.json()
-        status = TestingFarmResult(details.get("state"))
-        logger.info(f"The status is {status}")
-        if status == TestingFarmResult.running:
-            logger.info(f"Pipeline {run.pipeline_id} is still running")
-            continue
 
+        details = response.json()
         (
             project_url,
             ref,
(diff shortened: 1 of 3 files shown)


csomh:
On the long run: should we spend some time to separate state and result on our side, too, in order to avoid such confusions in the future?

packit_service/worker/build/babysit.py
```diff
@@ -76,6 +71,13 @@ def check_pending_testing_farm_runs() -> None:
             created,
         ) = Parser.parse_data_from_testing_farm(run, details)
 
+        logger.info(
+            f"Result for the testing farm pipeline {run.pipeline_id} is {result}."
+        )
+        if result == TestingFarmResult.running:
```
csomh, line 77: To me it seems that this should also include "new" and "queued", as these are also _not completed_ states. Opinions?

FrNecas: Good catch, using `result in (TestingFarmResult.new, TestingFarmResult.queued, TestingFarmResult.running)` makes more sense

csomh: Updated. PTAL.

tests/integration/test_babysit.py
```diff
@@ -295,14 +295,18 @@ def test_check_pending_testing_farm_runs(created):
...
 
-def test_check_pending_testing_farm_runs_timeout():
+@pytest.mark.parametrize(
+    "status",
+    [TestingFarmResult.new, TestingFarmResult.queued, TestingFarmResult.running],
+)
+def test_check_pending_testing_farm_runs_timeout(status):
     run = flexmock(
         pipeline_id=1,
         status=TestingFarmResult.running,
```
FrNecas, line 305: I may be missing something but I don't see the `status` argument used anywhere in the test. Should this rather be `status=status`?

csomh: Indeed. Forgot to update that line. Thanks for noticing!

FrNecas (approved):
LGTM

Issue #1217, opened 2021-09-20
Environment variables in the Testing Farm


lachmanfrantisek:
The task is to sent variables to our test runner -- Testing Farm -- when submitting the tests.

See the https://testing-farm.gitlab.io/api/ for the API definition (`requests` -> `environments` -> `variables`).

This is the code where we create the payload:
https://github.com/packit/packit-service/blob/ab0b84224de2535a3b500267e0e2e3f96e845620/packit_service/worker/testing_farm.py#L115-L134

We would like to send:
* [ ] Some pre-defined variables (project name, project URL, the git hash, ... please add more).
* [ ] User-defined env. vars (add new config option in the packit repo)

PR #1275, opened 2021-11-09
Pass environment variables to Testing Farm

FrNecas:
Fixes https://github.com/packit/packit-service/issues/1217

Related to

Merge after https://github.com/packit/packit/pull/1411 (though this solution should be backward compatible)

---
Environment variables from config are now passed to Testing Farm along with some more pre-defined variables (e.g. name of the project, URL, etc).

### DIFF
--- a/packit_service/worker/testing_farm.py
+++ b/packit_service/worker/testing_farm.py
@@ -126,9 +126,13 @@ class TestingFarmJobHelper(CoprBuildJobHelper):
         return artifact
 
     def _payload(
-        self, chroot: str, artifact: Optional[Dict[str, Union[List[str], str]]] = None
+        self,
+        chroot: str,
+        artifact: Optional[Dict[str, Union[List[str], str]]] = None,
+        build_id: Optional[int] = None,
     ) -> dict:
-        """
+        """Prepare a Testing Farm request payload.
+
         Testing Farm API: https://testing-farm.gitlab.io/api/
 
         Currently we use the same secret to authenticate both,
@@ -136,6 +140,10 @@ class TestingFarmJobHelper(CoprBuildJobHelper):
         and testing farm (when sending notification to packit service's webhook).
         We might later use a different secret for those use cases.
 
+        Args:
+            chroot: Target TF chroot.
+            build_id: ID of the related copr build.
+            artifact: Optional artifacts, e.g. list of package NEVRAs
         """
         distro, arch = self.chroot2distro_arch(chroot)
         compose = self.distro2compose(distro, arch)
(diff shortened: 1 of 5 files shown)


packit_service/worker/testing_farm.py
```diff
@@ -143,10 +143,29 @@ def _payload(
...
+        predefined_environment = {
+            "full_repo_name": self.project.full_repo_name,
+            "upstream_name": self.job_config.upstream_package_name,
+            "upstream_url": self.job_config.upstream_project_url,
+            "downstream_name": self.job_config.downstream_package_name,
+            "downstream_url": self.job_config.downstream_project_url
+            if self.job_config.downstream_package_name
+            else None,
+            "commit_sha": self.metadata.commit_sha,
+        }
```
lachmanfrantisek, line 155: I would use capital letters and maybe a prefix like `PACKIT_`. What do you think?

packit_service/worker/testing_farm.py
```diff
@@ -143,10 +143,29 @@ def _payload(
...
+        predefined_environment = {
+            k: v for k, v in predefined_environment.items() if v is not None
+        }
+        # User-defined variables have priority
+        metadata = self.job_config.metadata
+        env_variables = metadata.environment if hasattr(metadata, "environment") else {}
+        predefined_environment.update(env_variables)
```
lachmanfrantisek, line 162: I am not sure if overriding is a feature or can cause some problems..;)
(But I can't think of any issue with this for now.)

lachmanfrantisek (commented):
Nice!

jpopelka (commented):
looks OK

packit_service/worker/testing_farm.py
FrNecas: I was wondering about the same thing. The problem is that you somehow have to merge it, one way or the other. I feel like that if a user sets a variable in their config they expect it there so I'd rather not override it even if it clashes with our pre-defined variable. Your other suggestion for naming (capitalizing + prefixing `PACKIT_`) should almost completely remove the possibility of this hapening though :)

packit_service/worker/testing_farm.py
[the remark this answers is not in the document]
FrNecas: Added. Was not completely sure what's the best way to get NVR. Whether it's better to take the base name of SRPM from srpm_build (removing the extension) or constructing it based on a copr build. I ended up going with the latter for now.

packit_service/worker/testing_farm.py
```diff
@@ -126,27 +126,68 @@ def _artifact(
...
+
+        predefined_environment = {
+            "PACKIT_FULL_REPO_NAME": self.project.full_repo_name,
+            "PACKIT_UPSTREAM_NAME": self.job_config.upstream_package_name,
+            "PACKIT_UPSTREAM_URL": self.job_config.upstream_project_url,
+            "PACKIT_DOWNSTREAM_NAME": self.job_config.downstream_package_name,
+            "PACKIT_DOWNSTREAM_URL": self.job_config.downstream_project_url
+            if self.job_config.downstream_package_name
+            else None,
+            "PACKIT_PACKAGE_NAME": self.job_config.downstream_package_name,
```
lachmanfrantisek, line 172: As I see it now, we probably don't need both `PACKIT_DOWNSTREAM_NAME` and `PACKIT_PACKAGE_NAME`.

lachmanfrantisek (approved):
Thanks! Looks good to me. We can add more variables later if needed (and requested).

packit_service/worker/testing_farm.py
FrNecas: Which one do you find better? I think that `PACKIT_DOWNSTREAM_NAME` makes a bit more sense since we also have `PACKIT_UPSTREAM_NAME`. On the other hand, I don't see anything wrong with keeping both and letting the users decide whichever they prefer (some may prefer the term `downstream`, others `package`).

lachmanfrantisek: I agree, I would use `PACKIT_DOWNSTREAM_NAME` or both. I'll leave it up to you...;)

FrNecas: Lets keep both then. I am merging this PR and will hopefully be able to see on staging if everything is fine.

packit_service/worker/testing_farm.py
```diff
@@ -126,27 +126,68 @@ def _artifact(
...
         fmf = {"url": self.fmf_url}
         if self.fmf_ref:
             fmf["ref"] = self.fmf_ref
 
+        if build_id is not None:
+            copr_build = CoprBuildModel.get_by_build_id(build_id)
+            build_log_url = copr_build.build_logs_url
+            srpm_build = copr_build.get_srpm_build()
+            srpm_url = srpm_build.url
+            nvr_data = copr_build.built_packages[0]
```
FrNecas, line 159: Hm, `copr_build.built_packages` can apparently be None even if the `build_id` is specified, weird. Will try to debug what's going on tomorrow.

lbarcziova: From quickly looking into it, the problem can be we get the `copr_build` object from calling [`CoprBuildModel.get_by_build_id(build_id)`](https://github.com/packit/packit-service/blob/7accfee9b5a23583ca921616638ff067f2b5c6dc/packit_service/models.py#L940), but we don't specify the target, so the first matching `CoprBuildModel` is taken, which may not have set the `built_packages`. So the solution could be using directly `CoprBuildModel` object we already have or specify the target when  calling `get_by_build_id`.

Issue #1276, opened 2021-11-10
Fix handling with RunModel when there is no associated build


lbarcziova:
After a successful run of tests without the build phase in packit/hello-world#535, when trying to display pipelines this error occurs:
```
File "/usr/local/lib/python3.9/site-packages/packit_service/service/api/runs.py", line 80, in get
  result = process_runs(RunModel.get_merged_chroots(first, last)), referer: https://dashboard.stg.packit.dev/
File "/usr/local/lib/python3.9/site-packages/packit_service/service/api/runs.py", line 42, in process_runs
 "time_submitted": optional_timestamp(srpm_build.build_submitted_time)
 AttributeError: 'NoneType' object has no attribute 'build_submitted_time'
```

PR #1279, opened 2021-11-11
Process RunModel correctly if there is no associated build

lbarcziova:
Related to #1276 

---

N/A

### DIFF
--- a/packit_service/service/api/runs.py
+++ b/packit_service/service/api/runs.py
@@ -15,7 +15,10 @@ from packit_service.models import (
     optional_timestamp,
 )
 from packit_service.service.api.parsers import indices, pagination_arguments
-from packit_service.service.api.utils import get_project_info_from_build, response_maker
+from packit_service.service.api.utils import (
+    get_project_info_from_build,
+    response_maker,
+)
 
 logger = getLogger("packit_service")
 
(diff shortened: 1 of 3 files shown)


packit_service/service/api/runs.py
```diff
@@ -36,20 +36,25 @@ def process_runs(runs):
     result = []
 
     for pipeline in runs:
```
mfocko, line 38: just a heads up it's really temporary fix, cause you get already grouped by rows in `runs` ⇒ if you have two different projects triggering test without build in a row, they will get grouped together…

---

HOWEVER! Now that I'm thinking about it, we have `job_trigger` relation too, so it would probably make more sense to group them by the `job_trigger_id` (that must be set even for tests-only runs, right?). So that would resolve our issue with empty SRPM field.

`<already done in PR, I should try look ahead sometimes>`
And now we just have to decide where to get the `"trigger"` key from, but that is easy, since it works for all models (SRPM, Copr, Koji, TF), so it's just workflow like:

```
FOREACH job_type OF { copr, koji, testing_farm } DO
  FOREACH packit_id OF … DO
    // yada-yada-yada we do already
    IF trigger NOT IN response_dict THEN
      response_dict[trigger] <- get_project_info_from_build(row)
    FI
  OD
OD
```

and it also sounds like Friday work, since `get_project_info_from_build` supports testing farm too :upside_down_face:

`</already done in PR, I should try look ahead sometimes>`

packit_service/service/api/runs.py
```diff
@@ -64,6 +69,11 @@ def process_runs(runs):
                         "status": row.status,
                     }
                 )
+                if model_type == "test_run" and not srpm_build:
```
mfocko, line 72: oh you got it already… I would probably go with the condition above, so it's more generic and it would ideally catch the first build/test of the pipeline, might be helpful for SRPMs in Copr.

mfocko (approved):
lgtm

lachmanfrantisek (approved):
:+1: Thanks!

packit_service/service/api/runs.py
lbarcziova: thanks for looking into it, I changed the condition to be more generic, PTAL

mfocko (approved):

PR #1280, opened 2021-11-11
Update the enum types in the DB

csomh:
Some newer enum values are missing from AllowlistStatus and
TestingFramResult. Add a migration to update them according to the
definition in the Python-code.


lbarcziova (approved):
lgtm

PR #1281, opened 2021-11-11
Remove name of the app/service from check names

jpopelka:
The check run names are already quite long and I need to add even more info (test distro) to them.
I checked some checks of various apps and most of them don't contain the name of the app, so I think it's not needed.
We have a different icon for the staging app and in the Checks tab the checks are in sections with the app names.

Also, the job and target are newly separated with ':' instead of '-', because there already are more '-' in most target names and it's more readable this way.

---

The check run names are now shorter and easier to read.

### DIFF
--- a/packit_service/worker/parser.py
+++ b/packit_service/worker/parser.py
@@ -661,10 +661,18 @@ class Parser:
             return None
 
         check_name_job, check_name_target = None, None
-        # e.g. packit/rpm-build-fedora-34-x86_64
-        check_name_split = check_name.split("/", maxsplit=2)
-        if len(check_name_split) == 2:
-            check_name_job_info = check_name_split[1]
+        if ":" in check_name:
+            # e.g. "rpm-build:fedora-34-x86_64"
+            check_name_job, _, check_name_target = check_name.partition(":")
+            if check_name_job not in MAP_CHECK_PREFIX_TO_HANDLER:
+                logger.warning(
+                    f"{check_name_job} not in {list(MAP_CHECK_PREFIX_TO_HANDLER.keys())}"
+                )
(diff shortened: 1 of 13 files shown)


lbarcziova (approved):
lgtm, just not sure about the space after `:`, does it have any purpose?
also, after this change, validation script needs to be updated

jpopelka:
I thought, maybe, it'd be more readable by visually separating the job and target by the additional space, i.e.
`testing-farm: fedora-rawhide-x86_64` instead of `testing-farm:fedora-rawhide-x86_64`
but I see it's too radical :-)
Reverted to just ':'
Thanks for your opinion.

jpopelka:
> validation script needs to be updated

packit/deployment#280

packit_service/worker/parser.py
```diff
@@ -661,17 +661,25 @@ def parse_check_rerun_event(
...
-            check_name_job_info = check_name_split[1]
+        if ":" in check_name:
+            # e.g. "rpm-build:fedora-34-x86_64"
+            check_name_job, _, check_name_target = check_name.partition(":")
+            if check_name_job not in MAP_CHECK_PREFIX_TO_HANDLER:
+                logger.warning(
+                    f"{check_name_job} not in {list(MAP_CHECK_PREFIX_TO_HANDLER.keys())}"
+                )
+                check_name_job = None
+        elif "/" in check_name:
```
mfocko, line 672: I think you could add `TODO` for removal after some time, I doubt it makes sense to keep this in for longer than a month

mfocko (approved):

lachmanfrantisek (approved):

PR #1282, opened 2021-11-11
Improve pipeline grouping

lbarcziova:
If we group only by the SRPM build ID, all tests without build (without the SRPM build ID) are grouped together, which is incorrect. We discussed that until the DB schema is not improved, it is okay to display the tests without build as separate pipelines.

Related to #1276 

---

N/A

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -34,6 +34,8 @@ from sqlalchemy import (
     create_engine,
     desc,
     func,
+    null,
+    case,
 )
 from sqlalchemy.dialects.postgresql import array as psql_array
 from sqlalchemy.ext.declarative import declarative_base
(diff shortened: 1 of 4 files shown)


mfocko:
As I suggested in #1279, it might be better to group via `job_trigger_id` rather than `srpm_build_id`, since that must be set all the time and also I do not think we can have _run_ without any trigger :thinking:

lbarcziova:
But grouping via `job_trigger_id` would change the current behaviour, since e.g. for different pushes to PR there will be only one JobTriggerModel, so all pushes in one PR would be grouped together.

mfocko:
> But grouping via `job_trigger_id` would change the current behaviour, since e.g. for different pushes to PR there will be only one JobTriggerModel, so all pushes in one PR would be grouped together.

I thought it's nested in there, since job trigger model has only type and id to one of the specific trigger models... Will have a look in the morning

lachmanfrantisek (approved):

PR #1291, opened 2021-11-23
Use query.slice() to slice the query on server

jpopelka:
instead of in a client, as done previously.

https://docs.sqlalchemy.org/en/14/orm/query.html?highlight=order_by#sqlalchemy.orm.Query.slice

I built & pushed `stg` image, so you can check it in https://dashboard.stg.packit.dev
I was hoping to see much better speed improvement, but it's quite insignificant.
It's still a good thing to do this on the server instead of in a client, though.

---

N/A

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -276,10 +276,11 @@ class GitProjectModel(Base):
     @classmethod
     def get_projects(cls, first: int, last: int) -> Iterable["GitProjectModel"]:
         with get_sa_session() as session:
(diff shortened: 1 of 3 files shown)


TomasTomecek (approved):
neat

lbarcziova (approved):
nice

mfocko (approved):
I recall looking into it and finding out that `[:]` on the query object is actually using the `.slice` method, but better be explicit about it :)

PR #1299, opened 2021-11-28
Fix typo for getting comment object for GitLab

mfocko:
Fix typo that looks up GitLab MR by its internal ID from database rather
than IID which is used.

Fixes PACKIT-SERVICE-3TB

---


nikromen (approved):
:+1:

TomasTomecek (commented):
Could you please provide a bit of info what are the differences between the two and why IID is better?

mfocko:
> Could you please provide a bit of info what are the differences between the two and why IID is better?

IID is the classic MR id like 1299 here, I'm not exactly sure why they provide ID, but iirc it's from database (GitLab internal usage)

PR #1302, opened 2021-11-30
External contributors can't use internal services

lachmanfrantisek:
* Add user-specific pre-check for handlers to check permissions.
  * This allows us to restrict access for external contributors when using internal services.
* Restrict external contributor running a Test job when `user_internal_tf` is configured.
* All events that provide `actor` attribute are checked now.
---

External contributors can't trigger internal tests initially. Project maintainers need to trigger the action via `/packit test` comment to run the job.

### DIFF
--- a/packit_service/worker/handlers/testing_farm.py
+++ b/packit_service/worker/handlers/testing_farm.py
@@ -96,7 +96,8 @@ class TestingFarmHandler(JobHandler):
 
     def pre_check(self) -> bool:
         return not (
-            self.testing_farm_job_helper.skip_build and self.copr_build_comment_event()
+            self.testing_farm_job_helper.skip_build
+            and self.is_copr_build_comment_event()
         )
 
     @property
@@ -136,18 +137,18 @@ class TestingFarmHandler(JobHandler):
                 PullRequestGithubEvent.__name__,
                 MergeRequestGitlabEvent.__name__,
             )
-            or self.copr_build_comment_event()
+            or self.is_copr_build_comment_event()
         )
 
-    def comment_event(self) -> bool:
+    def is_comment_event(self) -> bool:
         return self.data.event_type in (
             PullRequestCommentGithubEvent.__name__,
             MergeRequestCommentGitlabEvent.__name__,
             PullRequestCommentPagureEvent.__name__,
         )
 
-    def copr_build_comment_event(self) -> bool:
-        return self.comment_event() and get_packit_commands_from_comment(
+    def is_copr_build_comment_event(self) -> bool:
+        return self.is_comment_event() and get_packit_commands_from_comment(
             self.data.event_dict.get("comment")
         )[0] in ("build", "copr-build")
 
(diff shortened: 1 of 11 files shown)


packit_service/worker/handlers/testing_farm.py
```diff
@@ -94,9 +94,21 @@ def __init__(
         self._db_trigger: Optional[AbstractTriggerDbType] = None
         self._testing_farm_job_helper: Optional[TestingFarmJobHelper] = None
 
+    def check_if_contributor_can_run_job_and_report(self, actor: str) -> bool:
+        if self.job_config.metadata.use_internal_tf and not self.project.can_merge_pr(
```
TomasTomecek, line 98: please put the docstring here and explain a bit what is happening here and why

packit_service/worker/jobs.py
```diff
@@ -299,6 +303,21 @@ def process_jobs(self, event: Event) -> List[TaskResults]:
...
+                    event,
+                    (
+                        PullRequestGithubEvent,
+                        PullRequestCommentGithubEvent,
+                        MergeRequestGitlabEvent,
+                        MergeRequestCommentGitlabEvent,
+                        PullRequestCommentPagureEvent,
+                        PullRequestPagureEvent,
+                    ),
+                ) and not handler.check_if_contributor_can_run_job_and_report(
```
TomasTomecek, line 316: I wouldn't mind a small explanation here as well :)

packit_service/worker/handlers/testing_farm.py
```diff
@@ -94,9 +94,21 @@ def __init__(
...
+            actor
+        ):
+            self.testing_farm_job_helper.report_status_to_tests(
+                description=f"{actor} can't run tests internally",
```
TomasTomecek, line 102: sounds like a good-enough solution to me, short-term

not sure if we can display text on the check page stating that the owner of the project needs to retrigger using `/packit test`

TomasTomecek (commented):
Thanks for such a quick fix!

```diff
-External contributors can't use an internal TF instance. Project maintainers need to trigger the action via commit to run the job.
+External contributors can't trigger internal tests initially. Project maintainers need to trigger the action via `/packit test` comment to run the job.
```

I was actually curious first about the `commit` part :D

packit_service/worker/handlers/testing_farm.py
lachmanfrantisek: It's in the parent, but I can add one here as well.

packit_service/worker/handlers/testing_farm.py
lachmanfrantisek: Good point, I'll take a look if we can show more info in the run's UI.

lachmanfrantisek: > short-term

@TomasTomecek that shouldn't be a short term solution...;) With this, we can easily do the same for internal Copr, or anything else. Do you have a better approach so we can do that now?

And regarding the runs UI, this will require edits in the reporter -- I'll send it as a follow-up.

lachmanfrantisek:
@TomasTomecek PTAL

packit_service/worker/handlers/testing_farm.py
TomasTomecek: I thought that long-term we want to figure out a permission system so that upstream maintainers can grant contributors ability to run CI on their own

TomasTomecek (approved):
LGTM, thank you for providing the detailed descriptions!

PR #1304, opened 2021-12-01
Improve EventData.db_trigger

nikromen:
TODO:

- [x] tests

Original thought how to improve this (TODO in previous solution) was about recreating original Event classes so then we can call `db_trigger` on them. I tried that but parameters of these classes are sometimes so different that the result was even more complex. But we can still recreate only their db trigger classes with necessary attributes and call `db_trigger` property on them.

pros:
- previous solution was hardcoded, this is not 
(but still, I can't figure out how to get rid of `needed_attributes_for_abstract_db_trigger` in `_recreate_original_db_trigger` - would be nice to get rid of this too 
    - what about assigning `EventData.__dict__` to `db_trigger_cls.__dict__` [here](https://github.com/packit/packit-service/pull/1304/files#diff-e8a39471e11af02ed27cca0023ecf56a5b2f5266579281b4bce81a370c1631ecR157)?)
- if we want to add new event, we can do this without touching EventData.db_trigger
- it is easier to read EventData.db_trigger property

cons:
- the solution is not as straightforward as previous solution (and maybe harder to debug so I added a few logs) 

My question is - do we want to do it this way? This is still in progress (without tests and maybe not correct) but I don't want to spend time on finishing it if we don't want it :grinning: 

---

N/A

### DIFF
--- a/packit_service/worker/events/event.py
+++ b/packit_service/worker/events/event.py
@@ -5,25 +5,36 @@
 Generic/abstract event classes.
 """
 import copy
+import inspect
 from datetime import datetime, timezone
 from logging import getLogger
-from typing import Dict, Iterable, Optional, Type, Union, Set, List
+from typing import Dict, Iterable, Optional, Type, Union, Set, List, Any
 
 from ogr.abstract import GitProject
 from packit.config import JobConfigTriggerType, PackageConfig
 
+import packit_service.worker.events
 from packit_service.config import PackageConfigGetter, ServiceConfig
-from packit_service.models import (
-    AbstractTriggerDbType,
-    CoprBuildTargetModel,
-    GitBranchModel,
-    IssueModel,
-    ProjectReleaseModel,
-    PullRequestModel,
-    TFTTestRunTargetModel,
+from packit_service.models import CoprBuildTargetModel, TFTTestRunTargetModel
+
+from packit_service.service.db_triggers import (
+    _AddDbTrigger,
+    AddReleaseDbTrigger,
+    AddPullRequestDbTrigger,
+    AddBranchPushDbTrigger,
+    AddIssueDbTrigger,
 )
+from packit_service.models import AbstractTriggerDbType
 
 logger = getLogger(__name__)
+
+
+AbstractAddDbTrigger = Union[
+    AddReleaseDbTrigger,
+    AddPullRequestDbTrigger,
+    AddIssueDbTrigger,
+    AddBranchPushDbTrigger,
+]
 
 
 MAP_EVENT_TO_JOB_CONFIG_TRIGGER_TYPE: Dict[Type["Event"], JobConfigTriggerType] = {}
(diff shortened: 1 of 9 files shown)


packit_service/worker/events/event.py
```diff
@@ -110,61 +107,55 @@ def project(self):
...
+            )
+            return None
+
+        # first object in list is class itself and last is `object` - let's skip it
+        for superclass in event_cls.mro()[1:-1]:
+            if _AddDbTrigger in superclass.__bases__:
+                logger.debug(
+                    f"AddDbTrigger class found: {superclass.__name__} in {event_cls.mro()}"
                 )
+                return superclass()
```
jpopelka, line 130:
```suggestion
                return superclass
```

packit_service/worker/events/event.py
```diff
@@ -110,61 +107,55 @@ def project(self):
...
+            for key, value in needed_attributes_for_abstract_db_trigger.items()
+            if value is not None
+        }
+
+        db_trigger_cls = self._get_db_trigger_class()
+        if db_trigger_cls is None:
+            return None
+
+        db_trigger_cls.__dict__ = non_none_attributes
+        return db_trigger_cls.db_trigger
```
jpopelka, line 158: The `get_or_create` is a class method so if `_get_db_trigger_class()` returns the base/super class we can call its `get_or_create()` and give it the existing attributes as arguments.

```suggestion
        return db_trigger.get_or_create(**non_none_attributes)
```

jpopelka (commented):
I see Github suggested me as a reviewer because the affected code seems to have been added by me, but that's just because I had moved some modules between directories.
I think this functionality/code was mostly modified by @lbarcziova and @lachmanfrantisek so I'd ask them to review :-)

But if you ask me: I'm not against the approach, if the methods are nicely commented then I think it's better than the hardcoded stuff we have there currently.

I did some suggestions, hopefully, it'll help, if not, ignore :-)

lachmanfrantisek (commented):
I really like that we can get rid of the hardcoded mappings but this feels a bit magical to me to be honest.

I am a bit worried that it will be hard for newcomers and also hard to maintain. (We've made an experience with requre on this low-level field of Python and it tends to change a bit from version to version.)

I am also wondering how we can be sure this works as expected and how to realise any possible change. At least, I would like to see some test cases.

What about others? Do you think the benefit is worth having a little magic in our code? :magic_wand:

nikromen:
I rethinked the implementation once more and reduced the magic in the code even more :D hopefully it's good enough now
mypy won't pass for now but current implementation is how I roughly see the solution now.
Of cource, there will be test enforcing the name scheme/relationship of EventData - AddDbTrigger mixins

@lachmanfrantisek PTAL

csomh:
I'm wondering what is the status of this? Does it need review? Does it need some more work?

PR #1305, opened 2021-12-02
Markdown content in reports

lachmanfrantisek:
Add optional `markdown_content` argument to report classes.

* Only GitHub check runs support this.
* It's ignored in other implementations.

---

N/A


lbarcziova (approved):
thanks!

PR #1315, opened 2021-12-17
Discard old SRPM build logs and artifact URLs

jpopelka:
We're doing this because the SRPM build logs take the most space in the database and we don't need the old logs.

Periodically (daily at 1AM) check SRPM builds in database and discard logs and urls to artifacts (uploaded .src.rpm files) at copr
of the builds which are older than `SRPMBUILDS_OUTDATED_AFTER_DAYS`.
The `SRPMBUILDS_OUTDATED_AFTER_DAYS` can be set also as env. var.

The artifact urls could be removed even sooner because copr deletes the artifacts after 2 weeks.
We could also do it similar way copr does, i.e. leave the latest (for each PR) srpm (build log, URL) indefinitely and discard older builds. But we might soon be building srpms in copr so it's probably not worth the effort at the moment.

---

Packit service now discards old (currently, this means 3 months) SRPM builds logs.

### DIFF
--- a/packit_service/worker/database.py
+++ b/packit_service/worker/database.py
@@ -1,0 +1,25 @@
+from datetime import timedelta, datetime
+from logging import getLogger
+from os import getenv
+
+from packit_service.constants import SRPMBUILDS_OUTDATED_AFTER_DAYS
+from packit_service.models import SRPMBuildModel
+
+logger = getLogger(__name__)
+
+
+def discard_old_srpm_build_logs():
+    """Called periodically (see celery_config.py) to discard logs of old SRPM builds."""
+    logger.debug("About to discard old SRPM build logs & artifact urls.")
+    outdated_after_days = getenv(
+        "SRPMBUILDS_OUTDATED_AFTER_DAYS", SRPMBUILDS_OUTDATED_AFTER_DAYS
+    )
+    ago = timedelta(days=int(outdated_after_days))
+    for build in SRPMBuildModel.get_older_than(ago):
+        age = datetime.utcnow() - build.build_submitted_time
+        logger.debug(
+            f"SRPM build {build.id}, age '{age}' is older than '{ago}'. "
+            "Discarding log and artifact url."
+        )
+        build.set_logs(None)
+        build.set_url(None)
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -35,7 +35,7 @@ services:
 
   postgres:
     container_name: postgres
-    image: quay.io/centos7/postgresql-12-centos7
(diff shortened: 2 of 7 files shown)


packit_service/worker/database.py
```diff
@@ -0,0 +1,23 @@
...
+    outdated_after_days = getenv(
+        "SRPMBUILDS_OUTDATED_AFTER_DAYS", SRPMBUILDS_OUTDATED_AFTER_DAYS
+    )
+    ago = timedelta(days=outdated_after_days)
+    for build in SRPMBuildModel.get_older_than(ago):
+        age = datetime.utcnow() - build.build_submitted_time
+        logger.debug(
+            f"SRPM build {build.id}, age '{age}' is older than '{ago}'. Discarding logs."
+        )
+        build.set_logs("")
```
lachmanfrantisek, line 23: How is the value represented in the database? (`None` vs `""` vs `"log removed"`)

lachmanfrantisek (approved):
It's that easy to add a periodic task? Nice!

packit_service/worker/database.py
jpopelka: Not sure I understand the question. [SRPMBuildModel.logs is a Text type](https://github.com/packit/packit-service/blob/c79d591709281503c11ae199c91598d3f471bcdf/packit_service/models.py#L1195) so I guess ""  is represented as an empty string in the database. I don't think we can put `None` there and if you're suggesting to put some explanation (e.g. "Logs have been removed") in place of the original logs, well, that's certainly doable.

lachmanfrantisek: Yes, I was thinking about something like that. (But am still not sure how the text is represented -- e.g. if  `""` and `"not available"` differs in disk space.)

If the explanation is doable, that would be nice. Or, use an empty string and show `Logs not available` in the dashboard (probably easier to do and saves some space).

lachmanfrantisek: Oh, you've already done that..;) Thanks

lachmanfrantisek (approved):

packit_service/worker/database.py
jpopelka: I completely forgot about the `NULL` value, which is what we want, I think.
I've added a commit, which sets `logs` and `url` to `NULL` for the outdated SRPM builds.
And we'll show some message in the dashboard if the log is missing (it already handles missing/empty `url`).

I'll squash those commits later if you agree it's better than storing empty string and/or "not available".

lachmanfrantisek (approved):
:+1:

jpopelka:
Thanks! Will merge next year :-)

PR #1322, opened 2022-01-12
Send periodic tasks to long-running queue

jpopelka:
They might take some time (if there are lots of tasks to babysit or lots of data to discard) and might take a lot of resources.

I'm not sure where's the line between a short running and a long running task, but I feel these should be classified as long running.

https://stackoverflow.com/questions/16962449/how-to-send-periodic-tasks-to-specific-queue-in-celery

---

N/A


lbarcziova (approved):
that makes sense, thanks!

PR #1331, opened 2022-01-28
docker-compose can be also used with podman

jpopelka:
https://fedoramagazine.org/use-docker-compose-with-podman-to-orchestrate-containers-on-fedora

This commit just amends:
- `docker-compose.yml` to work with `podman` as well
- `CONTRIBUTING.md` to not mention `podman-compose` since it's never worked with our `docker-compose.yml`

TODO:

N/A


lachmanfrantisek (approved):
Thanks Jirko!

Issue #74, opened 2019-08-01
create bodhi update for successful Koji build


msehnout:
The documentation says, that packit can build RPM in koji and propose updates in bodhi. Can packit-as-a-service do this automatically for us?
https://packit.dev/docs/cli/create-bodhi-update/

My idea is like this:

- I push a new tag
- New release on github is created
- koji build is triggered
- if successful, bodhi update is proposed

WDYT?

msehnout:
wontfix? If anything I would prefer postpone :)

jpopelka:
> wontfix?

renamed to `stale` and added `pinned` so that stale bot doesn't bother you in future

dhodovsk:
We plan to implement it in the future, however it is not a priority for us in the next weeks.

Somewhere in the process, there will have to be propose-update to include the changes from github release.

lachmanfrantisek:
TODO:

* [ ] create a new job type for updates
* [ ] create a new handler that will listen to the successful Koji build
    * get the correct state of the dist-git
    * trigger the bodhi update using the Packit's Python API

PR #1332, opened 2022-01-28
Bodhi update for koji build

lachmanfrantisek:
* We are reacting on successful production (=non-scratch) Koji builds.
* The bodhi update is created only when the job is configured for the fedora version (=branch) the build is for.
* I suggest going commit-by-commit when reviewing since there is a class rename making the overall diff a bit messy.

Fixes: #74 
Merge after: packit/packit#1466 (required for tests to pass)

---

There is a new job that you can use to automatically create a Bodhi update once a Koji Build successfully finishes.


packit_service/worker/handlers/bodhi.py
```diff
@@ -0,0 +1,130 @@
...
+    def koji_build_event(self):
+        if not self._koji_build_event:
+            self._koji_build_event = KojiBuildEvent.from_event_dict(
+                self.data.event_dict
+            )
+        return self._koji_build_event
+
+    def pre_check(self) -> bool:
+        """
+        We react only on finished builds (=KojiBuildState.open state)
```
jpopelka, line 68: It's strange that the completed builds are marked as `OPEN` (in fedmsg payload?), but I believe you.

packit_service/worker/handlers/bodhi.py
```diff
@@ -0,0 +1,130 @@
...
+            )
+            return False
+
+        if self.koji_build_event.git_ref not in (
+            configured_branches := get_branches(
+                # TODO: pick a better default since for rawhide,
+                #   update is created automatically
+                *self.job_config.metadata.dist_git_branches,
+                default="rawhide",
+                default_dg_branch="rawhide",
```
jpopelka, line 86: So, if someone omits `dist_git_branches` for `bodhi_update` job, then this handler will try to create an update for the rawhide build as well?
Then setting [`default=None` should make `get_branches()` return empty set](https://github.com/packit/packit/blob/dce0380bd05ad6df09f14a7ca0a23bb36a7d9932/packit/config/aliases.py#L152)?

```suggestion
                *self.job_config.metadata.dist_git_branches,
                # Don't use "rawhide" as default since update for rawhide is created automatically
                default=None,
```

jpopelka (approved):
Looks OK, with some suggestions. Thanks!

packit_service/worker/handlers/bodhi.py
lachmanfrantisek: I was more thinking about using `fedora-stable` alias as a default. What do you think?

TomasTomecek: +1 fedora-stable or require the list the be set

jpopelka: I'd vote for "require the list the be set", but `fedora-stable` is fine as well if it's explicitly stated in the documentation.

lachmanfrantisek: I'll use `fedora-stable` and document it.

lachmanfrantisek:
It looks like they have changed the structure a bit since I had started working on this. Now, it covers both options.

I've also implemented the default to be `fedora-stable`.

PTAL

packit_service/worker/handlers/bodhi.py
lbarcziova: maybe [this](https://fedora-fedmsg.readthedocs.io/en/latest/topics.html#buildsys-build-state-change) is the enum that should be used here? the currently used enum is for [buildsys-task-state-change](https://fedora-fedmsg.readthedocs.io/en/latest/topics.html#buildsys-task-state-change)

lbarcziova (commented):
Very nice!

packit_service/worker/handlers/bodhi.py
lachmanfrantisek: Ah, I might have missed that last time when I was checking that page. Thanks!

lachmanfrantisek:
@jpopelka @lbarcziova Can you please take another look?

jpopelka (approved):
walrus

packit_service/worker/events/koji.py
```diff
@@ -6,42 +6,209 @@
...
+
+    @property
+    def commit_sha(self) -> Optional[str]:  # type:ignore
+        # mypy does not like properties
+        return self._commit_sha
+
+    @property
+    def build_model(self) -> Optional[KojiBuildModel]:
+        if not super().build_model:
+            self._build_model = KojiBuildModel.create(
```
lbarcziova, line 132: would it make sense to have a new boolean field in the model such as scratch or production so that we can differentiate between scratch and non-scratch builds since e.g. status differs for them?

lbarcziova (approved):
only one question, otherwise LGTM!

packit_service/worker/events/koji.py
lachmanfrantisek: That's a good point. I'll send a follow-up if it is ok -- I have a few other small improvements in mind around this.

Issue #1327, opened 2022-01-19
Database: models for group of chroots


lachmanfrantisek:
Follow-up for: https://github.com/packit/research/pull/128

We need to store information about a group of related chroot-builds that were triggered together.

* [ ] Create a new model for each type (Koji, Copr, TF) with a clear name
* [ ] Provide a migration script that tries to group the already present chroot models (by Copr build id, timestamp, ...)
* [ ] Improve/enhance/fix the API endpoint to be able to show both chroot-info and grouped-info.
* [ ] Make sure that the newly created models are utilized in the existing code base

FrNecas:
Result from today's discussion on Fedora chat about model naming, there is a slight change from the initial research doc, the naming should be explicit so using `Group` inside the grouped model name seems to be the best way to go. For example, `CoprBuildTarget` will represent a single build in copr (the word target is used instead of chroot for more clarity) and `CoprBuildGroup` for the grouping model. Similarly for TF & Koji

nikromen:
When providing the database scheme (the first subtask) you can get inspired by ProposeDownstreamModel - ProposeDownstreamTargetModel scheme.

PR #1335, opened 2022-02-01
Improvements to DB model names

### LINKED ISSUE #1326: Database: better names
Follow-up for: https://github.com/packit/research/pull/128/files

Let's pick better names for our database models:

* [ ] Remove word `Model` from model names
* [ ] Rename `RunModel` to `Pipeline`
* [ ] Improve the build/test names (e.g. chroot-related models contain `Chroot` in the name, you can suggest a better rule, but have in mind that we will create grouped models as a next step)
* [ ] Optionally, suggest a better name for any other model
* [ ] Prepare the migration script using alembic (more info [here](https://github.com/packit/packit-service/blob/main/CONTRIBUTING.md#database).

FrNecas:
Summary of the changes:
 - Renamed `RunModel` to `PipelineModel`
 - Renamed `InstallationModel` to `GithubInstallationModel`
 - Renamed Koji, Copr and TF models, they now contain the word `Target` to emphasize that it's a single build/test run (to enable creation of group models, see comment in  #1327 )

Fixes #1326 

---
N/A

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -663,7 +663,7 @@ class JobTriggerModel(Base):
       (e.g. For each push to PR, there will be new PipelineModel, but same JobTriggerModel.)
     """
 
-    __tablename__ = "build_triggers"
+    __tablename__ = "job_triggers"
     id = Column(Integer, primary_key=True)  # our database PK
     type = Column(Enum(JobTriggerModelType))
     trigger_id = Column(Integer)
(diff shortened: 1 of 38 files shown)


packit_service/models.py
```diff
@@ -701,21 +693,21 @@ def get_trigger_object(self) -> Optional[AbstractTriggerDbType]:
...
-    * Each RunModel has to be connected to exactly one JobTriggerModel.
-    * There can be multiple RunModels for one JobTriggerModel.
-      (e.g. For each push to PR, there will be new RunModel, but same JobTriggerModel.)
+    * Each build/test model can be connected to multiple Pipelines (e.g. on retrigger).
+    * Each Pipeline has to be connected to exactly one JobTrigger.
+    * There can be multiple Pipelines for one JobTrigger.
+      (e.g. For each push to PR, there will be new Pipeline, but same JobTrigger.)
     """
 
     __tablename__ = "runs"
```
lachmanfrantisek, line 713: If you don't change the `__tablename__`, no change in a database is needed. Maybe, you need to rename only tables, run the alembic and then the class names so alembic can connect the models together.

In case of issues, I would ask @mfocko since he has done some renaming already:
https://github.com/packit/packit-service/blob/main/alembic/versions/70444197d206_rename_whitelist_to_allowlist.py

FrNecas: Hm, no luck with having alembic do the changes yet, it either fails or just removes the existing table and creates a new one. I wrote the renaming manually but looking at Mato's commit, I also should rename the indexes. Will finish this migration manually

mfocko: I would probably make a list of `(original_name, new_name)` and run it through the for-loop, there are some differences here and there (e.g. table vs index renaming), but it should be much less code and easier to understand

lachmanfrantisek (commented):
I know you will not like that but do you (=all) think the `Model` removal helped us? I am looking at the code thinking the explicitness wasn't bad at all. (I know we agreed on it as a team..;)

What specifically I see as a problem:
* Name collisions between Ogr classes and DB models.
* Name collisions with imported classes (e.g. `Allowlist`/`Bugzilla`) -- here I would suggest picking a different name, e.g. `AllowlistEntry` and `BugzillaTicket`.
* It's now not clear, what touches the database and what doesn't.

Nevertheless, thanks @FrNecas for fighting with the Alembic. I think only now, we can compare the naming for real so even if we don't change it, after all, the work wasn't useless.

nikromen:
I can see that keeping `Model` in the naming scheme can be useful sometimes (e.g. collision or easier navigation). But I also really like shorter names (explicit, but as short as possible). As I am looking into the code with new naming scheme I am wondering if it will bring confusion sometimes, so what about a compromise? Removing the `Model` suffix but replace it with something shorter but clear? e.g. with `DB` suffix? so we know that this class is working with database? It is descriptive but short (but in my opinion it is not so elegant :disappointed: )... WDYT?

TomasTomecek (commented):
I personally liked the `models.Table` notation.

It's true that in django world, every table is called `SomethingModel`

No strong preference from my side. My editor fills those anyway :)

lachmanfrantisek:
Just another two notes from my side:
* When writing some test, It's really handy to know what's touching database so I know I need to mock it.
* The problem is not only about a good class name, but also about good argument/property/variable names -- do we want to agree on something there as well? (As I'm thinking about that now, class names are not used so often compared to the attributes/properties/variables).

TomasTomecek (approved):
Giving :green_circle: to lift the rebase curse :)

PR #1339, opened 2022-02-04
Reduce db connections for events

lachmanfrantisek:
* [Make the db_trigger property of Events lazy](https://github.com/packit/packit-service/commit/f330e90d03ea939bef679d4763da91e7371080c0)
  * We use lazy property to touch database only when needed.
  * We don't expand the property when dumping the event to json to avoid creation of database models during event parsing.
    (We don't know if there will be any handler for this event so it might be a useless database entry.
    E.g. we don't need to save each and every finished Koji build to our database.)
* [Be able to hardcode a job_config_trigger_type for event class](https://github.com/packit/packit-service/commit/4c63a23df0fcbde4f9eb6e489e8680c58c707c7a)
  * For some events we know the config trigger
    and we don't need to do a database query for that.
    (And even create a new entry there.)

---

N/A

### DIFF
--- a/packit_service/worker/events/copr.py
+++ b/packit_service/worker/events/copr.py
@@ -66,8 +66,7 @@ class AbstractCoprBuildEvent(AbstractForgeIndependentEvent):
         self.pkg = pkg
         self.timestamp = timestamp
 
-    @property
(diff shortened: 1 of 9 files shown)


jpopelka (approved):
well done

PR #1341, opened 2022-02-11
babysit TF: some events don't have a matching DB runs

TomasTomecek:
Is this because of prod and stg? is prod getting stg events and vice
versa?

This commit logs when a DB trigger can't be found in the database.
Instead we have a log and the execution continues.

https://sentry.io/organizations/red-hat-0p/issues/3002679505/?environment=prod&project=1767823

### DIFF
--- a/packit_service/worker/events/event.py
+++ b/packit_service/worker/events/event.py
@@ -283,7 +283,7 @@ class Event:
         return self._db_trigger
 
     @property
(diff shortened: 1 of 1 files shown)


lachmanfrantisek (approved):
I hope this will not cause another issue in the next step where this property is used. But still much better -- we don't raise an exception and have logs.

PR #1342, opened 2022-02-11
add metrics for number of tasks in celery queue

TomasTomecek:
I realized that we don't need to use any of the external tools or
projects since we already gather metrics on our own: we can easily just
get the current number of celery tasks and record it in a push gateway.

I haven't actually try this code, yet.

TODO:

- [ ] get approval for the design (where should we record these?)

### DIFF
--- a/packit_service/worker/jobs.py
+++ b/packit_service/worker/jobs.py
@@ -12,6 +12,7 @@ from typing import List, Set, Type, Union
 from celery import group
 from packit.config import JobConfig, PackageConfig
 
+from packit_service.celerizer import celery_app
 from packit_service.config import ServiceConfig
 from packit_service.constants import TASK_ACCEPTED
 from packit_service.log_versions import log_job_versions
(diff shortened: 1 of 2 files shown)


packit_service/worker/jobs.py
```diff
@@ -207,6 +208,14 @@ def push_initial_metrics(event: Event, handler: JobHandler, build_targets_len: i
...
             pushgateway.copr_builds_queued.inc()
 
+    # record # of tasks in the celery queue
+    i = celery_app.control.inspect()
+    # >>> i.scheduled()
+    # {'celery@packit-worker-0': [{'eta': '2022-02-11T16:22:53.954247+00:00',...
+    pushgateway.celery_tasks_count = len(
+        [item for workers_tasks in i.scheduled().values() for item in workers_tasks]
+    )
+
```
lbarcziova, line 218: great that we can get this value in this very simple way! Regarding placing the code in `push_initial_metrics`, this function is called only when we create `Task accepted` status, was this intentional or do we want to record these at the beginning of each task?

when comparing the approaches, with the exporter approach I like that the metrics would be independent on running the tasks (e.g. here if I get it correctly, if in some corner case all the workers would be stuck with some tasks for a longer time, we wouldn't have the information about possibly increasing queue size, right?) and also possibility to have richer metrics, on the other hand, if we want to measure only the queue size, this is a simple way to begin with...WDYT?

also, could you please paste a link to the corresponding documentation?

TomasTomecek: you're making a good point: how tightly we want to couple this metric with our actual codebase - and especially, if things go bad, we may not find that out

(re documentation, I just copy-pasted some StackOverflow code O:-)

I checked those exporters you mentioned during your session and they are hooked to celery's events so the metrics should be updated realtime instead of here when the payload is being processed

I guess the solution I propose here is just too hacky and we shouldn't do that

TomasTomecek:
Thank you for your feedback, I'm going to close this one then, let's do it properly so we don't need to refactor in future :)

PR #1345, opened 2022-02-15
Rename tests_requre test suite and zuul job

lachmanfrantisek:
The suite contains also other tests (not only ones using requre)
so the name should make clear that it's about tests using the deployed openshift and running database.

---

N/A


lachmanfrantisek:
recheck

lachmanfrantisek:
recheck

TomasTomecek (approved):

PR #1346, opened 2022-02-15
Align with packit#1357 code

majamassarini:
Syncing of default files can now be disabled

Fixes packit/packit#1357

Merge after packit/packit#1483

PR #1349, opened 2022-02-15
Implement 1-n mapping between build and test targets

### LINKED ISSUE #1230: Enable users configure mapping between build and test targets

lbarcziova:
Fixes #1230 

---

TBD

### DIFF
--- a/packit_service/worker/build/build_helper.py
+++ b/packit_service/worker/build/build_helper.py
@@ -289,6 +289,28 @@ class BaseBuildJobHelper:
     def get_build_targets(self, configured_targets: Set[str]) -> Set[str]:
         """
         Return valid targets/chroots to build from configured targets.
+        If there are build_targets_override or tests_targets_override defined,
+        configured targets ∩ (build_targets_override ∪ mapped tests_targets_override)
+        will be returned.
+
+        Example:
+        build and test job configuration:
+          - job: build
+            trigger: pull_request
+            metadata:
+                targets:
+                      fedora-35-x86_64
+
+          - job: tests
+            trigger: pull_request
+            metadata:
+                targets:
+                      epel-7-x86_64:
+                        distros: [centos-7, rhel-7]
+
+        helper.build_targets_override: None
+        helper.test_targets_override = {"centos-7-x86_64"}
+        helper.get_build_targets({"fedora-35-x86_64", "epel-7-x86_64"})-> {"epel-7-x86_64"}
         """
         targets_override = set()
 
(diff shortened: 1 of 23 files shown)


lbarcziova:
test failure is related to the recent packit changes and not to this PR ( AttributeError: 'PackageConfig' object has no attribute 'synced_files'`, I assume it should be fixed by #1346 )

jpopelka (approved):
Wow, that's a lot of code. Very good. (even, I've never quite understood the targets-override and now it's even more of them :laughing:  )

packit_service/worker/build/build_helper.py
```diff
@@ -286,11 +295,52 @@ def build_targets(self) -> Set[str]:
...
+        )
+
+    @property
+    def build_targets_for_tests(self) -> Set[str]:
+        """
+        Return valid targets/chroots to build in needed to run testing farm.
+
+        (Used when submitting the tests and as a part of the commit status name.)
+        """
+        configured_targets = self.tests_targets_all
```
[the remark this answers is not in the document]
lbarcziova: Thanks for noticing, since they are properties, I left them separate but extracted the common code into a separate method, PTAL.

packit_service/worker/build/build_helper.py
```diff
@@ -299,13 +334,38 @@ def tests_targets(self) -> Set[str]:
...
+        """
+        Return all test targets defined for the build target
+        (from configuration or from default mapping).
+        """
+        raise NotImplementedError("Use subclass instead.")
+
+    def test_target2build_target(self, test_target) -> str:
+        """
+        Return build target to build in needed for testing in test target
+        (from configuration or from default mapping).
```
lachmanfrantisek, line 366:
```suggestion
        Return build target to be built for a given test target
        (from configuration or from default mapping).
```

packit_service/worker/build/copr_build.py
```diff
@@ -157,6 +161,65 @@ def tests_targets_all(self) -> Set[str]:
...
+    def test_target2build_target(self, test_target: str) -> str:
+        """
+        Return build target to build in needed for testing in test target
+        (from configuration or from default mapping).
+        """
+        distro, arch = test_target.rsplit("-", 1)
+        for chroot in self.job_config.metadata.targets:
+            distros_dict = self.job_config.metadata.targets_dict.get(chroot, {})
+            if distros_dict:
+                if distro in distros_dict.get("distros"):
```
lachmanfrantisek, line 208:
```suggestion
            if distros_dict and distro in distros_dict.get("distros"):
```

lachmanfrantisek (commented):
Thanks!

It was a bit hard for me to understand the concept of those build/test target methods and properties but now, it makes a lot of sense. I am wondering how to ease this (especially for newcomers) -- can we make the docstrings more descriptive? (Mention examples and context?)

Nevertheless, I really like the solution and you've fixed one issue as well. (When only test job is defined and we need to trigger a build. Currently, the reverse mapping is correctly used.) Thanks!

lbarcziova:
@lachmanfrantisek thanks for the feedback! I added some more comments and examples, let me know if you think it is more clear now

PR #1351, opened 2022-02-21
Improve instructions for updating allowlist

FrNecas:
As discussed today, lets make it a bit clearer in the README what the correct format is. @mfocko is there anything else you'd like to add?
---
None


TomasTomecek (approved):
definitely more clear

mfocko (approved):
`waiting` can be also interactive, since you can pick pending namespace and approve it from there, otherwise LGTM

FrNecas:
Thanks for the hint with interactive `waiting`. I adjusted the prompt a little bit (the message that was there seemed like a `yes/no` question when in fact I was to specify an integer) and added a mention of this to the README.

jpopelka (approved):

PR #1352, opened 2022-02-21
Use srpm deps from package config in Copr request

lbarcziova:
Use `srpm_build_deps` as an indicator to run SRPMs in Copr
and use them in the Copr API request.

---

You can now define `srpm_build_deps` in your Packit configuration file to state which RPM dependencies are needed for your `actions` to be run when building SRPM. This key will be an indicator to build your SRPM in Copr and the dependencies will be installed in the build environment.


TomasTomecek (approved):
wow, so simple

please also include a release note since this is significant

lachmanfrantisek:
@lbarcziova Can you please provide some context/goal in the release note? (Ideally, to motivate users to use it..;)

PR #1355, opened 2022-02-22
jwt.encode() in v2 returns string instead of bytes

jpopelka:
https://pyjwt.readthedocs.io/en/latest/changelog.html#jwt-encode-return-type

Will be needed once we bump the base image to F35 

---

N/A


lachmanfrantisek (approved):
Thanks!

Issue #1350, opened 2022-02-16
Process correctly Koji build events


lachmanfrantisek:
It looks like we are not parsing the `KojiBuildEvent` triggered by the change of state of non-scratch builds.

Here are some debugging notes:
* The message was received by the fedmsg listener pod:
```
[2022-02-16 15:36:33,768 INFO fedora_messaging.twisted.consumer] Consuming message from topic org.fedoraproject.prod.buildsys.build.state.change (message id b7c712ea-2d36-45df-822c-ad0e8791fd2e)
[2022-02-16 15:36:33,768 INFO packit_service_fedmsg.consumer] org.fedoraproject.prod.buildsys.build.state.change: None
[2022-02-16 15:36:33,770 DEBUG packit_service_fedmsg.consumer] Task UUID=57c39f45-ca29-4e3c-a6c0-0ef9fc150233 sent to Celery.
[2022-02-16 15:36:33,770 INFO fedora_messaging.twisted.consumer] Successfully consumed message from topic org.fedoraproject.prod.buildsys.build.state.change (message id b7c712ea-2d36-45df-822c-ad0e8791fd2e)
```
* The task was processed but no task was created:
```
[2022-02-16 15:36:33,771: INFO/MainProcess] Received task: task.steve_jobs.process_message[57c39f45-ca29-4e3c-a6c0-0ef9fc150233]  
[2022-02-16 15:36:33,772: INFO/ForkPoolWorker-1] 
Package Versions:
* OGR 0.34.1.dev13+g3829ba1
* Packit 0.45.1.dev4+geb8fac4
* Packit Service 0.1.dev1+g6fb443b.d20220214
* SQL Alchemy 1.3.24
[2022-02-16 15:36:33,772: INFO/ForkPoolWorker-1] Task task.steve_jobs.process_message[57c39f45-ca29-4e3c-a6c0-0ef9fc150233] succeeded in 0.00040541705675423145s: []
```
* There are no `INFO` logs and no exceptions in sentry and `[]` was returned.

lachmanfrantisek:
New try with debug turned on:

* Manual push to `main` on `python-ogr`: ad3142c6354063200eaa46b61e1214146ec55811
* fedmsg: https://apps.fedoraproject.org/datagrepper/v2/id?id=8a8436e3-377b-4927-bee4-dcce283064bf&is_raw=true&size=extra-large
* listener:
```
[...]

PR #1357, opened 2022-02-23
Parse koji events correctly

lachmanfrantisek:
Fixes: #1350

---

N/A


jpopelka (approved):
I'm glad you solved it. Good catch.

tests/unit/test_events.py
```diff
@@ -177,70 +177,6 @@ def distgit_commit(self):
...
-        ) as outfile:
-            return json.load(outfile)
-
-    @pytest.fixture()
-    def koji_build_completed_epel8(self):
-        with open(
-            DATA_DIR / "fedmsg" / "koji_build_completed_epel8.json", "r"
-        ) as outfile:
-            return json.load(outfile)
-
```
TomasTomecek, line 243: would it make sense to drop unused files from the repo?

TomasTomecek (approved):
wow, nicely spotted! lovely to see as well to have it covered with tests

tests/unit/test_events.py
lachmanfrantisek: The files are still in use. We just had the fixtures defined multiple times..;)

PR #1359, opened 2022-02-23
Improve examples of using packit-service locally

majamassarini:


CONTRIBUTING.md
```diff
@@ -51,8 +51,29 @@ You can also run only certain pieces of packit-service for local development
 
 When you are running service/httpd and making requests to it,
 make sure that `server_name` configuration file in `packit-service.yaml` is set.
-Then you **need** to make requests to httpd using that hostname
-(which can be done by creating a new entry in `/etc/hosts` on your laptop).
+
+### binding on localhost
+
+```yaml
+server_name: "service.localhost:8443"
```
jpopelka, line 58:
```suggestion
server_name: service.localhost:8443
```

jpopelka (commented):
Thank you for looking at this!
I confirm that I also see the `No route to host` when trying the old approach. It's most likely due to using `podman` with `docker-compose`.
I documented it in #1331 but had not properly tested that everything works OK, I'm sorry for the confusion it caused.
I confirm that your approach works and I think we can actually remove the `binding on other hosts` section to keep the description as simple as possible.

majamassarini:
> I confirm that your approach works and I think we can actually remove the `binding on other hosts` section to keep the description as simple as possible.

Do you think that we can move the description of the flask behaviour and configuration elsewhere? I think they can be handy when setting up a network with no binds to localhost?

jpopelka (approved):
> Do you think that we can move the description of the flask behaviour and configuration elsewhere? I think it can be handy when setting up a network with no binds to localhost?

Ok, then we can keep it as it is, I guess, thank you!

PR #1366, opened 2022-02-25
Set default loglevel for workers to debug

jpopelka:
It can still be changed with `LOGLEVEL` env. var.
We discussed that it's sometimes handy and we already have `debug: true` in all `packit-service.yaml`s to enable debug logs from `packit`.

---

N/A


lachmanfrantisek (approved):

lachmanfrantisek:
recheck

jpopelka:
recheck

Issue #1361, opened 2022-02-23
Suggest solution when propose-downstream fail on non-existent branch


johannbg:
***Edited by @lachmanfrantisek***:

User hits a `We were unable to push to dist-git: [remote rejected] (pre-receive hook declined)` error for propose-downstream when the dist-git branch doesn't exist.
So the solution is to check if that branch exists and if it's not, provide a meaningful error saying that one needs to request it with Fedora rel-eng.''

Example of the issue created because of this:
https://github.com/dracutdevs/dracut/issues/1713

***Original description***

Seriously can I get something to work with instead of a fracking bot that creates an issue in which it proposes to run '/packit propose-downstream' then gives me a thumbs up when in reality packit fails again o_O and that fracking bot creates a new issue proposing the exact same thing o_O

What's fubar in the process here upstream,packit, RH infra what? What needs to be done to actually resolve this issue???

TomasTomecek:
So the solution is to check if that branch exists and if it's not, provide a meaningful error saying that one needs to request it with Fedora rel-eng?

johannbg:
Well to me these are four seperated issues to look at being solved. 

First of all the bot should not be giving thumbs up when given commands  ( /packit propose-downstream ) I interpreted that as the command was successful but presumably you implemented that as command recieved by the bot.

Secondly the bot should not be creating new issue when being unsuccessful for the second ( or more ) time but comment that it was unsuccessful again on the existing issue he had already created. 

Thirdly the bot should not be proposing the exact same action that had already failed ( now twice ) again

Forthly the message provide by the bot should always be something actionanable by upstream not some cryptic error message that literaly means nothing to upstreams.

johannbg:
> So the solution is to check if that branch exists and if it's not, provide a meaningful error saying that one needs to request it with Fedora rel-eng?

Have the bot create the issue and request it since packit is already adding enough to the cost of developement to upstreams for literally no benefits.

Take this issue here for example, It costs an hour ( including the cost of distraction ) for upstreams to look at any issues related to packit, it has costed the packit team 2 hours ( one from each team member assuming that only two that commented here looked at this issue ), it will cost me more to file additional issue with fedora releng, and releng atleast one hour for someone there to look at this totaling ca 5 hours just to solve this failure. 

That's expensive on people paid ( for example one hour of my time costs 170 euros assuming everyone would get paid the same, it amounts to 850 euros ) or free time of their lifetime ( which people will never get back ). 

[...]

PR #1372, opened 2022-03-02
Use 'eyes' instead of 'thumbsup/+1' as a reaction to /packit comment

jpopelka:
https://github.com/packit/packit-service/issues/1361#issuecomment-1055790961 suggests using :rocket: instead of :+1: as a reaction to `/packit comment`.
The :+1: can be understood as "done", while :rocket: as "running".

WDYT?

---

TODO

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -17,4 +17,6 @@ Merge before/after
 
 ---
 
-<!-- release notes for changelog/blog follow -->
+RELEASE NOTES BEGIN
+￼
+RELEASE NOTES END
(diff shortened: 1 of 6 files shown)


lachmanfrantisek (approved):
It's tricky TBH. I am not sure if someone else will not say `rocket` means released/delivered.

I am not against it. Just not sure if there is anything really widely-understood. And the set is very limited. What about :eyes: meaning `we are looking into that`?

nikromen:
I think that if bot gives to command :+1: after cca 2 sec command was written it is more or less clear that this means `we are working on it`.

> I am not sure if someone else will not say rocket means released/delivered.
I agree with this, :rocket: may cause more confusion to this. I personally like the idea with :eyes: emoji.

TomasTomecek (approved):
+1 for eyes being more clear

PR #1376, opened 2022-03-02
Fix mapping of test targets to build targets

lbarcziova:
Don't try to reverse map the test targets, go through the not mapped test targets and check their mapped values for a match.

The problem in the previous implementation was, that for test targets that were hardcoded in our default build -> test mapping we picked the hardcoded build target, e.g. if test configuration had `centos-stream-8`, then `test_target2build_target(centos-stream-8-x86_64`) returned `epel-8-x86_64` ([see](https://github.com/packit/packit-service/blob/f2d1cd6adcd525dd4373c988f716a345e0514ede/packit_service/constants.py#L128)).

Tricky question in the end: is it a valid use case to test more builds with one test target? This would not work in that case correctly.

---

RELEASE NOTES BEGIN

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/build/copr_build.py
+++ b/packit_service/worker/build/copr_build.py
@@ -233,8 +233,10 @@ class CoprBuildJobHelper(BaseBuildJobHelper):
 
     def test_target2build_target(self, test_target: str) -> str:
         """
-        Return build target to build in needed for testing in test target
-        (from configuration or from default mapping).
+        Return build target to build in needed for testing in test target.
+        Go through the not mapped test targets (self.test_targets_all)
+        and for each check the mapped test targets, if test_target is
+        in mapped test targets, return the target that was mapped.
 
         Examples:
         configuration:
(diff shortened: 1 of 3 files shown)


packit_service/worker/build/copr_build.py
```diff
@@ -254,25 +256,25 @@ def test_target2build_target(self, test_target: str) -> str:
                 targets:
                       fedora-35-x86_64
 
+
```
lachmanfrantisek, line 259: > Tricky question in the end: is it a valid use case to test more builds with one test target? This would not work in that case correctly.

But they can still manually configure that themself, can't they?

lachmanfrantisek (approved):
Looks good to me. Thanks a lot for fixing that!

packit_service/worker/build/copr_build.py
lbarcziova: hmm, I don't think it is possible by the proper way, this question came to my mind looking into [this](https://github.com/sgallagher/sscg/blob/b9f757736f73db8c58bb9e422e018ab84eabd51f/.packit.yaml#L19) config, where there is build/test both for `epel-8` and `centos-stream-8` and `epel-8` will be in the end tested as well in `centos-stream-8` (I am not sure whether this is intended by the users or they just don't know about default mapping)...and the `test_target2build_target(centos-stream-8)` would return only on of the build targets, since we always talked only about 1 build: many tests and not the other way

lachmanfrantisek: The original ask for mappings was to allow reusing one build (mainly centos) in multiple test targets and avoiding a need to have all the centos clones like Alma or Rocky in Copr.

Regarding that specific config, there is no `epel` target so it needs to use something else like `centos-stream` or `rhel`.
And by this change, we will not change anything for them, will we?

lbarcziova: This change will change the behaviour for them, since `test_target2build_target(centos-stream-8)` will return always only one matching build target (first in the list).

PR #1381, opened 2022-03-04
Restart the httpd server every 8 hours

csomh:
Threads/processes get stuck with the following error message:

    OSError: Apache/mod_wsgi request data read error: Connection reset by peer.

Let's try to restart the processes every 8 hours in order to avoid it.


jpopelka (approved):
Can there be a link (as a comment in the file) to some docs where all the options are described?

csomh:
> Can there be a link (as a comment in the file) to some docs where all the options are described?

Yes, added.

PR #1385, opened 2022-03-04
Job identifiers (=monorepo support)

lachmanfrantisek:
This PR adds support for job identifiers.

It can be used to differentiate the results of two similar jobs and avoid check names collision.

There are various use-cases for this like building more packages from one repository (aka monorepo).

Fixes: packit/packit#1444

Requires: packit/packit#1506

---

RELEASE NOTES BEGIN
You can specify an identifier for your job to be able to configure one job multiple times.
For example, you can build multiple projects from one repository (known as monorepo concept)
or try multiple build options. Using identifiers allows Packit to avoid naming collisions in commit statuses
and default Copr project names.
RELEASE NOTES END


packit_service/worker/jobs.py
```diff
@@ -99,6 +99,13 @@ def get_handlers_for_event(
 
     matching_handlers: Set[Type["JobHandler"]] = set()
     for job in jobs_matching_trigger:
+        if (
+            isinstance(event, CheckRerunEvent)
+            and (event.job_identifier or job.identifier)
+            and event.job_identifier != job.identifier
```
jpopelka, line 105: I think just the second condition should be enough.
The first condition only makes sure that they are not both empty/undefined, which they can't be if they are different (the second condition). But I'm not against having it explicitly stated.
```suggestion
            and event.job_identifier != job.identifier
```

jpopelka (approved):

lbarcziova (approved):
LGTM

packit_service/worker/jobs.py
lachmanfrantisek: I wanted to be explicit, but it's not necessary as I am thinking about that now... I'll simplify that.

nikromen (approved):
lgtm

TomasTomecek (commented):
wh a a a a a t

I will try this with mock on Friday

PR #1386, opened 2022-03-07
Create a target to perform alembic revision

majamassarini:
I create a new Makefile target to be able to mount alembic files inside the service pod with the *host user permissions*.
Alembic is called by this target and runs inside the pod as the *local host user* generating a file belonging to the user. 

I used podman with the options: --user and --uidmap. 
I did not changed docker-compose since I thought that most of the time we do not want to run the service using our local user.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,8 @@ COV_REPORT ?= term-missing
 COLOR ?= yes
 SOURCE_BRANCH ?= $(shell git branch --show-current)
 CONTAINER_RUN_INTERACTIVE ?= -it
+COMPOSE ?= docker-compose
+MY_ID ?= `id -u`
 
 service: files/install-deps.yaml files/recipe.yaml
 	$(CONTAINER_ENGINE) pull $(BASE_IMAGE)
@@ -95,3 +97,29 @@ check-inside-toolbox:
 
 requre-purge-files:
 	pre-commit run requre-purge --all-files --verbose --hook-stage manual
+
+# run all the pods needed by the service pod
+# use docker-compose to update and run them
+# (postgres and redis)
+compose-for-migrate-db-up:
+	$(COMPOSE) up --build --force-recreate -d service
+
+# run alembic revision through another service pod
+# run processes as *local host user* inside the pod
+# preserve *local host user* files' uid inside the pod
+migrate-db: compose-for-migrate-db-up
+	podman run --rm -ti --user $(MY_ID) --uidmap=$(MY_ID):0:1 --uidmap=0:1:999 \
+	-e DEPLOYMENT=dev \
+	-e REDIS_SERVICE_HOST=redis \
+	-e POSTGRESQL_USER=packit \
+	-e POSTGRESQL_PASSWORD=secret-password \
+	-e POSTGRESQL_HOST=postgres \
+	-e POSTGRESQL_DATABASE=packit \
+	-v $(CURDIR)/alembic:/src/alembic:rw,z \
+	-v $(CURDIR)/packit_service:/usr/local/lib/python3.9/site-packages/packit_service:ro,z \
+	-v $(CURDIR)/secrets/packit/dev/packit-service.yaml:/home/packit/.config/packit-service.yaml:ro,z \
+	-v $(CURDIR)/secrets/packit/dev/fullchain.pem:/secrets/fullchain.pem:ro,z \
+	-v $(CURDIR)/secrets/packit/dev/privkey.pem:/secrets/privkey.pem:ro,z \
+	--network packit-service_default \
+	quay.io/packit/packit-service:dev alembic revision -m "$(CHANGE)" --autogenerate
+	$(COMPOSE) stop # stop previously started pods: service, postgres and redis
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -95,3 +97,25 @@ check-inside-toolbox:
...
+	-e DEPLOYMENT=dev \
+	-e REDIS_SERVICE_HOST=redis \
+	-e POSTGRESQL_USER=packit \
+	-e POSTGRESQL_PASSWORD=secret-password \
+	-e POSTGRESQL_HOST=postgres \
+	-e POSTGRESQL_DATABASE=packit \
+	-v $(CURDIR)/alembic:/src/alembic:rw,z \
+	-v $(CURDIR)/packit_service:/usr/local/lib/python3.9/site-packages/packit_service:ro,z \
+	-v $(CURDIR)/secrets/packit/dev/packit-service.yaml:/home/packit/.config/packit-service.yaml:ro,z \
+	-v $(CURDIR)/secrets/packit/dev/private-key.pem:/secrets/private-key.pem:ro,z \
```
jpopelka, line 116: not needed, see 15322825
```suggestion
```

Makefile
```diff
@@ -95,3 +97,25 @@ check-inside-toolbox:
 
 requre-purge-files:
 	pre-commit run requre-purge --all-files --verbose --hook-stage manual
+
+compose:
```
jpopelka, line 101: a comment about what's the purpose of this target would be nice

jpopelka (commented):
Can you update [CONTRIBUTING.md](https://github.com/packit/packit-service/blob/main/CONTRIBUTING.md#database) to use your new target?

Makefile
```diff
@@ -95,3 +97,31 @@ check-inside-toolbox:
...
 requre-purge-files:
 	pre-commit run requre-purge --all-files --verbose --hook-stage manual
+
+# run all the pods needed by the service pod
+# use docker-compose to update and run them all
+# stop the service pod to be able to re-run it
+# later with different options
+compose-for-migrate-db:
+	$(COMPOSE) up --build --force-recreate -d service
+	$(COMPOSE) stop service
```
jpopelka, line 107: The comment says: "stop the service pod to be able to re-run it later with different options", but in the `migrate-db`, we don't run the (docker-compose service) 'service', but a new podman container/pod, which I don't see how it's related to the previously up-ed & stop-ed 'service'.

If the purpose of this `compose-for-migrate-db` target is to make sure the `quay.io/packit/packit-service:dev` image is built and up-to-date, then I think `$(COMPOSE) build service` should be enough.

Makefile
```diff
@@ -95,3 +97,31 @@ check-inside-toolbox:
...
+	-e POSTGRESQL_HOST=postgres \
+	-e POSTGRESQL_DATABASE=packit \
+	-v $(CURDIR)/alembic:/src/alembic:rw,z \
+	-v $(CURDIR)/packit_service:/usr/local/lib/python3.9/site-packages/packit_service:ro,z \
+	-v $(CURDIR)/secrets/packit/dev/packit-service.yaml:/home/packit/.config/packit-service.yaml:ro,z \
+	-v $(CURDIR)/secrets/packit/dev/fullchain.pem:/secrets/fullchain.pem:ro,z \
+	-v $(CURDIR)/secrets/packit/dev/privkey.pem:/secrets/privkey.pem:ro,z \
+	--network packit-service_default \
+	quay.io/packit/packit-service:dev alembic revision -m "$(CHANGE)" --autogenerate
+	$(COMPOSE) down
```
jpopelka, line 127: I guess the purpose of this `$(COMPOSE) down` is to stop the container/pod started by podman in the previous step?
But the container/pod should exit/stop once the `alembic revision` finishes, so I don't see a point of this `$(COMPOSE) down`.

Makefile
majamassarini: > The comment says: "stop the service pod to be able to re-run it later with different options", but in the `migrate-db`, we don't run the (docker-compose service) 'service', but a new podman container/pod, which I don't see how it's related to the previously up-ed & stop-ed 'service'.

You are right, they are different pods and I need to fix the documentation, but since I am mapping the same ports (and maybe this is not necessary) for this reason I am closing the one created by docker-compose.

> If the purpose of this `compose-for-migrate-db` target is to make sure the `quay.io/packit/packit-service:dev` image is built and up-to-date, then I think `$(COMPOSE) build service` should be enough.

I need also the postgres (and probably the redis) pod; the service pod depends on them and I do not want to hard code this dependency here again. This is the reason why I used docker-compose to start them all.

Makefile
```diff
@@ -95,3 +97,29 @@ check-inside-toolbox:
 
 requre-purge-files:
 	pre-commit run requre-purge --all-files --verbose --hook-stage manual
+
+# run all the pods needed by the service pod
+# use docker-compose to update and run them
+# (postgres and redis)
+compose-for-migrate-db-up:
+	$(COMPOSE) up --build --force-recreate -d service
```
nikromen, line 105: Aren't you missing definition of COMPOSE -> docker-compose?

nikromen (commented):
good job! I look forward to use this! :partying_face:

Makefile
nikromen: oh, I am just blind :see_no_evil: it's there!

nikromen (approved):
LGTM! :rocket:

PR #1390, opened 2022-03-09
Parse pagure.pull-request.flag.* into PullRequestFlagPagureEvent

jpopelka:
Needed for packit/hardly#52

---

RELEASE NOTES BEGIN
￼
RELEASE NOTES END


packit_service/worker/events/enums.py
```diff
@@ -26,8 +26,15 @@ class IssueCommentAction(Enum):
     edited = "edited"
 
 
+class PagureFlagStatus(Enum):
```
csomh, line 29: Please make a comment with a URL where the documentation for all these names can be found.

jpopelka: Thanks, actually, it's forge-specific, so if it needs to be defined anywhere, then it should be in `ogr`.
I've removed it because we don't need it in the parsing-incoming-message-to-event code, the event handler should handle whatever value there is.

PR #1394, opened 2022-03-10
Add check for internal TF tests job for builds as well

lbarcziova:
Builds could trigger the test job that was not allowed, therefore
don't allow builds if there is an internal TF test job with the same trigger.

---

RELEASE NOTES BEGIN
￼
RELEASE NOTES END


lachmanfrantisek (approved):
Thanks!

nikromen (approved):
:rocket:

PR #1397, opened 2022-03-11
Report better message when we fail updating our chroot set

TomasTomecek:
The original problem is a limitation in Copr that you cannot update
chroots of a project while builds are running.

Unfortunately, Packit did not detect this well and reported confusing
message to users about permissions:

https://github.com/packit/hello-world/pull/677#issuecomment-1065106598

This commit fixes the particular case and provides a more helpful error
message to users how to proceed and what the problem is.

On top of it, if project chroots require update, the message now
contains a markdown-formatted diff.

Related: https://github.com/packit/packit/issues/1303

---

Example comment:

Settings of a Copr project packit/the-example-namespace-the-example-repo-342-stg need to be updated, but there was an issue while Packit performed the change.
You should be able to resolve the problem by recreating this pull request or running `/packit build` after all builds finished.

This was the change Packit tried to do:
| field | old value | new value |
| ----- | --------- | --------- |
| chroots | ['f30', 'f31'] | ['f31', 'f32'] |
| description | old | new |

Diff of chroots:
```diff
-f30
+f32
```

RELEASE NOTES BEGIN

Packit no longer provides a misleading comment when it fails to update a set of targets on its own Copr projects.
￼
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/build/copr_build.py
+++ b/packit_service/worker/build/copr_build.py
@@ -3,7 +3,7 @@
 
 import logging
 from datetime import datetime, timezone
-from typing import List, Optional, Set, Tuple
+from typing import List, Optional, Set, Tuple, Iterable
 
 from copr.v3 import CoprRequestException
 from ogr.abstract import GitProject
@@ -26,6 +26,7 @@ from packit_service.constants import (
     PG_BUILD_STATUS_SUCCESS,
     DEFAULT_MAPPING_INTERNAL_TF,
     DEFAULT_MAPPING_TF,
+    COPR_CHROOT_CHANGE_MSG,
 )
 from packit_service.models import (
     AbstractTriggerDbType,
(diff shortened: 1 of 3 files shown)


packit_service/worker/build/copr_build.py
```diff
@@ -620,6 +620,34 @@ def create_copr_project_if_not_exists(self) -> str:
...
+            if "chroots" in ex.fields_to_change:
+                old_chroots, new_chroots = ex.fields_to_change["chroots"]
+                chroots_diff = "Diff of chroots:\n```diff\n"
+                extra_chroots = set(old_chroots).difference(new_chroots)
+                missing_chroots = set(new_chroots).difference(old_chroots)
+                for e in extra_chroots:
+                    chroots_diff += f"-{e}\n"
+                for m in missing_chroots:
+                    chroots_diff += f"+{m}\n"
+                chroots_diff += "```\n"
```
lachmanfrantisek, line 633: It looks really nice!

Can you please move this to a separate function/method? This one is getting really long...

packit_service/worker/build/copr_build.py
```diff
@@ -620,6 +620,34 @@ def create_copr_project_if_not_exists(self) -> str:
...
+
+            if owner == "packit":
+                # the problem is on our side and user cannot fix it
+                msg = (
+                    f"Settings of a Copr project {owner}/{self.job_project} need to be updated, "
+                    "but there was an issue while Packit performed the change.\n"
```
lachmanfrantisek, line 639: Hmm, it looks to me like **we** have some unfixed issue. (Or is it unclear for a reason?)

```suggestion
                    f"Settings of a Copr project {owner}/{self.job_project} need to be updated, "
                    "but Packit can't do that when there are previous builds still in progress..\n"
```

lachmanfrantisek: Sorry for being a nitpicker, but can you please make a constant from this as we do for other comments?

lachmanfrantisek (approved):
Thanks a lot!

packit_service/worker/build/copr_build.py
TomasTomecek: well, there are many strings here that are not constants but I agree to do that for my case to not increase tech debt

lachmanfrantisek (approved):
Nicely done, thanks!

PR #1398, opened 2022-03-14
Add target to test database with docker compose

majamassarini:
Creates a Makefile target (check-db) ro run database tests using the network created by 
docker-compose.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -256,6 +256,16 @@ run only the unit tests use:
 
     TEST_TARGET=tests/unit make check-in-container
 
+#### **Database tests**
+
+Database tests can be run using a dedicated target.
+
+    make check-db
+
+To run them you need docker-compose.
+Otherwise you can run the same using _Openshift_ and following
+the instructions below.
+
 ### Openshift tests using requre
 
 This testsuite uses [requre project](https://github.com/packit/requre) project to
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -256,6 +256,16 @@ run only the unit tests use:
 
     TEST_TARGET=tests/unit make check-in-container
 
+#### **db tests**
```
TomasTomecek, line 259: nit: Can we call the section "Database tests" please? So that it's obvious after reading the header

TomasTomecek (approved):
LGTM, nicely done! this definitely does the trick

I'm personally not a fan of "random" sleeps but I understand that we need to wait for the DB to be up and serving

Also the last `$compose stop` seems odd - is the intention to wipe out the database and everything so that one can run the `check-db` command again?

majamassarini:
> I'm personally not a fan of "random" sleeps but I understand that we need to wait for the DB to be up and serving

Yes you are right but as you said I really do not know how to otherwise wait for the db to be up and running...  

> Also the last `$compose stop` seems odd - is the intention to wipe out the database and everything so that one can run the `check-db` command again?

Yes exactly, I not really know docker or compose so I found this solution but any other suggestion is welcome.

TomasTomecek (approved):
I'm still LGTM. The two points I raised could be improved but I don't think that's worth the time as long as the current solution works for you.

PR #1401, opened 2022-03-15
Add relationship between srcgit pr and distgit pr

majamassarini:
Create a new table to track relationship between source git pull request and dist git pull request in hardly code.

Related to packit/hardly#27

Merge before packit/hardly#51


jpopelka (commented):
I usually tend to avoid touching the db models, that's why I asked you to discuss it with @lachmanfrantisek first :-) but I don't see a reason why this should not work. So :+1: 

I however have one question/suggestion - [in the original issue I suggested adding the new model to hardly](https://github.com/packit/hardly/issues/27#issuecomment-1047043790), because it's not needed in packit-service. I understand that it's easier to add it here in packit-service because we already have `tests_openshift` here so one can test it easily and we don't have anything like that in hardly yet. I don't even mind to add it here initially, because it will just be empty table in packit-service, so no harm. But long term, I'd like to see it in hardly, so do you think we can at least create an issue in hardly to move it there eventually?

majamassarini:
> I however have one question/suggestion - [in the original issue I suggested adding the new model to hardly](https://github.com/packit/hardly/issues/27#issuecomment-1047043790), because it's not needed in packit-service. I understand that it's easier to add it here in packit-service because we already have `tests_openshift` here so one can test it easily and we don't have anything like that in hardly yet. I don't even mind to add it here initially, because it will just be empty table in packit-service, so no harm. But long term, I'd like to see it in hardly, so do you think we can at least create an issue in hardly to move it there eventually?

I spoke with @lachmanfrantisek about this and he suggested me to put things here for the moment.
I may be wrong but I feel like that the real problem here is not where to put the code but how to manage the database migrations.
Now database migrations are linear, if we diverge applying a migration just to hardly, then we need to duplicate all future migrations that are of interest both for hardly and packit-service or we should, somehow, manage a tree of migrations which maybe it is possible but I think it will require some work.

But as I said I could be wrong and I hope @lachmanfrantisek will correct me :-)

packit_service/models.py
```diff
@@ -1911,3 +1911,81 @@ def to_dict(self):
...
+                session.add(rel)
+            return rel
+
+    @classmethod
+    def get_by_id(cls, id_: int) -> Optional["SrcGitPRDistGitPRModel"]:
+        with get_sa_session() as session:
+            return session.query(SrcGitPRDistGitPRModel).filter_by(id=id_).all()
+
+    @classmethod
+    def get_by_source_git_id(cls, id_: int) -> Optional["SrcGitPRDistGitPRModel"]:
```
jpopelka, line 1976:
```suggestion
    def get_by_source_git_id(cls, id_: int) -> Optional[Iterable["SrcGitPRDistGitPRModel"]]:
```

jpopelka: the same applies to `get_by_dist_git_id()`

lachmanfrantisek:
@majamassarini @jpopelka 

I can't find the result from the discussion we had some time ago about that. Just [these](https://github.com/packit/research/tree/main/split-the-stream#database) options described when I was looking at that.

I found just this note in the arch document:

> Having empty tables 
> * lower maintenance, not a big problem
> * Tomas: configuration option to not create them in some envs 


But from what I remember, it's really not easy to structure/rebase/merge/maintain multiple schemas. It is not nice, but I don't see empty tables as a really big problem compared to the multi-schema solutions and all the possible issues. But if someone finds a better way how to do this nicely, I am really for it!

packit_service/models.py
```diff
@@ -1911,3 +1911,83 @@ def to_dict(self):
...
+
+class SrcGitPRDistGitPRModel(Base):
+    __tablename__ = "src_git_pr_dist_git_pr"
+    id = Column(Integer, primary_key=True)  # our database PK
+    src_git_pull_request_id = Column(Integer, ForeignKey("pull_requests.id"))
+    dist_git_pull_request_id = Column(Integer, ForeignKey("pull_requests.id"))
+    src_git_pull_request = relationship(
+        "PullRequestModel",
+        primaryjoin="SrcGitPRDistGitPRModel.src_git_pull_request_id==PullRequestModel.id",
+        uselist=False,
```
lachmanfrantisek, line 1924: Interesting. I haven't seen this syntax...

I see we are not adding the reverse way of the relation -- (? @jpopelka) is it ok to search through all of the entries when we need to find a matching PR for the given PR?

majamassarini: > Interesting. I haven't seen this syntax...

Yes, I struggled a little to find a way to have the two columns inside the same table reference the same keys... it seems to me that in this way it is working but I am not sure at all.

> I see we are not adding the reverse way of the relation -- (? @jpopelka) is it ok to search through all of the entries when we need to find a matching PR for the given PR?

Yes maybe I am missing something more, probably I can't figure it out which query you have in mind.

lachmanfrantisek: Depends on what we want..;)

* With the current version, we can create the mapping model in a single step (when we have both pull requests), but when we have just one and want to get the other (what I hope is the use-case we will need), we need to go through all the mapping models to get it.
* We can also create the relation in a way, that the relation to the mapping model is saved in the pull request so it can be found in one step. But the question is if it is really an issue. (From the database point of view, the current version does not contain circular dependencies so it looks cleaner.)

majamassarini: Ok, thanks, I think I got it! I can try to do that in another commit and then we can choose to take it or give it away.

majamassarini: Ok, for me it is the same.
If also @lachmanfrantisek agrees then I will remove the last commit.

jpopelka: Also [indexing](https://docs.sqlalchemy.org/en/14/core/constraints.html#sqlalchemy.schema.Index) the `dist_git_pull_request_id` and `src_git_pull_request_id` should speed the queries a lot.
Based on [this](https://stackoverflow.com/questions/107132/what-columns-generally-make-good-indexes): "columns that are generally increasing and unique make a good index" so if we make them unique (see [my other comment](https://github.com/packit/packit-service/pull/1401#discussion_r829444347)), there'll be no performance issues because the values will be increasing.

lachmanfrantisek: Yes, that was just a question if this is what we want. Jirka has a good point with the indexes, that's a clear solution with no speed downsides.

packit_service/models.py
```diff
@@ -1911,3 +1911,81 @@ def to_dict(self):
...
+        with get_sa_session() as session:
+            return session.query(SrcGitPRDistGitPRModel).filter_by(id=id_).first()
+
+    @classmethod
+    def get_by_source_git_id(cls, id_: int) -> Optional["SrcGitPRDistGitPRModel"]:
+        with get_sa_session() as session:
+            return (
+                session.query(SrcGitPRDistGitPRModel)
+                .filter_by(src_git_pull_request_id=id_)
+                .first()
```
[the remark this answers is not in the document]
majamassarini: I thought that the uselist=False into the relationship was enough to design a one to one relationship. But making I test I saw it was not. So I added the unique (and also the index) attribute to the columns.

packit_service/models.py
majamassarini: I removed the reverse relationship and added uniqueness and indexing to the columns.

I also wrote into the documentation a possible explanation for an error that can occur when using the migrate-db target.
It took me a while to realize what was happening; since the service pod runs into the background through compose there is no evidence of a failure in it.

If you have any idea of a check I can do to ensure everything  is properly setup let me know.

csomh:
Please use `source-git` or `source_git`, and `dist-git` or `dist_git` in all texts and variable names.

Grepping through the packit and packit-service codebase these two seem to be the current prominent naming patterns (with a few occurrences of `distgit`) and it would be awesome if we could stay consistent.

jpopelka (approved):
Would it be possible to merge the two alembic migrations into one?
Otherwise looks OK, I'll play with it a bit more on Monday since I need to test my hardly PR anyway and we can merge after moving stable branches.
Thank you! :tulip:

csomh (approved):

PR #1402, opened 2022-03-15
Fix propose downstream bugs

nikromen:
Each bug and solution is separated and properly described in commits.

Notes to the `Fix duplicate propose-downstream jobs in dashboard`:
according to [this answer](https://stackoverflow.com/questions/68915407/celery-retry-with-updated-arguments) and [celery docs](https://docs.celeryq.dev/en/stable/reference/celery.app.task.html#celery.app.task.Task.retry) (see the `**kwargs` bullet point in parameters) this solution should work (passing propose-downstream id to `retry`). But I don't know how to test it so I am not sure if it really works. Are there any `flexmock` tools how to run the retried task which was pushed back to the queue and check if the `propose_downstream_run_id` has correct value?

### DIFF
--- a/packit_service/worker/handlers/distgit.py
+++ b/packit_service/worker/handlers/distgit.py
@@ -272,6 +272,14 @@ class ProposeDownstreamHandler(JobHandler):
             for branch, model in zip(
                 branches, propose_downstream_model.propose_downstream_targets
             ):
+                # skip submitting a branch if we already did that (even if it failed)
+                if model.status not in [
+                    ProposeDownstreamTargetStatus.running,
+                    ProposeDownstreamTargetStatus.retry,
+                    ProposeDownstreamTargetStatus.queued,
+                ]:
+                    continue
+
                 model.set_status(status=ProposeDownstreamTargetStatus.running)
                 buffer, handler = gather_packit_logs_to_buffer(
                     logging_level=logging.DEBUG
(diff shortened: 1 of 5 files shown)


packit_service/worker/handlers/distgit.py
```diff
@@ -166,7 +171,13 @@ def sync_branch(self, branch: str) -> Optional[PullRequest]:
                     logger.info(f"Will retry for the {retries + 1}. time in {delay}s.")
                     # throw=False so that exception is not raised and task
                     # is not retried also automatically
-                    self.task.retry(exc=ex, countdown=delay, throw=False)
+                    self.task.retry(
+                        exc=ex,
+                        countdown=delay,
+                        args=(),
+                        # https://stackoverflow.com/questions/68915407/celery-retry-with-updated-arguments
+                        kwargs={"propose_downstream_run_id": model.id},
```
lachmanfrantisek, line 179: This reusing is just for the purpose of the not-found-source, isn't it? (Just comparing this to the build workflow where we create a new model in case of an error. But if it is just for waiting, that's ok.)

lachmanfrantisek (approved):
The code changes look good to me! Well spotted!

And regarding the tests, that's really hard. Maybe at least some unit tests with the mock of everything around?

Maybe we should somehow mock the celery worker so we can use this for other tests as well. The way we do that currently is really not ideal.
(You've spotted and named a key issue of our test suite I think.)

lachmanfrantisek (approved):
Thanks for the proper commit descriptions!

PR #1416, opened 2022-03-24
Don't add commit comments for non final states

### LINKED ISSUE #1411: Do not create status comments for non-final states
As you can see e.g. [here](https://github.com/osbuild/osbuild-composer/pull/2473), Packit created a couple of comments because we weren't able to use a commit status. That's correct (and the failure is on the GitHub side) but to avoid a big number of comments, Packit should create comments only for final (`success`/`failure`/`error`) states.

The related code is here:

https://github.com/packit/packit-service/blob/58ba3d1e3046ef8ee44eef35c2aa09be8586523f/packit_service/worker/reporting.py#L345-L350

https://github.com/packit/packit-service/blob/58ba3d1e3046ef8ee44eef35c2aa09be8586523f/packit_service/worker/reporting.py#L180-L196

jpopelka:
to avoid multiple comments.

The `_add_commit_comment_with_status()` is used as a fallback if setting commit status fails.

Fixes #1411

([I need this for hardly as well](https://github.com/packit/hardly/blob/159f66731a9241217a338bc39b6ced5218b27687/hardly/handlers/distgit.py#L193))

---

RELEASE NOTES BEGIN
N/A
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/reporting.py
+++ b/packit_service/worker/reporting.py
@@ -177,6 +177,14 @@ class StatusReporter:
                 markdown_content=markdown_content,
             )
 
+    @staticmethod
+    def is_final_state(state: BaseCommitStatus) -> bool:
+        return state in {
+            BaseCommitStatus.success,
+            BaseCommitStatus.error,
+            BaseCommitStatus.failure,
+        }
+
     def _add_commit_comment_with_status(
         self, state: BaseCommitStatus, description: str, check_name: str, url: str = ""
     ):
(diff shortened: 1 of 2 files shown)


packit_service/worker/reporting.py
```diff
@@ -190,6 +198,11 @@ def _add_commit_comment_with_status(
             )
             + f"\n\n{description}"
         )
+        if not self.is_final_state(state):
+            # To avoid multiple comments for non-final states
+            logger.debug(f"Not a final state -> not adding a comment with {body}")
```
lachmanfrantisek, line 203:
```suggestion
            logger.debug(f"Not a final state -> not adding a comment with body:\n{body}")
```

I am not sure if we want to log the whole body but it shouldn't happen so often so I am not against it.

lachmanfrantisek (approved):
Thanks!

packit_service/worker/reporting.py
mfocko: I wouldn't say that taking in account the current state of GitHub ;)

jpopelka: I've shortened it.

PR #1417, opened 2022-03-25
Support for packit instance configuration

lachmanfrantisek:
This will allow users to specify packit instances (=deployments) that will work on the configured jobs.

Nothing will change for the current users of the production instance, but the stage application will stop reacting to the stage jobs. And users need to specify that explicitly. => Do we want to start with some warning? Create an issue? Or, just fix that manually? (Note that not doing the filtering stops us from using stage instance for downstream.)

Next steps (after merging):
* [ ] update our projects to react to stage instances
* [ ] set up a stage FAS identity that will use prod Fedora instances (We can specify, which instance will handle what.)
* [ ] document the option

Requires: packit/packit#1531

RELEASE NOTES BEGIN
There is a new `packit_instances` key that you can use to specify the Packit instances you want to use for working on your jobs.
Nothing will change for our production users, but users of our stage instance need to use this key to preserve the support of the stage instance -- they can set both `stg` and `prod` in the `packit_instances` list to use both, or use just one. Just be careful with the downstream jobs where both instances work with the same services. This new option works like other Packit options so you can set it on the top level and/or (re)define it on the job level. More information about our staging instance can be found here: https://github.com/packit/packit/discussions/1530
RELEASE NOTES END


TomasTomecek (approved):
Thank you for cleaning up the tests!

LGTM

lachmanfrantisek:
recheck

lbarcziova (approved):
lgtm

lachmanfrantisek:
recheck

jpopelka (approved):

TomasTomecek (approved):

Issue #1409, opened 2022-03-22
downstream koji builds: notify when something's wrong


TomasTomecek:
We need a mechanism to notify our users when something's wrong when Packit automatically submits a dist-git build into koji.

Ideally we'd have a view for this in the dashboard.

Acceptance Criteria: user is notified when a koji build fails or fails to be submitted.

lachmanfrantisek:
It looks like we can hit a non-packit exception. The known issues should be covered by some `PackitException` and a commit comment should be created:

https://github.com/packit/packit-service/blob/4d67f8e9a06243a6b0dc60e4ad05c2d45edaa238/packit_service/worker/handlers/distgit.py#L358-L374

lachmanfrantisek:
> Ideally we'd have a view for this in the dashboard.

Regarding the build results, we agreed in #1310 that we'll leave it to Fedora notifications...

But I agree we should cover the failed submissions somehow.

lachmanfrantisek:
Oh, and `Commit comments are not supported on Pagure.`....;(

https://sentry.io/organizations/red-hat-0p/issues/3122397893/?environment=prod&project=1767823&query=is%3Aunresolved

lachmanfrantisek:
As a team, we agreed on adding a new config option to set up a project that will be used for creating issues.

* Project on any git forge supported by Packit can be used. (Pagure, GitHub and GitLab where packit has an identity.)
* By default, nothing is used to not spam our users.
* Like other config options, you can set it on the top level but (re)define on job level.
* Optionally, we can apply this for other use-cases where we create issues (e.g. failed propose-downstream).

PR #1421, opened 2022-03-28
Issue repository can be used for failures of downstream jobs

lachmanfrantisek:
Fixes: #1409

Requires: packit/packit#1536

---

RELEASE NOTES BEGIN
You can newly specify an `issue_repository` key in your config if you want to be
notified about errors of the downstream jobs (Koji build and Bodhi update).
This repository can be any GitHub/GitLab/Pagure repository
where issues are enabled and Packit has an identity on that instance.
(Let us know if you need some other instance to be supported.)
By default, no issue will be created and if the issue already exists,
a new comment will be added.
As usual, this new option works like other Packit options so you can set it
on the top level and/or (re)define on the job level.
RELEASE NOTES END

### DIFF
--- a/packit_service/config.py
+++ b/packit_service/config.py
@@ -272,17 +272,26 @@ class ServiceConfig(Config):
 class PackageConfigGetter:
     @staticmethod
     def create_issue_if_needed(
-        project: GitProject, title: str, message: str
+        project: GitProject,
+        title: str,
+        message: str,
+        comment_to_existing: Optional[str] = None,
     ) -> Optional[Issue]:
         # TODO: Improve filtering
         issues = project.get_issue_list()
         title = f"[packit] {title}"
 
-        if any(title in issue.title for issue in issues):
-            return None
+        for issue in issues:
+            if title in issue.title:
+                if comment_to_existing:
+                    issue.comment(body=comment_to_existing)
+                    logger.debug(f"Issue #{issue.id} updated: {issue.url}")
+                return None
 
         # TODO: store in DB
-        return project.create_issue(title=title, body=message)
+        issue = project.create_issue(title=title, body=message)
+        logger.debug(f"Issue #{issue.id} created: {issue.url}")
+        return issue
 
     @staticmethod
     def get_package_config_from_repo(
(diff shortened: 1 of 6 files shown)


lachmanfrantisek:
recheck

packit_service/config.py
```diff
@@ -272,17 +272,26 @@ def get_project_to_sync(self, dg_repo_name, dg_branch) -> Optional[ProjectToSync
...
         # TODO: Improve filtering
         issues = project.get_issue_list()
         title = f"[packit] {title}"
 
-        if any(title in issue.title for issue in issues):
-            return None
+        for issue in issues:
+            if title in issue.title:
+                if comment_to_existing:
+                    issue.comment(body=comment_to_existing)
```
mfocko, line 287: What happens with `message` here?

packit_service/config.py
```diff
@@ -272,17 +272,26 @@ def get_project_to_sync(self, dg_repo_name, dg_branch) -> Optional[ProjectToSync
 class PackageConfigGetter:
     @staticmethod
     def create_issue_if_needed(
-        project: GitProject, title: str, message: str
+        project: GitProject,
+        title: str,
+        message: str,
+        comment_to_existing: Optional[str] = None,
```
mfocko, line 278: I have only seen it used as `boolean` parameter, do you have any plans in mind with the `str`? ;)

mfocko (approved):

packit_service/config.py
lachmanfrantisek: It's used as a test for the comment. I wanted to differentiate the text used for issue description and issue comment.

packit_service/config.py
lachmanfrantisek: It's not used. I wanted to differentiate the text used for issue description and issue comment.

I know it's not used yet but we can use a different text when creating the issue (e.g. add contacts) and avoid this to be mentioned in each comment.

lbarcziova (approved):
:+1:

packit_service/config.py
lachmanfrantisek: @mfocko OK, I've removed the contact part from the comment since it will be in the description. (And now, it makes sense to have it different..;-)

Issue #1071, opened 2021-04-28
Filtering issues


mfocko:
Filtering of issues in case we want only one issue regarding configuration shared across multiple runs of p-s.

Current solution: filtering by the title of the issue
Tried solution: filtering by the author of the issue

Tried solution is not possible, since we use Tokman that provides unique tokens **per** repository and therefore service itself does not have any PyGithub object, that could return login of the currently authenticated "user".

1. We cannot guess the name, since:
   - GitHub: `app/packit-as-a-service`
   - GitLab: `packit-as-a-service`
2. We cannot directly touch `GitProject`'s pygithub instance, not compatible with GitLab
3. Extending `GitProject`'s interface with `get_username` and whatnot, does not make sense...
4. Extending `Tokman`'s interface, again... Does not relate, would not be compatible with GitLab

Possible solution: Tag our issues with `packit` label, but that requires users to have that label set up...

lachmanfrantisek:
> Possible solution: Tag our issues with `packit` label, but that requires users to have that label set up...

Can't we do that for them?

mfocko:
> Can't we do that for them?

We have a write access, so it should be possible to label the newly-created issue. Apparently we can create them too, but not sure how _nice_ it is to create label in someone's repository :thinking:

lachmanfrantisek:
Good point. What do the others do? Stale-bot/...

TomasTomecek:
We can also store the issue ID in our database.

mfocko:
> We can also store the issue ID in our database.

Yeah, I was thinking about that... IMO it requires some maintenance on our side (cleaning up closed issues), on the other hand we could probably do that before checking against the DB and creating new issue if necessary, but still results in "leftover" data in DB.

I like the label idea, but it seems a bit invasive :shrugs:

mfocko:
Still needed, so... let's be invasive? 😏 😂

lachmanfrantisek:
Can't we use this? https://github.com/packit/ogr/pull/655

mfocko:
> Can't we use this? [packit/ogr#655](https://github.com/packit/ogr/pull/655)

This issue is about the issues (pun intended) whereas the linked PR allows getting app that posted the check run.

The major issue here is that we cannot filter based on the user / service account, because in each service it has a different login (github: `app/packit-as-a-service`, gitlab: `packit-as-a-service`; diplomatically I'm leaving out filtering on Pagure…).

[...]

PR #1427, opened 2022-03-31
Do not duplicate issue for failed propose downstream

majamassarini:
TODO:
If we want, we can enhance the create_issue_if_needed method and save the issue in the database
instead of trying to guess if it already exist using string matching.
I feel like that the `title in issue.title` check can lead to some errors (even if I think this are rare)

Should fix packit-service#1380

---

RELEASE NOTES BEGIN
When multiple propose downstream attempts for the same PR fail the error messages are sent to to the same issue, as comments, instead of creating multiple new issues.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/handlers/distgit.py
+++ b/packit_service/worker/handlers/distgit.py
@@ -218,9 +218,11 @@ class ProposeDownstreamHandler(JobHandler):
             f"{msg_retrigger}\n"
         )
 
-        self.project.create_issue(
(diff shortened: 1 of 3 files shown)


lachmanfrantisek (approved):
:+1: Thanks!

I would merge this one as is and resolve the database elsewhere -- we even have an issue for that: #1071

TomasTomecek (approved):
what an easy fix and a ton of tests :D nicely done!

PR #1434, opened 2022-04-05
Do not search for downstream koji builds

majamassarini:
Initialize all the koji builds as scratch.
Do not perform a join looking for downstream koji builds,
since it costs too much cpu time,
and right now there are no downstream koji builds saved into the db.


lachmanfrantisek (approved):
Thanks!

majamassarini:
recheck

csomh:
recheck

PR #1435, opened 2022-04-06
Propose downstream reporting

### LINKED ISSUE #1293: Status for `propose-downstream` job
Use check-runs/commit-status to show users the state of the  `propose-downstream` job in the place it was triggered (release commit or a new commit for the branch push once #1284 is done).

For `propose-downstream` triggered by an issue comment, create a comment as well as the commit status.

Relates to packit/packit-service#1292 and packit/dashboard#155 (link the dashboard view once it's implemented).

lbarcziova:
Fixes #1293

As for the helper classes, I extracted the common logic to the `BaseJobHelper` and then added propose-downstream specific logic into `ProposeDownstreamJobHelper`. Since we already have also `TestingFarmJobHelper` which was placed directly in the `worker` directory, I created a dedicated `helpers` directory and moved these + the `build` dir including build helpers there. It is still not ideal (especially the testing farm inheriting from `BaseBuildJobHelper` but being placed outside of `build` dir),  but current class hierarchy is:
- BaseJobHelper
  - BaseBuildJobHelper
      - CoprBuildJobHelper
      - KojiBuildJobHelper
       - TestingFarmJobHelper
  - ProposeDownstreamHelper
   
We can still discuss some better structure.

---

RELEASE NOTES BEGIN
TBD
RELEASE NOTES END


mfocko:
If fails with `KeyError` for `copr_url` in `test_fedmsg`, ignore it, or cherry pick my commit and close my PR ;)

packit_service/worker/handlers/distgit.py
```diff
@@ -284,31 +300,63 @@ def run(self) -> TaskResults:
...
+                        "we were not able yet to download the archive. "
+                    )
                     model.set_status(staus=ProposeDownstreamTargetStatus.retry)
+                    self.propose_downstream_helper.report_status_to_branch(
+                        branch=branch,
+                        description="Propose downstream is being retried because "
+                        "we were not able yet to download the archive. ",
+                        state=BaseCommitStatus.pending,
+                        url=url,
+                    )
```
TomasTomecek, line 343: I like this: was thinking if we should do pending -> fail -> pending -> ...

but that would be confusing: pending -> pending with an accurate message makes more sense

packit_service/worker/jobs.py
```diff
@@ -233,6 +237,65 @@ def service_config(self):
             self._service_config = ServiceConfig.get_service_config()
         return self._service_config
 
+    def report_task_accepted(
```
TomasTomecek, line 240: glad you refactored the code into a new method

long term we should start breaking down jobs.py even further and come up with a clearer structure

TomasTomecek (approved):
checked only briefly

PR #1437, opened 2022-04-07
fix: Remove the option to override the specfile_path

csomh:
This possibility was added in order to support a different default
specfile_path in the v1 source-git repositories created for CentOS
Stream.

As those repositories are being decommissioned, this code can be
removed.


TomasTomecek (approved):
nice, very good catch

Issue #1439, opened 2022-04-07
Steve needs love


nikromen:
Currently, our code in `jobs.py` is not so well decomposed and it is really hard to know what are we doing - there are long statements and extra long methods/functions - and one can spend a long time figuring out what's going on. 
It would be nice to decompose/refactor long methods/functions/statements.

Mainly speaking of these methods and statements in them:

- [x] `get_handlers_for_event` 
- [ ] `SteveJobs.process_jobs`
- [ ] `SteveJobs.process_message`

TomasTomecek:
that friday label :D

Sounds like a great hackfest task to do

big +1

PR #1440, opened 2022-04-07
Refactor `get_handlers_for_event` function

nikromen:
It is hard to properly name functions and variables :D -> suggestions are welcomed!

Also - new argument `packit_comment_command_prefix` appeared without documentation so I added it there - can you check it if it's correct? :innocent: 

Related to #1439

---

RELEASE NOTES BEGIN
￼
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/jobs.py
+++ b/packit_service/worker/jobs.py
@@ -55,6 +55,86 @@ from packit_service.worker.result import TaskResults
 logger = logging.getLogger(__name__)
 
 
+def get_handlers_for_comment_and_rerun_event(
+    event: Event, packit_comment_command_prefix: str
+) -> Set[Type[JobHandler]]:
+    """
+    Get all handlers that can be triggered by comment (e.g. `/packit build`) or check rerun.
+
+    For comment events we want to get handlers mapped to comment commands. For check rerun event we
+     want to get handlers mapped to check name job.
+     These two sets of handlers are mutually exclusive.
+
+    Args:
+        event: Event which we are reacting to.
+        packit_comment_command_prefix: `/packit` for packit-prod or `/packit-stg` for stg.
+
+    Returns:
+        Set of handlers that are triggered by a comment or check rerun job.
+    """
+    handlers_triggered_by_job = None
+
+    if isinstance(event, AbstractCommentEvent):
+        handlers_triggered_by_job = get_handlers_for_comment(
+            event.comment, packit_comment_command_prefix
+        )
+
+        if handlers_triggered_by_job and not isinstance(
+            event, PullRequestCommentPagureEvent
+        ):
+            event.comment_object.add_reaction(COMMENT_REACTION)
+
+    if isinstance(event, CheckRerunEvent):
+        handlers_triggered_by_job = get_handlers_for_check_rerun(event.check_name_job)
+
+    return handlers_triggered_by_job
+
+
+def is_handler_matching_the_event(
+    event: Event,
+    handler: Type[JobHandler],
+    allowed_handlers: Set[Type[JobHandler]],
+) -> bool:
+    """
+    Decides whether handler matches to comment or check rerun job and given event supports handler.
+
+    Args:
+        event: Event which we are reacting to.
+        handler: Handler which we are observing whether it is matching to job.
+        allowed_handlers: Set of handlers that are triggered by a comment or check rerun
+         job.
+    """
+    handler_matches_to_comment_or_check_rerun_job = (
+        allowed_handlers is None or handler in allowed_handlers
+    )
+
+    return (
+        isinstance(event, tuple(SUPPORTED_EVENTS_FOR_HANDLER[handler]))
+        and handler_matches_to_comment_or_check_rerun_job
+    )
+
+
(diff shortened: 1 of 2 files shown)


packit_service/worker/jobs.py
```diff
@@ -53,24 +53,45 @@
...
 
 
-def get_handlers_for_event(
-    event: Event, package_config: PackageConfig, packit_comment_command_prefix: str
+def get_handlers_triggered_by_job(
+    event: Event, packit_comment_command_prefix: str
 ) -> Set[Type[JobHandler]]:
-    """
-    Get all handlers that we need to run for the given event.
+    handlers_triggered_by_job = None
```
TomasTomecek, line 59: docstring please :)

packit_service/worker/jobs.py
```diff
@@ -53,24 +53,45 @@
...
+        handlers_triggered_by_job = get_handlers_for_check_rerun(event.check_name_job)
+
+    return handlers_triggered_by_job
+
+
+def handler_matches_to_job(
+    event: Event,
+    handler: Type[JobHandler],
+    handlers_triggered_by_job: Set[Type[JobHandler]],
+) -> bool:
```
TomasTomecek, line 81: docstring please :)

TomasTomecek (commented):
+1

some basic unit tests would be great to have here since you decomposed the functions well

packit_service/worker/jobs.py
```diff
@@ -55,24 +55,75 @@
 logger = logging.getLogger(__name__)
 
 
-def get_handlers_for_event(
-    event: Event, package_config: PackageConfig, packit_comment_command_prefix: str
+def get_handlers_triggered_by_job(
```
lachmanfrantisek, line 58: I would be careful with using `job` here since we use it more for the implementation, e.g. `job_config`, `job_type`.

What about preserving the word `event`?

Also, it looks like this is only about comments and reruns.


What about something really explicit like `get_handlers_for_reaction_event` or `get_handlers_for_comment_and_rerun_event`?

packit_service/worker/jobs.py
```diff
@@ -55,24 +55,75 @@
...
+        ):
+            event.comment_object.add_reaction(COMMENT_REACTION)
+
+    if isinstance(event, CheckRerunEvent):
+        handlers_triggered_by_job = get_handlers_for_check_rerun(event.check_name_job)
+
+    return handlers_triggered_by_job
+
+
+def handler_matches_to_job(
```
lachmanfrantisek, line 93: Yes, naming is hard..;)

What about this to make it obvious it returns a boolean?

```suggestion
def is_handler_matching_the_event(
```

packit_service/worker/jobs.py
```diff
@@ -55,24 +55,75 @@
...
+    event: Event,
+    handler: Type[JobHandler],
+    handlers_triggered_by_job: Set[Type[JobHandler]],
```
lachmanfrantisek, line 96: This argument is really confusing to me, tbh.

What about something like `allowed_handlers`?

packit_service/worker/jobs.py
```diff
@@ -55,24 +55,75 @@
...
+        handlers_triggered_by_job is None or handler in handlers_triggered_by_job
+    )
+
+    return (
+        isinstance(event, tuple(SUPPORTED_EVENTS_FOR_HANDLER[handler]))
+        and handler_matches_to_comment_or_check_rerun_job
+    )
+
+
+def get_jobs_matching_trigger(
```
lachmanfrantisek, line 117: Maybe `get_jobs_matching_event` since we use `Event` across the codebase and also as an argument.

packit_service/worker/jobs.py
```diff
@@ -81,23 +132,36 @@ def get_handlers_for_event(
...
+    """
+    Get all handlers that we need to run for the given event.
 
-    if isinstance(event, CheckRerunEvent):
-        handlers_triggered_by_check_rerun = get_handlers_for_check_rerun(
-            event.check_name_job
-        )
+    We need to return all handler classes that:
+    - can react to the given event AND
+    - are configured in the package_config (either directly or as a required job)
```
lachmanfrantisek, line 146: I thought about this for some time, but what about returning pair of handler and job config so we don't need to do the similar search again in `get_config_for_handler_kls`.

(Yes, it does not need to be solved here...;)

lachmanfrantisek (commented):
Thanks a lot! Very nice improvement!


I needed to change my naming suggestions multiple times when going through the code. Naming is really hard...;)

packit_service/worker/jobs.py
nikromen: yes we can do that here but I would keep that function since we use it in `babysit.py` and it seems like it's doing its job there. Should I rather move the method to `babysit.py` then since it should be used only there or keep it in `jobs.py`? WDYT?

lachmanfrantisek: Ok, then it makes sense to preserve it... And you can leave it in the current space I think since it's related to the function you've refactored.

lachmanfrantisek (approved):
Thanks!

Issue #1375, opened 2022-03-02
`/packit retest-failed` and `/packit rebuild-failed` don't work as expected


nikromen:
Expected behavior is to restart/rebuild only failed tests/builds.
Currently these commands:
- `rebuild-failed` restarts everything - should restart only failed builds
- `retest-failed` restarts every tests - should restart only failed tests.

everything seems to working fine to [this point](https://github.com/packit/packit-service/blob/main/packit_service/worker/events/comment.py#L88). *_targets_override fills up with correct targets to rebuild/retest. Something after that might be broken. 

TODO:
- [ ] fix retest-failed
- [x] fix rebuild-failed
- [ ] fix - getting models from DB with commit is not enough

lachmanfrantisek:
It looks like it restarts everything for `Retry` button in the Checks UI as well...;(

I hit that here: https://github.com/packit/specfile/pull/25/checks?check_run_id=5656718240

nikromen:
Do you mean the `Re-run` button next to a failed job?
![Screenshot_20220323_224024](https://user-images.githubusercontent.com/70757578/159800439-d2f620f0-4ce4-4011-8d94-718d6c85669c.png)

lachmanfrantisek:
> Do you mean the `Re-run` button next to a failed job?

Yes, sorry about the bad text for that, but I really meant `Re-run` for a single result.

> Also Re-run for individual jobs works correctly... seems like there is something between which overrides the targets_override

I don't think so. Maybe I mixed that with something else but still think it triggered everything...

nikromen:
> I don't think so. Maybe I mixed that with something else (multi-job scenario, identifiers,...) but still think it triggered everything...

Oh you are right... again I tested it only for testing farm... It works... but for builds everything is restarted :/ 
It definitely worked when Laura implemented #1227 so something definitely overrides the `targets_override` :D

nikromen:
closed with #1479

PR #1441, opened 2022-04-07
Add option to filter builds/test from comment by a given statuses

nikromen:
Just to add to code more reusability - if we may want to filter models by their different statuses in some CommentEvent - you can now filter them by a given statuses.

And also the main point of this PR - I added some logs so I can see which targets passed to `targets_override` in logs because I am clueless when debugging #1375 :D 

---

RELEASE NOTES BEGIN
￼
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/events/comment.py
+++ b/packit_service/worker/events/comment.py
@@ -8,6 +8,9 @@ from logging import getLogger
 from typing import Dict, Optional, Set
 
 from ogr.abstract import Comment
+
+from packit_service.constants import PG_BUILD_STATUS_FAILURE
+from packit_service.models import TestingFarmResult
 from packit_service.service.db_triggers import (
     AddIssueDbTrigger,
     AddPullRequestDbTrigger,
@@ -88,14 +91,20 @@ class AbstractPRCommentEvent(AddPullRequestDbTrigger, AbstractCommentEvent):
     def build_targets_override(self) -> Optional[Set[str]]:
         if not self._build_targets_override and "rebuild-failed" in self.comment:
             self._build_targets_override = (
-                super().get_all_build_failed_targets() or None
+                super().get_all_build_targets_by_status([PG_BUILD_STATUS_FAILURE])
+                or None
             )
         return self._build_targets_override
 
     @property
     def tests_targets_override(self) -> Optional[Set[str]]:
         if not self._tests_targets_override and "retest-failed" in self.comment:
-            self._tests_targets_override = super().get_all_tf_failed_targets() or None
+            self._tests_targets_override = (
+                super().get_all_tf_targets_by_status(
+                    [TestingFarmResult.failed, TestingFarmResult.error]
+                )
+                or None
+            )
         return self._tests_targets_override
 
     def get_dict(self, default_dict: Optional[Dict] = None) -> dict:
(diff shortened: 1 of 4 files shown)


lbarcziova (commented):
The `FAILED_MODELS` constant looks a little bit fragile to me, could we use already existing and used in the implementation [TestingFarmResult](https://github.com/packit/packit-service/blob/9b26d91a8eeaa877a8ca351c67c02898d2b608c8/packit_service/models.py#L1494) enum states for TF and [PG_BUILD_STATUS_FAILURE](https://github.com/packit/packit-service/blob/9b26d91a8eeaa877a8ca351c67c02898d2b608c8/packit_service/constants.py#L55) for the build?

Also now looking into PG_BUILD_STATUS_FAILURE=`failure` and `failed` used in the implementation for rebuilding failed, couldn't this also cause the bugs with `packit rebuild-failed` ? :D (but since the retesting failed also doesn't work, there is probably still also something else wrong)

nikromen:
> Also now looking into PG_BUILD_STATUS_FAILURE=failure and failed used in the implementation for rebuilding failed, couldn't this also cause the bugs with packit rebuild-failed ?

Hmm yes that's another bug :D thank you! but it's not the current problem since the result should be opposite - it should rebuild less builds... not more (everything) :D

lbarcziova:
> Hmm yes that's another bug :D thank you! but it's not the current problem since the result should be opposite - it should rebuild less builds... not more (everything) :D

But as we already discussed, if you create a new copr build task where `build_targets_override` is None (because there are no builds with `failed` status, only with `failure`), builds will be run normally, for each target in the config.

nikromen:
Yes but also as we discussed user probably won't use rebuild-failed to all passed builds. Now we don't have functionality in `targets_override` to basically "skip" the building part. So the solution can be that overriding can be done with empty set:
```py
if targets_override is not None:
    override here
```
so we can override it with empty set. But as far as I tested - this still isn't an issue. I think that `targets_override` aren't even called (it's behaving like normal /packit build) - that's why I am adding these logs to code :D

lbarcziova:
> Now we don't have functionality in targets_override to basically "skip" the building part.

Yes, because it doesn't really make sense to create copr build handler task if we don't want to run any builds...

> Yes but also as we discussed user probably won't use rebuild-failed to all passed builds. 

That's still true. But since there is a bug and no builds are found in the DB because of wrong status, this scenario happens.

>  I think that targets_override aren't even called (it's behaving like normal /packit build) - that's why I am adding these logs to code :D

That's what I wrote :D it behaves like normal /packit build because there are no builds with 'failed' status found in the DB, therefore the build_targets_override is empty.

nikromen:
> That's still true. But since there is a bug and no builds are found in the DB because of wrong status, this scenario happens.

oh, now I get it :D makes sense, so can you please review the code once more so we can try it? :innocent:

packit_service/constants.py
```diff
@@ -207,3 +210,6 @@ def from_number(number: int):
     "*As a project maintainer, you can trigger the build and test jobs manually "
     "via `/packit build` comment or only test job via `/packit test` comment.*",
 )
+
+FAILED_TF_MODEL_STATUS: List[str] = [TestingFarmResult.failed, TestingFarmResult.error]
+FAILED_BUILD_MODEL_STATUS: List[str] = [PG_BUILD_STATUS_FAILURE]
```
lbarcziova, line 215: I meant these to use directly in the code but will leave it up to you (but I can see from the tests it probably causes circular imports).

lbarcziova (approved):
LGTM once tests are passing

Issue #1364, opened 2022-02-25
SPIKE: Improve the event processing time


lachmanfrantisek:
Last week, the service had troubles with our [SLO1: Changes to GitHub PRs receive a status update within 15 seconds in 99% of cases](https://github.com/packit/research/blob/main/error-budgets/SLOs-SLIs.md#slo1-changes-to-github-prs-receive-a-status-update-within-15-seconds-in-99-of-cases).

We scaled up but we want to take a deeper look at some optimizations and also to be sure that scaling matches the increase of the real usage.

The task of this spike/research is to:
* [x] Take a look at how we measure the indicators (if the current state is what we expect and if we can/want to get more measurements to have a better idea of where exactly we have problems). (#1446)
* [x] Identify the problematic parts of the process. (#1677)
* [x] Check that (healthy) scaling helps with this problem. ([research](https://github.com/packit/research/tree/main/workers-scaling))
* [ ] Provide suggestions for improvements. Here are the ideas we have discussed:
    * Can we uninstall a repo? - Not receive events, after some period of time when we didn’t produce any meaningful work.
    * If Packit is installed for all the repos, we receive events for repos where there is no packit.yaml. (Save this info to the database instead of checking GitHub.)
    * We could add higher prio to tasks from repos who are active users.

lachmanfrantisek:
We have other more specific cards for this and haven't been asked about this by users recently.. Let's close this and open more specific issues if/when needed.

PR #1446, opened 2022-04-12
Record the task accepted time right before reporting the statuses

lbarcziova:
Related to #1364

Currently, we record the metrics about setting the initial status right after setting them. But in the case of many targets to report to, interaction with Github API can take some seconds. This way, we will record the time just before setting the status for the first target, which is probably more precise.

---

RELEASE NOTES BEGIN
N/A
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/jobs.py
+++ b/packit_service/worker/jobs.py
@@ -259,12 +259,14 @@ def get_config_for_handler_kls(
 
 
 def push_initial_metrics(
-    event: Event, handler: JobHandler, build_targets_len: Optional[int] = None
+    task_accepted_time: datetime,
+    event: Event,
+    handler: JobHandler,
+    number_of_build_targets: Optional[int] = None,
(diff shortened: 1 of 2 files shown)


packit_service/worker/jobs.py
```diff
@@ -202,12 +202,14 @@ def get_config_for_handler_kls(
 
 
 def push_initial_metrics(
-    event: Event, handler: JobHandler, build_targets_len: Optional[int] = None
+    task_accepted_time: datetime,
+    event: Event,
+    handler: JobHandler,
+    build_targets_len: Optional[int] = None,
```
lachmanfrantisek, line 208: What about `number_of_build_targets` instead?

lachmanfrantisek (approved):
:+1:

That makes sense.

lbarcziova:
recheck

PR #1449, opened 2022-04-13
Propose downstream improvements

lbarcziova:
Accessing `local_project` currently requires cloning of the repo and we can get the GitProject directly. It doesn't make much sense to clone the repo only for getting the `default_dg_branch` for the creation of the Github checks (we need the `default_dg_branch` when getting branches to report for).

The `AbstractIssueCommentEvent` is used only for propose-downstream, so get the `commit_sha` the same way as for the release event. This will be used for reporting.

---

RELEASE NOTES BEGIN
N/A
RELEASE NOTES END

### DIFF
--- a/tests/unit/test_propose_downstream.py
+++ b/tests/unit/test_propose_downstream.py
@@ -6,8 +6,6 @@ from flexmock import flexmock
 
 from packit.config import PackageConfig, JobConfig, JobType
 from packit.config.job_config import JobMetadataConfig, JobConfigTriggerType
-from packit.distgit import DistGit
-from packit.local_project import LocalProject
 from packit_service.config import ServiceConfig
 from packit_service.worker.helpers.propose_downstream import ProposeDownstreamJobHelper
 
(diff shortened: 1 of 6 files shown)


tests/unit/test_propose_downstream.py
```diff
@@ -65,21 +63,18 @@
...
 )
 def test_branches(jobs, job_config_trigger_type, branches_override, branches):
+    project = flexmock(
+        default_branch="main",
+    )
+    service_config = ServiceConfig().get_service_config()
+    flexmock(ServiceConfig, get_project=lambda url: project)
     propose_downstream_helper = ProposeDownstreamJobHelper(
-        service_config=ServiceConfig(),
+        service_config=service_config,
```
lachmanfrantisek, line 72: Is this `service_config` instantiation here for some reason?

lachmanfrantisek (approved):
:+1:

tests/unit/test_propose_downstream.py
lbarcziova: fixed, thanks!

lbarcziova: fixed, thanks!

PR #1452, opened 2022-04-19
Add more reasonable logs for searching copr/tf targers by status

nikromen:
Adding more reasonable logs to this may finally tell what's wrong with `rebuild`/`retest` since regarding to current logs the bug will be 
- getting wrong commit sha or
-  getting wrong models according to commit sha or
-  wrong filtering of failed models

And ss you can see I cannot tell according to logs which of these three options works and which not and according to tests everything is working :D yes I should learn to write more reasonable logs

---

RELEASE NOTES BEGIN
￼
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/events/event.py
+++ b/packit_service/worker/events/event.py
@@ -473,7 +473,7 @@ class AbstractForgeIndependentEvent(Event):
         statuses_to_filter_with: List[str],
     ) -> Optional[Set[str]]:
         logger.info(
-            f"Trying to filter targets with possible status in: {statuses_to_filter_with}"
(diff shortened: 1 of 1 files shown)


packit_service/worker/events/event.py
```diff
@@ -489,6 +489,9 @@ def get_all_tf_targets_by_status(
         if self.commit_sha is None:
             return None
 
+        logger.debug(
+            f"Getting failed Testing Farm targets for commit sha: {self.commit_sha}"
```
lbarcziova, line 493: for consistency, I would log the repo name here as well

lbarcziova (approved):

PR #1454, opened 2022-04-19
Do not use stage fedora instances from service

lachmanfrantisek:
Both Packit prod and stage work with prod instances
so we can use them for our projects
and to have the same environment for both of them.


lachmanfrantisek:
recheck

jpopelka (approved):

PR #1457, opened 2022-04-21
Catch the correct exceptions when reporting in Github

lbarcziova:
Probably since packit/ogr#690 change, we wrap the exceptions and therefore don't catch the correct ones.

---

RELEASE NOTES BEGIN
￼
RELEASE NOTES END


packit_service/worker/reporting.py
```diff
@@ -5,10 +5,10 @@
 from enum import Enum
 from typing import Optional, Union, Dict
 
-import github
 import gitlab
```
mfocko, line 8: We should probably look into the reason for `gitlab` being here later on…

tests/unit/test_reporting.py
```diff
@@ -258,15 +257,15 @@ def test_set_status_github_check(
             "https://api.packit.dev/build/111/logs",
             CommitStatus.success,
             (
-                github.GithubException,
+                GithubAPIException,
                 # https://docs.python.org/3/library/inspect.html#inspect.signature
                 # to account for changes in positional arguments: pygithub 1.55 added headers
                 # as additional positional argument; this creates an iterable and sets None
-                # for every argument of GithubException.__init__ except for 'self'
+                # for every argument of GithubAPIException.__init__ except for 'self'
```
mfocko, line 264: I'm not sure about correctness of this part, since it's wrapped in ogr issue

tests/unit/test_reporting.py
```diff
@@ -415,15 +414,15 @@ def test_report_status_by_comment(
...
                 [
                     None
                     for param_name, param in signature(
-                        github.GithubException.__init__
+                        GithubAPIException.__init__
```
mfocko, line 425: same here…

looking at it again, it seems like it is some kind of a workaround for different versions of PyGithub? IMO it's no longer relevant :shrug: but I might be wrong :D

mfocko (approved):
apart from the inspection of GitHub constructor, lgtm

packit_service/worker/reporting.py
lbarcziova: similarly as for Github, we are catching `gitlab.exceptions.GitlabCreateError` for creating statuses and fallback to comments but if I checked the ogr PR correctly, this exceptions should not be wrapped, am I right?

mfocko: iirc we wrap both now… there is an easy way to find out tbh :D

lbarcziova: ok, will add it

mfocko: ya, it is…

_mf linking the Sentry issues_

lbarcziova:
Gitlab part added, PTAL

packit_service/worker/reporting.py
```diff
@@ -308,7 +306,7 @@ def set_status(
             self.project_with_commit.set_commit_status(
                 self.commit_sha, state_to_set, url, description, check_name, trim=True
             )
-        except gitlab.exceptions.GitlabCreateError as e:
+        except GitlabAPIException as e:
             # Ignoring Gitlab 'enqueue' error
             # https://github.com/packit-service/packit-service/issues/741
             if e.response_code != 400:
```
mfocko, line 312: How convenient that it's `.response_code` both in `GitlabCreateError` and `ogr.APIException` :D

mfocko (approved):
:rocket:

PR #1459, opened 2022-04-21
Fix rebuild-failed

nikromen:
Previously /packit rebuild-failed searched in DB for models with matching commit sha and project name. Searching by project name is redundant and it seems like project names in copr and repo's project names can differ (somehow) so result of searching is sometimes empty list.

Related to #1375 

---

RELEASE NOTES BEGIN
￼
RELEASE NOTES END

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -1013,6 +1013,15 @@ class CoprBuildTargetModel(ProjectAndTriggersConnector, Base):
             return query.all()
 
     @classmethod
+    def get_all_by_commit(
+        cls, commit_sha: str
+    ) -> Optional[Iterable["CoprBuildTargetModel"]]:
+        """Returns all builds that match a given commit sha"""
+        with get_sa_session() as session:
+            query = session.query(CoprBuildTargetModel).filter_by(commit_sha=commit_sha)
+            return query.all()
+
+    @classmethod
     def create(
         cls,
         build_id: str,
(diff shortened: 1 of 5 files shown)


packit_service/worker/events/comment.py
```diff
@@ -91,7 +91,9 @@ def comment_object(self) -> Optional[Comment]:
     def build_targets_override(self) -> Optional[Set[str]]:
         if not self._build_targets_override and "rebuild-failed" in self.comment:
             self._build_targets_override = (
-                super().get_all_build_targets_by_status([PG_BUILD_STATUS_FAILURE])
+                super().get_all_build_targets_by_status(
+                    statuses_to_filter_with=[PG_BUILD_STATUS_FAILURE]
```
mfocko, line 95: I have a feeling that the mess we have in enums might be the cause of this…

mfocko (approved):
Copr repository and git repository can definitely differ, since by default we build in `packit` namespace, but with the option to specify where to build.

And it is more than likely that `project_name` in context of `CoprBuild` is »Copr project« whereas you filtered with `self.project.repo` which was probably the »git repo«

packit_service/models.py
```diff
@@ -1012,6 +1012,15 @@ def get_all_by(
             )
             return query.all()
 
+    @classmethod
+    def get_all_by_commit(
+        cls, commit_sha: str
+    ) -> Optional[Iterable["CoprBuildTargetModel"]]:
+        """Returns all builds which matches a given commit sha"""
```
TomasTomecek, line 1019:
```suggestion
        """Returns all builds that match a given commit sha"""
```
sorry for picking nits

TomasTomecek (approved):

Issue #1257, opened 2021-10-26
Use fasjson for automatic allowlisting 


lbarcziova:
FAS2 (which we use [here](https://github.com/packit/packit-service/blob/f5f321f2a663bd797868d0ef17559309c9c7816c/packit_service/worker/allowlist.py#L8)) [is being retired](https://pagure.io/fedora-infrastructure/issue/10254) and doesn't provide up to date info about users , from what I read we can use [fasjson client](https://fasjson-client.readthedocs.io/en/latest/migration.html#people). 

TODO:
- [ ] use [fasjson client](https://fasjson-client.readthedocs.io/en/latest/usage.html) for the code of checking the FAS account
- [ ] implement the automatic check of Github/Gitlab username from the FAS account via fasjson API and check if is the same as the one who triggered the installation
- [ ] if the check is successful, store the FAS account in the allowlist in DB (check how it was done in the past in the git histoy) and don't create issue in notifications repo

lachmanfrantisek:
Relates to #997

csomh:
FAS2 seems to have stopped working for good.

PR #1460, opened 2022-04-21
Allowlisting improvements

lbarcziova:
- drop the FPCA requirement from the descriptions of issues created in the notifications repository
- since FAS2 stopped working and we want to use other library and also change the workflow for "auto"-allowlisting, drop the code using the retired library

Related to #1257

---

RELEASE NOTES BEGIN
N/A
RELEASE NOTES END


mfocko (commented):
lgtm

csomh:
Would this mean that there is no check if the requester has a Fedora ID? Is that okay to do?

lbarcziova:
> Would this mean that there is no check if the requester has a Fedora ID?

Yes, we wouldn't do any checks and directly create an issue in the notifications repo.

>  Is that okay to do?

Firstly, I think it already hasn't been working for some time since now we the library is retired and we can't log in to FAS to check it and when it was retired, it was readonly, so we got outdated info about users. Secondly, we discussed this check isn't robust and we want to implement the solution with checking the GH/GL username in FAS account via fasjson. Or am I missing something?

csomh:
> > Is that okay to do?
> 
> Firstly, I think it already hasn't been working for some time since now we the library is retired and we can't log in to FAS to check it and when it was retired, it was readonly, so we got outdated info about users. Secondly, we discussed this check isn't robust and we want to implement the solution with checking the GH/GL username in FAS account via fasjson. Or am I missing something?

Yes, all that is correct.

The effect of the old library not working was that we had to approve each request manually. Which also meant that we checked that the user had a Fedora account. If this change is deployed, installations would end up in the allow list without any verification (manual or automatic). Given that we sometimes receive some spam installations (someone playing with the GitHub API with malicious intent or just out of curiosity), I worry that we could receive a large number of spam workloads, which could hurt the experience of our users. This is why I thought, that until fasjson is adopted we will default to "allow manually". @lachmanfrantisek what do you think?

mfocko:
@csomh I had one more look to make sure and all installations should be added as `waiting`, unless it was previously installed and approved.

packit_service/worker/allowlist.py
```diff
@@ -74,66 +67,23 @@ def _strip_protocol_and_add_git(url: Optional[str]) -> Optional[str]:
...
-
-        logger.info(f"Cannot verify whether {account_login!r} signed FPCA.")
-        return False
-
-    def add_namespace(self, namespace: str, sender_login: Optional[str] = None) -> bool:
+    @staticmethod
+    def add_namespace(namespace: str) -> bool:
         """
         Add namespace to the allowlist. Namespace is set to `waiting` unless the
         `sender_login` matches packager account in FAS.
```
mfocko, line 74: No `sender_login` is passed :) I must've missedthat

lbarcziova: Because it was used only for checking the Fedora account and added to DB only if there was an automatic match.

mfocko: I meant that it's still mentioned in the comment.

lbarcziova:
> If this change is deployed, installations would end up in the allow list without any verification (manual or automatic). 

As Mato wrote, the installations will be added as `waiting`, so from the behaviour perspective, nothing will change.

packit_service/worker/allowlist.py
lbarcziova: aah, sorry :D will remove

csomh:
> > If this change is deployed, installations would end up in the allow list without any verification (manual or automatic).
> 
> As Mato wrote, the installations will be added as `waiting`, so from the behaviour perspective, nothing will change.

Oh, I missed that line. I didn't realize `add_namespace()`  sets also the status. Everything is fine, then.

csomh (approved):

csomh:
Please also update the message in https://github.com/packit/notifications/blob/main/.github/stale.yml once this is merged, and maybe mention that matching manually is a temporary solution until we automate things.

lbarcziova:
> Please also update the message in https://github.com/packit/notifications/blob/main/.github/stale.yml once this is merged, and maybe mention that matching manually is a temporary solution until we automate things.

Sure, it is already [opened](https://github.com/packit/notifications/pull/422), I will add the note mentioning it.

PR #1477, opened 2022-04-28
Steve refactor

lbarcziova:
TODO:
- [ ] check that these changes don't affect code in other repositories

Related to #1439 

---

RELEASE NOTES BEGIN
N/A
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/jobs.py
+++ b/packit_service/worker/jobs.py
@@ -331,47 +331,57 @@ class SteveJobs:
         signatures = []
         # we want to run handlers for all possible jobs, not just the first one
         for job_config in job_configs:
-            if self.service_config.deployment not in job_config.packit_instances:
-                logger.debug(
-                    f"Current deployment ({self.service_config.deployment}) "
-                    f"does not match the job configuration ({job_config.packit_instances}). "
-                    "The job will not be run."
-                )
-                continue
-
-            handler = handler_kls(
-                package_config=self.event.package_config,
-                job_config=job_config,
-                event=self.event.get_dict(),
-            )
-            if not handler.pre_check():
-                continue
-
-            if self.event.actor and not handler.check_if_actor_can_run_job_and_report(
-                actor=self.event.actor
+            if self.should_task_be_created_for_job_config_and_handler(
+                job_config, handler_kls
             ):
-                # For external contributors, we need to be more careful when running jobs.
-                # This is a handler-specific permission check
-                # for a user who trigger the action on a PR.
-                # e.g. We don't allow using internal TF for external contributors.
-                continue
-
-            self.report_task_accepted(handler=handler, job_config=job_config)
-
-            signatures.append(
-                handler_kls.get_signature(event=self.event, job=job_config)
-            )
-            processing_results.append(
-                TaskResults.create_from(
-                    success=True,
-                    msg="Job created.",
-                    job_config=job_config,
-                    event=self.event,
+                signatures.append(
+                    handler_kls.get_signature(event=self.event, job=job_config)
                 )
-            )
+                processing_results.append(
+                    TaskResults.create_from(
+                        success=True,
+                        msg="Job created.",
+                        job_config=job_config,
+                        event=self.event,
+                    )
+                )
         # https://docs.celeryproject.org/en/stable/userguide/canvas.html#groups
         group(signatures).apply_async()
         return processing_results
+
+    def should_task_be_created_for_job_config_and_handler(
+        self, job_config: JobConfig, handler_kls: Type[JobHandler]
+    ):
+        """
+        Check whether a new task should be created for job config and handler.
+        """
+        if self.service_config.deployment not in job_config.packit_instances:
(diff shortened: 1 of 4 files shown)


lachmanfrantisek (commented):
I like that approach. With that, we can refactor the code more easily, so `+1` from me on continuing this way.

Thanks!

packit_service/worker/jobs.py
```diff
@@ -420,109 +470,100 @@ def process_jobs(self, event: Event) -> List[TaskResults]:
...
+            )
 
         return processing_results
 
-    def process_message(
-        self, event: dict, topic: str = None, source: str = None
+    def create_tasks(
+        self, event: Event, job_configs: List[JobConfig], handler_kls: Type[JobHandler]
     ) -> List[TaskResults]:
         """
```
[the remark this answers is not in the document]
lbarcziova: I decomposed the create_task, PTAL if you think it should be improved even more.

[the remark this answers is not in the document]
lachmanfrantisek: @lbarcziova Everything looks really good, but I am still inclined to have `create_task` instead of `create_tasks`.

Otherwise, it's really a huge improvement and the code is now much easier to read.

packit_service/worker/jobs.py
```diff
@@ -212,97 +91,77 @@ def get_handlers_for_check_rerun(check_name_job: str) -> Set[Type[JobHandler]]:
...
+        if not (event_object and event_object.pre_check()):
+            return []
 
-    if isinstance(handler, CoprBuildHandler) and number_of_build_targets:
-        for _ in range(number_of_build_targets):
-            pushgateway.copr_builds_queued.inc()
+        return SteveJobs(event_object).process()
 
-    pushgateway.push()
+    def process(self) -> List[TaskResults]:
```
nikromen, line 128: For me it's little bit unclear what `process` means. I get that it's something like `main` method for Steve so maybe... but I won't recommend changing namings since I am not good at this :D

packit_service/worker/jobs.py
```diff
@@ -343,12 +202,12 @@ def report_task_accepted(
...
-                db_trigger=event.db_trigger,
+                package_config=self.event.package_config,
+                project=self.event.project,
+                metadata=EventData.from_event_dict(self.event.get_dict()),
+                db_trigger=self.event.db_trigger,
                 job_config=job_config,
-                branches_override=event.branches_override,
+                branches_override=self.event.branches_override,
             )
             task_accepted_time = datetime.now(timezone.utc)
```
nikromen, line 212: I would move this `task_accepted_time` before if-statement since it's present in if and elif block so we can use it there only once.

packit_service/worker/jobs.py
```diff
@@ -212,97 +91,77 @@ def get_handlers_for_check_rerun(check_name_job: str) -> Set[Type[JobHandler]]:
...
-    def service_config(self):
-        if self._service_config is None:
-            self._service_config = ServiceConfig.get_service_config()
-        return self._service_config
+        return processing_results
 
-    def report_task_accepted(
-        self, event: Event, handler: JobHandler, job_config: JobConfig
-    ):
+    def report_task_accepted(self, handler: JobHandler, job_config: JobConfig):
```
nikromen, line 156: This method is still pretty long. Can we at least separate creation of `job_helper` or something like this?
I was thinking about something like this in separate method
```py
...

params = {
    service_config=self.service_config,
    package_config=self.event.package_config,
    project=self.event.project,
    metadata=EventData.from_event_dict(self.event.get_dict()),
    db_trigger=self.event.db_trigger,
    job_config=job_config,
}

if propose_downstream:
    other_params = {branches_override=self.event.branches_override}
else:
    other_params = {...}

return helper_cls(params | other_params)
```

nikromen (commented):
It's a really huge improvement from the previous version! Thank you!

packit_service/worker/jobs.py
lbarcziova: Good point! I split it, PTAL

packit_service/worker/jobs.py
lbarcziova: I also wasn't able to think about something better :D I am opened to suggestions

csomh: Isn't this name confusing, b/c it's not clear _what_ is processed (ie. there are no arguments, besides `self`)? If so, that becomes obvious when looking at how it's used, I think: `SteveJobs(event_object).process()`.

How about adding a docstring here, so if someone reads only this method, they can immediately have some context of what is processed, without looking up the usage?

nikromen (approved):
Thanks!

Since you decomposed Steve really well I am thinking about some _really_ basic unit test for these functions/methods (otherwise if someone will break some of these methods it will be hard to find where the problem is :D). 
So maybe it wold be nice to create card/issue about this since you spent a lot of time on this (and we want our cards to be well splitted beacuse we do Kanban :grinning:) WDYT?

PR #1479, opened 2022-04-28
Filter most recent targets

nikromen:
for `/packit rebuild-retest-failed` we want to search for most recent copr/tf models in DB. Currently we search only for models which matches commit sha.
But what if some builds fails and after `/packit rebuild-failed` all builds succeeds? 
- we still have that failed model connected to commit sha so after another `rebuild-failed` we rebuild the previously failed build even if the latest build for corresponding target passed.

aditional functionality of filtering the most recent targets from iterable of models should solve this issue

Related to #1375 and maybe finally closes it? :D

---

RELEASE NOTES BEGIN
￼
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/events/event.py
+++ b/packit_service/worker/events/event.py
@@ -507,8 +507,9 @@ class AbstractForgeIndependentEvent(Event):
 
         return list(most_recent_models.values())
 
-    @staticmethod
+    @classmethod
     def _filter_most_recent_models_targets_by_status(
+        cls,
         models: Union[
             Optional[Iterable[CoprBuildTargetModel]],
             Optional[Iterable[TFTTestRunTargetModel]],
@@ -519,7 +520,7 @@ class AbstractForgeIndependentEvent(Event):
             f"Trying to filter targets with possible status: {statuses_to_filter_with} in {models}"
         )
         failed_models_targets = set()
-        for model in models:
+        for model in cls.get_most_recent_targets(models):
             if model.status in statuses_to_filter_with:
                 failed_models_targets.add(model.target)
 
--- a/packit_service/worker/events/event.py
+++ b/packit_service/worker/events/event.py
@@ -465,7 +465,50 @@ class AbstractForgeIndependentEvent(Event):
         return package_config
 
     @staticmethod
-    def _filter_failed_models_targets(
+    def _get_submitted_time_from_model(
+        model: Union[CoprBuildTargetModel, TFTTestRunTargetModel]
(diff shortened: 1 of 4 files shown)


nikromen:
recheck

nikromen:
recheck

nikromen:
recheck

packit_service/worker/events/event.py
```diff
@@ -465,7 +465,47 @@ def get_package_config(self) -> Optional[PackageConfig]:
...
+        Args:
+            models: Copr ot TF models
+
+        Returns:
+            Dictionary - target as a key and corresponding most recent submission time as a value.
+        """
+        target_submission_time: Dict[str, datetime] = {}
+        for model in models:
+            submitted_time = cls._get_submitted_time_from_model(model)
+            if (
```
lbarcziova, line 498: it took me a while to get what this function really does :D would it work to have Dict[str, "TargetModel"] and then return directly the values of the dictionary and in the `_filter..` methods no filtering on time would be needed anymore? 
(in the check we would compare then `cls._get_submitted_time_from_model(the_dictionary.get(model.target))` with the current submitted time) Then the method name could stay how it is now

lbarcziova (commented):
> Related to https://github.com/packit/packit-service/issues/1375 and maybe finally closes it? :D

I believe that this will fix it, thanks for working on it!

packit_service/worker/events/event.py
```diff
@@ -465,7 +465,51 @@ def get_package_config(self) -> Optional[PackageConfig]:
...
+        for model in models:
+            submitted_time_of_current_model = cls._get_submitted_time_from_model(model)
+            if (
+                most_recent_models.get(model.target) is None
+                or cls._get_submitted_time_from_model(most_recent_models[model.target])
+                < submitted_time_of_current_model
+            ):
+                most_recent_models[model.target] = model
+
+        return most_recent_models
```
lbarcziova, line 508: is there a reason to not return only the most recent target models (values)? Is/will this method be reused somewhere?

lbarcziova (approved):

packit_service/worker/events/event.py
nikromen: I hope the method will be reused in the future for similar purposes.  But you are right, returning only values should be enough.

TomasTomecek (approved):
LGTM

let's try on stg

I wonder what would be the performance impact of these changes: on the other hand, network traffic definitely takes more time than our local cluster database operations.

PR #1483, opened 2022-05-02
Allowlist automatic check

### LINKED ISSUE #1257: Use fasjson for automatic allowlisting

lbarcziova:
I don't know how to make the testing nicer, since the [fasjson_client library](https://github.com/fedora-infra/fasjson-client) has  dynamic api methods so I can't mock the `get_user` directly, any suggestions are welcome :D (I tested the get_user locally and was following [their docs](https://github.com/fedora-infra/fasjson-client/blob/dev/docs/migration.rst#person_by_username))

Fixes #1257 

---

RELEASE NOTES BEGIN
TODO
RELEASE NOTES END

### DIFF
--- a/files/install-deps-worker.yaml
+++ b/files/install-deps-worker.yaml
@@ -39,6 +39,7 @@
           - fedpkg-stage
           - postgresql # pg_dump
           - python3-boto3 # AWS (S3)
+          - python3-fasjson-client
         state: present
     - import_tasks: tasks/install-packit-deps.yaml
     - import_tasks: tasks/install-ogr-deps.yaml
--- a/files/install-deps-worker.yaml
+++ b/files/install-deps-worker.yaml
@@ -16,7 +16,6 @@
           # changes b/w major versions and we want to pin to the specific
           # version to be sure it works well - it's 12.0.1 for now
           # - python3-kubernetes  # for sandcastle
-          - python3-fedora # to access FAS
           - python3-requests
           - python3-alembic
           - python3-prometheus_client
@@ -36,7 +35,6 @@
           - dnf-plugins-core
           # oc rsync /tmp/sandcastle -> sandcastle pod
           - rsync
-          - fedpkg-stage
           - postgresql # pg_dump
           - python3-boto3 # AWS (S3)
           - python3-fasjson-client
--- a/packit_service/worker/allowlist.py
+++ b/packit_service/worker/allowlist.py
@@ -4,12 +4,15 @@
 import logging
 from typing import Any, Iterable, Optional, Union, Callable, List, Tuple, Dict
 
+from fasjson_client import Client
+from fasjson_client.errors import APIError
 from ogr.abstract import GitProject
 
+from packit.api import PackitAPI
 from packit.config.job_config import JobConfig
-from packit.exceptions import PackitException
+from packit.exceptions import PackitException, PackitCommandFailedError
 from packit_service.config import ServiceConfig
-from packit_service.constants import FAQ_URL
+from packit_service.constants import FAQ_URL, FASJSON_URL
 from packit_service.models import AllowlistModel, AllowlistStatus
 from packit_service.worker.events import (
     EventData,
@@ -51,6 +54,9 @@ UncheckedEvent = Union[
 
 
 class Allowlist:
+    def __init__(self, service_config: Optional[ServiceConfig] = None):
+        self.service_config = service_config
+
     @staticmethod
     def _strip_protocol_and_add_git(url: Optional[str]) -> Optional[str]:
         """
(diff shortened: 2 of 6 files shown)


TomasTomecek:
> I don't know how to make the testing nicer, since the [fasjson_client library](https://github.com/fedora-infra/fasjson-client) has dynamic api methods so I can't mock the `get_user` directly, any suggestions are welcome :D (I tested the get_user locally and was following [their docs](https://github.com/fedora-infra/fasjson-client/blob/dev/docs/migration.rst#person_by_username))

Would the daily validation job be a good candidate?

TomasTomecek (approved):
pretty straightforward

files/install-deps-worker.yaml
```diff
@@ -39,6 +39,7 @@
           - fedpkg-stage
           - postgresql # pg_dump
           - python3-boto3 # AWS (S3)
+          - python3-fasjson-client
```
jpopelka, line 42: Do we still need the `python3-fedora`? The comment (20 lines above) says "to access FAS".

tests/integration/test_installation.py
```diff
@@ -32,7 +32,7 @@ def test_installation():
 
     flexmock(GithubInstallationModel).should_receive("create").once()
     flexmock(Allowlist).should_receive("add_namespace").with_args(
-        "github.com/packit-service"
+        "github.com/packit-service", "jpopelka"
```
jpopelka, line 35: :open_mouth:

jpopelka (approved):

files/install-deps-worker.yaml
lbarcziova: you are right, I think we shouldn't need it anymore

jpopelka: OTOH, we've probably depended on it to pull in also `fedpkg` and `python3-bodhi-client` as I don't see those in the list. But it's not an excuse to have it there, because it pulls in other dependencies as well.

jpopelka:
The tests are expected to fail when there's a change in the list of dependencies because we use a pre-built test image built from the last commit in `main`.
For me, it's OK if you state that they pass for you locally.
Or, alternatively, if you need to see green here, you can push your locally built image to [packit-service-tests](https://quay.io/repository/packit/packit-service-tests) and restart the tests in Zuul.

files/install-deps-worker.yaml
lbarcziova: hmm, so you propose to replace `python3-fedora` with `fedpkg` and `python3-bodhi-client`? or do I get it wrong? :D

lbarcziova:
> The tests are expected to fail when there's a change in the list of dependencies because we use a pre-built test image built from the last commit in main.
For me, it's OK if you state that they pass for you locally.
Or, alternatively, if you need to see green here, you can push your locally built image to [packit-service-tests](https://quay.io/repository/packit/packit-service-tests) and restart the tests in Zuul.
EDIT: And merge right after it passes, so that you don't break tests for other PRs.

Ok, thanks for the clarification, I will play with it locally and see.

files/install-deps-worker.yaml
jpopelka: yes, exactly, and then `make build-test-image && make check-in-container` to see if we need something else.

lbarcziova: I just did `make build-test-image && make check-in-container` only with removing `python3-fedora` (and not adding the 2 packages) and it is weird that the tests passed even without those :thinking:

packit_service/worker/allowlist.py
```diff
@@ -67,22 +73,90 @@ def _strip_protocol_and_add_git(url: Optional[str]) -> Optional[str]:
...
+        Args:
+            fas_account: FAS account for which we will get the account info.
+            sender_login: Login of the user that will be checked for be match
+                            against info from FAS.
+
+        Returns:
+            Whether there was a match found.
+        """
+        if not self.init_kerberos_ticket():
+            return False
```
lachmanfrantisek, line 109: I would at least mention in docstring that failure means `False` as well. (For me, it probably does not fit the method name `Is GitHub username from FAS account matching?`.)


I understand why it is like this... Maybe, we can check the errors outside of the method? (Just thinking out loud...;)

lachmanfrantisek (approved):
Nice!

PR #1486, opened 2022-05-04
feat(tf): pass Copr project and RPMs

### LINKED ISSUE #1045: Enable users identify installed RPMs during testing in TF
We've just a request from the kdump maintainer so they can identify what are the exact RPMs that were installed in the test environment so they can manipulate them. There are two unsolved problems here:
1. Since the builds are being picked from COPR, users also need to know the copr project name
2. One SRPM can easily produce multiple binary RPMs - we especially need to track those for this use case

I'm wondering if the easiest solution here is to download the binary RPMs, place them in a pre-defined path and document this.

mfocko:
Fixes #1045

---

RELEASE NOTES BEGIN
Packit now passes `PACKIT_COPR_PROJECT` and `PACKIT_COPR_RPMS` variables to the Testing Farm. `PACKIT_COPR_PROJECT` holds Copr project in format `owner/project` and `PACKIT_COPR_RPMS` space-separated RPMs that were built in Copr.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/helpers/testing_farm.py
+++ b/packit_service/worker/helpers/testing_farm.py
@@ -219,6 +219,12 @@ class TestingFarmJobHelper(CoprBuildJobHelper):
             "PACKIT_SOURCE_URL": self.source_project_url,
             "PACKIT_TARGET_URL": self.target_project_url,
             "PACKIT_PR_ID": self.pr_id,
+            "PACKIT_COPR_PROJECT": f"{build.owner}/{build.project_name}"
+            if build
+            else None,
+            "PACKIT_COPR_RPMS": " ".join(artifact["packages"])
+            if artifact and artifact.get("packages")
+            else None,
         }
         predefined_environment = {
             k: v for k, v in predefined_environment.items() if v is not None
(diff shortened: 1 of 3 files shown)


packit_service/worker/helpers/testing_farm.py
```diff
@@ -219,6 +219,12 @@ def _payload(
             "PACKIT_SOURCE_URL": self.source_project_url,
             "PACKIT_TARGET_URL": self.target_project_url,
             "PACKIT_PR_ID": self.pr_id,
+            "PACKIT_COPR_PROJECT": f"{build.owner}/{build.project_name}"
+            if build
+            else None,
+            "PACKIT_COPR_RPMS": " ".join(artifact["packages"])
+            if artifact and artifact.get("packages")
+            else None,
```
lachmanfrantisek, line 227: What about just `PACKIT_RPMS`? Or, do we want tight that to Copr?

TomasTomecek: I guess this is about the future: if we fill the variable from other build systems (especially koji) - should we create a new variable or only use a single one?

I personally lean towards `PACKIT_RPMS` but it's true that `PACKIT_COPR_RPMS` is very explicit about the content.

Could be worth brainstorming on architecture. We can also set both :D

PR #1487, opened 2022-05-04
Packit verify fas

### LINKED ISSUE #1476: Implement `/packit verify-fas username` issue comment command
Followup of #1257.

TODO:
- [ ] react only for the comments on issues in the `packit/notifications` repository (make sure Packit Service is installed on this repository)
- [ ] implement getting the organization's name from the issue where `/packit verify-fas username` comes 
- [ ] implement a new handler for automatic check of the FAS account and storing it in the DB that will be triggered by this comment
- [ ] update the description of issues created in the packit/notifications and stale bot comments to include precise instructions on how to trigger the automatic check

lbarcziova:
TODO:
- [x] Update documentation in `packit/packit.dev` (packit/packit.dev#453).

Fixes #1476

---

RELEASE NOTES BEGIN
TODO
RELEASE NOTES END


packit_service/worker/handlers/forges.py
```diff
@@ -90,3 +96,119 @@ def run(self) -> TaskResults:
...
+
+        else:
+            msg = (
+                f"We were not able to find a match between GitHub Username field in FAS account"
+                f" `{fas_account}` and Github user `{self.sender_login}`. Please, check that you "
+                f"have set the field correctly and try again or contact "
+                f"[us](https://packit.dev/#contact)."
+            )
+            logger.debug(msg)
+            self.issue.comment(msg)
```
lachmanfrantisek, line 212: What about giving :+1: / :-1: to the original comment?  (Definitely not needed..;)

lachmanfrantisek:
Nice work, btw!

packit_service/worker/handlers/forges.py
lbarcziova: I wanted to be consistent with the behaviour for the other Packit comment commands, so implemented only :eyes: (https://github.com/packit/packit-service/blob/e5897986314acf955944829385898947070ca62f/packit_service/worker/jobs.py#L576). Although thinking about it now, I am not sure whether that won't be confusing for newcomers who have never used Packit before :D

packit_service/worker/handlers/forges.py
```diff
@@ -90,3 +96,119 @@ def run(self) -> TaskResults:
...
+        command_parts = get_packit_commands_from_comment(
+            self.comment, self.service_config.comment_command_prefix
+        )
+        # we expect ["verify-fas", "fas-account"]
+        if len(command_parts) != 2:
+            msg = "Incorrect format of the Packit verification comment command."
+            logger.debug(msg)
+            self.issue.comment(
+                f"{msg} The expected format: `/packit verify-fas my-fas-account`"
+            )
```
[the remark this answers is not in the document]
lbarcziova: this code should be executed only if there is a command /packit verify-fas in the comment, but either not followed by any "argument" or followed by more than one. Is that ok?

packit_service/worker/jobs.py
```diff
@@ -561,6 +567,18 @@ def process_message(
...
+        ) and self.is_fas_verification_comment(event_object.comment):
+            if GithubFasVerificationHandler(
+                package_config=None, job_config=None, event=event_object.get_dict()
+            ).pre_check():
+                event_object.comment_object.add_reaction(COMMENT_REACTION)
+                GithubFasVerificationHandler.get_signature(
+                    event=event_object, job=None
+                ).apply_async()
+            # should we comment about not processing if the comment is not
+            # on the issue created by us or not in packit/notifications?
```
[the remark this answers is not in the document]
lbarcziova: it is here because we can't process this as a "normal" jobs since there is no configuration with job which we can handle, so the behaviour is little bit like for GithubAppInstallationHandler above, but in that case if the event is an installation, we can directly tell to assign GithubAppInstallationHandler, in case of verification we need to check also the content of an issue comment, so it made sense to me to have it here and not try to initialise verification handler (which should be initialised only when we know that it is the handler we need for processing the event, since it ) if it can be just any other packit command comment/ any comment :thinking. Does it make sense?

packit_service/worker/handlers/forges.py
```diff
@@ -90,3 +96,119 @@ def run(self) -> TaskResults:
...
+        Returns:
+            TaskResults
+        """
+        logger.debug(
+            f"Going to run verification of FAS account triggered by comment:"
+            f" {self.comment}"
+        )
+        # e.g. User Bebaabeni needs to be approved.
+        account_type, account_login, _ = self.issue.title.split(maxsplit=2)
+        namespace = f"github.com/{account_login}"
```
TomasTomecek, line 169: nit: account_login can also be an organization, right?

packit_service/worker/handlers/forges.py
```diff
@@ -90,3 +96,119 @@ def run(self) -> TaskResults:
...
+        fas_account = command_parts[1]
+
+        allowlist = Allowlist(service_config=self.service_config)
+        if allowlist.is_approved(namespace):
+            msg = (
+                f"Invalid namespace! {account_type} `{account_login}` already approved."
+            )
+            logger.debug(msg)
+            self.issue.comment(msg)
+            return TaskResults(success=False, details={"msg": msg})
```
TomasTomecek, line 191: I am trying to understand how could this happen (someone did it manually on CLI?) and if it's actually a problem. Sounds like someone resolved the ask in the background :D and is not an error.

packit_service/worker/handlers/forges.py
lbarcziova: yes, therefore the `account_type` before which can be `organization` or `user`...or what am I missing? :D

packit_service/worker/handlers/forges.py
lbarcziova: yes, that's true, it is not very likely. So do you propose to not handle this situation at all or just referencing to change `success` to True?

TomasTomecek: A counterquestion for you: if you saw a sentry issue for this, would you investigate it?

I don't personally think we need a sentry issue here, the comment in the issue should be sufficient. Can we also make it more polite, please? Something like `Namespace {account_login} was already approved`.

packit_service/worker/handlers/forges.py
TomasTomecek: oh, right! I got confused about the naming of the variables, also 6pm on Fri - very little brain power :) thanks for explaining!

packit_service/worker/handlers/forges.py
lbarcziova: > A counterquestion for you: if you saw a sentry issue for this, would you investigate it?

Ok, totally forgot that these come to Sentry :D makes sense, will change it including the message, thanks!

TomasTomecek (approved):
:clap:

packit_service/worker/handlers/forges.py
csomh: I think that's okay. It means, that Packit noticed the command and will do something about it. It's also a good indication for us and can help with debugging issues.

csomh (approved):

PR #1493, opened 2022-05-08
Fix a mocked call of 'get_aliases()'

csomh:
When creating an expectation for a function call with the help of
flexmock, the module mocked should be the one in which the function is
imported and used, in order to be properly replaced.

In this particular example although 'get_aliases()' is _defined_ in
'packit.config.aliases', when used, it's imported in
'packit_service.worker.helpers.build.copr_build' with

    from packit.config.aliases import get_aliases

and so it becomes a name in 'copr_build'. This is why it's 'copr_build'
that needs to be mocked, instead of 'packit.config.aliases'.

### DIFF
--- a/tests/unit/test_copr_build.py
+++ b/tests/unit/test_copr_build.py
@@ -2122,9 +2122,9 @@ def test_run_copr_build_from_source_script(github_pr_event):
 
 
 def test_get_latest_fedora_stable_chroot(github_pr_event):
-    flexmock(packit.config.aliases).should_receive("get_aliases").and_return(
-        {"fedora-stable": ["fedora-34", "fedora-35"]}
-    )
+    flexmock(packit_service.worker.helpers.build.copr_build).should_receive(
+        "get_aliases"
+    ).and_return({"fedora-stable": ["fedora-34", "fedora-35"]})
     flexmock(packit_service.worker.helpers.build.copr_build).should_receive(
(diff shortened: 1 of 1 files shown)


TomasTomecek (approved):
The code change was fairly unclear to me but the explanation cleared it up pretty well.

The only sentence I'd improve would be:

> so now it's a name in that module, and so that module needs to be used
> when creating the expectation for its call.

I'd replace some pronouns to be completely explicit what they were referring to:

> so now it's a name in the `copr_build` module, and so that module needs to be used
> when creating the expectation for the `get_aliases` call.

If I got the references correctly :)

csomh:
> The code change was fairly unclear to me but the explanation cleared it up pretty well.
> 
> The only sentence I'd improve would be:
> 
> > so now it's a name in that module, and so that module needs to be used
> > when creating the expectation for its call.
> 
> I'd replace some pronouns to be completely explicit what they were referring to:
> 
> > so now it's a name in the `copr_build` module, and so that module needs to be used
> > when creating the expectation for the `get_aliases` call.
> 
> If I got the references correctly :)

Thanks for the suggestion! Improved it accordingly.

TomasTomecek (approved):
Very nice! Explicitly clear now. :clap:

jpopelka:
:+1: Thanks, I was looking at that on Friday but couldn't spot the cause.

Issue #1490, opened 2022-05-06
Do not run Koji build for proven packagers


lachmanfrantisek:
Based on a reaction of Miro Hrončok on fedora-devel to our announcement:

> Hey František.
>
> I read that "If Packit sees a new commit in the configured dist-git branch, it
submits a new build in Koji like maintainers usually do."
>
> Does that mean that if a provenpackager (e.g. me) commits to rpms/python-ogr in
order to rebuild the package in a side tag (e.g. f37-python for Python 3.11
rebuild), packit will build it in the regular target, preventing the side tag
build from happening (the same NVR cannot be built multiple times in Koji)?
>
> -- 
> Miro Hrončok

I  am thinking about multiple options here:

1) configurable allow-list / block list for committers (for the user and/or for the whole service)
* For our projects, we would allow just the `packit` user that we use for submitting the dist-git pull requests.
* We could probably react to just `packit` commits for all the users but I don't like this all-or-nothing approach. (That people are required to use all the Packit's features or none.)

2) check the committer for not being a proven packager (Packit is not a proven packager and we don't want to change that. We would like to have as few permissions as possible.)
* I don't see any technical problem with the implementation here and personally see this as a possible solution.

3) use a commit message to skip/trigger the build
* We can't inform everyone about this feature so skipping the build by commit command is probably not a good idea.
* And forcing the commit structure to trigger the build makes it hard to use.

We can probably think about this more, but the second option looks the best for me so far. Automation is not reduced and work for proven packagers is untouched. What do you think? Would this work for you? Or, do you have any better idea/solution?

TomasTomecek:
>   2. Automation only happens when a packit PR was merged*

I really like this one.

TomasTomecek:
Setting high-prio since this blocks the side-tag proven-packager workflow. Let's refine and plan it next week.

I propose to fix this in 2 stages:
1. Packit only reacts to changes done by itself to immediately unblock the workflow (should be an easy fix).
2. People can actually configure a list of FAS accounts when the automatic behaviour is triggered.

TomasTomecek:
The first part is now implemented, merged, deployed on stg and will be deployed to production on Tue.

We have also prioritized the long term task and hopefully someone will pick it up next week.

hroncok:
The first part means "Packit only reacts to changes done by itself to immediately unblock the workflow"?

PR #1498, opened 2022-05-10
submit koji builds and bodhi updates only for our own actions

TomasTomecek:
TODO:

- [x] CI green (testing it out now)
- [x] Drop the bodhi commit that is not needed
- [x] Update release notes
- [x] Update or write new documentation in `packit/packit.dev`. https://github.com/packit/packit.dev/pull/455

Packit should only submit production koji builds for its own dist-git commits.
This plays nicely with the provenpackager workflow: their changes are not being
acted upon so they can safely build in sidetags or kick off builds as they
need.

This PR implements the first solution so the provenpackager workflow is possible. We should then continue with an actual allowlist.

Related: https://github.com/packit/packit-service/issues/1490

---

RELEASE NOTES BEGIN
Packit now builds only its own dist-git commits. Other commits are not being acted upon. For more info see https://github.com/packit/packit-service/issues/1490
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/handlers/distgit.py
+++ b/packit_service/worker/handlers/distgit.py
@@ -421,6 +421,20 @@ class DownstreamKojiBuildHandler(JobHandler):
                     f"Koji build configured only for '{configured_branches}'."
                 )
                 return False
+
+            # Packit should only build its own contributions
+            # We need to preserve the proven packager workflow - pushes not done by Packit
+            # should be built by those that do it, e.g. rebuilds in a side tag.
+            # prod/stg filtering happens in jobs.py -> packit_instances
+            if not (
+                self.data.event_dict["email"] == "hello@packit.dev"
+                and self.data.event_dict["name"] == "Packit"
+            ):
+                logger.info(
+                    f"Push event {self.data.identifier} ({self.data.event_dict['name']} "
+                    f"<{self.data.event_dict['email']}>) not done by us."
+                )
+                return False
         return True
 
     def run(self) -> TaskResults:
--- a/packit_service/worker/events/pagure.py
+++ b/packit_service/worker/events/pagure.py
@@ -38,6 +38,8 @@ class PushPagureEvent(AddBranchPushDbTrigger, AbstractPagureEvent):
         git_ref: str,
         project_url: str,
         commit_sha: str,
+        name: str,
+        email: str,
     ):
         super().__init__(project_url=project_url)
         self.repo_namespace = repo_namespace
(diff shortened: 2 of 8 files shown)


packit_service/worker/handlers/bodhi.py
```diff
@@ -96,6 +102,12 @@ def pre_check(self) -> bool:
                 f"Bodhi update configured only for '{configured_branches}'."
             )
             return False
+        if self.koji_build_event.owner != self.expected_build_owner:
+            logger.debug(
+                f"Build {self.koji_build_event.build_id} submitted "
+                f"by {self.koji_build_event.owner}, skipping."
+            )
+            return False
```
lbarcziova, line 110: for Koji builds, we already do the filtering in the fedmsg [here](https://github.com/packit/packit-service-fedmsg/blob/3883a95f0f67ff56ffa3ad61f3a9822b9b7c3fcb/packit_service_fedmsg/consumer.py#L131)

packit_service/worker/handlers/koji.py
```diff
@@ -124,6 +124,14 @@ def pre_check(self) -> bool:
                     f"Push configured only for '{configured_branch}'."
                 )
                 return False
+        elif self.data.event_type == PushPagureEvent.__name__:
+            # Packit should only build its own contributions
+            if not self.data.email.endsiwth("packit.dev"):
```
lbarcziova, line 129: can we check here explicitly both name and full email?

lbarcziova (commented):
thanks for working on this, lgtm, just 2 notes

packit_service/worker/handlers/bodhi.py
TomasTomecek: Nice! so I can drop that commit.

lbarcziova (approved):

packit_service/worker/handlers/distgit.py
```diff
@@ -421,6 +421,20 @@ def pre_check(self) -> bool:
                     f"Koji build configured only for '{configured_branches}'."
                 )
                 return False
+
+            # Packit should only build its own contributions
+            # We need to preserve the proven packager workflow - pushes not done by Packit
+            # should be built by those that do it, e.g. rebuilds in a side tag.
+            # prod/stg filtering happens in jobs.py -> packit_instances
+            if not (
```
mfocko, line 429: What if maintainer pushes to the PR?

mfocko (approved):

packit_service/worker/handlers/distgit.py
TomasTomecek: I actually don't know how pagure/fedora-messaging do it - it would still be a PR coming from Packit so I presume the event would contain Packit's credentials and would be acted upon - this is my expectation.

The long-term solution should definitely be an allowlist of FAS accounts where builds should be done automatically.

TomasTomecek: (discussed in person)

TomasTomecek:
uhhh, I thought it would be merged already so I can play with it after meetings

nope

TomasTomecek:
regate

PR #1502, opened 2022-05-12
Verify fas improvements

lbarcziova:
I tested `/packit verify-fas` in packit/notifications#429, works just fine, only the message needs to be adjusted for staging.

Also, I added the check so that only the person who triggered the installation can be verified.

---

RELEASE NOTES BEGIN
N/A
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/handlers/forges.py
+++ b/packit_service/worker/handlers/forges.py
@@ -168,6 +168,10 @@ class GithubFasVerificationHandler(JobHandler):
         )
         # e.g. User Bebaabeni needs to be approved.
         _, account_login, _ = self.issue.title.split(maxsplit=2)
+        original_sender_login = GithubInstallationModel.get_by_account_login(
+            account_login
+        ).sender_login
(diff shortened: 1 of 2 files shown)


packit_service/worker/handlers/forges.py
```diff
@@ -181,6 +187,22 @@ def run(self) -> TaskResults:
...
+        if original_sender_login != self.sender_login:
+            msg = (
+                "Packit verification comment command not created by the person who "
+                "installed the application."
+            )
+            logger.debug(msg)
+            self.issue.comment(msg)
+            return TaskResults(success=True, details={"msg": msg})
+
+        return self.verify(namespace=namespace, fas_account=fas_account)
```
TomasTomecek, line 199: agreed that the use case for doing the verification for someone else sounds iffy

TomasTomecek (approved):
+1 No need to update the release notes.

lbarcziova:
recheck

lbarcziova:
regate

PR #1508, opened 2022-05-18
Provide links to the FAS settings to set the GitHub account

lachmanfrantisek:
There are various Fedora pages so it might be good to suggest the right one to the user
when doing the verification.

---

RELEASE NOTES BEGIN
When doing an automatic FAS account verification, Packit now suggests the URL where the GitHub account needs to be set.
RELEASE NOTES END


lbarcziova (approved):
thanks!

PR #1509, opened 2022-05-19
Koji allowlist

### LINKED ISSUE #1490: Do not run Koji build for proven packagers

lbarcziova:
TODO:

- [x] Write more tests.
- [ ] enable the new topic in fedmsg configuration

Fixes #1490 

For the merged PR pagure event, the db trigger model is currently GitBranchModel, since this is easier from the jobs mapping side for downstream koji build handler, although if we wanted to utilize this event for some other jobs, it could be a problem. Let me know what you think, should we use PRModel?

---

RELEASE NOTES BEGIN
TODO
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/handlers/distgit.py
+++ b/packit_service/worker/handlers/distgit.py
@@ -38,6 +38,7 @@ from packit_service.models import (
 )
 from packit_service.service.urls import get_propose_downstream_info_url
 from packit_service.utils import gather_packit_logs_to_buffer, collect_packit_logs
+from packit_service.worker.events.pagure import PullRequestMergedPagureEvent
 from packit_service.worker.helpers.propose_downstream import ProposeDownstreamJobHelper
 from packit_service.worker.events import (
     PushPagureEvent,
@@ -384,12 +385,12 @@ class ProposeDownstreamHandler(JobHandler):
 
 @configured_as(job_type=JobType.koji_build)
 @reacts_to(event=PushPagureEvent)
+@reacts_to(event=PullRequestMergedPagureEvent)
 class DownstreamKojiBuildHandler(JobHandler):
     """
     This handler can submit a build in Koji from a dist-git.
     """
 
-    topic = "org.fedoraproject.prod.git.receive"
     task_name = TaskName.downstream_koji_build
 
     def __init__(
@@ -408,8 +409,11 @@ class DownstreamKojiBuildHandler(JobHandler):
         self.dg_branch = event.get("git_ref")
 
     def pre_check(self) -> bool:
-        if self.data.event_type in (PushPagureEvent.__name__,):
-            if self.data.git_ref not in (
+        if self.data.event_type in (
+            PushPagureEvent.__name__,
+            PullRequestMergedPagureEvent.__name__,
+        ):
+            if self.dg_branch not in (
                 configured_branches := get_branches(
                     *self.job_config.dist_git_branches,
                     default="main",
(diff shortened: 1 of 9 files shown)


lbarcziova:
recheck

TomasTomecek (approved):
This is what we agreed on.

I wonder if we could set some commit statuses in src.fp.o and state that a build is not initiated b/c FAS is not on the list.

packit_service/worker/handlers/distgit.py
```diff
@@ -408,33 +409,58 @@ def __init__(
...
+                and self.data.event_dict["pr_author"]
+                not in self.job_config.allowed_pr_authors
+            ):
+                logger.info(
+                    f"Pull request merged event {self.data.identifier} for PR created by"
+                    f" {self.data.event_dict['pr_author']} that is not allowed in project "
+                    f"configuration: {self.job_config.allowed_pr_authors} and merged by"
+                    f" {self.data.event_dict['committer']} that is also not allowed in "
+                    f"project configuration: {self.job_config.allowed_committers}."
+                )
```
lachmanfrantisek, line 461: Can we add something to make it clear that one of these conditions needs to be fulfilled?
Something like this?

```suggestion
                logger.info(
                    f"Pull request merged event {self.data.identifier} for PR created by"
                    f" {self.data.event_dict['pr_author']} that is not allowed in project "
                    f"configuration: {self.job_config.allowed_pr_authors} and merged by"
                    f" {self.data.event_dict['committer']} that is also not allowed in "
                    f"project configuration: {self.job_config.allowed_committers}. "
                    "One of these conditions needs to be fulfilled."
                )
```

lachmanfrantisek (approved):
Nice and clear! Well done!

lbarcziova:
We discussed with @lachmanfrantisek that there is also a condition that the downstream pushes are being reacted to only if there are specfile changes and with reacting to the pull request closed message we can't get this information directly from the message. As we would probably need to utilize API, which was why we didn't implement the initial solution with getting PR info via API, I think we can use the initial solution. I opened a PR for it: #1520 (mostly the same code except of the PR check)

PR #1513, opened 2022-05-19
Fix default Copr project name for releases

### LINKED ISSUE #1503: Incorrect constructing of default Copr project name from Copr event for release trigger
If there is a Copr build triggered by a release that uses our default Copr project naming (for releases `packit/namespace-repo-releases`), when we parse the Copr build event (start/end of a build) and check whether we want to process the message, it looks like we construct the default project name we check incorrectly with the tag name suffix resulting in `packit/namespace-repo-tag`).

This occurred in fedora-infra/fedora-messaging [here](https://github.com/fedora-infra/fedora-messaging/runs/6408433212) and resulted in the status hanging in a pending state. The message in logs:
`The Copr project packit/fedora-infra-fedora-messaging-releases does not match the configuration (packit/fedora-infra-fedora-messaging-3.0.1 expected).`

The construction of the Copr project name happens [here](https://github.com/packit/packit-service/blob/54b1d121cdb6a1498a62da2d4a92cac2880fbd77/packit_service/worker/helpers/build/copr_build.py#L106).

lbarcziova:
When processing the message from copr, we didn't have the `tag_name` in the `metadata`, therefore the `else` branch was executed. Checking directly the db trigger should be more straigtforward.

Fixes #1503

---

RELEASE NOTES BEGIN
The construction of the default Copr project name for releases which caused incorrect handling of the release Copr builds was fixed.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/helpers/build/copr_build.py
+++ b/packit_service/worker/helpers/build/copr_build.py
@@ -34,6 +34,7 @@ from packit_service.models import (
     AbstractTriggerDbType,
     CoprBuildTargetModel,
     SRPMBuildModel,
+    JobTriggerModelType,
 )
 from packit_service.service.urls import (
     get_copr_build_info_url,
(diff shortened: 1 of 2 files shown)


lachmanfrantisek (approved):
Thanks!

PR #1520, opened 2022-05-25
Implement DownstreamKojiBuildHandler pre-check

### LINKED ISSUE #1490: Do not run Koji build for proven packagers

lbarcziova:
Fixes #1490

---

RELEASE NOTES BEGIN
You can now override which dist-git commits we will create Koji builds for by specifying
`allowed_pr_authors` and/or `allowed_committers` in the job metadata (see [the documentation](https://packit.dev/docs/configuration/#koji_build)).
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/handlers/distgit.py
+++ b/packit_service/worker/handlers/distgit.py
@@ -12,7 +12,7 @@ import shutil
 from typing import Optional, Dict, List
 
 from celery import Task
-from ogr.abstract import PullRequest
+from ogr.abstract import PullRequest, PRStatus
 from packit.exceptions import PackitException
 
 from packit.api import PackitAPI
@@ -406,6 +406,23 @@ class DownstreamKojiBuildHandler(JobHandler):
         )
         self.task = task
         self.dg_branch = event.get("git_ref")
+        self._pull_request: Optional[PullRequest] = None
+
+    @property
+    def pull_request(self):
+        if not self._pull_request and self.data.event_dict["committer"] == "pagure":
+            prs = [
+                pr
+                for pr in self.project.get_pr_list(status=PRStatus.merged)
+                if pr.head_commit == self.data.commit_sha
+            ]
+            if prs:
+                self._pull_request = prs[0]
+        return self._pull_request
+
+    def get_pr_author(self):
+        """Get the login of the author of the PR (if there is any corresponding PR)."""
+        return self.pull_request.author if self.pull_request else None
 
     def pre_check(self) -> bool:
         if self.data.event_type in (PushPagureEvent.__name__,):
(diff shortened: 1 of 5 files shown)


packit_service/worker/handlers/distgit.py
```diff
@@ -422,19 +443,35 @@ def pre_check(self) -> bool:
...
-                and self.data.event_dict["name"] == "Packit"
-            ):
-                logger.info(
-                    f"Push event {self.data.identifier} ({self.data.event_dict['name']} "
-                    f"<{self.data.event_dict['email']}>) not done by us."
-                )
-                return False
+        if self.data.event_type == PushPagureEvent.__name__:
+            if self.data.event_dict["committer"] == "pagure":
+                closed_by = self.get_pr_closed_by()
```
lachmanfrantisek, line 448: When I am looking at this now, I am not sure if this can't cause some unexpected builds. (Actually, it's not anyhow committed by that user..;)

@TomasTomecek @lbarcziova Do you think we can leave it like this or introduce the third option to make it really explicit?

lachmanfrantisek (commented):
Nicely done. I just have one question. (Sorry for bringing this up so late.)

packit_service/worker/handlers/distgit.py
lbarcziova: I don't really have a strong opinion, do you have any example of a use-case where it would be needed? :thinking: so that we don't have 3 configuration options if it is not really needed.

lachmanfrantisek: The question is if we want to allow configuration for those who performed the merge at all. What about removing this check? So, in the case of PRs, we will check just an author, for pushes, will check committers.

I am really not sure. I am just worried that it might be confusing to have a config option for `committers` and use it for something else....

lbarcziova: Ok, I understand now. So I can remove the check for the person who merged the PR (and update the docs accordingly) and in case of someone in the future requests it, it shouldn't be a problem to implement the third option.

lachmanfrantisek (approved):

mfocko (approved):

packit_service/worker/handlers/distgit.py
```diff
@@ -406,6 +406,23 @@ def __init__(
...
+    @property
+    def pull_request(self):
+        if not self._pull_request and self.data.event_dict["committer"] == "pagure":
+            prs = [
+                pr
+                for pr in self.project.get_pr_list(status=PRStatus.merged)
+                if pr.head_commit == self.data.commit_sha
+            ]
+            if prs:
+                self._pull_request = prs[0]
```
TomasTomecek, line 420: should there be an explicit `else:` path here to indicate that the commit is not coming from a PR?

TomasTomecek (commented):
I'm also wondering if pagure's commits that come from merged PRs are always fast-forward or if there actually can be a merge commit.

packit_service/worker/handlers/distgit.py
TomasTomecek: update: discussed this over chat; status: resolved :)

PR #1525, opened 2022-05-26
Don't add duplicated commit comments

jpopelka:
The commit comments are added only as a fallback when adding a commit status fails
(because of permissions or forge issues).
It can happen that we want the set a status that a commit already has.
With commit statuses it's not a problem, but in the fallback mode
when we add comments instead, it results in duplicated comments.

Fixes https://github.com/packit/hardly/issues/29
Merge after https://github.com/packit/ogr/pull/717

### DIFF
--- a/packit_service/worker/reporting.py
+++ b/packit_service/worker/reporting.py
@@ -183,6 +183,15 @@ class StatusReporter:
             BaseCommitStatus.failure,
         }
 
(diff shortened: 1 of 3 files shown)


lachmanfrantisek (approved):

mfocko (approved):

jpopelka:
recheck

PR #1530, opened 2022-05-31
feat!: do not default to fedora-stable for Copr builds

### LINKED ISSUE #1499: Discover targets when building in users Copr repo
It's pretty cumbersome for users who are using their own Copr repositories to keep the list of chroots in sync with the target lists in the jobs.

This could be improved if the service could be configured to set the targets for jobs based on what chroots are enabled in the user provided Copr repo.

Something like:

```
jobs:
- job: copr_build
  trigger: release
  owner: "@cockpit"
  project: "cockpit-preview"
  targets: copr_chroots
```

Or going further: `targets` could become optional when a user provided Copr repo is used, with the default behaviour to get the list from the Copr repo.

(The example above uses the [`metadata`-less config variant](https://github.com/packit/packit/issues/1515), which is not yet released.)

mfocko:

TODO:

- [x] Fix status reporting

Fixes #1499

---

RELEASE NOTES BEGIN
_TODO_
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/helpers/build/build_helper.py
+++ b/packit_service/worker/helpers/build/build_helper.py
@@ -84,6 +84,14 @@ class BaseBuildJobHelper(BaseJobHelper):
         if self.job_tests and not self.job_tests.skip_build:
             targets.update(self.job_tests.targets)
 
+        if (
+            self.job_type_build == JobType.copr_build
+            and (self.job_build and not self.job_build.targets)
+            and self.is_custom_copr_project_defined()  # type: ignore
+        ):
+            copr_targets = self.get_configured_targets()  # type: ignore
+            targets.update(copr_targets)
+
         return targets or {DEFAULT_VERSION}
 
     @property
--- a/packit_service/worker/helpers/build/copr_build.py
+++ b/packit_service/worker/helpers/build/copr_build.py
@@ -853,3 +853,13 @@ class CoprBuildJobHelper(BaseBuildJobHelper):
             raise ex
 
         return owner
+
+    def get_configured_targets(self) -> Set[str]:
+        """
+        Get configured targets of the custom Copr project.
+
+        Returns:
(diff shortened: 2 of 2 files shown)


packit_service/worker/helpers/build/koji_build.py
```diff
@@ -67,20 +71,28 @@ def __init__(
     def is_scratch(self) -> bool:
         return self.job_build and self.job_build.scratch
 
+    @staticmethod
+    def configured_or_default(targets: Set[str]) -> Set[str]:
+        return targets or {DEFAULT_VERSION}
```
lachmanfrantisek, line 76:
```suggestion
    @staticmethod
    def configured_targets_or_default(targets: Set[str]) -> Set[str]:
        return targets or {DEFAULT_VERSION}
```

lachmanfrantisek (approved):

packit_service/worker/helpers/build/koji_build.py
csomh: Is this method really needed? Using `*(self.configured_{build,tests}_targets or {DEFAULT_VERSION})` bellow would result in less code.

mfocko: I'm going to look into it… Probably going to prepare some PRs to `hello-world` for Friday, to put it on stage and test it out, cause with the tests the targets are non-trivial to think about :D

Or I could just fetch the Copr targets and do not care about the fact that Copr can do it by itself :thinking: (sounds easier tbf)

csomh: Sorry, I don't get it :) This method takes an argument and returns it if it's true-ish or returns a default otherwise. My question was whether this simple expression really needs to be wrapped in a method, instead of just using it straight away where it's needed.

PR #1531, opened 2022-06-01
Tee worker/beat output to /logs/*.log

jpopelka:
Needed for packit/deployment#347

### DIFF
--- a/files/run_worker.sh
+++ b/files/run_worker.sh
@@ -20,6 +20,9 @@ DEFAULT_CELERY_COMMAND="worker"
 # Whether to run Celery worker or beat (task scheduler)
 CELERY_COMMAND="${CELERY_COMMAND:-$DEFAULT_CELERY_COMMAND}"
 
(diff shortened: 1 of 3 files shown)


files/run_worker.sh
```diff
@@ -20,6 +20,9 @@ DEFAULT_CELERY_COMMAND="worker"
 # Whether to run Celery worker or beat (task scheduler)
 CELERY_COMMAND="${CELERY_COMMAND:-$DEFAULT_CELERY_COMMAND}"
 
+# https://stackoverflow.com/questions/49509264/explain-the-bash-command-exec-tee-log-file-21
+exec > >(tee -a "/logs/${HOSTNAME}.log") 2>&1
```
TomasTomecek, line 24: I hoped that we'd do this in celery instead of bash sorcery. Although CKI is doing something similar.

jpopelka: Celery can log either to stdout or to a logfile, not both.

TomasTomecek: @_@

PR #1539, opened 2022-06-07
Implement reporting when no FMF metadata found and skip_build option …

### LINKED ISSUE #1501: Fix reporting when no metadata for TF found when `skip_build=True`
If a user configures tests with `skip_build` option and then we can' find any FMF metadata, the user is not informed about this and the checks stay stuck in the `pending` state with the message `Submitting the tests...`, because in this [if-else code](https://github.com/packit/packit-service/blob/d228409d5324c186fa8b8aa3120939bc856c5767/packit_service/worker/helpers/testing_farm.py#L439) we check for the presence of metadata and then end without reporting if there is no metadata found.

lbarcziova:
Fixes #1501

---

RELEASE NOTES BEGIN
We improved the reporting for the test job if we are not able to find any FMF metadata the `skip_build` option is enabled.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/helpers/testing_farm.py
+++ b/packit_service/worker/helpers/testing_farm.py
@@ -452,9 +452,13 @@ class TestingFarmJobHelper(CoprBuildJobHelper):
         elif not self.is_fmf_configured() and not self.skip_build:
             payload = self._payload_install_test(int(build.build_id), target)
         else:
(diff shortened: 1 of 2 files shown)


mfocko (approved):
umm, that's it? :) ( apart from the 90% of tests /o\ )

PR #1553, opened 2022-06-17
Avoid duplicating ProposeDownstreamTargetModel when retrying the task

lbarcziova:
When retrying the propose downstream task, we pass the ID of ProposeDownstreamModel so that
it is created only once, but for each retry we created new ProposeDownstreamTargetModel instances
which in combination with zipping with branches created incorrect records in the database (incorrect statuses,
mismatch between the model branch and the actual branch ([example](https://dashboard.packit.dev/results/propose-downstream/683)), running propose downstream for a branch that already
succeeded). With these changes, for each branch there is created ProposeDownstreamTargetModel only once, when creating the ProposeDownstreamModel.

Related to #1504

---

RELEASE NOTES BEGIN
We have fixed the incorrect displaying of propose downstream results in our dashboard.
RELEASE NOTES END


TomasTomecek (approved):
nice, this is really clever

nikromen (approved):
lgtm :rocket:

PR #1565, opened 2022-06-28
Correctly report results from Testing Farm

### LINKED ISSUE #1535: Fix reporting from multiple TF jobs
When having multiple TF jobs defined using `identifier` attribute, there is an issue with reporting back from the TF, since there is no knowledge on the service-side which job has finished.

One of the proposed solutions by @lachmanfrantisek is to pass the check-name (which can be easily obtained as of now) to the TF as an environment variable that could be retrieved from the payload received after the job is done.

FrNecas:
Previously, if there were multiple TF jobs differentiated by their
identifier, Steve would create a celery task for each of the jobs,
regardless of their identifier for the single results event. To overcome
this, save the identifier in the DB and obtain it when a result arrives.
Check for identifier match when creating celery tasks based on jobs.

TODO:

Fixes #1535 (or at least I hope, will need to be verified on stg, I have a simple reproducer https://github.com/FrNecas/packit-reproducer/pull/1 )

---

RELEASE NOTES BEGIN
Results from Testing Farm are now correctly reported when multiple jobs with different identifier are defined.
RELEASE NOTES END


packit_service/worker/jobs.py
```diff
@@ -473,10 +474,17 @@ def is_project_public_or_enabled_private(self) -> bool:
...
-        Get list of non-duplicated all jobs that matches with event's trigger.
+        Get list of non-duplicated all jobs that matches with event's trigger and identifier.
         """
         jobs_matching_trigger = []
         for job in self.event.package_config.jobs:
+            # FIXME: perhaps this should be done more "abstractly"?
+            #        same for CheckRerunEvent in get_handlers_for_event
+            if (
+                isinstance(self.event, TestingFarmResultsEvent)
+                and self.event.identifier != job.identifier
```
lbarcziova, line 485: would it make sense to run this in the `pre_check` method of the `TestingFarmResultsHandler`? otherwise, LGTM

TomasTomecek (approved):
👏🏻

packit_service/worker/jobs.py
FrNecas: So testing turned out to be more complicated than I expected due to where `pre_check` is called -- it is necessary to test `create_tasks` which uses a lot of utility functions (dumping package config, etc) + celery stuff. I tried to remove most of this through mocks and focus on the number of tasks created. Could someone please re-review the changes?

FrNecas:
recheck

lachmanfrantisek (approved):
:+1: Looks good to me!

FrNecas:
recheck

PR #1573, opened 2022-07-07
Go through all PRs when getting distgit PR author for koji builds

lbarcziova:
This should solve the problem when the fedmsg message about merged PR is created sooner than
the PR is acutally merged which may result in not being able to get the PR author from API
if we go only through the merged distgit PRs.

---

RELEASE NOTES BEGIN
We adjusted the way we check the author of the PR for PRs related to dist-git commits that trigger Koji build jobs. This should fix the race condition causing not creating Koji builds in some cases.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/handlers/distgit.py
+++ b/packit_service/worker/handlers/distgit.py
@@ -409,9 +409,13 @@ class DownstreamKojiBuildHandler(JobHandler):
     @property
     def pull_request(self):
         if not self._pull_request and self.data.event_dict["committer"] == "pagure":
+            logger.debug(
+                f"Getting pull request with head commit {self.data.commit_sha}"
+                f"for repo {self.project.namespace}/{self.project.repo}"
+            )
             prs = [
                 pr
-                for pr in self.project.get_pr_list(status=PRStatus.merged)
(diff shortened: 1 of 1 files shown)


packit_service/worker/handlers/distgit.py
```diff
@@ -409,9 +409,13 @@ def __init__(
     @property
     def pull_request(self):
         if not self._pull_request and self.data.event_dict["committer"] == "pagure":
+            logger.debug(
```
majamassarini, line 412:
```suggestion
            logger.info(
```

majamassarini: I think we can not see debug messages in production and this seems not to be happening in stage. For this reason I will put it as an info message.

majamassarini (approved):

majamassarini:
Should a release note be added in the hope that the problem is resolved?

packit_service/worker/handlers/distgit.py
lbarcziova: We should see the debug messages both in prod and staging.

lachmanfrantisek (approved):

PR #1575, opened 2022-07-12
Mention that FAS account can't be private for verification

lbarcziova:
---

RELEASE NOTES BEGIN
￼
RELEASE NOTES END


packit_service/worker/handlers/forges.py
```diff
@@ -83,7 +83,8 @@ def run(self) -> TaskResults:
                     " field in the settings of the FAS account (if you don't have it set already)"
                     f" and provide it in a comment in this issue as "
                     f"`{self.service_config.comment_command_prefix} verify-fas "
-                    "<my-fas-username>`. We automatically check for the match between the `GitHub"
+                    "<my-fas-username>` (please, also make sure your profile is not private). "
```
jpopelka, line 86: We already have one "please" in the sentence and the whole message is quite long so every saved word counts.

```suggestion
                    "<my-fas-username>` (make sure your profile is not private). "
```

jpopelka (approved):

packit_service/worker/allowlist.py
```diff
@@ -132,6 +132,7 @@ def is_github_username_from_fas_account_matching(self, fas_account, sender_login
             )
             return github_username == sender_login
 
+        logger.debug("github_username not set or account private")
```
TomasTomecek, line 135: I wonder if we should log the whole `user_info` here to help with debugging. Or at least `github_username`.

TomasTomecek (approved):
Forgot to submit the review in the morning 🤦‍♂

These changes would be really helpful to diagnose the issue from yesterday.

packit_service/worker/allowlist.py
lbarcziova: I was thinking that logging the whole `user_info` isn't really needed and would just take up too many log lines. `github_username` is logged above if it is not None, I just added this line so that we can be sure when it was None

TomasTomecek: oki, that's true

there is a `is_private` field in the payload:
```python
{'username': 'commonjavabot',
 'surname': 'Bot',
 'givenname': 'Commonjava',
 'human_name': 'Commonjava Bot',
 'emails': ['exd-guild-mpakt@redhat.com'],
 'ircnicks': None,
 'locale': 'en-US',
 'timezone': 'UTC',
 'gpgkeyids': None,
 'sshpubkeys': None,
 'certificates': None,
 'creation': '2017-06-16T16:39:39',
 'is_private': False,
 'locked': False,
 'github_username': 'commonjava-bot',
 'gitlab_username': None,
 'website': None,
 'pronouns': None,
 'rhbzemail': None,
 'uri': 'http://fasjson.fedoraproject.org/v1/users/commonjavabot/'}
```

I'm wondering if we can retrieve the userinfo if the account is private :D

so yeah, let's not log that, we can always check logs or retrieve ourselves

lbarcziova:
@TomasTomecek except for the changes we discussed, I implemented it in the way that if the namespace was once approved and then reinstalled, we won't create a new notifications issue and request new verification, WDYT?

packit_service/worker/handlers/forges.py
```diff
@@ -64,11 +64,22 @@ def run(self) -> TaskResults:
...
         allowlist = Allowlist(self.service_config)
-        if not allowlist.add_namespace(
-            f"github.com/{self.account_login}", self.sender_login
+        namespace = f"github.com/{self.account_login}"
+        # if the namespace was not on our allowlist (in any state) or
+        # the app was reinstalled and the namespace was not approved,
+        # create an issue in notifications repo
+        if not allowlist.add_namespace(namespace, self.sender_login) or (
+            is_reinstalled and not allowlist.is_approved(namespace)
         ):
```
TomasTomecek, line 83: I'm confused because I understand the `if` statement as "if the namespace is reinstalled and not approved, create a new issue", which makes perfect sense to me (alternatively we could recycle the old issue but could become messy)

PR #1579, opened 2022-07-13
Copr error retries

### LINKED ISSUE #1549: Retry job in case of Copr error/outage
* [ ] In case of the Copr error, put the job back into the queue.  Combine/leave that with/to the auto-retry mechanism we already have:
  https://github.com/packit/packit-service/blob/307974f7bba1f349d5b0f9d0c0d3eaa6e988d2b4/packit_service/worker/tasks.py#L60-L65
* [ ] behave nicely after X retries (you can get inspired by the `propose-downstream` source download retires)
* [ ] in case of permission error, let the user know and put the job back in the queue

*part of the https://github.com/packit/packit-service/issues/927 epic*

lbarcziova:
TODO:

- [x] agree on the interval (and create a constant for it) and types of exceptions

Fixes #1549 

---

RELEASE NOTES BEGIN
During creating Copr builds, on Copr errors, Packit will retry the task multiple times
in case there is a Copr outage.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/handlers/abstract.py
+++ b/packit_service/worker/handlers/abstract.py
@@ -7,6 +7,7 @@ This file defines generic job handler
 import enum
 import logging
 import shutil
+from celery import Task
 from collections import defaultdict
 from datetime import datetime
 from os import getenv
@@ -22,6 +23,7 @@ from packit.constants import DATETIME_FORMAT
 from packit.local_project import LocalProject
 
 from packit_service.config import ServiceConfig
+from packit_service.constants import DEFAULT_RETRY_LIMIT
 from packit_service.models import (
     AbstractTriggerDbType,
 )
@@ -207,6 +209,54 @@ def get_packit_commands_from_comment(
             return packit_command
 
     return []
+
+
+class CeleryTask:
+    """
+    Class wrapping the Celery task object with methods related to retrying.
+    """
+
+    def __init__(self, task: Task):
+        self.task = task
+
+    @property
+    def retries(self):
+        return self.task.request.retries
+
+    def is_last_try(self) -> bool:
+        """
+        Returns True if the current celery task is run for the last try.
+        More info about retries can be found here:
+        https://docs.celeryq.dev/en/latest/userguide/tasks.html#retrying
+        """
+        return self.retries >= self.get_retry_limit()
+
+    def get_retry_limit(self) -> int:
+        """
+        Returns the limit of the celery task retries.
+        (Packit uses this env.var. in HandlerTaskWithRetry base class
+        to set `max_retries` in `retry_kwargs`.)
+        """
+        return int(getenv("CELERY_RETRY_LIMIT", DEFAULT_RETRY_LIMIT))
+
+    def retry(self, ex: Exception, delay: Optional[int] = None) -> None:
+        """
+        Retries the celery task.
+        Argument `throw` is set to False to not retry
+        the task also because of the `autoretry_for` mechanism.
+
+        More info about retries can be found here:
+        https://docs.celeryq.dev/en/latest/userguide/tasks.html#retrying
+
+        Args:
+            ex: Exception which caused the retry (will be logged).
+            delay: Number of seconds the task will wait before being run again.
+        """
+        retries = self.retries
+        delay = delay if delay is not None else 60 * 2**retries
+        logger.info(f"Will retry for the {retries + 1}. time in {delay}s.")
+        kargs = self.task.request.kwargs.copy()
+        self.task.retry(exc=ex, countdown=delay, throw=False, args=(), kwargs=kargs)
 
 
 class TaskName(str, enum.Enum):
@@ -395,6 +445,23 @@ class JobHandler(Handler):
                 "event": event.get_dict(),
             },
         )
+
+    def run(self) -> TaskResults:
+        raise NotImplementedError("This should have been implemented.")
+
+
+class RetriableJobHandler(JobHandler):
+    def __init__(
+        self,
+        package_config: PackageConfig,
+        job_config: JobConfig,
+        event: dict,
+        celery_task: Optional[Task] = None,
+    ):
+        super().__init__(
+            package_config=package_config, job_config=job_config, event=event
+        )
+        self.celery_task = CeleryTask(celery_task) if celery_task else None
 
     def run(self) -> TaskResults:
         raise NotImplementedError("This should have been implemented.")
--- a/packit_service/worker/handlers/copr.py
+++ b/packit_service/worker/handlers/copr.py
@@ -5,7 +5,7 @@ import logging
 from datetime import datetime, timezone
 from typing import Optional
 
-from celery import signature
+from celery import signature, Task
 from ogr.services.github import GithubProject
 from ogr.services.gitlab import GitlabProject
 from packit.config import (
@@ -54,6 +54,7 @@ from packit_service.worker.handlers.abstract import (
     required_for,
     run_for_comment,
     run_for_check_rerun,
+    RetriableJobHandler,
 )
 from packit_service.worker.monitoring import measure_time
 from packit_service.worker.reporting import BaseCommitStatus
@@ -77,7 +78,7 @@ logger = logging.getLogger(__name__)
 @reacts_to(CheckRerunPullRequestEvent)
 @reacts_to(CheckRerunCommitEvent)
 @reacts_to(CheckRerunReleaseEvent)
-class CoprBuildHandler(JobHandler):
+class CoprBuildHandler(RetriableJobHandler):
     task_name = TaskName.copr_build
 
     def __init__(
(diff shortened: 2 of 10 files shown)


packit_service/worker/handlers/abstract.py
```diff
@@ -209,6 +211,49 @@ def get_packit_commands_from_comment(
     return []
 
 
+class CeleryTask:
```
FrNecas, line 214: My initial thought was to put the relevant methods into `JobHandler` but I like your solution (new class wrapping celery's `Task`) more :+1:

packit_service/worker/handlers/abstract.py
```diff
@@ -209,6 +211,49 @@ def get_packit_commands_from_comment(
...
+    def retries(self):
+        return self.task.request.retries
+
+    def is_last_try(self) -> bool:
+        """
+        Returns True if the current celery task is run for the last try.
+        More info about retries can be found here:
+        https://celeryproject.readthedocs.io/en/latest/userguide/tasks.html#retrying
+        """
+        return self.retries == self.get_retry_limit()
```
FrNecas, line 228: I'd update this to `self.retries >= self.get_retry_limit()` just to be a bit safer.

packit_service/worker/handlers/abstract.py
```diff
@@ -209,6 +211,49 @@ def get_packit_commands_from_comment(
...
+
+    @property
+    def retries(self):
+        return self.task.request.retries
+
+    def is_last_try(self) -> bool:
+        """
+        Returns True if the current celery task is run for the last try.
+        More info about retries can be found here:
+        https://celeryproject.readthedocs.io/en/latest/userguide/tasks.html#retrying
```
FrNecas, line 226: The links are no longer up to date. The correct link is:

https://docs.celeryq.dev/en/latest/userguide/tasks.html#retrying

Same below

packit_service/worker/handlers/abstract.py
```diff
@@ -209,6 +211,49 @@ def get_packit_commands_from_comment(
...
+    def retry(self, ex: Exception, delay: Optional[int] = None) -> None:
+        """
+        Retries the celery task.
+        Argument `throw` is set to False to not retry
+        the task also because of the `autoretry_for` mechanism.
+
+        More info about retries can be found here:
+        https://celeryproject.readthedocs.io/en/latest/userguide/tasks.html#retrying
+
+        :param ex: Exception needs to be specified.
```
FrNecas, line 247: Since refactoring is being done, could we make the docstring consistent with the rest of the code? Most methods I see use the `Args:` notation. Also a suggestion for rewording:

```
Args:
    ex: Exception which caused the retry (will be logged).
    delay: Number of seconds the task will wait before being run again.
```

packit_service/worker/handlers/copr.py
```diff
@@ -85,13 +86,14 @@ def __init__(
...
         event: dict,
+        celery_task: Optional[CeleryTask] = None,
     ):
         super().__init__(
             package_config=package_config,
             job_config=job_config,
             event=event,
         )
-
+        self.celery_task = celery_task
```
FrNecas, line 96: Since 2 handlers already contain this initialization, what about creating another class doing this? Something like

```python
class RetriableJobHandler(JobHandler):
    def __init__(
        self,
        package_config: PackageConfig,
        job_config: JobConfig,
        event: dict,
        celery_task: Optional[CeleryTask] = None,
    ):
    super().__init__(
            package_config=package_config,
            job_config=job_config,
            event=event,
       )
       self.celery_task = celery_task
```

(sorry for the formatting, github editor makes it hard to manipulate)

This way, it's more obvious which handlers actually retry

packit_service/worker/helpers/build/copr_build.py
```diff
@@ -569,6 +574,26 @@ def handle_build_submit_error(self, ex):
         """
         Handle errors when submitting Copr build.
         """
+        if (
+            # don't retry the task if user action needed
+            not (
+                MISSING_PERMISSIONS_TO_BUILD_IN_COPR in str(ex)
+                or NOT_ALLOWED_TO_BUILD_IN_COPR in str(ex)
+            )
+            and not self.celery_task.is_last_try()
```
FrNecas, line 583: Shouldn't we on the last retry notify the user through an issue if an issue repository is configured? Similarly to propose-downstream/bodhi.

In my implementation, I refactored `notify_user_about_failure` into `JobHandler` and made it a bit more general like so:

```python
    def notify_user_about_failure(self, title: str, body: str) -> None:
        """
        If user configures `issue_repository`,
        Packit will create there an issue with the details.
        If the issue already exists and is opened, comment will be added
        instead of creating a new issue.

        The issue will be a place where to re-trigger the job.

        Args:
            title: Title of the issue to create.
            body: Content to be sent to the user (comment or issue description,
                for issue description a contact footer is added).
        """
        if not self.job_config.issue_repository:
            logger.debug(
                "No issue repository configured. User will not be notified about the failure."
            )
            return

        logger.debug(
            f"Issue repository configured. We will create "
            f"a new issue in {self.job_config.issue_repository}"
            "or update the existing one."
        )
        issue_repo = self.service_config.get_project(
            url=self.job_config.issue_repository
        )
        PackageConfigGetter.create_issue_if_needed(
            project=issue_repo,
            title=title,
            message=body
                    + f"\n\n---\n\n*Get in [touch with us]({CONTACTS_URL}) if you need some help.*",
            comment_to_existing=body,
        )
```

FrNecas (commented):
Nice changes, a few suggestions for improvements

packit_service/worker/handlers/abstract.py
```diff
@@ -398,3 +448,20 @@ def get_signature(cls, event: Event, job: Optional[JobConfig]) -> Signature:
...
         raise NotImplementedError("This should have been implemented.")
+
+
+class RetriableJobHandler(JobHandler):
+    def __init__(
+        self,
+        package_config: PackageConfig,
+        job_config: JobConfig,
+        event: dict,
+        celery_task: Optional[CeleryTask] = None,
```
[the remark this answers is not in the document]
lbarcziova: `None` needs to be allowed, since we initialize the handlers already in the processing task to call some pre-check methods and there we don't want to pass the processing celery task since we retry only the actual handler task.

We discussed this with @FrNecas in chat, to resolve the None-correctness, a bigger refactoring needs to be done and Franta will create an issue with the requirements for that.

lachmanfrantisek:
> @lachmanfrantisek the issue says that if there is a permission error, the task should be put back in the queue as well, but since users are easily able to retrigger the build (as opposed to bodhi updates) after resolving the permissions issue, I implemented it in the way that for these exceptions the task will not be retried, WDYT? Also, I was thinking what should be the interval for retrying if there is a Copr error, now it is set to 1 minute.

Is the retrigger easy also for the commit or release trigger? (Is the `re-run` button reliable and easy to use?)

I would say if we can be clever, why not do it. Do you see any problem in the implementation or performance?

packit_service/worker/handlers/abstract.py
```diff
@@ -209,6 +211,54 @@ def get_packit_commands_from_comment(
     return []
 
 
+class CeleryTask:
+    """
+    Class wrapping the Celery task object with methods related to retrying.
+    """
```
lachmanfrantisek, line 217: Wow, thanks for that!

packit_service/worker/helpers/build/copr_build.py
```diff
@@ -569,6 +574,26 @@ def handle_build_submit_error(self, ex):
...
+        ):
+            interval = 1
```
lachmanfrantisek, line 585: What about some backoff? In case of an outage, increasing the interval might be useful to not overwhelm Copr and ourselves.

lachmanfrantisek (commented):
Looks really good! I would maybe go with the retries for permission issues as well 
(just with bigger intervals like with Bodhi).

lbarcziova:
> I would say if we can be clever, why not do it. Do you see any problem in the implementation or performance?

@lachmanfrantisek My point was more that via check or comment retries users can trigger the builds when they are done with resolving the permission issue, which e.g. may be sooner than our interval is set. From the implementation POV I don't see a problem.

In that case, can the interval for the permission issues where a user action is required 10 minutes (the same as for Bodhi) and for others the one minute with exponential backoff?

lbarcziova:
I was looking into the Copr exceptions and also our Sentry issues and implemented retrying for `CoprRequestException` with the message "Unable to connect" and `CoprTimeoutException` (and the mentioned permission-related issues). Let me know if you think the exceptions we retry on should be extended, but my point was not to retry for errors where it may not make sense, such as `CoprNoResultException`, `CoprNoConfigException`,..

packit_service/worker/helpers/build/copr_build.py
```diff
@@ -524,6 +527,7 @@ def submit_copr_build(self, script: Optional[str] = None) -> Tuple[int, str]:
                     f"({permissions_url})"
                     " and retrigger the build.",
```
lachmanfrantisek, line 528:
```suggestion
                    " and retrigger the build by a `/packit build` pull-request comment "
                    "or click on a `Re-run` button.",
```

lachmanfrantisek (approved):
Thanks!

lbarcziova:
recheck

PR #1584, opened 2022-07-15
Allow specifying concurrency (autoscale & pool)

jpopelka:
The default autoscale is 1,1 which is equal to `--concurrency=1`
The default pool is `prefork`, which also stays the same.

So nothing changes by default, but it allows us to experiment with concurrency as suggested in https://github.com/packit/research/pull/153

### DIFF
--- a/files/run_worker.sh
+++ b/files/run_worker.sh
@@ -30,14 +30,22 @@ elif [[ "${CELERY_COMMAND}" == "worker" ]]; then
     DEFAULT_QUEUES="short-running,long-running"
     QUEUES="${QUEUES:-$DEFAULT_QUEUES}"
 
+    # Min,max number of concurrent worker processes/threads executing tasks.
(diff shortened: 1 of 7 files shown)


files/run_worker.sh
```diff
@@ -22,22 +22,30 @@ CELERY_COMMAND="${CELERY_COMMAND:-$DEFAULT_CELERY_COMMAND}"
 
 if [[ "${CELERY_COMMAND}" == "beat" ]]; then
     # when using the database backend, celery beat must be running for the results to be expired.
-    # https://docs.celeryproject.org/en/stable/userguide/periodic-tasks.html#starting-the-scheduler
+    # https://docs.celeryq.dev/en/stable/userguide/periodic-tasks.html#starting-the-scheduler
```
TomasTomecek, line 25: 😐 curious what happened there

TomasTomecek (approved):

Issue #1583, opened 2022-07-14
Build SRPMs in Copr for newly onboarded projects


lbarcziova:
Since we want to fully move to building SRPMs in Copr (#1245), we want to enforce newly onboarded projects to build their SRPMs in Copr. This should work in a way that for projects with corresponding installations after a certain date we will always build the SRPMs in Copr and do not consider the `srpm_build_deps` configuration key that is currently the indicator.

TODO:
- [ ] reimplement [the condition](https://github.com/packit/packit-service/blob/abeed0ac23501594d3a9f05f906a5d8d602b9395/packit_service/worker/handlers/copr.py#L136) for building SRPMs in Copr by utilising the [GithubInstallationModel](https://github.com/packit/packit-service/blob/abeed0ac23501594d3a9f05f906a5d8d602b9395/packit_service/models.py#L1852) for getting information about the date of the installation of Packit application for the corresponding project (installation will be got by the account login corresponding to the namespace of the repository that is being processed)
- [ ] make sure the documentation leads to configuring the `srpm_build_deps` and doesn't promote building SRPMs in our sandbox anymore

PR #1585, opened 2022-07-18
Installation added repositories

lbarcziova:
When creating #1583, I was looking into the related code and noticed we handle also events when repositories are being added to the app installation, but we were not updating the `repositories` in the DB  (until my changes I made in #1575, which were meant mainly for reinstallation to have the new `sender_login` and `created_at` data and the `repositories` is only rewritten by the added repos which is not correct). Also, we don't react to events when repositories are being removed from the app installation.  We can discuss whether we want to react to both `added` and `removed` events and actually update the `repositories` in the DB properly, but with the current codebase, I think it doesn't make sense to react to these, but maybe somebody can correct me what was the reason for handling these events.

---

RELEASE NOTES BEGIN
￼
RELEASE NOTES END


packit_service/worker/handlers/forges.py
```diff
@@ -64,22 +64,24 @@ def run(self) -> TaskResults:
         match.
         :return: TaskResults
```
mfocko, line 65:
```suggestion

        Returns:
            Result of the run task.
```

mfocko (commented):
lgtm

TomasTomecek (approved):
I appreciate the detailed write-up in the PR description, could you please add that info into a commit description so we have it in git as well? Please don't be afraid to break down compound sentences into smaller ones - the opening one is really complex, I had to read it 3 times :D

LGTM changes - it feels like it would be a much bigger effort to have 100% clarity in which repos Packit is installed. That doesn't seem worth it since we actually care about repos that actively use Packit.

mfocko:
> I appreciate the detailed write-up in the PR description, could you please add that info into a commit description so we have it in git as well? Please don't be afraid to break down compound sentences into smaller ones - the opening one is really complex, I had to read it 3 times :D

should be included in the merge commit :P

TomasTomecek:
ok, if that's the case - the info will be in the merge commit by GitHub

lbarcziova:
@TomasTomecek but you are right that it is more clear to have it directly in the commit containing the changes, so I updated it there as well...sorry for the long sentence, that's my weakness :D

jpopelka (approved):

Issue #1541, opened 2022-06-09
Implement retriggering Bodhi updates via dist-git pull-request comments


lbarcziova:
- [ ] retriggering will be done via dedicated Packit command - e.g `/packit create-update`
- [ ] we need to listen to `org.fedoraproject.prod.pagure.pull-request.comment.added` messages in fedmsg ([example message](https://apps.fedoraproject.org/datagrepper/v2/id?id=85d284e9-fa80-4f7c-8db6-cc0210f40941&is_raw=true&size=extra-large))
- [ ] user who triggers the jobs should have write access to the dist-git repository
- [ ] get the last build for the given branch

TomasTomecek:
I spoke with @thozza yesterday and he said he's looking forward to this functionality :)

thozza:
Yes, this would be awesome for al ofl the Image Builder components 😍 
We are releasing them in upstream by scheduled automation and rely on Packit to handle most of the Fedora workflow.

lbarcziova:
This is in progress.

PR #1586, opened 2022-07-18
Retrigger downstream koji builds

### LINKED ISSUE #1540: Implement retriggering downstream Koji build via dist-git pull-request comments
- [ ] retriggering will be done via dedicated Packit command - e.g `/packit koji-build`
- [ ] as for the running the build, there are multiple options, discuss these with the team:
  1. we will always run the build for the latest commit in the given branch regardless of the PR head commit
  2. we will get the head commit of the PR and check it is the same as the latest commit of the given branch and run the build only if the commits equal

majamassarini:
Through a Pagure comment in a PR

Fixes #1540

Merge after packit/packit-service-fedmsg#70

---

RELEASE NOTES BEGIN
A downstream *koji-build* can now be re-triggered commenting in a PR like this: `/packit koji-build`.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/events/pagure.py
+++ b/packit_service/worker/events/pagure.py
@@ -90,14 +90,14 @@ class PullRequestCommentPagureEvent(AbstractPRCommentEvent, AbstractPagureEvent)
         return result
 
     def get_base_project(self) -> GitProject:
-        fork = self.project.service.get_project(
+        project = self.project.service.get_project(
             namespace=self.base_repo_namespace,
             repo=self.base_repo_name,
             username=self.base_repo_owner,
-            is_fork=True,
+            is_fork=False,
         )
-        logger.debug(f"Base project: {fork} owned by {self.base_repo_owner}")
-        return fork
+        logger.debug(f"Base project: {project} owned by {self.base_repo_owner}")
+        return project
 
 
 class PullRequestPagureEvent(AddPullRequestDbTrigger, AbstractPagureEvent):
--- a/packit_service/worker/handlers/distgit.py
+++ b/packit_service/worker/handlers/distgit.py
@@ -10,6 +10,8 @@ from celery import Task
 from datetime import datetime
 from typing import Optional, Dict
 
+from fasjson_client import Client
+from fasjson_client.errors import APIError
 from ogr.abstract import PullRequest, PRStatus
 
 from packit.api import PackitAPI
@@ -23,6 +25,7 @@ from packit_service import sentry_integration
 from packit_service.config import PackageConfigGetter, ProjectToSync
 from packit_service.constants import (
     CONTACTS_URL,
+    FASJSON_URL,
     FILE_DOWNLOAD_FAILURE,
     MSG_RETRIGGER,
 )
@@ -39,6 +42,7 @@ from packit_service.worker.events import (
     ReleaseEvent,
     AbstractIssueCommentEvent,
     CheckRerunReleaseEvent,
+    PullRequestCommentPagureEvent,
 )
 from packit_service.worker.handlers.abstract import (
     JobHandler,
@@ -380,7 +384,9 @@ class ProposeDownstreamHandler(RetriableJobHandler):
 
 
 @configured_as(job_type=JobType.koji_build)
+@run_for_comment(command="koji-build")
 @reacts_to(event=PushPagureEvent)
+@reacts_to(event=PullRequestCommentPagureEvent)
 class DownstreamKojiBuildHandler(RetriableJobHandler):
     """
     This handler can submit a build in Koji from a dist-git.
@@ -404,6 +410,7 @@ class DownstreamKojiBuildHandler(RetriableJobHandler):
         )
         self.dg_branch = event.get("git_ref")
         self._pull_request: Optional[PullRequest] = None
+        self._packit_api = None
 
     @property
     def pull_request(self):
@@ -421,9 +428,37 @@ class DownstreamKojiBuildHandler(RetriableJobHandler):
                 self._pull_request = prs[0]
         return self._pull_request
 
+    @property
+    def packit_api(self):
+        if not self._packit_api:
+            self._packit_api = PackitAPI(
+                self.service_config,
+                self.job_config,
+                downstream_local_project=self.local_project,
+            )
+        return self._packit_api
+
     def get_pr_author(self):
         """Get the login of the author of the PR (if there is any corresponding PR)."""
         return self.pull_request.author if self.pull_request else None
+
+    def is_packager(self, user):
+        """Check that the given FAS user
+        is a packager
+
+        Args:
+            user (str) FAS user account name
+        Returns:
+            true if a packager false otherwise
+        """
+        self.packit_api.init_kerberos_ticket()
+        client = Client(FASJSON_URL)
+        try:
+            groups = client.list_user_groups(username=user)
+        except APIError:
+            logger.debug(f"Unable to get groups for user {user}.")
+            return False
+        return "packager" in [group["groupname"] for group in groups.result]
 
     def pre_check(self) -> bool:
         if self.data.event_type in (PushPagureEvent.__name__,):
@@ -460,6 +495,17 @@ class DownstreamKojiBuildHandler(RetriableJobHandler):
                         f"configuration: {self.job_config.allowed_committers}."
                     )
                     return False
+        elif self.data.event_type in (PullRequestCommentPagureEvent.__name__,):
+            commenter = self.data.actor
+            logger.debug(
+                f"Triggering downstream koji build through comment by: {commenter}"
+            )
+            if not self.is_packager(commenter):
+                logger.info(
+                    f"koji-build retrigger comment event on PR identifier {self.data.pr_id} "
+                    f"done by {commenter} which is not a packager."
+                )
+                return False
 
         return True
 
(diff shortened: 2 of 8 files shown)


packit_service/worker/handlers/distgit.py
```diff
@@ -461,6 +464,19 @@ def pre_check(self) -> bool:
                         f"configuration: {self.job_config.allowed_committers}."
                     )
                     return False
+        elif self.data.event_type in (PullRequestCommentPagureEvent.__name__,):
```
nikromen, line 467: I can see that we are doing similar thing here (I am implementing #1541). Can you please move this `elif` block to the method into `JobHandler` superclass so I can later use it in `CreateBodhiUpdateHandler`? Just moving this block of code to separated method in the superclass is enough... I will update this method later to fit the needs of both classes.

(Or if you want, replacing the `koji builds` part in the stings with some variable should be enough)

packit_service/worker/events/pagure.py
```diff
@@ -90,11 +90,18 @@ def get_dict(self, default_dict: Optional[Dict] = None) -> dict:
         return result
 
     def get_base_project(self) -> GitProject:
+        if isinstance(self, PullRequestCommentPagureEvent):
```
nikromen, line 93: personal preference: I would use ternary operator here

packit_service/worker/jobs.py
```diff
@@ -483,6 +483,18 @@ def get_jobs_matching_event(self) -> List[JobConfig]:
             ):
                 jobs_matching_trigger.append(job)
 
+        if not jobs_matching_trigger and isinstance(
+            self.event, PullRequestCommentPagureEvent
+        ):
+            if (
+                job.type == JobType.koji_build
```
nikromen, line 490: personal preference: It would be awesome if you named this longer condition so that it's clear to a person what it means without a few extra seconds of thinking (yes, I am a lazy person :D). What about reusing the comment you posted bellow "A koji_build can be retriggered by a Pagure comment in a PR"... so something like `can_koji_build_be_retriggered`?

packit_service/worker/jobs.py
```diff
@@ -483,6 +483,18 @@ def get_jobs_matching_event(self) -> List[JobConfig]:
...
+                and job.trigger == JobConfigTriggerType.commit
+                and self.event.job_config_trigger_type
+                == JobConfigTriggerType.pull_request
+            ):
+                "A koji_build can be retriggered by a Pagure comment in a PR"
```
nikromen, line 495:
```suggestion
                # A koji_build can be retriggered by a Pagure comment in a PR
```

nikromen (commented):
I left a one comment about checking whether user who triggers the jobs have write access to the dist-git repository so we don't write duplicated code. Feel free to ignore the other suggestions since they are just nitpicks :D

packit_service/worker/handlers/distgit.py
```diff
@@ -460,6 +463,19 @@ def pre_check(self) -> bool:
...
+            commenter = self.data.actor
+            project_url = self.data.event_dict["project_url"]
+            logger.debug(
+                f"Triggering downstream koji build through comment by: {commenter}"
+            )
+            if commenter not in self.job_config.allowed_committers:
```
lbarcziova, line 472: since this is an explicit comment, I think the commenter doesn't need to be allowed in the project configuration but only has a write access to the dist-git repo...but maybe we can discuss it on standup

packit_service/worker/handlers/distgit.py
```diff
@@ -479,9 +495,14 @@ def run(self) -> TaskResults:
             self.job_config,
             downstream_local_project=self.local_project,
         )
+        branch = (
+            self.project.get_pr(self.data.pr_id).source_branch
```
lbarcziova, line 499: we are interested in the branch the changes were merged into, so I think this should be the `target_branch`

packit_service/worker/jobs.py
```diff
@@ -483,6 +483,18 @@ def get_jobs_matching_event(self) -> List[JobConfig]:
...
 
+        if not jobs_matching_trigger and isinstance(
+            self.event, PullRequestCommentPagureEvent
+        ):
+            if (
+                job.type == JobType.koji_build
+                and job.trigger == JobConfigTriggerType.commit
+                and self.event.job_config_trigger_type
+                == JobConfigTriggerType.pull_request
+            ):
```
lbarcziova, line 494: I think this wouldn't work since we don't run this condition for all jobs in a cycle :thinking:..but looking into Steve code, couldn't we use [`get_handlers_for_comment`](https://github.com/packit/packit-service/blob/da4a797985708b814e744a7c19610f00384d6729/packit_service/worker/jobs.py#L64) (which uses the mapping via `@run_for_comment` decorator you correctly used) instead of hardcoding this entirely? (maybe some hardcoding will still be needed, but it should make things simpler)

packit_service/worker/events/pagure.py
lbarcziova: maybe I am missing something, but since this is a method of `PullRequestCommentPagureEvent`, can the `else` block happen?

packit_service/worker/handlers/distgit.py
majamassarini: When I have fixed it, as Laura wrote maybe I am doing the wrong check now, I will move it in a separate method.

packit_service/worker/events/pagure.py
majamassarini: Yes you are right! I was thinking of being in a more general PagureEvent...

packit_service/worker/handlers/distgit.py
majamassarini: Ok, so if I understand it correctly: re-triggering a koji-build through a comment should be allowed to more developers.
Some of them, which may be not allowed to do a koji-build through propose-downstream are instead allowed to re-trigger someone else propose-downstream koji-builds, right?

packit_service/worker/handlers/distgit.py
majamassarini: Ok, thanks. I was not sure. Just a thing I do not understand: if I do checkout the `target_branch` I think I do not have all the commits in the PR and thus I perform a koji-build with not all commits. Am I wrong or this is ok for the koji-build we want?

packit_service/worker/handlers/distgit.py
lbarcziova: my thinking was that with the configurable `allowed_committers` and `allowed_pr_authors` we wanted to preserve the proven packager workflow, but if someone explicitly creates a comment to retrigger the build, it doesn't make that much sense to me to require this user to be allowed in the configuration...@lachmanfrantisek what do you think? should we discuss this in a meeting?

packit_service/worker/jobs.py
lbarcziova: ok, I understand :+1:

packit_service/worker/handlers/distgit.py
lbarcziova: the conclusion from the architecture discussion: we will only require the user who retriggers the koji build by a dist-git comment to be a packager

packit_service/worker/handlers/distgit.py
majamassarini: I moved the check in a separate method, but in the same class. I am not really sure you should reuse this check. I think that not every packager (as is for the `koji-build`) should be able to do a `bodhi-update` but I may be wrong.

packit_service/worker/handlers/distgit.py
```diff
@@ -425,6 +431,29 @@ def get_pr_author(self):
...
+        Args:
+            user (str) FAS user account name
+        Returns:
+            true if a packager false otherwise
+        """
+        packit_api = PackitAPI(
+            self.service_config,
+            self.job_config,
+            downstream_local_project=self.local_project,
+        )
```
lachmanfrantisek, line 447: We can maybe move the API instance to a (lazy-evaluated) property.

packit_service/worker/jobs.py
```diff
@@ -471,6 +471,27 @@ def is_project_public_or_enabled_private(self) -> bool:
 
         return True
 
+    def allow_non_matching_events(self):
+        """Do explicit exceptions for non matching event/jobs triggers
+
+        Returns:
+            list of jobs
+        """
```
lachmanfrantisek, line 479: I am still wondering why is this needed and if we can't fix this elsewhere. This is because of the `commit` vs `pull_request` trigger, isn't it?

TBH, I was a bit confused by the method name and docstring. Could you please rename it to something like `check_manual_mapping`/`check_explicit_matching`/.. or something like that? (From the name, I thought we are allowing any combination...;)  And then, I've seen the word `exception` and was confused completely..;)

lachmanfrantisek (commented):
Well done! Thanks!

packit_service/worker/jobs.py
majamassarini: Yes this is because of the `commit` vs `pull_request` trigger. 
I may be wrong but I don't see how to allow this without this hack.
But we could probably ask to the users to put in their configuration a job with trigger `pull_request` for the `koji_build` if they want to trigger it through a comment in a `pull_request`.
I copied the config file from osbuilder project and they do not have such a job in their config, they only have a `koji_build` job triggered by a `commit`and I want my code to be able to trigger a build with their configuration.
But I can remove this lines if we agree that a change in the config file should be needed.

lachmanfrantisek: No, I don't think we should ask users to add another job. (It's same as with the `propose-downstream` where nothing is needed to be able to retrigger the job by an issue comment.)

We can leave it as is and try to refactor it later.

packit_service/worker/handlers/distgit.py
```diff
@@ -460,6 +495,17 @@ def pre_check(self) -> bool:
                         f"configuration: {self.job_config.allowed_committers}."
                     )
                     return False
+        elif self.data.event_type in (PullRequestCommentPagureEvent.__name__,):
+            commenter = self.data.actor
+            logger.debug(
+                f"Triggering downstream koji build through comment by: {commenter}"
+            )
+            if not self.is_packager(commenter):
```
lbarcziova, line 503: in this case, it could be nice to notify user about this, but we can discuss if and how we want to do it (PR comment?) as a follow-up and be also consistent with retriggering bodhi updates

lbarcziova (approved):
thanks! let's test this on stage

PR #1589, opened 2022-07-20
Retry copr build in case of Git forge errors

### LINKED ISSUE #1548: Retry copr-build job in case of the git forge error
* [ ] Make sure that the job is retried for  `ogr.exceptions.GitForgeInternalError` and `ogr.exceptions.OgrNetworkError`. There is already an auto-retry mechanism we already have:
  https://github.com/packit/packit-service/blob/307974f7bba1f349d5b0f9d0c0d3eaa6e988d2b4/packit_service/worker/tasks.py#L60-L65
* [ ] react nicely after X retries (you can get inspired by the `propose-downstream` retries for source download).
* [ ] check that the copr-build job can be retried in any state and make the job idempotent if needed

*part of the #927 epic*

FrNecas:
TODO:

Looking for feedback on the length of the delay. I feel like that in case of Github outage, the behaviour should be the same as for Copr (minutes), however in case of a 500 from Github, I think we can try earlier (hence the 10 seconds used for now). What do you think?

Fixes #1548 

Related to

Merge before/after

---

RELEASE NOTES BEGIN
When creating Copr builds, Packit will now retry multiple times in case there is a Github outage or internal error.
RELEASE NOTES END


FrNecas:
Looking at sentry, I've only found `500 null` errors in case of `GitForgeInternalError` which is not very helpful in diagnosing if we can retry quickly or should wait longer... :/

packit_service/worker/helpers/build/copr_build.py
```diff
@@ -576,25 +578,43 @@ def handle_build_submit_error(self, ex) -> TaskResults:
...
-            interval = (
-                BASE_RETRY_INTERVAL_IN_MINUTES_FOR_OUTAGES
-                * 2**self.celery_task.retries
-            )
+        if not self.celery_task.is_last_try() and (
+            possible_copr_outage_exc or forge_outage_exc or forge_internal_error
+        ):
+            what_failed = "Copr" if possible_copr_outage_exc else "Git forge"
+            if forge_internal_error:
+                # Internal error is delayed in seconds
```
lachmanfrantisek, line 589: I would probably treat git forges as another external service.

We should maybe change the structure of the retry intervals -- have an exponential back-off, cover a few retries in seconds but still be able to survive outage in minutes...
(=> Have more retries.)

But we can increase this gradually so we don't flood our queues.

lachmanfrantisek: What do you think? Do you think we are able to cover all external services with the same retries or have a different retry strategy for each service?

FrNecas: We are using exponential backoff but since we are currently using 2 retries at maximum, it isn't as significant (we multiply the wait time by `2 ** retry_number`, so the waiting will be 1 minute and then 2 minutes). The reasoning why I split the logic this way, is as follows:
 - `OgrNetworkError` means a `requests` error when performing a request, which probably signals an outage (hence the same handling as for copr - I'd go with the same handling for all the services when an outage occurs).
 - `GitForgeInternalError` means a 5XX code from GitHub which could be anything, maybe something just went wrong when processing the request, so I feel like we can retry sooner.

TomasTomecek:
I naively thought this is already merged & released since I saw 2 related sentry alerts today. Silly me.

packit_service/worker/helpers/build/copr_build.py
TomasTomecek: I'm personally in favour of trying more times (e.g. up to 5) given the number of network flakes and 5xx we are seeing.

lbarcziova: Since we already discussed with @lachmanfrantisek that for outages it would be nice to retry more times (than the default max_retries=2), I quickly looked up [the Celery docs](https://docs.celeryq.dev/en/stable/userguide/tasks.html#automatic-retry-for-known-exceptions) but they are not so clear to me. Firstly, in this sentence, I am not sure whether `internal` means the automatic retries: `If you want to specify custom arguments for an internal retry() call, pass retry_kwargs argument to app.task() decorator` (we specify `retry_kwargs` [here](https://github.com/packit/packit-service/blob/0d3d960e590165fd37b85a93f8a002798095eb32/packit_service/worker/tasks.py#L62)). Also, directly [the retry method](https://docs.celeryq.dev/en/stable/reference/celery.app.task.html#celery.app.task.Task.retry) has the `max_retries` argument which should override the default, so we could maybe utilize that.

FrNecas: Yes, even according to the implementation, it overrides. I've adjusted the implementation to retry 5 times in case of an outage.

lbarcziova (approved):
Thanks! I would say we could go with the 5 retries for both categories of errors, but I will leave it up to you or we can tweak it if needed.

TomasTomecek (approved):

lachmanfrantisek (approved):
:+1: :rocket:

FrNecas:
Oh, I forgot to add the label, whoops. Will rebase and merge :D

FrNecas:
recheck

Issue #1448, opened 2022-04-13
Don't add reaction on comment when `/packit(-stg)` is not in `packit_instances` 


nikromen:
We give :eyes: reaction to comments that are using packit instance which is not in `packit_instances`. Then - as expected - we do nothing.

This is caused because reaction to comment is given when we are processing Events but `packit_instances` is checked after that - when processing Jobs.

Example: user don't have enabled `/packit-stg` in `packit_instances` but if user uses `/packit-stg` command, bot gives :eyes: reaction and then the bot does nothing. This can be really confusing because user may expect some action (since we reacted on comment).

You can try it by commenting to [this PR](https://github.com/packit/hello-world/pull/494) with `/packit-stg` prefix.

Possible ways to solve this:

- give reaction later (after checking `packit_instances`) - this may cause larger delay between adding a comment and reaction
- delete reaction if nothing should happen (after checking `packit_instances`) - but it's weird to give a reaction and later remove it
- do the step above (delete reaction) and then add :-1: to tell the user that something is wrong

Deleting the reaction should be easy since it's already implemented in OGR and the usage looks like this:
```py
reaction = comment_object.add_reaction("+1")  # adds reaction to comment
reaction.delete()  # deletes the reaction from comment
```

TomasTomecek:
ahaaaaa, so it's about packit_instances, did not realize that; thank you for the explanation

danmyway:
Hi!
Would this also solve a situation, when the test job is requested whilst builds as dependencies are not finished?
The bot reacts to the comment/command (`/packit test`), but the test jobs won't get triggered, when build tasks finish.
For context all the test jobs are defined as manual in the config file, build jobs are set for the commit/push trigger.

pcahyna:
As an alternative, would it work to add a reaction ("I saw it") and then an error message as a reply to the comment explaining why the command is ignored? The situations where this happens look to me like errors that should be reported somehow and not silently ignored.

pcahyna:
An error message would be more useful than not adding a reaction, which would be a bit less confusing, but still would not really help the user with debugging. The user would be still left wondering why the reaction does not appear, like now they are wondering why nothing happens. One really needs explanatory error messages, not just an emoji. (Reminds me of the only way `ed` the standard text editor reported errors: a lone `?`. https://news.ycombinator.com/item?id=7727206 )

PR #1593, opened 2022-07-25
Don't require packit config when the command isn't supported by packit

### LINKED ISSUE #1592: Packit complains that our notifications repo doesn't have a config
https://github.com/packit/notifications/issues/453

This breaks the automatic verifications.

nikromen:
Don't require a packit config when we don't have matching handlers for it (non-existing commands - typos in commands etc.). With this, when user writes a typo,  `self.event.fail_when_config_file_missing` won't to False so `self.event.package_config` won't raise anything on a missing repo. The previous condition was checking for a `/packit whatever` commands :grin:  

With this we should hopefully ignore every non-existing packit commands (e.g. typos in them) but the issue that we firstly give them :eyes: even when we ignore them remains - #1448 (I was trying to do it since it's easy but... too much tests to fix :grinning: )

Fixes #1592

### DIFF
--- a/packit_service/worker/jobs.py
+++ b/packit_service/worker/jobs.py
@@ -290,14 +290,13 @@ class SteveJobs:
         Returns:
             Whether the Packit configuration is present in the repo.
         """
-        if isinstance(
-            self.event, AbstractCommentEvent
-        ) and get_packit_commands_from_comment(
+        if isinstance(self.event, AbstractCommentEvent) and get_handlers_for_comment(
             self.event.comment,
             packit_comment_command_prefix=self.service_config.comment_command_prefix,
         ):
             # we require packit config file when event is triggered by /packit command
             self.event.fail_when_config_file_missing = True
+
         if not self.event.package_config:
             # this happens when service receives events for repos which don't have packit config
             # success=True - it's not an error that people don't have packit.yaml in their repo
--- a/tests/integration/test_pr_comment.py
+++ b/tests/integration/test_pr_comment.py
@@ -1509,3 +1509,74 @@ def test_pr_test_command_handler_skip_build_option_no_fmf_metadata(
         event=event_dict,
         job_config=job_config,
     )
+
+
+@pytest.mark.parametrize(
+    "mock_pr_comment_functionality",
+    (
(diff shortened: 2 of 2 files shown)


packit_service/worker/jobs.py
```diff
@@ -290,11 +290,9 @@ def is_packit_config_present(self) -> bool:
...
             Whether the Packit configuration is present in the repo.
         """
-        if isinstance(
-            self.event, AbstractCommentEvent
-        ) and get_packit_commands_from_comment(
-            self.event.comment,
-            packit_comment_command_prefix=self.service_config.comment_command_prefix,
+        if (
+            isinstance(self.event, AbstractCommentEvent)
+            and self._search_for_handlers_for_comment_commands()
```
lbarcziova, line 295: would it make sense to only call `get_handlers_for_comment` here? The name `_search_for_handlers_for_comment_commands` currently doesn't match with what the method does - it searches handlers for both comments and check reruns and we should not need to search handlers for check reruns here

lbarcziova (commented):
thank you for looking into this! could we have also a simple test case so that we make sure we don't break this in the future?

packit_service/worker/jobs.py
nikromen: nice catch! reruns aren't comment commands :grin:

tests/integration/test_pr_comment.py
```diff
@@ -1509,3 +1509,74 @@ def test_pr_test_command_handler_skip_build_option_no_fmf_metadata(
...
+            FileNotFoundError
+        ),  # no config
+        get_files=lambda ref, filter_regex: ["the-specfile.spec"],
+        get_web_url=lambda: "https://github.com/the-namespace/the-repo",
+    )
+
+    ServiceConfig.get_service_config().comment_command_prefix = "/packit"
+    pr_embedded_command_comment_event["comment"][
+        "body"
+    ] = "/packit 10minutesOfImplementing3HoursOfTesting"
```
lbarcziova, line 1570: :grinning:

lbarcziova (approved):
thanks!

PR #1598, opened 2022-07-29
Always allow explicit job/event matching

majamassarini:
Packit config has many other matching jobs,
do not skip the koji-build job just because there are other matching jobs.

related to packit-service#1540

### DIFF
--- a/packit_service/worker/jobs.py
+++ b/packit_service/worker/jobs.py
@@ -500,8 +500,7 @@ class SteveJobs:
             ):
                 jobs_matching_trigger.append(job)
 
-        if not jobs_matching_trigger:
-            jobs_matching_trigger.extend(self.check_explicit_matching())
+        jobs_matching_trigger.extend(self.check_explicit_matching())
 
         return jobs_matching_trigger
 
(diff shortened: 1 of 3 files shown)


lachmanfrantisek (approved):

lachmanfrantisek (commented):
Oh, maybe one thing. Can we add this use-case to the tests?

edit: Probably here: https://github.com/packit/packit-service/blob/main/tests/unit/test_jobs.py

majamassarini:
> Oh, maybe one thing. Can we add this use-case to the tests?
> 
> edit: Probably here: https://github.com/packit/packit-service/blob/main/tests/unit/test_jobs.py

The test already exists: https://github.com/packit/packit-service/blob/main/tests/unit/test_jobs.py#L2362

With the previous version of the code one more test was needed to catch the problem I met today. With this new version of the code this is the only needed test, I think :-).

lachmanfrantisek:
I meant the test for the use-case covered by this PR (so it fails before this change and successes after it).

So, exactly what you had in the description: `Packit config has many other matching jobs, do not skip the koji-build job just because there are other matching jobs.`

lbarcziova (approved):

majamassarini:
recheck

PR #1600, opened 2022-08-02
Improve configuration tests

csomh:
When trying to fix failures in packit/packit#1663, it turned out some configuration tests were overzealous, and were also testing functionality within _packit_. Re-factor these tests to limit their scope only to functionality in _packit_service_.

Other tests needed a 'specfile_path' to be explicitly defined in order to work with changes from the PR above.

Additionally, add a new Makefile-target to easily build a revdep-test-image for local testing. Document its usage.


lachmanfrantisek (approved):
Thanks!

PR #1602, opened 2022-08-05
Use proper source branch for ogr and specfile

nforro:


TomasTomecek (approved):
both builds passed for me locally; let's try stg! :)

files/install-deps-worker.yaml
```diff
@@ -52,6 +52,7 @@
         state: present
     - import_tasks: tasks/install-packit-deps.yaml
     - import_tasks: tasks/install-ogr-deps.yaml
+    - import_tasks: tasks/install-specfile-deps.yaml
```
jpopelka, line 55: I see that for worker we first install `packit` deps and then `ogr` & `specfile` deps, while for service (`files/install-deps.yaml`) we first install `ogr` & `specfile` deps and then `packit` deps.
Can't it possibly result in a different package set installed?
I know it already was like that, and that you just added the `specfile` deps into the set, I just wanted to ask whether you think it's OK.

nforro: Good point, I think it can't hurt to unify this. I suppose installing `ogr` and `specfile` deps first makes more sense, right?

jpopelka: Yes, I'd install `packit` after `ogr` & `specfile`.

nforro: Changed in https://github.com/packit/packit-service/pull/1603.

PR #1611, opened 2022-08-11
Make test_pr_comment runnable alone

FrNecas:
Found while working on group models but since it's unrelated and a simple fix, creating a simple PR.

Previously, the test only worked when run as a part of the whole suite
because other tests set the DASHBOARD_URL. Make the test self-contained
by mocking the dashboard url there as well.

Previously, `TEST_TARGET="tests/integration/test_pr_comment.py" make check-in-container` would result in a failed test (whereas `make check-in-container` passes).


FrNecas:
recheck

Issue #1601, opened 2022-08-03
Copr builds fail for epel-7 and epel-6


danmyway:
Since yesterday 2022-08-02, approximately 14:30 packages for epel-6 and epel-7 fail to be build for [Convert2RHEL](https://github.com/oamg/convert2rhel) PR's.
Building packages for epel-8 works fine.
See https://copr.fedorainfracloud.org/coprs/g/oamg/convert2rhel/build/4698072/
There were no issues with packages being built for `main` at that time
See https://copr.fedorainfracloud.org/coprs/g/oamg/convert2rhel/build/4698179/

```
Start: buildsrpm
Start: rpmbuild -bs
error: bad date in %changelog: Tue Aug 02 02:31:02 UTC 2022 Packit <hello@packit.dev> - 0.26-1.20220802143102718182.pr451.28.g33894f1
Building target platforms: x86_64
Building for target x86_64
Finish: rpmbuild -bs
INFO: chroot_scan: 3 files copied to /var/lib/copr-rpmbuild/results/chroot_scan
INFO: /var/lib/mock/epel-6-x86_64-1659450720.702706/root/var/log/dnf.rpm.log
/var/lib/mock/epel-6-x86_64-1659450720.702706/root/var/log/dnf.librepo.log
/var/lib/mock/epel-6-x86_64-1659450720.702706/root/var/log/dnf.log
Finish: buildsrpm
ERROR: Exception(/var/lib/copr-rpmbuild/workspace/workdir-6hm9s8u6/convert2rhel/convert2rhel.spec) Config(epel-6-x86_64) 0 minutes 57 seconds
INFO: Results and/or logs in: /var/lib/copr-rpmbuild/results
INFO: Cleaning up build root ('cleanup_on_failure=True')
Start: clean chroot
INFO: unmounting tmpfs.
Finish: clean chroot
ERROR: Command failed: 
 # bash --login -c /usr/bin/rpmbuild -bs  --target x86_64 --nodeps /builddir/build/SPECS/convert2rhel.spec

Copr build error: Mock build failed
```

@TomasTomecek edit:
Task list for this issue is below: https://github.com/packit/packit-service/issues/1601#issuecomment-1207923372

TomasTomecek:
do I get it correctly that the date is the problem?

```
error: bad date in %changelog: Tue Aug 02 02:31:02 UTC 2022 Packit <hello@packit.dev> - 0.26-1.20220802143102718182.pr451.28.g33894f1
```

[...]

PR #1616, opened 2022-08-16
Install packit in container images from copr

nforro:
Related to packit/packit-service#1601.

Merge after packit/deployment#367.

---

RELEASE NOTES BEGIN
Tooling in container images is now installed from Packit Copr repositories rather than from source using pip.
RELEASE NOTES END


nforro:
Just a note, sandcastle is still [being installed](https://github.com/packit/packit-service/blob/52fb387b7514d321a9b73bb22355e806a5091ad9/files/install-deps-worker.yaml#L56) from git using pip. Should we do something about it?

files/tasks/install-packit-deps.yaml
```diff
@@ -1,49 +0,0 @@
-# Copyright Contributors to the Packit project.
-# SPDX-License-Identifier: MIT
-
-# Install dependencies for packit
-- name: Download script setupcfg2rpm.py
-  ansible.builtin.get_url:
-    url: https://raw.githubusercontent.com/packit/deployment/main/scripts/setupcfg2rpm.py
```
jpopelka, line 7: Is the `setupcfg2rpm.py` still used anywhere? If not, can it be deleted as well?

jpopelka (approved):

jpopelka:
> Just a note, sandcastle is still being installed from git using pip. Should we do something about it?

We'd need to package it first and it might now be that easy because there's [one pinned dependency](https://github.com/packit/sandcastle/blob/main/setup.cfg#L31).
And given that we're moving away from building via sandcastle to building in copr, I think it's not worth the effort.

files/tasks/install-packit-deps.yaml
nforro: I don't think it is still used, but there is a reference to it in [tokman containerfile](https://github.com/packit/tokman/blob/8db9af20f24f7e7eb855f25e8431f723c061f955/Containerfile#L6-L7), not sure why.

nforro: I'm blind, it [is being used](https://github.com/packit/tokman/blob/8db9af20f24f7e7eb855f25e8431f723c061f955/Containerfile#L18) there. Should I keep it or try to remove it?

TomasTomecek (approved):
nice and clean!! 👋👋

files/tasks/install-packit-deps.yaml
jpopelka: Definitely keep.

Issue #1588, opened 2022-07-19
API queries take a long time to return results


jpopelka:
For example, querying 
https://prod.packit.dev/api/srpm-builds?page=1000&per_page=50
takes about 25 seconds while it should be more or less just 

    # EXPLAIN (ANALYZE) SELECT * FROM srpm_builds ORDER BY srpm_builds.id LIMIT 50 OFFSET 50000;
    Execution Time: 61.416 ms

or https://prod.packit.dev/api/copr-builds?page=1000&per_page=50
takes 30 seconds while

    # EXPLAIN (ANALYZE) SELECT * FROM copr_build_targets LIMIT 50 OFFSET 50000;
    Execution Time: 11.480 ms

So what (the heck) takes the API so long when it should be done in less than a second?
Do we add so much overhead on top of the simple sequential scan?
Or are the SQLAlchemy-generated queries so inefficient?

There are ways how to debug SQLAlchemy:
- https://stackoverflow.com/questions/27748053/how-to-log-sql-statements-and-rows-returned-in-sqlalchemy-to-aid-in-debugging
- https://docs.sqlalchemy.org/en/14/faq/performance.html#query-profiling
- https://skyportal.io/docs/query_analysis.html

If you want to run some code directly in the packit-service's pod's terminal, you need to increase its memory limit, because I was getting 'Killed' when trying to run some code in ipython3, while on stage it was OK because it has more free memory.

jpopelka:
I just tried locally with the imported prod db dump and those queries now finish in seconds.
So either #1617 fixed it or it can't be reproduced locally (I hadn't tried locally before filling this issue).
We'll see next week after #1617 gets to prod.

jpopelka:
More than twice faster now :rocket: 
```bash
$ time http 'https://prod.packit.dev/api/srpm-builds?page=1000&per_page=50'
real	0m9.058s
$ time http 'https://prod.packit.dev/api/copr-builds?page=1000&per_page=50'
real	0m12.520s
```

PR #1617, opened 2022-08-17
Don't commit (& rollback) read-only queries

jpopelka:
See the commit messages.

Might help with #1588

I temporarily redirected staging service/worker image streams to [my builds](https://quay.io/user/jpopelka), let it process some jobs and did some queries via API and everything looked OK.

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -47,8 +47,6 @@ from packit.exceptions import PackitException
 from packit_service.constants import ALLOWLIST_CONSTANTS
 
 logger = logging.getLogger(__name__)
-# SQLAlchemy session, get it with `get_sa_session`
-session_instance = None
 
 
 def get_pg_url() -> str:
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -215,10 +215,9 @@ class ProjectAndTriggersConnector:
 
     def get_issue_id(self) -> Optional[int]:
         trigger_object = self.get_trigger_object()
-        if not isinstance(trigger_object, IssueModel):
-            return None
-
-        return trigger_object.issue_id
+        if isinstance(trigger_object, IssueModel):
+            return trigger_object.issue_id
+        return None
 
     def get_branch_name(self) -> Optional[str]:
         trigger_object = self.get_trigger_object()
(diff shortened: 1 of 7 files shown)


lbarcziova (approved):
thanks!

packit_service/worker/helpers/build/babysit.py
```diff
@@ -141,7 +141,7 @@ def check_copr_build(build_id: int) -> bool:
         bool: Whether the run was successful, False signals the need to retry.
     """
     logger.debug(f"Getting copr build ID {build_id} from DB.")
-    builds = CoprBuildTargetModel.get_all_by_build_id(build_id)
+    builds = list(CoprBuildTargetModel.get_all_by_build_id(build_id))
```
mfocko, line 144: It doesn't look like we need a list here, cause it calls `update_copr_builds` and is used there just in 2 `for-each` loops

packit_service/worker/helpers/testing_farm.py
```diff
@@ -410,11 +410,11 @@ def get_latest_copr_build(
             owner=self.job_owner,
             target=target,
         )
-        if not copr_builds:
+        try:
+            return next(iter(copr_builds))
+        except StopIteration:
             return None
```
mfocko, line 416: _Python ships whole universe with the interpreter, yet it doesn't have `first_or_none`_ /o\

packit_service/models.py
```diff
@@ -58,14 +63,15 @@ def get_pg_url() -> str:
     )
 
 
-engine = create_engine(get_pg_url())
-ScopedSession = scoped_session(sessionmaker(bind=engine))
+# To log SQL statements, set echo=True
```
mfocko, line 66: :eyes:

mfocko (approved):
Thanks a lot :) Just out of interest, have you tried it locally with the dump of the DB?

packit_service/worker/helpers/build/babysit.py
jpopelka: Yes, but we first want to check whether it's empty and to do that I'd either need to do
- `next(iter()))` or
- call `.first()`, which would make `mypy` angry because `error: "Iterable[CoprBuildTargetModel]" has no attribute "first"`
So I decided that this would be the easiest.

packit_service/worker/helpers/testing_farm.py
jpopelka: Actually, `.first()` does that, i.e. it returns `None` in case of empty results, but mypy screams because
`error: "Iterable[CoprBuildTargetModel]" has no attribute "first"`
So I did it this way to make it happy because I was lazy to add `CoprBuildTargetModel.get_first_by()`.

mfocko: I have a feeling the `.first()` comes from SQLAlchemy and is not supported for any iterable /o\

jpopelka:
thanks for reviews!

> have you tried it locally with the dump of the DB?

Yes, but there's only a small subset of things I'm able to test locally so as the https://github.com/packit/packit-service/pull/1617#issue-1341902833 says I tried that on stg as well.

packit_service/worker/helpers/testing_farm.py
jpopelka: Oh, you didn't mean the SQLAlchemy's first() but something which would do that `try next(iter())) except StopIteration` stuff. I see.
That, or at least [has_next()](https://stackoverflow.com/questions/1966591/has-next-in-python-iterators) would be nice.

jpopelka:
Anyway, https://dashboard.stg.packit.dev/jobs feels now (to me) ever faster than after your packit/dashboard#189

PR #1628, opened 2022-08-23
Update checking Testing Farm composes

### LINKED ISSUE #1612: Use composes endpoint for internal TF
There is a new endpoint (`https://api.dev.testing-farm.io/v0.1/composes/redhat`) that can be used to check the available composes publically. Let's use it before our custom mapping so users can pick anything explicitly. So, the task is to:
* [ ] When running tests in internal TF, check the presence of a compose by using the `https://api.dev.testing-farm.io/v0.1/composes/redhat` endpoint. If it is there, use it.
* [ ] Do the mapping, if it does not match, return an error.
* [ ] + Update the public endpoint to `https://api.dev.testing-farm.io/v0.1/composes/public`.

lbarcziova:
The endpoints for getting the TF composes were updated: 'https://api.dev.testing-farm.io/v0.1/composes/redhat' for
internal TF and 'https://api.dev.testing-farm.io/v0.1/composes/public' for the public. Update these in the code
and check the existence of compose also for internal TF. Implement also the reporting if the distro
cannot be mapped to a compose.

Fixes #1612

---

RELEASE NOTES BEGIN
We have implemented checking the available composes before submitting the tests for both internal and public Testing Farm. 
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/helpers/testing_farm.py
+++ b/packit_service/worker/helpers/testing_farm.py
@@ -179,6 +179,7 @@ class TestingFarmJobHelper(CoprBuildJobHelper):
     def _payload(
         self,
         target: str,
+        compose: str,
         artifact: Optional[Dict[str, Union[List[str], str]]] = None,
         build: Optional["CoprBuildTargetModel"] = None,
     ) -> dict:
@@ -197,7 +198,6 @@ class TestingFarmJobHelper(CoprBuildJobHelper):
             build: The related copr build.
         """
         distro, arch = target.rsplit("-", 1)
-        compose = self.distro2compose(distro, arch)
         fmf = {"url": self.fmf_url}
         if self.fmf_ref:
             fmf["ref"] = self.fmf_ref
@@ -283,7 +283,7 @@ class TestingFarmJobHelper(CoprBuildJobHelper):
             },
         }
 
-    def _payload_install_test(self, build_id: int, target: str) -> dict:
+    def _payload_install_test(self, build_id: int, target: str, compose: str) -> dict:
         """
         If the project doesn't use fmf, but still wants to run tests in TF.
         TF provides 'installation test', we request it in ['test']['fmf']['url'].
@@ -291,7 +291,6 @@ class TestingFarmJobHelper(CoprBuildJobHelper):
         """
         copr_build = CoprBuildTargetModel.get_by_build_id(build_id)
         distro, arch = target.rsplit("-", 1)
-        compose = self.distro2compose(distro, arch)
         return {
             "api_key": self.service_config.testing_farm_secret,
             "test": {
@@ -331,7 +330,7 @@ class TestingFarmJobHelper(CoprBuildJobHelper):
         except FileNotFoundError:
             return False
 
-    def distro2compose(self, distro: str, arch: str) -> str:
+    def distro2compose(self, target: str) -> Optional[str]:
         """
         Create a compose string from distro, e.g. fedora-33 -> Fedora-33
         https://api.dev.testing-farm.io/v0.1/composes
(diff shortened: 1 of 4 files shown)


packit_service/worker/helpers/testing_farm.py
```diff
@@ -354,40 +358,61 @@ def distro2compose(self, distro: str, arch: str) -> str:
...
+            elif compose == "RHEL-6":
+                compose = "RHEL-6-LatestReleased"
+            elif compose == "RHEL-7":
+                compose = "RHEL-7-LatestReleased"
+            elif compose == "RHEL-8":
+                compose = "RHEL-8.5.0-Nightly"
+            elif compose == "Oracle-Linux-7":
+                compose = "Oracle-Linux-7.9"
+            elif compose == "Oracle-Linux-8":
+                compose = "Oracle-Linux-8.6"
```
jpopelka, line 404: I don't see Oracle-Linux-X.Y in https://api.dev.testing-farm.io/v0.1/composes/redhat so I'm wondering what's the status. @thrix ?

packit_service/worker/helpers/testing_farm.py
```diff
@@ -455,14 +480,19 @@ def run_testing_farm(
...
             description=f"{'Build succeeded. ' if not self.skip_build else ''}"
             f"Submitting the tests ...",
             target=target,
         )
 
+        compose = self.distro2compose(target)
+
+        if not compose:
+            msg = "We were not able to map distro to TF compose."
+            return TaskResults(success=False, details={"msg": msg})
```
lachmanfrantisek, line 494: Would it be possible to let the user know about it?

E.g. `The compose XY is not available in the {internal/public} Testing Farm infrastructure.`

Ideally, use markdown content for more details:
* List available composes or link the endpoint.
* Suggest checking the config.
* Suggest contacting the team if the valid compose is not available and I, as a user, think it should be there.

lachmanfrantisek (commented):
Nice! Just a UX suggestion... Otherwise LGTM.

(Maybe a side note -- I wonder how we can clarify what's `compose` and what's `target`...)

packit_service/worker/helpers/testing_farm.py
lbarcziova: Updated, I haven't mentioned the mapping because pointing to the mapping code can get outdated and I wasn't able to figure out a clear message describing it (and since people will probably contact us following the message, we can resolve this with them individually maybe?).

packit_service/worker/helpers/testing_farm.py
```diff
@@ -354,40 +358,73 @@ def distro2compose(self, distro: str, arch: str) -> str:
...
+            elif compose == "Oracle-Linux-7":
+                compose = "Oracle-Linux-7.9"
+            elif compose == "Oracle-Linux-8":
+                compose = "Oracle-Linux-8.6"
+
+        if compose not in composes:
+            msg = (
+                f"The compose {compose} (from target {distro}) is not in the list of "
+                f"available composes {composes}. "
+            )
```
lachmanfrantisek, line 410:
```suggestion
            composes_str = "\n".join(f"* {c}" for c in composes)
            msg = (
                f"The compose {compose} (from target {distro}) is not in the list of "
                f"available composes:\n{composes_str}\n"
            )
```

packit_service/worker/helpers/testing_farm.py
lachmanfrantisek: I agree linking the code is not a good idea. The list might be long but that's ok if it is in the markdown detail.

lachmanfrantisek (approved):

PR #1631, opened 2022-08-24
Use singleton SQLAlchemy session for workers

### LINKED ISSUE #1591: TimeoutError: QueuePool limit of size 5 overflow 10 reached, connection timed out, timeout 30.00 (Background on...
Happened on stg when there was a short-running worker with more threads (worker --autoscale=1,16 --pool=gevent)

```
TimeoutError: QueuePool limit of size 5 overflow 10 reached, connection timed out, timeout 30.00 (Background on this error at: https://sqlalche.me/e/14/3o7r)
(16 additional frame(s) were not displayed)
```

Sentry Issue: [PACKIT-SERVICE-4HQ](https://sentry.io/organizations/red-hat-0p/issues/3447050554/?referrer=github_integration)

https://docs.sqlalchemy.org/en/14/errors.html#queuepool-limit-of-size-x-overflow-y-reached-connection-timed-out-timeout-z

jpopelka:
The [Contextual/Thread-local session](https://docs.sqlalchemy.org/en/14/orm/contextual.html) registry, as created by `scoped_session()`, is needed for `httpd` (packit-service), which uses the traditional threads.

It's however not compatible with alternative forms of concurrency control, i.e. greenlets as provided by the eventlet/gevent libraries, which celery uses for alternative execution pool implementation.
Running multi-(green)threaded worker with it results in [QueuePool limit of size x overflow y reached, connection timed out](
https://docs.sqlalchemy.org/en/14/errors.html#queuepool-limit-of-size-x-overflow-y-reached-connection-timed-out-timeout-z).

Using a single session, kept as a global variable, is the only way I've found that doesn't throw a bunch of other errors (like 'Instance not bound to a session' or 'Object already attached to a session') and doesn't look more like a workaround rather than a clean code.
We actually used it like this before 631a4a9f so it should be fine per se.

Fixes #1591

I tried it on staging and didn't see any problems, but we need to watch it carefully as they might appear later.

---

RELEASE NOTES BEGIN
N/A
RELEASE NOTES END

### DIFF
--- a/files/run_worker.sh
+++ b/files/run_worker.sh
@@ -30,7 +30,8 @@ elif [[ "${CELERY_COMMAND}" == "worker" ]]; then
     DEFAULT_QUEUES="short-running,long-running"
     QUEUES="${QUEUES:-$DEFAULT_QUEUES}"
 
-    # Min,max number of concurrent worker processes/threads executing tasks.
+    # Max,min number of concurrent worker processes/threads executing tasks.
+    # pool support: prefork, gevent
     # https://docs.celeryq.dev/en/stable/userguide/workers.html#autoscaling
     DEFAULT_AUTOSCALE="1,1"
     AUTOSCALE="${AUTOSCALE:-$DEFAULT_AUTOSCALE}"
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -10,6 +10,7 @@ import logging
 import os
 from contextlib import contextmanager
 from datetime import datetime, timedelta, timezone
+from pathlib import Path
 from typing import (
     Dict,
     Iterable,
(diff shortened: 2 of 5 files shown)


lachmanfrantisek (approved):
Thanks for working on this. I believe you..;)

lachmanfrantisek:
Btw. Can I ignore the `CeleryWorkerOffline` alerts I am receiving (the last one 3 minutes ago)?

lbarcziova (approved):

jpopelka:
> Can I ignore the `CeleryWorkerOffline` alerts

tl;dr yes

On staging, we run only the common `packit-worker-0` and when I want to play with the concurrency I scale it down and scale up one short-running and one long-running. After some time you get alert that the `packit-worker-0` is down.
And once I stop my experiments and turn on the `packit-worker-0` and turn off the `packit-worker-short-running-0`/`packit-worker-long-running-0` you get also alerts about them being down ;-)
It stops after some time :)

PR #1632, opened 2022-08-24
Raise Exception for triggering celery autoretries

majamassarini:
If `autoretry_for` some Exceptions is not specified then
celery autoretries are not used
and `retry_backoff` and `retry_backoff_max`
are ignored.

Fixes packit/packit#1450

### DIFF
--- a/packit_service/worker/tasks.py
+++ b/packit_service/worker/tasks.py
@@ -43,8 +43,13 @@ from packit_service.worker.helpers.build.babysit import (
 )
 from packit_service.worker.jobs import SteveJobs
 from packit_service.worker.result import TaskResults
+from packit.exceptions import PackitException
 
 logger = logging.getLogger(__name__)
+
+
+class PackitCoprBuildTimeoutException(PackitException):
+    """Copr build has timed out"""
 
 
 @after_setup_logger.connect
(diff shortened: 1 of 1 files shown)


lbarcziova (commented):
nice idea! would it make sense to create a specific exception for this?

I know it's hard to write tests for this, were you able to test it locally in the end?

majamassarini:
> nice idea! would it make sense to create a specific exception for this?

Ok

> I know it's hard to write tests for this, were you able to test it locally in the end?

Yes, I have somehow reproduced it locally
```python
from celery import Celery, Task
from celery.utils.log import get_task_logger

host = "localhost"
port = "6379"
broker_url = f"redis://{host}:{port}/0"
backend_url = f"redis://{host}:{port}/1"

logger = get_task_logger(__name__)

app = Celery('retry_app', broker=broker_url, backend=backend_url)

@app.task(name='print_hello',
    bind=True,
    autoretry_for=(Exception, ),
    retry_backoff=1,
    retry_backoff_max=32,
    max_retries=14,
    retry_jitter=False,
)
def print_hello(self):
    logger.error("hello world")
    raise Exception

if __name__ == "__main__":
    result = print_hello.apply_async()
    str = result.get(timeout=32)
    print(str)
```

If the `raise Exception` is substituted with the `self.retry` (and `autoretry_for` is not specified) then the `retry_backoff` values are not used. Instead it is used a standard 180s value. For this reason I am quite confident that this change will made the trick.

lbarcziova (approved):
thanks!

jpopelka (commented):
Good catch :rocket:

PR #1634, opened 2022-08-25
Run pre_check for TestingFarmResultsHandler called from babysit task

### LINKED ISSUE #1630: Stray jobs reported with private/public packit mixture
Reported jobs for  https://github.com/teemtee/tmt/pull/1463 seems a bit off:
You can see that "testing-farm:fedora-36-x86_64:internal-plugins — idorax can't run tests internally" 
however for "testing-farm:centos-stream-8-x86_64:internal-plugins - Test passed..."

When I opened "Details" I saw that all testing-farm:centos-stream-8-x86_64 jobs were pointing to the same Dashboard / Testing farm link. So in the end packit ran only "testing-farm:centos-stream-8-x86_64" but reported it as a success for "internal-plugins" and "internal-wow" as well. 

Currently .packit is set to execute only fedora-latest-stable for internal jobs.

Tl;dr: More jobs are reported than executed.

image

PR is under development so jobs will change... But IIRC stray jobs appear after some time (fresh results are for just correct jobs, few hours later the stray centos jobs appear).

lbarcziova:
In the pre_check, the equality of identifiers is checked. Without running it,
the reporting of the TF results can be incorrect.

Fixes #1630

---

RELEASE NOTES BEGIN
We have fixed the reporting when multiple Testing Farm jobs with identifiers are configured.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/handlers/testing_farm.py
+++ b/packit_service/worker/handlers/testing_farm.py
@@ -403,4 +403,11 @@ class TestingFarmResultsHandler(JobHandler):
         return TaskResults(success=True, details={})
 
     def pre_check(self) -> bool:
-        return self.data.identifier == self.job_config.identifier
+        if self.data.identifier != self.job_config.identifier:
+            logger.debug(
+                f"Skipping reporting, identifiers don't match "
+                f"(identifier of the test job to report: {self.data.identifier}, "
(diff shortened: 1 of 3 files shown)


jpopelka (approved):
:+1:

PR #1636, opened 2022-08-30
SRPMs are built in Copr by default for installations since September 6

### LINKED ISSUE #1583: Build SRPMs in Copr for newly onboarded projects

lachmanfrantisek:

TODO:

- [x] Write a new test.
- [x] Update the old tests.
- [x] Update documentation: packit/packit.dev#518

Fixes #1583

---

RELEASE NOTES BEGIN
SRPMs for Copr Builds are built in Copr by default for Packit GitHub app installations since September 6, 2022. For older installations, you can set the  `srpm_build_deps` config option to use Copr as a builder. Let us know if you hit any issue with the new implementation. We are going to slowly decommission the old implementation and are happy to help with the transition.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/handlers/copr.py
+++ b/packit_service/worker/handlers/copr.py
@@ -18,6 +18,7 @@ from packit.config.package_config import PackageConfig
 from packit_service.constants import (
     COPR_API_SUCC_STATE,
     COPR_SRPM_CHROOT,
+    DATE_OF_DEFAULT_SRPM_BUILD_IN_COPR,
     PG_BUILD_STATUS_FAILURE,
     PG_BUILD_STATUS_SUCCESS,
     INTERNAL_TF_BUILDS_AND_TESTS_NOT_ALLOWED,
@@ -25,6 +26,7 @@ from packit_service.constants import (
 from packit_service.models import (
     AbstractTriggerDbType,
     CoprBuildTargetModel,
+    GithubInstallationModel,
     SRPMBuildModel,
 )
 from packit_service.worker.events import (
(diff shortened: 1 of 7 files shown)


packit_service/worker/handlers/copr.py
```diff
@@ -135,8 +137,20 @@ def check_if_actor_can_run_job_and_report(self, actor: str) -> bool:
...
+                namespace=self.project.namespace, repo_name=self.project.repo
+            )
+        ):
+            return installation.created_at
+        return None
+
     def run(self) -> TaskResults:
-        if self.package_config.srpm_build_deps is not None:
+        installed_at = self.get_packit_github_installation_time()
+        if self.package_config.srpm_build_deps is not None or (
```
jpopelka, line 151: [srpm_build_deps](https://github.com/packit/packit/blob/4b30b85f3d943ef068d8cdd5d3bf6fbac17e823a/packit/config/common_package_config.py#L147) is a list, so it can be empty, right?
```suggestion
        if self.package_config.srpm_build_deps or (
```

lachmanfrantisek: Since we use it as an enabler, I would accept an empty string (=non-default value / =user interaction happened ) to build in Copr as well.

lachmanfrantisek: When thinking about this I realised that I need to transform `None` into an empty list for new installations that do not define `srpm_build_deps` at all...

lachmanfrantisek: I've just added test cases to cover that.

lbarcziova (approved):
nice! should we mention in the release notes that this is only about SRPMs for Copr builds and for Koji builds, everything is the same or do you think it is clear?

lachmanfrantisek:
> should we mention in the release notes that this is only about SRPMs for Copr builds and for Koji builds, everything is the same or do you think it is clear?

Good point. I'll rather mention that.

PR #1639, opened 2022-08-30
More indexes

### LINKED ISSUE #1588: API queries take a long time to return results

jpopelka:
Fixes #1588 

---

RELEASE NOTES BEGIN
More indexes added to the database have further improved API/dashboard response times.
RELEASE NOTES END


lachmanfrantisek (approved):
Wow! Thanks!

lbarcziova (approved):
:+1:

PR #1644, opened 2022-09-02
Solo execution pool & --autoscale -> --concurrency

jpopelka:
Works OK on staging.

### DIFF
--- a/files/run_worker.sh
+++ b/files/run_worker.sh
@@ -30,13 +30,13 @@ elif [[ "${CELERY_COMMAND}" == "worker" ]]; then
     DEFAULT_QUEUES="short-running,long-running"
     QUEUES="${QUEUES:-$DEFAULT_QUEUES}"
 
-    # Number of concurrent worker processes/threads executing tasks.
-    # https://docs.celeryq.dev/en/stable/userguide/workers.html#autoscaling
+    # Number of concurrent worker threads executing tasks.
     DEFAULT_CONCURRENCY="1"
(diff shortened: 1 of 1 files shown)


files/run_worker.sh
```diff
@@ -30,14 +30,13 @@ elif [[ "${CELERY_COMMAND}" == "worker" ]]; then
...
-    DEFAULT_AUTOSCALE="1,1"
-    AUTOSCALE="${AUTOSCALE:-$DEFAULT_AUTOSCALE}"
+    # Number of concurrent worker threads executing tasks.
+    DEFAULT_CONCURRENCY="1"
+    CONCURRENCY="${CONCURRENCY:-$DEFAULT_CONCURRENCY}"
 
     # Options: prefork | eventlet | gevent | solo
-    DEFAULT_POOL="prefork"
+    # https://www.distributedpython.com/2018/10/26/celery-execution-pool/
+    DEFAULT_POOL="solo"
```
TomasTomecek, line 39: Had to read the description, thanks for the pointer!

> The solo pool is a bit of a special execution pool. Strictly speaking, the solo pool 
> is neither threaded nor process-based. And more strictly speaking, the solo pool is
> not even a pool as it is always solo. And even more strictly speaking, the solo pool 
> contradicts the principle that the worker itself does not process any tasks.

> The solo pool runs inside the worker process. It runs inline which means there is no 
> bookkeeping overhead. Which makes the solo worker fast. But it also blocks the worker 
> while it executes tasks. Which has some implications when remote-controlling workers.

TomasTomecek (approved):
okay, I understand now: default = solo, but can be overridden

PR #1652, opened 2022-09-12
Refactor functionality for timezone-awareness checking

lbarcziova:


lachmanfrantisek (approved):
I am wondering if we can't get rid of this and switch all of the dates to `UTC`. Or is it too risky and we want to be sure?

jpopelka (approved):

lbarcziova:
> I am wondering if we can't get rid of this and switch all of the dates to UTC. Or is it too risky and we want to be sure?

I think we should not replace the timezone if the datetime already has one (even if it is very likely we work with UTC in most cases)

PR #1654, opened 2022-09-14
Retrigger bodhi update via dist-git PR comment

### LINKED ISSUE #1541: Implement retriggering Bodhi updates via dist-git pull-request comments

nikromen:
I'm sorry it took so long. I couldn't find the time to come back to this issue recently :disappointed: 

note: OGR doesn't have `has_write_access` method. Not yet. I am going to implement it :)

also help needed :grin::  `test_bodhi_update_retrigger_via_dist_git_pr_comment` won't pass because because the `CreateBodhiUpdateHandler.run` method is not called. Even though the `task.bodhi_update` task gets the signature of the Celery task. This means that the `JobHandler.get_signature` method is executed and according to its documentation the `run_bodhi_update` task should be run, but this obviously does not happen. Any idea why?

Fixes #1541

Merge after [org#742](https://github.com/packit/ogr/pull/742) [packit#1729](https://github.com/packit/packit/pull/1729)

---

RELEASE NOTES BEGIN
Packit now allows to re-trigger Bodhi update via dist-git PR comment `/packit create-update`.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/handlers/bodhi.py
+++ b/packit_service/worker/handlers/bodhi.py
@@ -36,6 +36,7 @@ from packit_service.worker.handlers.abstract import (
     configured_as,
     reacts_to,
     RetriableJobHandler,
+    run_for_comment,
 )
 from packit_service.worker.result import TaskResults
 
@@ -51,6 +52,8 @@ class KojiBuildEventWrapper:
 
 
 @configured_as(job_type=JobType.bodhi_update)
+@run_for_comment(command="create-update")
+@reacts_to(event=PullRequestCommentPagureEvent)
 @reacts_to(event=KojiBuildEvent)
 class CreateBodhiUpdateHandler(RetriableJobHandler):
     """
@@ -75,7 +78,9 @@ class CreateBodhiUpdateHandler(RetriableJobHandler):
         )
 
         # lazy properties
-        self._koji_build_event_data: Optional[KojiBuildEvent] = None
+        self._koji_build_event_data: Optional[KojiBuildEventWrapper] = None
+        self._packit_api: Optional[PackitAPI] = None
+        self._local_project: Optional[LocalProject] = None
 
     def _build_from_event_dict(self) -> KojiBuildEventWrapper:
         koji_build_event = KojiBuildEvent.from_event_dict(self.data.event_dict)
@@ -114,6 +119,52 @@ class CreateBodhiUpdateHandler(RetriableJobHandler):
 
         return self._koji_build_event_data
 
+    @property
+    def packit_api(self) -> Optional[PackitAPI]:
+        if self._packit_api is None:
+            self._packit_api = PackitAPI(
+                self.service_config,
+                self.job_config,
+                downstream_local_project=self.local_project,
+            )
+
+        return self._packit_api
+
+    @property
+    def local_project(self) -> LocalProject:
+        if self._local_project is None:
+            self._local_project = LocalProject(
+                git_project=self.project,
+                working_dir=self.service_config.command_handler_work_dir,
+                cache=RepositoryCache(
+                    cache_path=self.service_config.repository_cache,
+                    add_new=self.service_config.add_repositories_to_repository_cache,
+                )
+                if self.service_config.repository_cache
+                else None,
+            )
+
+        return self._local_project
+
+    def _write_access_to_dist_git_repo_is_not_needed_or_satisfied(self) -> bool:
+        if self.data.event_type != PullRequestCommentPagureEvent.__name__:
+            # not needed
+            return True
+
+        user = self.data.actor
+        logger.debug(
+            f"Bodhi update will be re-triggered via dist-git PR comment by user {user}."
+        )
+
+        if not self.project.has_write_access(user=user):
+            logger.info(
+                f"Re-triggering Bodhi update via dist-git comment in PR#{self.data.pr_id}"
+                f" and project {self.project.repo} is not allowed for the user: {user}."
+            )
+            return False
+
+        return True
+
     def pre_check(self) -> bool:
         """
         We react only on finished builds (=KojiBuildState.complete)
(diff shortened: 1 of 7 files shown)


lbarcziova:
Thanks for finding time to implement this! I will have a closer look at the PR later. But as for the question:

> also help needed :grin: test_bodhi_update_retrigger_via_dist_git_pr_comment won't pass because because the CreateBodhiUpdateHandler.run method is not called. Even though the task.bodhi_update task gets the signature of the Celery task. This means that the JobHandler.get_signature method is executed and according to its documentation the run_bodhi_update task should be run, but this obviously does not happen. Any idea why?

Since there is no real Celery in the tests, we call some methods which would Celery do  => call the `run_..` ourselves :D yes, there is definitely a place for improvement of the tests, we already discussed splitting the integration tests to not test the whole use case at once. You can see how this should be done e.g. [here](https://github.com/packit/packit-service/blob/main/tests/integration/test_bodhi_update.py#L89-L103).

majamassarini (commented):
I tried your changes locally with `podman-compose`.
I have this error and maybe it could be somehow related with the tests (but I am not sure)
```
worker              |   File "/usr/local/lib/python3.10/site-packages/packit_service/worker/handlers/bodhi.py", line 175, in pre_check
worker              |     if self.koji_build_event_data.state != KojiBuildState.complete:
worker              |   File "/usr/local/lib/python3.10/site-packages/packit_service/worker/handlers/bodhi.py", line 118, in koji_build_event_data
worker              |     self._koji_build_event_data = self._build_koji_event_data()
worker              |   File "/usr/local/lib/python3.10/site-packages/packit_service/worker/handlers/bodhi.py", line 111, in _build_koji_event_data
worker              |     return self._build_from_koji_api()
worker              |   File "/usr/local/lib/python3.10/site-packages/packit_service/worker/handlers/bodhi.py", line 96, in _build_from_koji_api
worker              |     nvr = DistGit.get_latest_build_in_tag(
worker              | TypeError: DistGit.get_latest_build_in_tag() missing 1 required positional argument: 'self'
```
I have not fixed it because I am not really sure that it could be helpful.
I may be wrong (and maybe I am missing something else) but I think that in the commit named `Feat(bodhi hangler): Get last koji build from koji api` we expect a `KojiBuildEvent` even if in this case we are reacting to a `PullRequestCommentPagureEvent` which has not the needed Koji data.
Should we take the last built `Koji` package from the db instead?

nikromen:
I'll take it in points:
- > I have this error and maybe it could be somehow related with the tests (but I am not sure)
  - well the reason why you get this error is that the `get_latest_build_in_tag` method should be static, because the `self` parameter is not used. Only I forgot to make a PR in the packit repository that removes the `self` parameter :D thanks for pointing to it
- > Should we take the last built Koji package from the db instead?
  - I'm not sure if I get your point. However, I tried to recreate the `KojiBuildEvent` (because I didn't know that koji has python API) using data from the database. I ended up adding a some new stuff to the database and code in general, and the result was a lot of extra code. On top of that, most of the tests were crashing :D
- > but I think that in the commit named Feat(bodhi hangler): Get last koji build from koji api we expect a KojiBuildEvent even if in this case we are reacting to a PullRequestCommentPagureEvent which has not the needed Koji data
  - I don't think that `KojiBuildEvent` is relevant in this context. We only need to extract a minimal amount of information from it, so I have created a wrapper that contains the minimal amount of information we need to re-trigger a bodhi update. Basically, the user wants to re-trigger the bodhi update and we are not interested in `KojiBuildEvent` (besides the above mentioned amount of information to re-trigger the bodhi update). Compared to creating a whole `KojiBuildEvent`, I don't think it's worth it in the current situation. (Although it's possible that the koji API has tools to get all the information (for example `ClientSession().getBuild()`). However, I haven't read that much documentation and also `getBuild` method doesn't return everything (I think).
- > I tried your changes locally with podman-compose.
  - Wow that's cool, I could use it at least 50 times already... How? :D I didn't know we are able to run packit service locally. Do we have it documented somewhere? How you do it? :D

nikromen:
> I think you don't need to do that; [here](https://github.com/packit/packit/blob/41a74aefaa6a7d07f2559513c0d454d082b1efdd/packit/api.py#L1206) you can see that the list of Koji builds is optional.

Yes, I know that if I'll set the `koji_builds` parameter to `None` it gets the latest build. However the problem is that we need some data for logging, creating error messages and generally in the `pre-check` method. So this means we still have to get the data somehow.

> then the latest Koji builds are searched in the same way you are doing

Unfortunately yes. But at least I am trying to use at least `get_latest_build_in_tag` method from packit. As I described above - we need this data not only for `create_update` but for logs, etc...

But if you have an idea how to get around it without searching for the data in the way I do it then I am open to suggestions :innocent:

packit_service/worker/handlers/bodhi.py
```diff
@@ -63,15 +78,92 @@ def __init__(
...
+        )
+
+    def _build_from_koji_api(self) -> KojiBuildEventWrapper:
+        dist_git_branch = self.project.get_pr(self.data.pr_id).target_branch
+        nvr = DistGit.get_latest_build_in_tag(
+            downstream_package_name=self.project.repo, dist_git_branch=dist_git_branch
+        )
+
+        session = ClientSession(baseurl=KOJI_BASEURL)
+        build = session.getBuild(buildInfo=nvr)
```
lbarcziova, line 101: I think that we could refactor the packit `get_latest_build_in_tag` method to return the whole build dictionary and not only nvr [here](https://github.com/packit/packit/blob/0d8c2b41e22e4477dfa7550723df0ab0cfd0a38b/packit/distgit.py#L449) and then we would not need to call the Koji API here the second time

packit_service/worker/handlers/bodhi.py
```diff
@@ -63,15 +78,92 @@ def __init__(
...
+        if self.data.event_type != PullRequestCommentPagureEvent.__name__:
+            # not needed
+            return True
+
+        user = self.data.actor
+        logger.debug(
+            f"Bodhi update will be re-triggered via dist-git PR comment by user {user}."
+        )
+
+        if not self.project.has_write_access(user=user):
```
lbarcziova, line 159: we just discussed with Franta and Tomas that here we should check both whether the user is a packager and has write access to the repo, so here we need to call the `is_packager`(https://github.com/packit/packit-service/blob/85b02fc86ea7a3591cbce3306a807eff261da616/packit_service/worker/handlers/distgit.py#L443)

lbarcziova (commented):
thanks for the implementation! will you be able to finish this (I only added 2 notes) or should we help you somehow? Also, we do not have any reporting in case pre-check fails (e.g.when the person who commented doesn't have permissions), I am wondering whether we should solve this here or do it together for retriggering Koji builds by comment where this is missing as well

lachmanfrantisek:
> will you be able to finish this (I only added 2 notes) or should we help you somehow?

We have discussed with Jirka that we will probably meet next Thursday to finish this. (And that we can help if needed.)

TomasTomecek:
> I am wondering whether we should solve this here or do it together for retriggering Koji builds by comment where this is missing as well

I would open a followup to that :) so we can get this merged finally

PR #1656, opened 2022-09-15
Deprecate production_build in favour of upstream_koji_build

lachmanfrantisek:
Deprecate also the `build` alias since it's too generic.

Fixes: packit/packit#1658
Requires: packit/packit#1728
Documentation: packit/packit.dev#527

---

RELEASE NOTES BEGIN
There are two changes in the naming of the service jobs:
The `build` job type name has been deprecated. It aimed to be an alias when Packit supported just one build type.
There are currently more types of builds and just `build` can be misleading. Please, be explicit and use `copr_build` instead.
The `production_build` name for upstream Koji build is misleading because it is not used to run production/non-scratch builds and because it can be confused with the `koji_build` job that is triggered for dist-git commits.
(The `koji_build` job can trigger both scratch and non-scratch/production builds.)
To be explicit, use `upstream_koji_build` for builds triggered in upstream and `koji_build` for builds triggered in downstream.
Users will get a neutral status describing the change when the old names are in use. The status will become a warning starting in November and the old names will be removed by the end of the year.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/jobs.py
+++ b/packit_service/worker/jobs.py
@@ -13,8 +13,10 @@ import celery
 
 from ogr.exceptions import GithubAppNotInstalledError
 from packit.config import JobConfig, JobType, JobConfigTriggerType
+from packit.config.job_config import DEPRECATED_JOB_TYPES
 from packit_service.config import ServiceConfig
 from packit_service.constants import (
+    DOCS_CONFIGURATION_URL,
     TASK_ACCEPTED,
     COMMENT_REACTION,
     PACKIT_VERIFY_FAS_COMMAND,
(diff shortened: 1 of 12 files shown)


packit_service/worker/jobs.py
```diff
@@ -427,6 +428,19 @@ def should_task_be_created_for_job_config_and_handler(
             # e.g. We don't allow using internal TF for external contributors.
             return False
 
+        if deprecation_msg := DEPRECATED_JOB_TYPES.get(job_config.type):
+            job_helper = self.initialize_job_helper(handler, job_config)
+            job_helper.status_reporter.report(
+                state=BaseCommitStatus.neutral,  # TODO: change to warning after one month
```
TomasTomecek, line 434:
```suggestion
                state=BaseCommitStatus.neutral,  # TODO: change to warning in Nov 2022
```

TomasTomecek (approved):

lbarcziova (approved):
nice! would it make sense to write somewhere down how to do the deprecation (the neutral and error statuses) so that we behave consistently in the future? (although, we also want to work on packit/research#159, so I am not sure how relevant it will be)

lachmanfrantisek:
> nice! would it make sense to write somewhere down how to do the deprecation (the neutral and error statuses) so that we behave consistently in the future? (although, we also want to work on [packit/research#159](https://github.com/packit/research/issues/159), so I am not sure how relevant it will be)

Ideally, I would go with the automated fixed next time we need to change anything.

lachmanfrantisek:
BTW, I've fixed the tests, atest casew test-case covering the deprecation and restructured the changes across commits to make it more clear... @TomasTomecek @lbarcziova Could you please take another look?

TomasTomecek (approved):
very nice work with the commit split

PR #1658, opened 2022-09-15
Implement triggering TF with build from other PR

### LINKED ISSUE #1544: Triggering testing farm tests with artifacts from another project
leapp requires 2 copr builds to be passed to testing farm - one of [leapp](https://github.com/oamg/leapp) and another of [leapp-repository](https://github.com/oamg/leapp-repository) .
In simple cases you can get away with the-copr-build-from-the-pr and master version of another project, but for certain cases it is essential that we use the-copr-build-from-the-user-specified-pr-of-that-other-project instead of master.

For example:
A '/rerun 4242' comment is issued by a developer under some upstream leapp PR. Our [current copr copr-build workflow](https://github.com/oamg/leapp/blob/master/.github/workflows/reuse-copr-build.yml#L31) will create both leapp PR's copr build, leapp-repository\*PR4242\* copr build, and pass both artifacts to the [tmt-tests workflows](https://github.com/oamg/leapp/blob/master/.github/workflows/tmt-tests.yml#L18) as copr_artifacts.

We would love to have the same mechanism with packit as well.

lbarcziova:
TODO:
- [x] Polish the code.
- [x] Fix openshift tests.
- [x] Improve doc-strings.
- [x] Update or write new documentation in `packit/packit.dev` (packit/packit.dev#528).
- [ ] Update view of the TF results in dashboard (I will open a dashboard PR once this is in staging)

Fixes #1544

Since this is a bigger change, maybe we can try the pair-review if anyone would be interested :D

---

RELEASE NOTES BEGIN
We have added support for running the tests with Copr builds built by Packit in another pull request (in a different repository). You can read more about this feature in [our documentation](https://packit.dev/docs/testing-farm/#running-tests-with-builds-from-another-pull-request).
RELEASE NOTES END

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -547,6 +547,20 @@ class PullRequestModel(BuildsAndTestsConnector, Base):
             return pr
 
     @classmethod
+    def get(
+        cls, pr_id: int, namespace: str, repo_name: str, project_url: str
+    ) -> Optional["PullRequestModel"]:
+        with sa_session_transaction() as session:
+            project = GitProjectModel.get_or_create(
+                namespace=namespace, repo_name=repo_name, project_url=project_url
+            )
+            return (
+                session.query(PullRequestModel)
+                .filter_by(pr_id=pr_id, project_id=project.id)
+                .first()
+            )
+
+    @classmethod
     def get_by_id(cls, id_: int) -> Optional["PullRequestModel"]:
         return sa_session().query(PullRequestModel).filter_by(id=id_).first()
 
@@ -685,7 +699,6 @@ AbstractTriggerDbType = Union[
     GitBranchModel,
     IssueModel,
 ]
-
 
 MODEL_FOR_TRIGGER: Dict[JobTriggerModelType, Type[AbstractTriggerDbType]] = {
     JobTriggerModelType.pull_request: PullRequestModel,
@@ -1564,7 +1577,7 @@ class TFTTestRunTargetModel(ProjectAndTriggersConnector, Base):
         commit_sha: str,
         status: TestingFarmResult,
         target: str,
-        run_model: "PipelineModel",
+        run_models: List["PipelineModel"],
         web_url: Optional[str] = None,
         data: dict = None,
         identifier: Optional[str] = None,
@@ -1580,19 +1593,20 @@ class TFTTestRunTargetModel(ProjectAndTriggersConnector, Base):
             test_run.data = data
             session.add(test_run)
 
-            if run_model.test_run:
-                # Clone run model
-                new_run_model = PipelineModel.create(
-                    type=run_model.job_trigger.type,
-                    trigger_id=run_model.job_trigger.trigger_id,
-                )
-                new_run_model.srpm_build = run_model.srpm_build
-                new_run_model.copr_build = run_model.copr_build
-                new_run_model.test_run = test_run
-                session.add(new_run_model)
-            else:
-                run_model.test_run = test_run
-                session.add(run_model)
+            for run_model in run_models:
+                if run_model.test_run:
+                    # Clone run model
+                    new_run_model = PipelineModel.create(
+                        type=run_model.job_trigger.type,
+                        trigger_id=run_model.job_trigger.trigger_id,
+                    )
+                    new_run_model.srpm_build = run_model.srpm_build
+                    new_run_model.copr_build = run_model.copr_build
+                    new_run_model.test_run = test_run
+                    session.add(new_run_model)
+                else:
+                    run_model.test_run = test_run
+                    session.add(run_model)
 
             return test_run
 
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -20,6 +20,8 @@ from typing import (
     Tuple,
     Type,
     Union,
+    Set,
+    overload,
 )
 from urllib.parse import urlparse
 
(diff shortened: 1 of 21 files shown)


packit_service/worker/helpers/testing_farm.py
```diff
@@ -495,29 +618,61 @@ def run_testing_farm(
...
             )
+
+        additional_build = None
+        if self.copr_builds_from_other_pr and not (
+            additional_build := self.copr_builds_from_other_pr.get(chroot)
+        ):
+            self.report_status_to_test_for_test_target(
+                state=BaseCommitStatus.failure,
+                description="No latest successful Copr build from the other PR found.",
+                target=target,
```
TomasTomecek, line 629: what would be the target name for these special builds?

TomasTomecek (commented):
Very nice work with the refactoring!

I also like how straightforward this is.

packit_service/worker/helpers/testing_farm.py
lbarcziova: for now, I haven't changed the check's names so they would not reflect that the tests are run with the builds from other PR...but I was also thinking about including the pr argument somehow, but I am not sure whether to go that direction from the beginning...WDYT?

TomasTomecek: I am sure there will be a discussion (and issues created) about the naming once this feature is live.

I'd go for keeping the name equivalent, as you implemented it, and treat it as a future RFE if people want it configurable or have a different default.

Would that work for you?

lbarcziova: agreed :+1:

packit_service/models.py
```diff
@@ -133,6 +134,80 @@ def optional_timestamp(datetime_object: Optional[datetime]) -> Optional[int]:
...
+) -> List[Union["CoprBuildTargetModel", "TFTTestRunTargetModel"]]:
+    """
+    Gets most recent models from an iterable (regarding submission time).
+
+    Args:
+        models: Copr or TF models - if there are any duplicates in them then use the most
+         recent model
+
+    Returns:
+        Dictionary - target as a key and corresponding most recent model as a value.
```
mfocko, line 162: Returns list, not a dictionary.

packit_service/models.py
```diff
@@ -133,6 +134,80 @@ def optional_timestamp(datetime_object: Optional[datetime]) -> Optional[int]:
...
+    return list(most_recent_models.values())
+
+
+def filter_most_recent_target_models_by_status(
+    models: Union[
+        Iterable["CoprBuildTargetModel"],
+        Iterable["TFTTestRunTargetModel"],
+    ],
+    statuses_to_filter_with: List[str],
+) -> Optional[Set[Union["CoprBuildTargetModel", "TFTTestRunTargetModel"]]]:
```
mfocko, line 185: Given the input, I guess you meant

```suggestion
) -> Optional[Union[Set["CoprBuildTargetModel"], Set["TFTTestRunTargetModel"]]]:
```

packit_service/models.py
```diff
@@ -133,6 +134,80 @@ def optional_timestamp(datetime_object: Optional[datetime]) -> Optional[int]:
...
+    logger.info(
+        f"Trying to filter targets with possible status: {statuses_to_filter_with} in {models}"
+    )
+    filtered_target_models = set()
+    for model in get_most_recent_targets(models):
+        if model.status in statuses_to_filter_with:
+            filtered_target_models.add(model)
```
mfocko, line 192:
```suggestion
filtered_target_models = set(filter(
    lambda model: model.status in statuses_to_filter_with,
    get_most_recent_targets(models),
))
```

Doesn't matter that much though :)

packit_service/worker/helpers/testing_farm.py
```diff
@@ -220,6 +297,17 @@ def _payload(
         else:
             build_log_url = nvr = srpm_url = None
 
+        packit_copr_rpms = (
+            [
+                package
+                for artifact in artifacts
+                if artifact.get("packages")
```
mfocko, line 301: I'd be careful, cause this doesn't necessarily mean they are Copr builds. You can also have Koji or URL to RPM repo.

```suggestion
                if artifact.get("type") == "fedora-copr-build" and artifact.get("packages")
```

_Edit_: not valid anymore, since only they get generated only on our side

packit_service/worker/helpers/testing_farm.py
```diff
@@ -320,16 +408,28 @@ def _payload_install_test(self, build_id: int, target: str, compose: str) -> dic
...
+                description="We were not able to get any Copr builds for given additional PR. "
+                "Please, make sure the comment command is in correct format "
+                "`/packit test repo/namespace#pr_id`",
+                state=BaseCommitStatus.error,
+                url="",
+            )
+            return False
 
-    def is_fmf_configured(self) -> bool:
+        return True
```
mfocko, line 423:
```suggestion
        if self.copr_builds_from_other_pr:
            return True

        self.report_status_to_tests(
            description="We were not able to get any Copr builds for given additional PR. "
            "Please, make sure the comment command is in correct format "
            "`/packit test repo/namespace#pr_id`",
            state=BaseCommitStatus.error,
            url="",
        )
        return False
```

↑ This diff looks very weird :eyes:

packit_service/worker/helpers/testing_farm.py
```diff
@@ -463,6 +563,33 @@ def get_latest_copr_build(
...
+        self,
+        chroot: str,
+        build: CoprBuildTargetModel,
+        additional_build: Optional[CoprBuildTargetModel],
+    ) -> List[Dict]:
+        """
+        Get the artifacts list from the build (if the skip_build option is not defined)
+        and additional build (from other PR) if present.
+        """
+        artifacts = []
```
mfocko, line 573: ~~What happens to predefined artifacts?~~ Never mind, I thought we added them with some of the PRs for Leapp

packit_service/worker/helpers/testing_farm.py
```diff
@@ -539,103 +698,28 @@ def run_testing_farm(
...
         )
 
-        if not req:
-            msg = "Failed to post request to testing farm API."
-            if not self.celery_task.is_last_try():
-                return self.retry_on_submit_failure(msg)
+        if not response:
+            return self._handle_tf_submit_no_response(target=target, payload=payload)
 
-            logger.error(f"{msg} {self._payload_without_token(payload)}")
```
mfocko, line 554: oh my… :eyes:

mfocko (approved):
pls tell me I didn't spend 30 minutes on that second commit /o\

packit_service/models.py
lbarcziova: I haven't written this code, but yes, the types were wrong and not only here...I tried to correct it (1. commit), but I don't know how to explain some things to mypy (line 226 in models.py :D)..I would be glad for any advice

packit_service/worker/helpers/testing_farm.py
```diff
@@ -165,6 +178,59 @@ def _pr(self) -> Optional[PullRequest]:
...
+            '/packit test namespace/repo#pr' -> ["test", "namespace/repo#pr"]
+        """
+        if not self._comment_command_parts and (
+            comment := self.metadata.event_dict.get("comment")
+        ):
+            self._comment_command_parts = get_packit_commands_from_comment(
+                comment,
+                packit_comment_command_prefix=self.service_config.comment_command_prefix,
+            )
+        return self._comment_command_parts
```
r0x0d, line 197: I know it's a bite late for this comment (and probably I'm not the most qualified person to suggest this), but this method looks a bit strange, specially if because there is a mutation happening to the `self._comment_command_parts` class-property, and we are returning the value we just assigned to a class-property that can be accessed later. 

Ideally, I think it would be more readable and concise if this method doesn't return any result in it, but instead, override the value if necessary of the class-property and that's it. 

My suggestion for this one would be removing the return statement at the end of the method and leaving the mutation where it is right now, especially because the class initializes the property as `None`. 

```suggestion
    def comment_command_parts(self) -> Optional[List[str]]:
        """
        List of packit comment command parts if the testing farm was triggered by a comment.
        
        Note: 
            If the `self._comment_command_parts is not None, we will let the value as it is for now. 
            This method will only override the contents of the class-property if the same is `None`, and 
            we detect a comment from the `self.metadata` 
            
        Example:
            '/packit test' -> ["test"]
            '/packit test namespace/repo#pr' -> ["test", "namespace/repo#pr"]
        """
        if not self._comment_command_parts and (
            comment := self.metadata.event_dict.get("comment")
        ):
            self._comment_command_parts = get_packit_commands_from_comment(
                comment,
                packit_comment_command_prefix=self.service_config.comment_command_prefix,
            )
```

If you choose that, then you necessarily don't need to transform this method into a property, but rather, leave it as a method. 

If you would still prefer to have it as a property because that would trigger a lot of changes in other parts of the code, then, I think something like this could also be good: 

```suggestion
    def comment_command_parts(self) -> Optional[List[str]]:
        """
        List of packit comment command parts if the testing farm was triggered by a comment.
        
        Note: 
            If the `self._comment_command_parts is not None, we will let the value as it is for now. 
            This method will only override the contents of the class-property if the same is `None`, and 
            we detect a comment from the `self.metadata` 
            
        Example:
            '/packit test' -> ["test"]
            '/packit test namespace/repo#pr' -> ["test", "namespace/repo#pr"]
        """
        if comment := self.metadata.event_dict.get("comment"):
            return get_packit_commands_from_comment(
                comment,
                packit_comment_command_prefix=self.service_config.comment_command_prefix,
            )
       else: 
           return None
```

I'm recommending this especially because I can see that `self._comment_command_parts: Optional[List[str]] = None` is not used anywhere else other than here, so, probably, it would be easier to not have this class-property and have it either the method or the method-property instead. 

But of course, again, I may not be the best person to be proposing this right now, especially because the PR is already merged. Feel free to just close this comment if you think it's not worth it.

P.S: In my opinion, I lean towards the second suggestion.

PR #1669, opened 2022-09-20
Improve testing farm reporting

FrNecas:
This fixes an issue that I recently ran across, which occurs for example when a copr build takes too long and a user manually runs `/packit test`. In such case, packit would report an error on the commit flag since the build has not finished which is IMHO confusing. This should hopefully improve this UX.

As a part of this PR, I've slightly refactored the status to use an enum since it was used all over the place.

TODO:

- [x] Figure out if a migration for str -> Enum[str] is needed (I haven't found anything in alembic docs yet, may need to search a bit more). Openshift tests will probably fail if this is necessary.

Fixes

Related to

Merge before/after

---

RELEASE NOTES BEGIN
Packit now reports a pending state rather than an error on Testing Farm runs in case the related copr build has not finished yet.
RELEASE NOTES END


FrNecas:
Looks like it works without a migration. The failing test is caused by the fact that there is a test setting status to `"awesome"` which is not a member of the enum. Do you see any reason for keeping statuses as a free string? It does not make much sense in my opinion, enum is much more clear but I'd appreciate any other opinions on this matter. If we agree to use an enum, I will just drop the test setting status to `"awesome"`

lbarcziova (approved):
nice improvement!

I think we are good to go with the enum

jpopelka:
> Looks like it works without a migration.

I myself wouldn't bet on that:
* https://stackoverflow.com/questions/50338814/postgres-update-column-with-enum-from-string
* https://stackoverflow.com/questions/15655820/upgrading-a-varchar-column-to-enum-type-in-postgresql

FrNecas:
> I myself wouldn't bet on that

Thanks for the links. I will try creating a migration which adds the enum and casts the type of the column, hopefully that will work.

jpopelka (approved):

FrNecas:
Hopefully my latest commit fixes the failing tests. Enum does not seem to play well with `array_agg` (not sure why) but `json_agg` seems to handle it (or at least it does locally) - https://stackoverflow.com/questions/58445632/applying-array-agg-on-enum-column-returns-is-not-among-the-defined-enum-val

PR #1671, opened 2022-09-21
Move the check for event and job identifiers equality

### LINKED ISSUE #1625: Identifier ignored for re-run?
Hi,
in tmt we have 3 fedora-36 jobs. 
One is from target: fedora-all, other two from target: fedora-latest-stable. To have them distinguishable the later contain  `identifier: XXX`.

From reporting PoV all is good - I can see which job is which. However if I click 're-run' next to `testing-farm:fedora-36-x86_64:internal-wow` all 3 fedora-36 jobs are re-run.

How can one reschedule only selected one?

lbarcziova:
Move the check whether identifier of the event we are reacting to
 is equal to the identifier of the currently processed job
to get_jobs_matching_event method. Previously, the check was in the get_handlers_for_event method  which calls the get_jobs_matching_event. This was not enough, because get_jobs_matching_event is being called 2 times, once more after running get_handlers_for_event and therefore handler was triggered with more job configs than correct.

Fixes #1625

---

RELEASE NOTES BEGIN
Retriggering tasks via re-run button in Github commit checks when there are configured identifiers for jobs should now work correctly.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/jobs.py
+++ b/packit_service/worker/jobs.py
@@ -496,6 +496,10 @@ class SteveJobs:
         for job in self.event.package_config.jobs:
             if (
                 job.trigger == self.event.job_config_trigger_type
+                and (
+                    not isinstance(self.event, CheckRerunEvent)
+                    or self.event.job_identifier == job.identifier
+                )
                 and job not in jobs_matching_trigger
             ):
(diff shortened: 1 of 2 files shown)


jpopelka (approved):
:rocket:

PR #1672, opened 2022-09-22
Fix instructions for using live data

FrNecas:
Just some improvements to our contribution docs, binary format does not seem to work (it probably breaks the format somewhere in the process of fetching and output redirection). + dropping sudo from podman command since we recommend rootless podman in the same document. I was able to get API with stage data running locally following this.


jpopelka (approved):
Thanks!

PR #1673, opened 2022-09-22
Deduce copr targets

### LINKED ISSUE #1499: Discover targets when building in users Copr repo

mfocko:
TODO:

Fixes #1499

---

RELEASE NOTES BEGIN
Packit now deduces Copr targets for Copr builds when you have set your custom Copr project to be used.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/helpers/build/build_helper.py
+++ b/packit_service/worker/helpers/build/build_helper.py
@@ -83,6 +83,14 @@ class BaseBuildJobHelper(BaseJobHelper):
         if self.job_tests and not self.job_tests.skip_build:
             targets.update(self.job_tests.targets)
 
+        if (
+            self.job_type_build == JobType.copr_build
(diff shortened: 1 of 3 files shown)


TomasTomecek (approved):
👏🏻

PR #1675, opened 2022-09-23
Tf composes

### LINKED ISSUE #1653: Improve getting of the Testing Farm composes
When getting a TF compose, we sanitize the target in [distro2compose method](https://github.com/packit/packit-service/blob/dad5a6d1c5eea1feb15a95652804cfc84083c0ca/packit_service/worker/helpers/testing_farm.py#L336). This breaks some usecases, e.g. having `RHEL-7.8-ZStream`  in the configuration results in `RHEL-7.8-Zstream` compose  (because of `title()`)  that is not in the list of available composes and therefore the test run fails ([example](https://github.com/r0x0d/leapp-packit-poc/pull/3/checks?check_run_id=8308160405)).

We could firstly check the raw distro (the target without arch) and if it is in the list of available composes, don't sanitize it. Or we could try to improve the sanitization itself, but we need to make sure we don't break the other usecases.

mfocko:
TODO:

- [X] Update or write new documentation in `packit/packit.dev`.

Fixes #1653

---

RELEASE NOTES BEGIN
We have improved mapping of Testing Farm Composes, if you have set your own custom mapping and the TF Compose is available, it will be used as is without any additional modifications we do (version, etc.).
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/helpers/testing_farm.py
+++ b/packit_service/worker/helpers/testing_farm.py
@@ -231,6 +231,25 @@ class TestingFarmJobHelper(CoprBuildJobHelper):
             self._copr_builds_from_other_pr = self.get_copr_builds_from_other_pr()
         return self._copr_builds_from_other_pr
 
+    @property
+    def available_composes(self) -> Optional[Set[str]]:
+        """
+        Fetches available composes from the Testing Farm endpoint.
+
+        Returns:
+            Set of all available composes or `None` if error occurs.
+        """
+        endpoint = (
+            f"composes/{'redhat' if self.job_config.use_internal_tf else 'public' }"
+        )
+
+        response = self.send_testing_farm_request(endpoint=endpoint)
+        if response.status_code != 200:
+            return None
+
+        # {'composes': [{'name': 'CentOS-Stream-8'}, {'name': 'Fedora-Rawhide'}]}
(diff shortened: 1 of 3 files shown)


packit_service/worker/helpers/testing_farm.py
```diff
@@ -354,6 +373,20 @@ def distro2compose(self, target: str) -> Optional[str]:
             compose if we were able to map the distro to compose present
             in the list of available composes, otherwise None
         """
+        composes = self.available_composes
+        if composes is None:
+            msg = "We were not able to get the available TF composes."
+            logger.error(msg)
+            self.report_status_to_test_for_test_target(
+                state=BaseCommitStatus.error,
```
lachmanfrantisek, line 381: :+1: Errors will make it possible to retrigger via the check-run button.

lachmanfrantisek (approved):
Thanks!

lachmanfrantisek:
> TODO? do we need it in the release notes?

I think we can mention that if you pick anything from the list, we will not touch it, otherwise there is a mapping happening (= explain the behaviour and let users know that we've fixed a bug here).

lbarcziova (approved):
:+1:

mfocko:
regate

PR #1678, opened 2022-09-29
Move job handler's pre-checks in a new class

### LINKED ISSUE #1581: Refactor Handlers for `None`-correctness
Follow up of the discussion with @lbarcziova in https://github.com/packit/packit-service/pull/1579#discussion_r921034064

Currently, handlers are instantiated in 2 times, once in `Steve` and then in the celery task corresponding to the handler. However, the use-cases for these two initializations are different:
 - `Steve` just uses `pre_check()` and perhaps some other stuff but doesn't run it
 - celery task runs the handler

This causes problems with regards to `None` correctness as explained in the comment linked above. Ideally, a task that can be retried should not have its `CeleryTask` argument set to `None` (without the task attached, retrying is not possible and does not make sense). However, the initialization inside `Steve` does not have a celery task yet to attach, hence it must be set to `None` (a bit of a chicken-egg problem).

A solution that I assume could work (however, feel free to come up with anything else):
 - split each handler (including the abstract classes) into 2 parts (naming is just a suggestion):
    - BaseHandler - implements just the `pre_check` and similar stuff
    - RunnableHandler - also implements `run` and inherits from `BaseHandler`
 - inside `Steve` initialize just the BaseHandler which does not need a `CeleryTask` even for a retriable job handler.
 - inside tasks, initialize RunnableHandler which in the case of retriable handlers will have non-Optional `CeleryTask` argument

This is a blocker for #1433 (the implementation from #1579 would not pass mypy's strict checks -- however it does not make sense to block it on this rather large refactoring)

majamassarini:
Job handlers should be retriable and may have a CeleryTask reference.
If a job handler is instantiated just for calling the pre-check method
then it can not be given a CeleryTask reference.

The pre_check method is now a classmethod for any job handler.
The job handler's pre_check method will invoke the pre_check method of
a list of new checker classes.
A checker class needs all the parameters needed by a job handler class
but it does not need the CeleryTask reference.

All the common code between checkers and handlers has been moved
in mixin classes.

Fixes #1581

### DIFF
--- a/packit_service/worker/checker/abstract.py
+++ b/packit_service/worker/checker/abstract.py
@@ -34,7 +34,6 @@ class ActorChecker(Checker):
         ...
 
     def pre_check(self) -> bool:
-        if self.actor:
-            return self._pre_check()
-        else:
+        if not self.actor:
             return False
+        return self._pre_check()
--- a/packit_service/worker/checker/abstract.py
+++ b/packit_service/worker/checker/abstract.py
@@ -1,0 +1,40 @@
+from abc import abstractmethod
+from typing import Optional
+
+from packit.config import JobConfig
+from packit.config.package_config import PackageConfig
+
+from packit_service.worker.events import EventData
+from packit_service.worker.mixin import ConfigMixin, PackitAPIWithDownstreamMixin
+
+
+class Checker(ConfigMixin, PackitAPIWithDownstreamMixin):
+    def __init__(
+        self,
+        package_config: PackageConfig,
+        job_config: JobConfig,
+        event: dict,
+    ):
+        self.package_config = package_config
+        self.job_config = job_config
+        self.data = EventData.from_event_dict(event)
+
+    @abstractmethod
+    def pre_check(self) -> bool:
+        ...
+
+
+class ActorChecker(Checker):
+    @property
+    def actor(self) -> Optional[str]:
+        return self.data.actor
+
+    @abstractmethod
+    def _pre_check(self) -> bool:
+        ...
+
+    def pre_check(self) -> bool:
+        if self.actor:
+            return self._pre_check()
+        else:
+            return False
--- a/packit_service/worker/checker/bodhi.py
+++ b/packit_service/worker/checker/bodhi.py
@@ -13,7 +13,7 @@ from packit_service.worker.handlers.mixin import GetKojiBuildEventMixin
 logger = logging.getLogger(__name__)
 
 
-class IsKojiBuildComplete(Checker, GetKojiBuildEventMixin):
+class IsKojiBuildCompleteAndBranchConfigured(Checker, GetKojiBuildEventMixin):
     def pre_check(self) -> bool:
         """Check if builds are finished (=KojiBuildState.complete)
         and branches are configured.
--- a/packit_service/worker/checker/bodhi.py
+++ b/packit_service/worker/checker/bodhi.py
@@ -1,0 +1,42 @@
+# Copyright Contributors to the Packit project.
+# SPDX-License-Identifier: MIT
+
+import logging
+
+from packit.config.aliases import get_branches
+
+from packit_service.constants import KojiBuildState
+
+from packit_service.worker.checker.abstract import Checker
+from packit_service.worker.handlers.mixin import GetKojiBuildEventMixin
+
+logger = logging.getLogger(__name__)
+
+
+class IsKojiBuildComplete(Checker, GetKojiBuildEventMixin):
+    def pre_check(self) -> bool:
+        """Check if builds are finished (=KojiBuildState.complete)
+        and branches are configured.
+        By default, we use `fedora-stable` alias.
+        (Rawhide updates are already created automatically.)
+        """
+        if self.koji_build_event.state != KojiBuildState.complete:
+            logger.debug(
+                f"Skipping build '{self.koji_build_event.build_id}' "
+                f"on '{self.koji_build_event.git_ref}'. "
+                f"Build not finished yet."
+            )
+            return False
+
+        if self.koji_build_event.git_ref not in (
+            configured_branches := get_branches(
+                *(self.job_config.dist_git_branches or {"fedora-stable"}),
+                default_dg_branch="rawhide",  # Koji calls it rawhide, not main
+            )
+        ):
+            logger.info(
+                f"Skipping build on '{self.data.git_ref}'. "
+                f"Bodhi update configured only for '{configured_branches}'."
+            )
+            return False
+        return True
--- a/packit_service/worker/checker/copr.py
+++ b/packit_service/worker/checker/copr.py
@@ -80,8 +80,7 @@ class CoprBuildPermission(Checker, GetCoprBuildJobHelperForIdMixin):
 
 
 class CanActorRunJob(ActorChecker, GetCoprBuildJobHelperMixin):
-    """
-    For external contributors, we need to be more careful when running jobs.
+    """For external contributors, we need to be more careful when running jobs.
     This is a handler-specific permission check
     for a user who trigger the action on a PR.
     """
--- a/packit_service/worker/checker/copr.py
+++ b/packit_service/worker/checker/copr.py
@@ -1,0 +1,106 @@
+# Copyright Contributors to the Packit project.
+# SPDX-License-Identifier: MIT
+
+import logging
+
+from packit_service.worker.checker.abstract import ActorChecker, Checker
+from packit_service.worker.events.enums import GitlabEventAction
+from packit_service.worker.events import (
+    MergeRequestGitlabEvent,
+    PushGitHubEvent,
+    PushGitlabEvent,
+    PushPagureEvent,
+)
+from packit_service.worker.handlers.mixin import (
+    GetCoprBuildJobHelperForIdMixin,
+    GetCoprBuildJobHelperMixin,
+)
+from packit_service.constants import (
+    INTERNAL_TF_BUILDS_AND_TESTS_NOT_ALLOWED,
+)
+from packit_service.worker.reporting import BaseCommitStatus
+
+logger = logging.getLogger(__name__)
+
+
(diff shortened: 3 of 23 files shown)


FrNecas:
We will try pair review on this PR with Maja

lachmanfrantisek (commented):
Wow. I really like the way this is going! (Very clean and Pythonish... :snake: )

packit_service/worker/jobs.py
```diff
@@ -216,14 +216,14 @@ def initialize_job_helper(
             "job_config": job_config,
         }
 
-        if isinstance(handler, ProposeDownstreamHandler):
+        if handler_kls == ProposeDownstreamHandler:
```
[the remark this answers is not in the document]
majamassarini: I can see we are already using the equality operator instead of the identity check in different places like [here](https://github.com/packit/packit-service/blob/65c129fd94cf0ace80a19c9ea18e1f67f407bf9c/packit_service/worker/jobs.py#L620).
I think that rewriting this code using the identity check will make it less readable and I think that we have no reason to override the `__eq__` operator in this objects, so probably we should not fear it.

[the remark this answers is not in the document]
FrNecas: You are right, that `in` operator matches using `==`, so yes, you are right, let's keep it the way it is

packit_service/worker/checker/copr.py
```diff
@@ -0,0 +1,106 @@
...
+)
+from packit_service.constants import (
+    INTERNAL_TF_BUILDS_AND_TESTS_NOT_ALLOWED,
+)
+from packit_service.worker.reporting import BaseCommitStatus
+
+logger = logging.getLogger(__name__)
+
+
+class Permission(Checker, GetCoprBuildJobHelperMixin):
```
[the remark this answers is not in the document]
majamassarini: I like the idea of using namespaces like this `checker.copr.Permission` but we are not really doing this inside our code so to preserve our style I think you are right I should add something more to that names.

packit_service/worker/checker/bodhi.py
```diff
@@ -0,0 +1,42 @@
...
+logger = logging.getLogger(__name__)
+
+
+class IsKojiBuildComplete(Checker, GetKojiBuildEventMixin):
+    def pre_check(self) -> bool:
+        """Check if builds are finished (=KojiBuildState.complete)
+        and branches are configured.
+        By default, we use `fedora-stable` alias.
+        (Rawhide updates are already created automatically.)
+        """
```
lbarcziova, line 22: e.g.here the checker class name doesn't correspond to the docstring and to what the pre_check checks, could we either rename it or split it to check the completeness and whether the branch we are reacting to is configured?

packit_service/worker/checker/copr.py
```diff
@@ -0,0 +1,105 @@
...
+                "Custom Copr owner/project set. "
+                "Checking if this GitHub project can use this Copr project."
+            )
+            if not self.copr_build_helper.check_if_custom_copr_can_be_used_and_report():
+                return False
+
+        return True
+
+
+class CoprBuildPermission(Checker, GetCoprBuildJobHelperForIdMixin):
```
lbarcziova, line 66: could we be more explicit with the naming? If I only had a quick look, `PermissionOnCopr` and `CoprBuildPermission` are not clear to me

packit_service/worker/checker/copr.py
```diff
@@ -0,0 +1,105 @@
...
+
+        logger.debug(
+            f"The Copr project {self.copr_event.owner}/{self.copr_event.project_name} "
+            f"does not match the configuration "
+            f"({self.copr_build_helper.job_owner}/{self.copr_build_helper.job_project} expected)."
+        )
+        return False
+
+
+class CanActorRunJob(ActorChecker, GetCoprBuildJobHelperMixin):
```
lbarcziova, line 82: I would rename this so that it is clear this is related to internal tests

packit_service/worker/checker/koji.py
```diff
@@ -0,0 +1,65 @@
...
+    KOJI_PRODUCTION_BUILDS_ISSUE,
+    PERMISSIONS_ERROR_WRITE_OR_ADMIN,
+)
+from packit_service.worker.reporting import BaseCommitStatus
+from packit_service.worker.handlers.mixin import GetKojiBuildJobHelperMixin
+
+logger = logging.getLogger(__name__)
+
+
+class PermissionOnKoji(Checker, GetKojiBuildJobHelperMixin):
```
lbarcziova, line 25: I would say we check multiple things here: closed merge requests, whether we react to configured branch, the user who triggered the build and whether scratch build is configured: does it make sense to split this or at least rename since this is not only about permissions?

lbarcziova (commented):
very nice job with the abstraction!! only few comments about splitting/renaming the checkers

packit_service/worker/checker/abstract.py
```diff
@@ -0,0 +1,40 @@
...
+
+    @abstractmethod
+    def _pre_check(self) -> bool:
+        ...
+
+    def pre_check(self) -> bool:
+        if self.actor:
+            return self._pre_check()
+        else:
+            return False
```
mfocko, line 40:
```suggestion
        return self.actor and self._pre_check()
```

or with just early return
```suggestion
        if not self.actor:
            return False
        return self._pre_check()
```

If you prefer it the current way, I don't mind :)

packit_service/worker/checker/copr.py
```diff
@@ -0,0 +1,106 @@
...
+    This is a handler-specific permission check
+    for a user who trigger the action on a PR.
+    """
+
+    def _pre_check(self) -> bool:
+        test_job = self.copr_build_helper.job_tests
+        if (
+            test_job
+            and test_job.use_internal_tf
+            and not self.project.can_merge_pr(self.actor)
```
mfocko, line 93: I would probably add admins here, so it is easier for us to debug any issues.

```suggestion
            and not self.project.can_merge_pr(self.actor)
            and self.actor not in self.service_config.admins
```

packit_service/worker/checker/koji.py
```diff
@@ -0,0 +1,65 @@
...
+        ):
+            configured_branch = self.koji_build_helper.job_build_branch
+            if self.data.git_ref != configured_branch:
+                logger.info(
+                    f"Skipping build on '{self.data.git_ref}'. "
+                    f"Push configured only for '{configured_branch}'."
+                )
+                return False
+
+        if self.data.event_type == PullRequestGithubEvent.__name__:
```
mfocko, line 47: Are we missing `MergeRequestGitlabEvent`? :eyes: /o\

packit_service/worker/checker/testing_farm.py
```diff
@@ -0,0 +1,81 @@
...
+    This is a handler-specific permission check
+    for a user who trigger the action on a PR.
+
+    The job is not allowed for external contributors when using internal TF.
+    """
+
+    def _pre_check(self) -> bool:
+        if self.job_config.use_internal_tf and not self.project.can_merge_pr(
+            self.actor
+        ):
```
mfocko, line 66:
```suggestion
        ) and self.actor not in self.service_config.admins:
```
same here

packit_service/worker/handlers/abstract.py
```diff
@@ -357,14 +345,30 @@ def _clean_workplace(self):
             else:
                 shutil.rmtree(item)
 
-    def pre_check(self) -> bool:
-        """
-        Implement this method for those handlers, where you want to check if the properties are
-        correct. If this method returns False during runtime, execution of service code is skipped.
+    @staticmethod
+    def get_checkers() -> Tuple:
```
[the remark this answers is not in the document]
mfocko: I think you should be able to go with `Sequence[Type[Checker]]` too… And since this comment is 8 days old, I'm not sure if it was the best idea to go commit-by-commit :D

mfocko (commented):
very nice :eyes:

packit_service/worker/checker/bodhi.py
majamassarini: You are right, they need to be splitted. There is also some more code, in the handlers, that can be removed using the Mixins. I have not done it yet, because I think that I have already touched too much code (the one needed to remove the celery argument from __init__) and it will be difficult to fix any bug introduced with this changes. I think that more refactoring should follow but we should do it gradually. So if you agree, for now, I just adjust the names and then I will create some more cards to finish all the refactor.

packit_service/worker/checker/koji.py
majamassarini: I don't know, but I checked and the original code was like this one. For the moment I will leave it as it is, and if everything will work we can try to add it later. What do you think?

mfocko: Fine by me, I'll open a follow-up issue to track this and make sure we don't forget about it :)

mfocko: Created #1709

packit_service/worker/checker/forges.py
```diff
@@ -0,0 +1,33 @@
...
+
+import logging
+
+from packit_service.worker.checker.abstract import Checker
+from packit_service.worker.mixin import GetIssueMixin
+
+logger = logging.getLogger(__name__)
+
+
+class PermissionOnForge(Checker, GetIssueMixin):
```
FrNecas, line 12: Could we make this name a bit more specific? I feel like it's not really about permissions. Maybe something like `IsIssueInNotificationRepoChecker`? But that's just a suggestion

packit_service/worker/checker/koji.py
FrNecas: Isn't this comment still valid (I see it marked as resolved)?

FrNecas (approved):
Just a couple of last nitpicks (some separate comments, one comment in this review). Otherwise looks good to me - let's get out of this rebasing hell. We can improve things further later on. We will also probably need to watch stg a bit more closely after merging :)

PR #1716, opened 2022-10-19
Do not repeat custom copr restriction message

### LINKED ISSUE #1686: Restricting custom copr build message written multiple times
Service is writing multiple times the same message (on restricting custom copr builds): https://github.com/facebook/netconsd/pull/8

On [this method](https://github.com/majamassarini/packit-service/blob/09744908b931ec384dce845fc1fdd6b8e87f9294/packit_service/worker/helpers/build/copr_build.py#L375) we probably need to read what is already written into the github thread and comment only if the same message is not the last written message.

FrNecas:
Some slight refactoring of duplicate message check into status reporter using an optional argument + then using it in appropriate places.

TODO:

Fixes #1686 

RELEASE NOTES BEGIN
Packit now won't repeatedly comment in pull requests about the need to migrate configuration of allowed forge projects to Copr.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/handlers/testing_farm.py
+++ b/packit_service/worker/handlers/testing_farm.py
@@ -393,6 +393,7 @@ class TestingFarmResultsHandler(JobHandler):
         status_reporter = StatusReporter.get_instance(
             project=self.project,
             commit_sha=self.data.commit_sha,
+            packit_user=self.service_config.get_github_account_name(),
             trigger_id=trigger.id if trigger else None,
             pr_id=self.data.pr_id,
         )
--- a/packit_service/worker/reporting.py
+++ b/packit_service/worker/reporting.py
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: MIT
 
 import logging
-from enum import Enum
+from enum import Enum, auto
 from typing import Optional, Union, Dict
 
 from ogr.abstract import CommitStatus, GitProject
@@ -33,6 +33,17 @@ class BaseCommitStatus(Enum):
     error = "error"
 
 
+class DuplicateCheckMode(Enum):
+    """Enum of possible behaviour for handling duplicates when commenting."""
+
+    # Do not check for duplicates
+    do_not_check = auto()
+    # Check only last comment from us for duplicate
+    check_last_comment = auto()
+    # Check the whole comment list for duplicate
+    check_all_comments = auto()
+
+
 MAP_TO_COMMIT_STATUS: Dict[BaseCommitStatus, CommitStatus] = {
     BaseCommitStatus.pending: CommitStatus.pending,
     BaseCommitStatus.running: CommitStatus.running,
@@ -59,6 +70,7 @@ class StatusReporter:
         self,
         project: GitProject,
         commit_sha: str,
+        packit_user: str,
         trigger_id: int = None,
         pr_id: Optional[int] = None,
     ):
@@ -67,6 +79,7 @@ class StatusReporter:
         )
         self.project: GitProject = project
         self._project_with_commit: Optional[GitProject] = None
+        self._packit_user = packit_user
 
         self.commit_sha: str = commit_sha
         self.trigger_id: int = trigger_id
@@ -77,6 +90,7 @@ class StatusReporter:
         cls,
         project: GitProject,
         commit_sha: str,
+        packit_user,
         trigger_id: Optional[int] = None,
         pr_id: Optional[int] = None,
     ) -> "StatusReporter":
@@ -90,7 +104,7 @@ class StatusReporter:
             reporter = StatusReporterGitlab
         elif isinstance(project, PagureProject):
             reporter = StatusReporterPagure
-        return reporter(project, commit_sha, trigger_id, pr_id)
+        return reporter(project, commit_sha, packit_user, trigger_id, pr_id)
 
     @property
     def project_with_commit(self) -> GitProject:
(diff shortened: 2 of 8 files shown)


jpopelka:
The failing openshift tests are my fault, sorry, I'm working on the fix.

jpopelka:
recheck

packit_service/constants.py
```diff
@@ -56,6 +56,7 @@
     "in the linked issue.\n\n"
 )
 NOTIFICATION_REPO = "https://github.com/packit/notifications"
+PACKIT_APP_PREFIX = "packit-as-a-service"
```
lbarcziova, line 59: would it make sense to instead use [this method](https://github.com/packit/packit-service/blob/50e1c9cd220ab4c3f875b1bda15dd701bd98180e/packit_service/config.py#L268)?

packit_service/worker/reporting.py
```diff
@@ -240,7 +241,37 @@ def report_status_by_comment(
...
+        """Reports status by adding a comment.
+
+        If the instance is tied to a PR, the comment is added to the PR,
+        otherwise, it is added to the tied commit (if the forge supports
+        commit comments).
+
+        Args:
+            body: The comment text.
+            check_for_duplicate: If set to True, the comment won't be added if
+                it is already present in the PR/commit as the last comment.
```
lbarcziova, line 270: nice refactoring! I wonder whether we want to check only the last comment or avoid duplicity at all or have options to do either, WDYT? E.g. for the copr restriction messages, it could be enough to comment once per event.

packit_service/constants.py
FrNecas: It may behave slightly differently, if both stg and prod were enabled in the repo, then both will comment if we use the linked method. But I think that's what we want. I will modify my code.

packit_service/worker/reporting.py
lbarcziova: that sounds perfect

packit_service/worker/reporting.py
```diff
@@ -240,7 +254,56 @@ def report_status_by_comment(
     def get_statuses(self):
         self.project_with_commit.get_commit_statuses(commit=self.commit_sha)
 
-    def comment(self, body: str):
+    def _has_identical_comment(self, body: str, mode: DuplicateCheckMode) -> bool:
+        """Checks if the body is the same as the last or any comment based on mode."""
+        if self.pr_id:
```
FrNecas, line 259: Does anyone possibly see a better way of doing this? Since the interface of PR and commit comment is different, all solutions became more complicated than this slight duplication of logic, so I ended up doing it this way.

jpopelka:
recheck

packit_service/worker/handlers/testing_farm.py
```diff
@@ -393,6 +393,7 @@ def run(self) -> TaskResults:
         status_reporter = StatusReporter.get_instance(
             project=self.project,
             commit_sha=self.data.commit_sha,
+            packit_user=self.service_config.get_github_account_name(),
```
mfocko, line 396: Could you please check what is the account name? IIRC it might be prefixed with `app/`, but on GitLab, we »do not« have that prefix.

packit_service/worker/reporting.py
```diff
@@ -240,7 +254,56 @@ def report_status_by_comment(
...
+            commit_comments = self.project.get_commit_comments(self.commit_sha)
+            for comment in reversed(commit_comments):
+                if comment.author.startswith(self._packit_user):
+                    if mode == DuplicateCheckMode.check_last_comment:
+                        return body == comment.comment
+                    elif (
+                        mode == DuplicateCheckMode.check_all_comments
+                        and body == comment.comment
+                    ):
+                        return True
```
mfocko, line 281:
```suggestion
        comments = (
            self.project.get_pr(pr_id=self.pr_id).get_comments(reverse=True)
            if self.pr_id
            else reversed(self.project.get_commit_comments(self.commit_sha))
        )
        for comment in comments:
            if comment.author.startswith(self._packit_user):
                if mode == DuplicateCheckMode.check_last_comment:
                    return body == comment.body
                elif (
                    mode == DuplicateCheckMode.check_all_comments
                    and body == comment.body
                ):
                    return True
```

aaaa /o\ I see it now… I'd just add a property (and ideally deprecate the current one) to the commit comment to unify it, since it's the only one we haven't touched during the ogr refactoring.

It looks like we use `body` for the refactored issue/PR comments and it's also used in other objects (like `Release`)

mfocko (commented):
lgtm

packit_service/worker/reporting.py
FrNecas: Yes, that's what I wanted to do initially but then realized the different interfaces. I will create an ogr PR to change `comment` to `body`

packit_service/worker/handlers/testing_farm.py
FrNecas: I am not sure if I follow what you mean by the `app/` prefix. The [method](https://github.com/packit/packit-service/blob/50e1c9cd220ab4c3f875b1bda15dd701bd98180e/packit_service/config.py#L268) just returns one of `packit-as-a-service[bot]`, `packit-as-a-service-stg[bot]`, `packit-as-a-service-dev[bot]` depending on the deployment used. This should match the commenter name in the Github PRs/commits

packit_service/worker/reporting.py
```diff
@@ -33,6 +33,17 @@ class BaseCommitStatus(Enum):
...
 
+class DuplicateCheckMode(Enum):
+    """Enum of possible behaviour for handling duplicates when commenting."""
+
+    # Do not check for duplicates
+    do_not_check = auto()
+    # Check only last comment from us for duplicate
+    check_last_comment = auto()
+    # Check the whole comment list for duplicate
+    check_all_comments = auto()
```
lachmanfrantisek, line 44: I am wondering if we want to switch between the modes and don't be consistent but I like it after all..;) really nicely implemented!

lachmanfrantisek (approved):

FrNecas:
Updated the access to commit's comment body to match https://github.com/packit/ogr/pull/748 .

FrNecas:
recheck

lbarcziova (approved):
nicely done!

PR #1729, opened 2022-10-28
Feat(bodhi): Retrigger bodhi update via dist-git PR command

### LINKED ISSUE #1541: Implement retriggering Bodhi updates via dist-git pull-request comments

majamassarini:
I have rebased the Jirka K. work and I have used Mixins to manage Koji data needed to create a new bodhi update.

This comes from #1654

Fixes #1541

Merge after ogr#742 packit#1729

- [ ] call `is_packager` as aked by Laura: https://github.com/packit/packit-service/pull/1654#discussion_r994522524

---

RELEASE NOTES BEGIN
Packit now allows to re-trigger Bodhi update via dist-git PR comment `/packit create-update`.
RELEASE NOTES END


packit_service/worker/handlers/bodhi.py
```diff
@@ -167,3 +171,51 @@ def notify_user_about_failure(self, body: str) -> None:
...
+    def get_checkers() -> Tuple[Type[Checker], ...]:
+        """We react only on finished builds (=KojiBuildState.complete)
+        and configured branches.
+        """
+        logger.debug("Bodhi update will be re-triggered via dist-git PR comment.")
+        return (
+            HasAuthorWriteAccess,
+            IsAuthorAPackager,
+            IsKojiBuildCompleteAndBranchConfiguredCheckService,
+        )
```
TomasTomecek, line 221: Will Packit provide any feedback back to the user if one of these checkers fails?

TomasTomecek (commented):
I like the Checkers design! This is nicely done, just one question about the UX

packit_service/worker/handlers/mixin.py
```diff
@@ -80,6 +82,86 @@ def koji_build_helper(self) -> KojiBuildJobHelper:
...
+                )
+        return self._build
+
+    @property
+    def nvr(self) -> str:
+        return self.build["nvr"]
+
+    @property
+    def build_id(self) -> int:
+        return self.build["build_id"]
```
lbarcziova, line 154: could you please maybe add a link to the documentation where can one find what the build dictionary looks like?

lbarcziova (commented):
nicely done!

lbarcziova:
> Note: since I am using [this method](https://github.com/packit/packit-service/blob/96a59e7b1ca4ecf2c68633ec64ff4f01a9f5cea0/packit_service/worker/checker/abstract.py#L39) if for any reason the actor is None then the body update will succeed, I don't know if this could be a problem.

We should definitely make sure the actor is set, but looking into the code [here](https://github.com/packit/packit-service/blob/96a59e7b1ca4ecf2c68633ec64ff4f01a9f5cea0/packit_service/worker/events/event.py#L107) and [here](https://github.com/packit/packit-service/blob/96a59e7b1ca4ecf2c68633ec64ff4f01a9f5cea0/packit_service/worker/parser.py#L1291) (we could finally unify so that all events have the `actor` and not `user_login`, this is not very straightforward :D), I think this should be ok.

packit_service/worker/handlers/bodhi.py
majamassarini: You are right, there is no feedback for the users when checks fail. For none of the them.
If you agree I will create a new card to add the feedback, I think there is some work on it. Right now there is no `BodhiUpdateJobHelper` class and for the other services is this kind of helper class that handles the feedback.

TomasTomecek: agreed! we should absolutely have something to comment back so users know what's happening in Packit

majamassarini: > agreed! we should absolutely have something to comment back so users know what's happening in Packit

https://github.com/packit/packit-service/issues/1747

PR #1731, opened 2022-10-31
Fix tests with identifiers retriggering

### LINKED ISSUE #1730: Incorrect default Copr project name when identifiers are used
When reacting to test retriggering (via commit checks or test comment commands), the job config used for the construction of the default Copr project name is the test job config, therefore if there is for that test job an identifier defined (and there is also a build job defined in the config), default Copr project name is different than the one that was created in the Copr build job and therefore, the build is triggered as well (since there is no build in the DB for that Copr project) and the whole reporting becomes incorrect.

lbarcziova:
Fixes  #1730 

---

RELEASE NOTES BEGIN
Retriggering of the tests when there is a build job and test job with an identifier configured was fixed.
RELEASE NOTES END

### DIFF
--- a/packit_service/utils.py
+++ b/packit_service/utils.py
@@ -40,7 +40,7 @@ class only_once(object):
 
 
 # wrappers for dumping/loading of configs
-def load_package_config(package_config: PackageConfig):
+def load_package_config(package_config: dict):
     return PackageConfigSchema().load(package_config) if package_config else None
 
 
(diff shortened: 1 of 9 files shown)


lachmanfrantisek (approved):

lbarcziova:
@lachmanfrantisek PTAL :pray:

lachmanfrantisek (approved):

PR #1733, opened 2022-11-01
Skip actor pre_check if actor not present

lbarcziova:
This should make the behaviour the same as it was previously (before handler pre_checks refactoring
https://github.com/packit/packit-service/blob/da94cc9b511843d73e9b94d20677f0b1d1426bf8/packit_service/worker/jobs.py#L422): if the actor is None, the actor pre_check will be skipped (e.g. pushes to branch, releases).

---

RELEASE NOTES BEGIN

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/checker/abstract.py
+++ b/packit_service/worker/checker/abstract.py
@@ -1,3 +1,4 @@
+import logging
 from abc import abstractmethod
 from typing import Optional
 
@@ -6,6 +7,8 @@ from packit.config.package_config import PackageConfig
 
 from packit_service.worker.events import EventData
 from packit_service.worker.mixin import ConfigMixin, PackitAPIWithDownstreamMixin
+
+logger = logging.getLogger(__name__)
 
 
 class Checker(ConfigMixin, PackitAPIWithDownstreamMixin):
@@ -35,5 +38,6 @@ class ActorChecker(Checker):
 
     def pre_check(self) -> bool:
         if not self.actor:
-            return False
+            logger.debug("Actor not set for this event, skipping the actor check.")
+            return True
         return self._pre_check()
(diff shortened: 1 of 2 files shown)


mfocko:
But someone »does« the push or release, don't they?

lbarcziova:
If I recall correctly, we previously did the check only for PRs since if a person commits or releases, he/she has also write access to the repo.

mfocko (approved):

packit_service/worker/checker/abstract.py
```diff
@@ -35,5 +38,6 @@ def _pre_check(self) -> bool:
 
     def pre_check(self) -> bool:
         if not self.actor:
-            return False
+            logger.debug("Actor not set for this event, skipping the actor check.")
+            return True
         return self._pre_check()
```
lachmanfrantisek, line 43: Aren't we checking also other non-actor conditions? (I went through the code really quickly and am not sure about the `pre_check` / `_pre_check` call chain...)

lachmanfrantisek (approved):
Thanks!

packit_service/worker/checker/abstract.py
lachmanfrantisek: e.g. project allowed for internal TF,..

lbarcziova: I think only `_pre_check`s in subclasses [here](https://github.com/packit/packit-service/blob/d126fe4839f54c7b4ab7e6bd4232d2e709930208/packit_service/worker/checker/copr.py#L90) or [here](https://github.com/packit/packit-service/blob/d126fe4839f54c7b4ab7e6bd4232d2e709930208/packit_service/worker/checker/testing_farm.py#L63) are called, but I would wait for Maja's approval since she knows the code the best

majamassarini: Thanks Laura for fixing this. 
I did not realize I had changed the actor checker behaviour. You are right, before my changes it would have returned `False` only with a valid `actor` field.

Before my changes this code were called for every handler, but most of them just returned `True`, now it is called just for this two handlers Laura linked here.

majamassarini (approved):

PR #1736, opened 2022-11-02
Don't calculate the retrigger message during __init__()

csomh:
This way it can be avoided to have further properties calculated during __init__().

This way we can avoid calculating the 'job_build' property, which caused issues after TestingFarmJobHelper.get_request_details() was modified in e666087f to pass None as package_config, instead of a dummy PackageConfig object.

Test this by not mocking the job_build attribute of TestingFarmJobHelper in test_get_request_details().


lachmanfrantisek (approved):

TomasTomecek (approved):

csomh:
Missed updating `KojiBuildJobHelper` :facepalm:

PR #1737, opened 2022-11-04
 Drop _commit_comment_already_added()

jpopelka:
and reuse `comment()` which runs `_has_identical_comment()` which can also check commit comments.
`comment()` now has a new `to_commit` switch which forces it to operate (check and add) on commit comments even when the instance is tied to a PR.

### DIFF
--- a/packit_service/worker/reporting.py
+++ b/packit_service/worker/reporting.py
@@ -71,7 +71,7 @@ class StatusReporter:
         project: GitProject,
         commit_sha: str,
         packit_user: str,
-        trigger_id: int = None,
+        trigger_id: Optional[int] = None,
         pr_id: Optional[int] = None,
     ):
         logger.debug(
@@ -90,7 +90,7 @@ class StatusReporter:
         cls,
         project: GitProject,
         commit_sha: str,
-        packit_user,
+        packit_user: str,
         trigger_id: Optional[int] = None,
         pr_id: Optional[int] = None,
     ) -> "StatusReporter":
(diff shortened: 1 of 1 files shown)


packit_service/worker/reporting.py
```diff
@@ -254,8 +256,11 @@ def report_status_by_comment(
     def get_statuses(self):
         self.project_with_commit.get_commit_statuses(commit=self.commit_sha)
 
-    def _has_identical_comment(self, body: str, mode: DuplicateCheckMode) -> bool:
+    def _pr_has_identical_comment(self, body: str, mode: DuplicateCheckMode) -> bool:
```
FrNecas, line 259: This method is not only about PR comments, it also checks commit comments if `commit_sha` is set so I think that the new name is not very precise.

packit_service/worker/reporting.py
```diff
@@ -254,8 +256,11 @@ def report_status_by_comment(
...
         """Checks if the body is the same as the last or any comment based on mode."""
+        if not self._packit_user:
+            raise PackitException("packit_user is required for identical comment check")
```
FrNecas, line 262: I am just wondering if there isn't a better way of doing this. Currently the exception is not caught anywhere so it would go to the user, I assume, even though it is just our internal error caused by the fact that we forgot to set `packit_user` in some environment (e.g. hardly in this case). And I can see that sometimes in the future, someone is going to add `status_reporter.comment(..., duplicate_check=DuplicateCheckMode.check_last_comment)` without realizing that it is not valid. I don't really see a good way now

packit_service/worker/reporting.py
jpopelka: I didn't realize it can operate on commit comments as well.
In that case, we should merge it with `_commit_comment_already_added()`.
I changed the approach, please review it again.

packit_service/worker/reporting.py
jpopelka: Yes, I realized that as well. I'll properly set it in hardly.

packit_service/worker/reporting.py
```diff
@@ -270,35 +268,32 @@ def _has_identical_comment(self, body: str, mode: DuplicateCheckMode) -> bool:
...
                 ):
                     return True
-
         return False
 
     def comment(
         self,
         body: str,
         duplicate_check: DuplicateCheckMode = DuplicateCheckMode.do_not_check,
+        to_commit: Optional[bool] = False,
```
FrNecas, line 277: Is there any reason for `Optional[bool]` that I am missing? I think that just `bool` with a default value of `False` should be sufficient

packit_service/worker/reporting.py
```diff
@@ -254,12 +244,20 @@ def report_status_by_comment(
     def get_statuses(self):
         self.project_with_commit.get_commit_statuses(commit=self.commit_sha)
 
-    def _has_identical_comment(self, body: str, mode: DuplicateCheckMode) -> bool:
-        """Checks if the body is the same as the last or any comment based on mode."""
+    def _has_identical_comment(
+        self, body: str, mode: DuplicateCheckMode, commit: Optional[bool] = False
```
FrNecas, line 248: nitpick: I'd rename `commit` to `check_commit` to make it clear that is is not some kind of git commit object

FrNecas (commented):
Nice, I like this more! Some last nitpicks

packit_service/worker/reporting.py
jpopelka: :facepalm:

packit_service/worker/reporting.py
```diff
@@ -254,12 +244,20 @@ def report_status_by_comment(
     def get_statuses(self):
         self.project_with_commit.get_commit_statuses(commit=self.commit_sha)
 
-    def _has_identical_comment(self, body: str, mode: DuplicateCheckMode) -> bool:
-        """Checks if the body is the same as the last or any comment based on mode."""
+    def _has_identical_comment(
+        self, body: str, mode: DuplicateCheckMode, check_commit: Optional[bool] = False
```
FrNecas, line 248: Extra optional here as well for `check_commit`

FrNecas (approved):

lbarcziova (approved):

jpopelka:
recheck

PR #1744, opened 2022-11-08
Do bump release if release_suffix is None

majamassarini:
We want to preserve release only when release_suffix is empty in config.

### DIFF
--- a/packit_service/worker/helpers/build/build_helper.py
+++ b/packit_service/worker/helpers/build/build_helper.py
@@ -432,7 +432,7 @@ class BaseBuildJobHelper(BaseJobHelper):
         self._srpm_model.set_start_time(datetime.datetime.utcnow())
 
         bump_version = (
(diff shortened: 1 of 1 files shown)


packit_service/worker/helpers/build/build_helper.py
```diff
@@ -432,7 +432,9 @@ def _create_srpm(self):
         self._srpm_model.set_start_time(datetime.datetime.utcnow())
 
         bump_version = (
-            self.job_config.release_suffix  # do not modify version if release suffix is ''
+            (
+                self.job_config.release_suffix is None or self.job_config.release_suffix
+            )  # do not modify version if release suffix is ''
```
jpopelka, line 437: just a suggestion

```suggestion
            self.job_config.release_suffix != ""
```

jpopelka (approved):

PR #1745, opened 2022-11-08
Add shellcheck to pre-commit config

jpopelka:
And a few other polishing commits.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ version: "2"
 
 services:
   redis:
-    image: quay.io/centos7/redis-5-centos7
+    image: quay.io/sclorg/redis-6-c9s
     container_name: redis
     ports:
       - 6379:6379
@@ -35,7 +35,7 @@ services:
 
   postgres:
     container_name: postgres
-    image: quay.io/centos7/postgresql-13-centos7
+    image: quay.io/sclorg/postgresql-13-c9s
     environment:
       POSTGRESQL_USER: packit
       POSTGRESQL_PASSWORD: secret-password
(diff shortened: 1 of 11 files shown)


docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: "2"
 
 services:
   redis:
-    image: quay.io/centos7/redis-5-centos7
+    image: quay.io/sclorg/redis-6-c9s
```
TomasTomecek, line 5: sneaky

files/run_tests.sh
```diff
@@ -2,6 +2,7 @@
 
 set -x
 
+# shellcheck source=/dev/null
 source /src/files/setup_env_in_openshift.sh
```
TomasTomecek, line 6: huh, is this telling shellcheck to not try to access that file?

TomasTomecek (approved):
Your completely unrelated commits are always so amusing to review 😁

files/run_tests.sh
jpopelka: Yes, there's no such file in the repository (it is in the image/container) so shellcheck can't access it.
Should I add a comment with a link to https://www.shellcheck.net/wiki/SC1091 ?

jpopelka:
> Your completely unrelated commits are always so amusing to review

I see, just tell me if you'd rather prefer 5 separate PRs :-)

lbarcziova (approved):
:+1:

files/run_tests.sh
TomasTomecek: it would definitely be welcoming to people who see this line for the first time, thank you for the explanation

FrNecas (approved):

PR #1751, opened 2022-11-10
New hotness event

### LINKED ISSUE #1577: Packit listens to NewHotness messages
For #1576, we need to be notified about a new upstream version.

Michal (zlopez) suggested using `hotness.update.bug.file` messages because it contains all the needed info and some pre-filtering is done before creating the bug.

* Message schema:
  * https://the-new-hotness.readthedocs.io/en/stable/message-schema.html#message-topics
  * https://github.com/fedora-infra/the-new-hotness-messages/blob/main/hotness_schema/messages.py#L147
  * https://the-new-hotness-messaging-schema.readthedocs.io/en/latest/topics.html
  * Example: https://apps.fedoraproject.org/datagrepper/v2/id?id=82f2e960-bab0-4c76-9be4-033a47f4589d&is_raw=true&size=extra-large

When working with fedmsg messages in the parser, be aware of this:
https://github.com/packit/packit-service/blob/0afd541a922dd25751d3e968d8807b8d5d27d0e1/packit_service/worker/parser.py#L105-L108

The task here is to:
* [ ] packit/packit-service-fedmsg#71
* [ ] create a new event class to represent the new release event
* [ ] implemnt parsing of the `hotness.update.bug.file` message into the newly created event class instances
* [ ] make sure the Packit config can be gathered for the event. (probably `main` dist-git branch)
* [ ] add tests for the parsing

*part of #1576 epic*

lbarcziova:
TODO:

- [ ] TODOs in the code :D
- [x] more tests

Fixes #1577

---

RELEASE NOTES BEGIN
N/A
RELEASE NOTES END


packit_service/constants.py
```diff
@@ -287,3 +287,5 @@ def from_number(number: int):
     "grant `{packit_user}` user `Developer` permissions on the `{namespace}/{repo}`"
     " repository. You can do so [here]({url}/-/project_members)."
 )
+
+DISTGIT_URL = "https://src.fedoraproject.org"
```
jpopelka, line 291: There's already a [`DISTGIT_URL` environment variable](https://github.com/packit/deployment/blob/main/openshift/packit-worker.yml.j2#L68) set to [`https://src.fedoraproject.org/` by default](https://github.com/packit/deployment/blob/main/playbooks/deploy.yml#L45).

There's also a [`PROD_DISTGIT_URL` defined in packit/constants.py](https://github.com/packit/packit/blob/main/packit/constants.py#L33)

I'd suggest using it like [here](https://github.com/packit/packit-service/blob/b147091ca383ac8ba884b8d647651abad200cc39/packit_service/worker/parser.py#L893)

lbarcziova: I haven't searched properly, thanks for the suggestion! I will change it

packit_service/worker/events/new_hotness.py
```diff
@@ -0,0 +1,102 @@
...
+        )
+
+    @property
+    def commit_sha(self):
+        # this is used for getting the package configuration, by default the method
+        # gets it from the default branch of the repository which is what we want
+        return None
+
+    def pre_check(self):
+        if not self.package_config:
```
TomasTomecek, line 81: Could you please add a docstring here to sum up what exactly is being checked? :)

packit_service/worker/parser.py
```diff
@@ -1292,3 +1295,25 @@ def parse_pagure_pull_request_comment_event(
...
+
+        logger.info(
+            f"New hotness update event for package: {package_name}, version: {version}"
+        )
+
+        return NewHotnessUpdateEvent(
+            package_name=package_name,
+            version=version,
+            distgit_project_url=distgit_project_url,
+        )
```
TomasTomecek, line 1319: Since version and tag can be different, should we also populate the tag as well?

tests/unit/test_events.py
```diff
@@ -1500,6 +1506,92 @@ def test_parse_check_rerun_release(self, check_rerun):
...
+                True,
+                "no-version-tag",
+                "redis-namespace",
+                "redis",
+            ),
+            (
+                "https://github.com/redis-namespace/redis",
+                "v{version}",
+                True,
+                "v7.0.3",
```
TomasTomecek, line 1533: okay, this probably answers my question from above :D

although I'm contemplating if the tag name is already present in the message event and if it's, we should probably use it

lbarcziova: no, it is not in the message (there is an example message in the test data). We discussed with @lachmanfrantisek  that if there is `upstream_tag_template` defined in config, we will use it to get the tag and if it is not, take the version as the tag.

TomasTomecek: understood: you solved it well

TomasTomecek (approved):

packit_service/worker/events/event.py
```diff
@@ -445,10 +445,12 @@ def get_package_config(self) -> Optional[PackageConfig]:
             fail_when_missing=self.fail_when_config_file_missing,
         )
 
+        # TODO do we need to do this?
         # job config change note:
         #   this is used in sync-from-downstream which is buggy - we don't need to change this
-        if package_config:
+        if package_config and self.__class__.__name__ != "NewHotnessUpdateEvent":
```
lachmanfrantisek, line 451: Would it be possible to revert this check (`self.__class__.__name__  == ... `) so we don't need to enhance this when adding any other event?

packit_service/worker/events/new_hotness.py
```diff
@@ -0,0 +1,102 @@
...
+        package_name: str,
+        version: str,
+        distgit_project_url: str,
+    ):
+        super().__init__()
+        self.package_name = package_name
+        self.version = version
+        self.project_url = distgit_project_url
+
+        self._upstream_repo_namespace_and_name: Optional[Tuple] = None
```
lachmanfrantisek, line 25: We use `full_name` in OGR for this, but that's just a detail.

packit_service/worker/events/new_hotness.py
```diff
@@ -0,0 +1,102 @@
...
+    def get_repo_namespace_and_name(self) -> Optional[Tuple]:
+        path = urlparse(self.upstream_project_url).path
+        if not path:
+            return None, None
+        # remove leading /
+        path = path[1:]
+        path_split = path.rsplit("/", maxsplit=1)
+        if len(path_split) != 2:
+            return None, None
+        return path_split[0], path_split[1]
```
lachmanfrantisek, line 40: We have a class for this..;)

https://github.com/packit/ogr/blob/67470efc7e29942ddcfa99ac653d225f16455a6e/ogr/parsing.py#L8

packit_service/worker/events/new_hotness.py
```diff
@@ -0,0 +1,102 @@
...
+    def pre_check(self):
+        if not self.package_config:
+            logger.warning(
+                "Package config not present in the dist-git repository with version update."
+            )
+            return False
+
+        if not self.package_config.upstream_project_url:
+            logger.warning("upstream_project_url not set in the package config.")
+            return False
```
lachmanfrantisek, line 89: OK for now, but it would be nice to react to such a situation to let a user know that something needs to be fixed in the config.

packit_service/worker/parser.py
```diff
@@ -1292,3 +1295,25 @@ def parse_pagure_pull_request_comment_event(
             comment=comment,
             comment_id=comment_id,
         )
+
+    @staticmethod
+    def parse_new_hotness_update_event(event) -> Optional[NewHotnessUpdateEvent]:
+        if "hotness.update.bug.file" not in event.get("topic", ""):
+            return None
+
+        package_name = event.get("package")
```
lachmanfrantisek, line 1304: Shouldn't we use `package_name` from a Fedora subdictionary instead? Or provide both. This one is probably what we call `upstream_name`.

packit_service/worker/parser.py
```diff
@@ -1292,3 +1295,25 @@ def parse_pagure_pull_request_comment_event(
...
+        dg_base_url = getenv("DISTGIT_URL", PROD_DISTGIT_URL)
+
+        distgit_project_url = f"{dg_base_url}rpms/{package_name}"
```
lachmanfrantisek, line 1307: I don't want to break more things than needed, but wouldn't be better to have the `/` here, instead of the constant/env-var?

lachmanfrantisek (commented):
Nice!

Just a few suggestions...

packit_service/worker/events/new_hotness.py
lbarcziova: great! I will replace it with this

packit_service/worker/parser.py
lbarcziova: I also wasn't sure and therefore checked the code how 'package' is set [here](https://github.com/fedora-infra/the-new-hotness/blob/363acd33623dadd5fc3b60a83a528926c7c21fc1/hotness/hotness_consumer.py#L385) and [here](https://github.com/fedora-infra/the-new-hotness/blob/363acd33623dadd5fc3b60a83a528926c7c21fc1/hotness/hotness_consumer.py#L444-L455) so it should be the Fedora package name...but I will double check it

packit_service/worker/parser.py
lbarcziova: it probably would, but this is already used across repositories, [see](https://github.com/search?p=1&q=org%3Apackit+DISTGIT_URL&type=Code), so changing the constant would not be enough

packit_service/worker/events/event.py
lbarcziova: sorry, I am not getting it :sweat_smile: could you please explain how would the change look like?

lachmanfrantisek: With your change, we are checking if it is not a newly added event. We will need to do this again when adding another event. I was suggesting checking for a case where this is required (so, name the events that require this).

So, instead of `self.__class__.__name__ != Something` (and `self.__class__.__name__ not in {"NewHotnessUpdateEven", "AnotherEventAddedInTheFuture"}` in the future), use `self.__class__.__name__  == "SomeEvent"` or `self.__class__.__name__  in {"SomeEvent", "AnotherEvent"}`

lachmanfrantisek: Also, could you please use `NewHotnessUpdateEvent.__name__` so type checks are able to realise problems?

packit_service/worker/parser.py
lachmanfrantisek: Ok, we can easily fix this if it is not true. (Could you please add  a comment about this to the parser so we don't forget why it's like that?)

packit_service/worker/parser.py
lachmanfrantisek: I mainly meant the constant but you are right that we should be consistent between the config and env.var...

packit_service/worker/events/event.py
lbarcziova: I wanted to, but that would cause circular import

lbarcziova: resolved via chat: we will not need to update the condition for each new event in future, this is specific for NewHotnessUpdateEvent, since in its case we will rely on `upstream_project_url` found in the config and do not want to rewrite it

PR #1762, opened 2022-11-18
Fix a typo in the failure message

r0x0d:
A typo was introduced in packit/packit-service#1658 where the `namespace` and `repo` were in the wrong order. 

Also removed the parameter `url` as the default value for `url` in `report_status_to_tests` is already an empty string.


lbarcziova (approved):
Thanks!! Please, rebase and we can merge this

r0x0d:
Oh, I clicked the `update branch` through the github ui. Is that a problem? I can undo and do a rebase locally.

Issue #1728, opened 2022-10-27
Reporting initial status time: 20.890889 seconds.


jpopelka:
Yesterday, I noticed on staging several cases of the initial status time being set after 15 seconds, which is the limit for the [SLO1 error budget](https://github.com/packit/research/blob/main/error-budgets/SLOs-SLIs.md#slo1-changes-to-github-prs-receive-a-status-update-within-15-seconds-in-99-of-cases).

Looking at the short-running worker logs, you can see that there was a window of 17 seconds between the `process_message` parsed the event and started to get the package config where (as it looks to me) it was just slacking.
```
[12:34:19,375: INFO/MainProcess] Task task.steve_jobs.process_message[e4622d3e-786a-4821-bfb2-f0069bf63fa8] received
[12:34:19,376: DEBUG/MainProcess] TaskPool: Apply <function fast_trace_task at 0x7f7aa5d28af0> (args:('task.steve_jobs.process_message', 'e4622d3e-786a-4821-bfb2-f0069bf63fa8',... )
[12:34:19,377: INFO/MainProcess] Github PR#2 comment: '/packit-stg build' id#1291965878 'created' event.
[12:34:19,378: INFO/MainProcess] Target repo: lbarcziova/hello-world.
[12:34:20,967: DEBUG/MainProcess] pidbox received method enable_events() [reply_to:None ticket:None]
[12:34:25,968: DEBUG/MainProcess] pidbox received method enable_events() [reply_to:None ticket:None]
[12:34:30,968: DEBUG/MainProcess] pidbox received method enable_events() [reply_to:None ticket:None]
[12:34:35,968: DEBUG/MainProcess] pidbox received method enable_events() [reply_to:None ticket:None]
[12:34:36,079: DEBUG/MainProcess] Getting package_config: ...
[12:34:36,646: DEBUG/MainProcess] Found a config file '.packit.yaml' on ref '...' of the 'lbarcziova/hello-world' repository.
...
[12:34:40,704: DEBUG/MainProcess] Reporting initial status time: 20.890889 seconds.
...
[12:34:44,298: INFO/MainProcess] Task task.steve_jobs.process_message[e4622d3e-786a-4821-bfb2-f0069bf63fa8] succeeded in 24.920801570871845s: ...
```

When checking the `Average Task Runtime at Worker` graph in the `Packit boards / Celery Monitoring with Flower` 
[...]

PR #1766, opened 2022-11-22
debug(response_time): logger.error when response > 15s

TomasTomecek:
...which violates our SLO1

We don't have enough data to understand why these spikes happen
(sometimes it takes up to 20s to send the initial status).

This commit simply just `logger.error()` when the response takes more
than 15s to do. Hence we'll get all logs for this in sentry. in sentry.
in sentry. in sentry.

Related: https://github.com/packit/packit-service/issues/1728


lachmanfrantisek (approved):
Are we ok with new sentry issues being created for this?

lbarcziova (approved):

jpopelka (approved):

PR #1768, opened 2022-11-22
fix(build logs URL formatting): strings → int cast

TomasTomecek:
For some reason, Copr build models use build_id as str but in code, we
expect it to be int.

This commit fixes formatting of build_id to construct a build logs URL.

Fixes PCKT-002-PACKIT-SERVICE-4WT


lbarcziova (commented):
thanks for such a quick fix!

lbarcziova (approved):

PR #1771, opened 2022-11-23
Testing farm related improvements

### LINKED ISSUE #1725: Improve the actor check for running internal TF
Since for test jobs, builds may be required, we additionally don't allow builds if there is an internal TF test job with the same trigger configured (builds could trigger the test job that was not allowed). The reason for doing this is that we don't store the `actor` anywhere and therefore cannot run the check directly when running tests if there was a required build before. A solution to this would be to store the `actor` in the DB. This way we could run the actor check directly when running the test job.

TODO:
- [ ] store the `actor` from the webhook in the DB
- [ ] remove [this check class](https://github.com/packit/packit-service/blob/d61623d0129a7423bfe325c80cb16b1ba9fee8d6/packit_service/worker/checker/copr.py#L84) for `CoprBuildHandler`
- [ ] for `TestingFarmHandler` remove [this condition](https://github.com/packit/packit-service/blob/main/packit_service/worker/checker/testing_farm.py#L64-L70)

lbarcziova:
The first commit solves the issue that if there are multiple test jobs defined (and no build job, I found this setup only in teemtee/tmt so far), after the identifier-related changes (that fixed the use case one copr build + more testing farm jobs defined) we have been triggering TF  more than needed (for each Copr build, we have been running all test jobs).

The second commit implements storing the `actor` for PRs in DB and simplifying actors check for internal TF. The code around the internal TF checks was changed multiple times and has not been the best UX (builds have not been run for external contributors if any internal TF job is defined) so it was just easier to implement it using DB.

Fixes #1725 

---

RELEASE NOTES BEGIN
The check that prevented external contributors to trigger internal tests was simplified and other test jobs and builds required for tests will be not affected anymore.
RELEASE NOTES END


lbarcziova:
Update: I am thinking about removing the first commit after our discussion on architecture where we decided we would like to start requiring an explicit definition of build job for tests (will attach an issue) and maybe opening PR for teemtee to add build job for their configuration as well.

PR #1776, opened 2022-11-25
SLO1: raise limit from 15s to 30s

### LINKED ISSUE #1728: Reporting initial status time: 20.890889 seconds.

TomasTomecek:
If a job defines a lot of targets (10+), we can easily run out of SLO1
    because it takes 1s to set one check. With the current implementation,
    we measure how long it took to set every single status line - even the
    last one.
    
    We want to preserve this to be sure all are set in a reasonable time.
    
Precise description:
https://github.com/packit/packit-service/issues/1728#issuecomment-1326534470
    
Fixes #1728

RELEASE NOTES BEGIN
We have changed the limit for our [SLO1](https://packit.dev/docs/service-level-objectives/#slo1-changes-to-github-prs-receive-a-status-update-within-15-seconds-in-99-of-cases): it was increased from 15s to 30s to account for setting all statuses.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/monitoring.py
+++ b/packit_service/worker/monitoring.py
@@ -56,9 +56,9 @@ class Pushgateway:
             registry=self.registry,
         )
 
-        self.no_status_after_15_s = Counter(
-            "no_status_after_15_s",
-            "Number of PRs/commits with no commit status for more than 15s",
+        self.no_status_after_30_s = Counter(
+            "no_status_after_30_s",
+            "Number of PRs/commits with no commit status for more than 30s",
             registry=self.registry,
         )
 
--- a/packit_service/worker/jobs.py
+++ b/packit_service/worker/jobs.py
@@ -668,8 +668,8 @@ class SteveJobs:
         )
         logger.debug(f"Reporting initial status time: {response_time} seconds.")
         pushgateway.initial_status_time.observe(response_time)
-        if response_time > 15:
-            pushgateway.no_status_after_15_s.inc()
+        if response_time > 30:
+            pushgateway.no_status_after_30_s.inc()
             # https://github.com/packit/packit-service/issues/1728
             # we need more info why this has happened
             logger.debug(f"Event dict: {self.event}.")


jpopelka:
> We have two options here:
> 
>     1. Change SLO1 to account for this (e.g. increase to 1 minute?)
>     2. Change how we measure SLO1
> 
> I went for (2) because I understand SLO1 as:
> 
> > @psss writes to our chat that Packit did not set initial statuses
> > after 2 minutes of him pushing to a PR

Thank you for the explanation. I, however, suggest a variant of (1) for the very same reason.
What I suggest is to simply be less strict and lower the SLO1 from 99% to let's say 95%. (At the moment of writing this, the SLO1 is at 98,6%)
For the whole explanation see https://github.com/packit/packit-service/issues/1728#issuecomment-1326534470

lachmanfrantisek (commented):
Why `40`?..;) What about `30`? or `60` so we can easily describe this to our users?

packit_service/worker/monitoring.py
```diff
@@ -56,9 +56,9 @@ def __init__(self):
             registry=self.registry,
         )
 
-        self.no_status_after_15_s = Counter(
-            "no_status_after_15_s",
-            "Number of PRs/commits with no commit status for more than 15s",
+        self.no_status_after_40_s = Counter(
+            "no_status_after_40_s",
+            "Number of PRs/commits with no commit status for more than 40s",
```
lbarcziova, line 61: I am wondering if we want to still measure the count specifically, since if I am not mistaken, currently everywhere (all the graphs/alerts,..) the `initial_status_time` is used (even though probably I introduced both of these metrics:D)

TomasTomecek:
> Why `40`?..;) What about `30`? or `60` so we can easily describe this to our users?

I don't follow how's 30 or 60 easier to be described :D they are just integers.

> * `30` is strict but the user will most probably see the update after performing the action.

The reason I picked 40 was that, as you say, 30 felt too strict to me. Even though we haven't hit that right now, folks with 10 targets+ (x2 because of build & test jobs) can hit it

> * `60` is still ok, but I wouldn't consider this as an immediate response.

As long as it's more than ~25, which seems like the current top, I'm okay with it. 40 felt more generous than 30

packit_service/worker/monitoring.py
TomasTomecek: oh, you're right, I haven't noticed, thank you.

TomasTomecek: well since I already made the changes, I'd just keep it :D unless you won't me to drop it

lbarcziova (approved):

TomasTomecek:
Spoke to Franta over PM and we concluded to go for 30s instead.

PR #1778, opened 2022-11-25
Add gitlab release and tag_push events

### LINKED ISSUE #1739: Support propose downstream on GitLab
We don't accept Release/Tag events, nor have our events, etc.

Implement the parsing class.

majamassarini:
Fixes #1739

### DIFF
--- a/packit_service/worker/events/gitlab.py
+++ b/packit_service/worker/events/gitlab.py
@@ -6,6 +6,7 @@ from ogr.abstract import GitProject, Comment
 from packit_service.service.db_triggers import (
     AddPullRequestDbTrigger,
     AddBranchPushDbTrigger,
+    AddReleaseDbTrigger,
 )
 from packit_service.worker.events.comment import (
     AbstractIssueCommentEvent,
@@ -198,3 +199,47 @@ class PipelineGitlabEvent(AbstractGitlabEvent):
         self.commit_sha = commit_sha
         self.source = source
         self.merge_request_url = merge_request_url
+
+
+class ReleaseGitlabEvent(AddReleaseDbTrigger, AbstractGitlabEvent):
+    def __init__(
+        self,
+        repo_namespace: str,
+        repo_name: str,
+        git_ref: str,
+        tag_name: str,
+        project_url: str,
+        commit_sha: str,
+    ):
+        super().__init__(project_url=project_url)
+        self.repo_namespace = repo_namespace
+        self.repo_name = repo_name
+        self.git_ref = git_ref
+        self.tag_name = tag_name
+        self._commit_sha = commit_sha
+
+    @property
+    def commit_sha(self):
+        return self._commit_sha
+
+
+class TagPushGitlabEvent(AddBranchPushDbTrigger, AbstractGitlabEvent):
+    def __init__(
+        self,
+        repo_namespace: str,
+        repo_name: str,
+        actor: str,
+        git_ref: str,
+        project_url: str,
+        commit_sha: str,
+        title: str,
+        message: str,
+    ):
+        super().__init__(project_url=project_url)
+        self.repo_namespace = repo_namespace
+        self.repo_name = repo_name
+        self.actor = actor
+        self.git_ref = git_ref
+        self.commit_sha = commit_sha
+        self.title = title
+        self.message = message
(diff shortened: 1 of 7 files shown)


TomasTomecek (commented):
I'm pretty sure we'll need to add `@reacts_to(event=TheseNewHandlers)` to the propose_downstream handlers.

@mfocko is the best to review this :)

majamassarini:
> I'm pretty sure we'll need to add `@reacts_to(event=TheseNewHandlers)` to the propose_downstream handlers.

Sure

packit_service/worker/handlers/distgit.py
```diff
@@ -113,6 +115,8 @@ class AbortProposeDownstream(Exception):
 @run_for_comment(command="propose-update")  # deprecated
 @run_for_check_rerun(prefix="propose-downstream")
 @reacts_to(event=ReleaseEvent)
+@reacts_to(event=ReleaseGitlabEvent)
+@reacts_to(event=TagPushGitlabEvent)
```
TomasTomecek, line 119: I am not familiar with GitLab enough to know what would be the best experience here: if Packit should react to both or to just one of these (or even make it configurable?)

Does anyone know if tags & releases work the same way on GitLab as they work on GitHub?

TomasTomecek (approved):
LGTM, just one question

I'm sure others can give you a more helpful code review :)

packit_service/worker/handlers/distgit.py
mfocko: I would keep it consistent with GitHub and that would be just the `ReleaseGitlabEvent`.

mfocko: Thinking about it, this could also cause some issues with `copy_upstream_release_description` or whatever that setting for populating changelog is called.

mfocko (approved):
LGTM :)

packit_service/worker/handlers/distgit.py
TomasTomecek: that sounds perfect! let users configure what they send to us :)

mfocko: > that sounds perfect! let users configure what they send to us :)

Given that we have `copy_upstream_release_description`, this could be an issue with the tags, so unless they use it, I guess those two triggers could be interchangeable, but it sounds a bit fragile tbf

packit_service/worker/events/gitlab.py
```diff
@@ -198,3 +198,43 @@ def __init__(
         self.commit_sha = commit_sha
         self.source = source
         self.merge_request_url = merge_request_url
+
+
+class ReleaseGitlabEvent(AddBranchPushDbTrigger, AbstractGitlabEvent):
```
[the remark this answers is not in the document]
majamassarini: I don't see a celery attribute anywhere in this class, but maybe I am not following you 🤔

packit_service/worker/handlers/distgit.py
majamassarini: If I should decide I will go with the 2nd, but I do not really understand the drawbacks of this. So I wait for someone to tell me what to do 😅. Thanks!

packit_service/worker/events/gitlab.py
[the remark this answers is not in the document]
mfocko: For the retries, once handler takes over and runs the `propose-downstream`, if an error occurs it should go back to the celery queue to be retried.

Release event for github just calls super and adds commit hash, so I guess that's all that's needed.

packit_service/worker/handlers/distgit.py
majamassarini: Now I have seen the Tomas preference *let users configure what they send to us*. 
Then, I try to update the documentation.

mfocko: tl;dr if you kept both `@reacts_to` and ticked both tag and releases in gitlab's webhook settings, we would get both of the events, and if tag push and release was done in a close proximity of each other, we could end up with a race condition (and very likely duplicit PRs on downstream)

majamassarini: Personally I think that if we end up duplicating the PR in distgit, upon a wrong user configuration, this is not too bad. But the race condition could lead to bad exceptions which could be more difficult to understand.

So, should we trust our users? Should we keep both events? Or should we just keep the `Release` event and maybe create another card to study a better solution to introduce the `TagPush` event?

mfocko: I think it would make sense to react just to releases right now and open some low-prio card for the tag push (on both github and gitlab), but as of now I am not aware of anyone requesting that, simpler compared to release, solution.

I'm also against going with tag push to prod rightaway. With `release` I am not worried that much.

majamassarini: https://github.com/packit/packit-service/issues/1785

Issue #1665, opened 2022-09-20
Failed test (testing farm job) not shown in packit dashboard


majamassarini:
While I was analyzing issue packit-service#1664 I realized that the failed job `testing-farm:fedora-37-x86_64` was not shown in packit dashboard, only the successfull testing farm jobs were shown.

lbarcziova:
The reason is that we create the models in DB just after successful submitting the tests to TF, so in this case, we don't create a DB record for the particular target. IIRC, we do the same with Copr builds - create the DB records just after submitting the builds in Copr. Do you think we should adjust this behaviour?

majamassarini:
If we think it is doable to enhance it probably it is worth spending some time on it.
I use the dashboard as a overall view to see if everything is going smooth... if problems are not shown then probably it becomes less usefull, at least for me.

lbarcziova:
I get your point. This is definitely something to discuss as a team because the current implementation also makes sense if you think about it in the way that the state of the job is tracked only after successful submission => failure/error states mean some actual failure in the execution of build/tests and if the job is not in the database it means that the job was not executed at all because of some failed preconditions (invalid configuration, ...).

FrNecas:
Outcome from today's arch discussion: we should store the model in DB even if TF submission fails (same for other types of targets). As for retries, we should have one model instance per test (avoid creating a new model when retrying). Henceforth, we need to track the retry status in the model as well (something similar should already be in propose-downstream most likely).

I will try to take care of this issue since it is a blocker for #1633

PR #1787, opened 2022-11-29
Create group models

FrNecas:
This is a good candidate for a pair review, here are just some pointers:
- The diff looks scary but a lot of it is just tests and migrations which contain definition of models
- We now create a model before sending a request to TF/Copr/Koji, the model will have an empty pipeline ID/web url since it is not known yet. This will be filled in later, once the request succeeds. For handling retries, the handlers now accept a group ID which they fetch rather than creating a new model to avoid creating unnecessary models on failures.
- As discussed at last arch, I added an explicit M:N mapping between copr and TF. This was necessary because by introducing groups, a pipeline now contains a group of runs and a group of tests and mapping a test to a single build (e.g. f35 test to f35 build) would otherwise be impossible.
- Testing farm changes are probably the most prone to breaking (lots of merge conflicts, I tried to make groups work with Laura's recent work), koji and copr seemed a lot less complex.
- I also made some changes to documentation and some other improvements along the way
- The migrations are not really backwards compatible, even though I tried my best. When you merge something, splitting it back is complicated. Also see my TODO comment in the migrations `            # TODO: should we merge the groups? This would result in deletion and possibly some mess`. I feel like removing existing pipelines is not the way to go so I rather went with "duplicating data"

TODO:

- [x] Update or write new documentation in `packit/packit.dev`. [nothing to document]
- [x] Add API endpoints for getting group data by ID (may not be necessary for now, the existing API should be more-or-less unimpacted, but it may be useful for the future)
- [x] Try running the migration on production data to see how fast it runs and if there is any need for trying to optimize (the 4 migrations take roughly a minute for stage data on my local machine).
- [x] Openshift tests?

Fixes:  #1327 
Fixes: #1665 

---

RELEASE NOTES BEGIN
Packit now groups related builds and test runs (e.g. triggered by the same event, just different chroots) together. In the future, this will allow better presentation of the overall pipelines (e.g. in the dashboard). We now also create database entries before sending requests to external service (e.g. Testing Farm), therefore failures will be more visible in the dashboard.
RELEASE NOTES END

### DIFF
--- a/alembic/versions/68ce6a306b30_add_koji_build_groups.py
+++ b/alembic/versions/68ce6a306b30_add_koji_build_groups.py
@@ -1,0 +1,449 @@
+"""Add Koji build groups
+
+Revision ID: 68ce6a306b30
+Revises: 9fe0e970880f
+Create Date: 2022-11-29 10:24:00.722554
+
+"""
+import collections
+import enum
+import itertools
+from datetime import datetime
+from typing import TYPE_CHECKING, List
+
+from packit_service.models import ProjectAndTriggersConnector
+
+from alembic import op
+import sqlalchemy as sa
+import sqlalchemy.orm
+from sqlalchemy import (
+    Column,
+    Integer,
+    String,
+    DateTime,
+    Enum,
+    Table,
+    ForeignKey,
+    JSON,
+    Text,
+    Boolean,
+)
+from sqlalchemy.orm import relationship
+from sqlalchemy.ext.declarative import declarative_base
+
+if TYPE_CHECKING:
+    Base = object
+else:
+    Base = declarative_base()
+
+
+# revision identifiers, used by Alembic.
+revision = "68ce6a306b30"
+down_revision = "9fe0e970880f"
+branch_labels = None
+depends_on = None
+
+
+tf_copr_association_table = Table(
+    "tf_copr_build_association_table",
+    Base.metadata,  # type: ignore
+    Column("copr_id", ForeignKey("copr_build_targets.id"), primary_key=True),
+    Column("tft_id", ForeignKey("tft_test_run_targets.id"), primary_key=True),
+)
+
+
+class JobTriggerModelType(str, enum.Enum):
+    pull_request = "pull_request"
+    branch_push = "branch_push"
+    release = "release"
+    issue = "issue"
+
+
+class JobTriggerModel(Base):
+    __tablename__ = "job_triggers"
+    id = Column(Integer, primary_key=True)  # our database PK
+    type = Column(Enum(JobTriggerModelType))
+    trigger_id = Column(Integer, index=True)
+
+    runs = relationship("PipelineModel", back_populates="job_trigger")
+
+
+class PipelineModel(Base):
+    __tablename__ = "pipelines"
+    id = Column(Integer, primary_key=True)  # our database PK
+    # datetime.utcnow instead of datetime.utcnow() because it's an argument to the function,
+    # so it will run when the model is initiated, not when the table is made
+    datetime = Column(DateTime, default=datetime.utcnow)
+
+    job_trigger_id = Column(Integer, ForeignKey("job_triggers.id"))
+    job_trigger = relationship("JobTriggerModel", back_populates="runs")
+
+    srpm_build_id = Column(Integer, ForeignKey("srpm_builds.id"), index=True)
+    srpm_build = relationship("SRPMBuildModel", back_populates="runs")
+    copr_build_group_id = Column(
+        Integer, ForeignKey("copr_build_groups.id"), index=True
+    )
+    copr_build_group = relationship("CoprBuildGroupModel", back_populates="runs")
+    koji_build_group_id = Column(
+        Integer, ForeignKey("koji_build_groups.id"), index=True
+    )
+    koji_build_group = relationship("KojiBuildGroupModel", back_populates="runs")
+    koji_build_id = Column(Integer, ForeignKey("koji_build_targets.id"), index=True)
+    koji_build = relationship("KojiBuildTargetModel", back_populates="runs")
+    test_run_group_id = Column(
+        Integer, ForeignKey("tft_test_run_groups.id"), index=True
+    )
+    test_run_group = relationship("TFTTestRunGroupModel", back_populates="runs")
+    sync_release_run_id = Column(
+        Integer, ForeignKey("sync_release_runs.id"), index=True
+    )
+    sync_release_run = relationship("SyncReleaseModel", back_populates="runs")
+
+
+class TestingFarmResult(str, enum.Enum):
+    __test__ = False
+
+    new = "new"
+    queued = "queued"
+    running = "running"
+    passed = "passed"
+    failed = "failed"
+    skipped = "skipped"
+    error = "error"
+    unknown = "unknown"
+    needs_inspection = "needs_inspection"
+
+
+class TFTTestRunTargetModel(ProjectAndTriggersConnector, Base):
+    __tablename__ = "tft_test_run_targets"
+    id = Column(Integer, primary_key=True)
+    pipeline_id = Column(String, index=True)
+    identifier = Column(String)
+    commit_sha = Column(String)
+    status = Column(Enum(TestingFarmResult))
+    target = Column(String)
+    web_url = Column(String)
+    # datetime.utcnow instead of datetime.utcnow() because its an argument to the function
+    # so it will run when the model is initiated, not when the table is made
+    submitted_time = Column(DateTime, default=datetime.utcnow)
+    data = Column(JSON)
+    tft_test_run_group_id = Column(Integer, ForeignKey("tft_test_run_groups.id"))
+    copr_builds = relationship(
+        "CoprBuildTargetModel",
+        secondary=tf_copr_association_table,
+        backref="tft_test_run_targets",
+    )
+
+    group_of_targets = relationship(
+        "TFTTestRunGroupModel", back_populates="tft_test_run_targets"
+    )
+
+
+class BuildStatus(str, enum.Enum):
+    success = "success"
+    pending = "pending"
+    failure = "failure"
+    error = "error"
+    waiting_for_srpm = "waiting_for_srpm"
+
+
+class CoprBuildTargetModel(ProjectAndTriggersConnector, Base):
+    __tablename__ = "copr_build_targets"
+    id = Column(Integer, primary_key=True)
+    build_id = Column(String, index=True)  # copr build id
+
+    # commit sha of the PR (or a branch, release) we used for a build
+    commit_sha = Column(String, index=True)
+    # what's the build status?
+    status = Column(Enum(BuildStatus))
+    # chroot, but we use the word target in our docs
+    target = Column(String)
+    # URL to copr web ui for the particular build
+    web_url = Column(String)
+    # url to copr build logs
+    build_logs_url = Column(String)
+    # for monitoring: time when we set the status about accepted task
+    task_accepted_time = Column(DateTime)
+    # datetime.utcnow instead of datetime.utcnow() because its an argument to the function
+    # so it will run when the copr build is initiated, not when the table is made
+    build_submitted_time = Column(DateTime, default=datetime.utcnow)
+    build_start_time = Column(DateTime)
+    build_finished_time = Column(DateTime)
+
+    # project name as shown in copr
+    project_name = Column(String)
+    owner = Column(String)
+
+    # metadata for the build which didn't make it to schema yet
+    # metadata is reserved to sqlalch
+    data = Column(JSON)
+
+    # info about built packages we get from Copr, e.g.
+    # [
+    #   {
+    #       "arch": "noarch",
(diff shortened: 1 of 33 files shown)


packit_service/models.py
```diff
@@ -1269,18 +1426,8 @@ def create(
...
-                    trigger_id=run_model.job_trigger.trigger_id,
-                )
-                new_run_model.srpm_build = run_model.srpm_build
-                new_run_model.koji_build = build
-                session.add(new_run_model)
-            else:
-                run_model.koji_build = build
-                session.add(run_model)
+            koji_build_group.koji_build_targets.append(build)
+            session.add(koji_build_group)
```
TomasTomecek, line 1430: nicely simplified

packit_service/service/api/runs.py
```diff
@@ -83,29 +85,37 @@ def process_runs(runs):
...
-                    {
-                        "packit_id": packit_id,
-                        "target": row.target,
-                        "status": row.status,
-                    }
+                group_row = Model.get_by_id(packit_id)
+                target_models = (
+                    group_row.grouped_targets
+                    if isinstance(group_row, GroupModel)
+                    else [group_row]
```
TomasTomecek, line 97: what else could it be?

packit_service/models.py
```diff
@@ -366,6 +379,42 @@ def get_release_tag(self) -> Optional[str]:
...
+        return self.group_of_targets.get_branch_name()
+
+    def get_release_tag(self) -> Optional[str]:
+        return self.group_of_targets.get_release_tag()
+
+
+class GroupModel:
+    @property
+    def grouped_targets(self):
+        raise NotImplementedError
```
TomasTomecek, line 415: please add docstrings for this one too

packit_service/models.py
FrNecas: Just moved elsewhere, to be honest, the duplicating logic is still necessary inside group :D

packit_service/service/api/runs.py
FrNecas: Right, this can now be removed. I added this check because I was doing one type of model at a time, so previously it was a mix of group and target models.

packit_service/worker/handlers/testing_farm.py
```diff
@@ -170,7 +213,17 @@ def run_with_copr_builds(self, targets: List[str], failed: Dict):
             event_data["build_targets_override"] = list(targets_without_builds)
             self.run_copr_build_handler(event_data, len(targets_without_builds))
 
-        for target, copr_build in targets_with_builds.items():
+        if not targets_with_builds:
```
FrNecas, line 216: TODO: if there are copr builds, the handler will trigger for each copr build end, resulting in multiple groups. We should group these together

FrNecas:
TODO: add API endpoints for getting group info

alembic/versions/68ce6a306b30_add_koji_build_groups.py
```diff
@@ -0,0 +1,443 @@
...
+    op.add_column(
+        "pipelines", sa.Column("koji_build_group_id", sa.Integer(), nullable=True)
+    )
+    op.create_foreign_key(
+        None, "pipelines", "koji_build_groups", ["koji_build_group_id"], ["id"]
+    )
+
+    # Group by the same SRPM
+    srpm_ids = collections.defaultdict(list)
+    for koji_build in session.query(KojiBuildTargetModel).all():
```
jpopelka, line 379: The `Query` is already iterable so you don't need `all()` (to create a list from it first) if you are going to just iterate over it, see [for example](https://github.com/packit/packit-service/blob/6f648528db0535b9bf798b814efb74a5fc895139/packit_service/models.py#L1210)

```suggestion
    for koji_build in session.query(KojiBuildTargetModel):
```

packit_service/models.py
TomasTomecek: noticed after I commented :D but still, this method is simpler

FrNecas:
Rebased, adjusted based on feedback, added API. I still need to go through the openshift test failures

tests_openshift/database/test_models.py
```diff
@@ -793,18 +804,17 @@ def test_merged_runs(clean_before_and_after, few_runs):
         merged_run = PipelineModel.get_merged_run(run_id)
         srpm_build_id = merged_run.srpm_build_id
 
-        # for different_pr (i=1) there are Copr builds twice, since the second
-        # run of TFT from the same Copr build produces new row with same SRPM
-        # and Copr IDs, but different Testing Farm IDs
-        # ^ handled in API by iterating over set of IDs instead of list
-        assert len(merged_run.copr_build_id) == 2 * i
+        # Since the introduction of build groups, the builds are grouped
+        assert len(merged_run.copr_build_group_id) == 1
```
FrNecas, line 808: I had to change this test because now the pipelines creation/copying works differently. Do you think that this is fine? @mfocko I see you were the author of the change, I'd appreciate your opinion

FrNecas:
Hm, one last failing openshift test that did not run locally in `make check-db`. Hopefully it will be a simple fix

tests_openshift/database/test_models.py
mfocko: I'd say it's fine.

FrNecas:
Rebased again, some minor merge conflicts, had to adjust @lachmanfrantisek 's usage queries a bit. Let's see if everything passes now.

FrNecas:
I've tried running the migrations on prod data locally in the container. It took ~30 minutes, the peak memory usage was around 3GB (we will surely have to increase limits for moving stable). The first migration (keeping track of related copr builds in TF model) took 10 minutes, grouping copr builds took 15 minutes, the remaining 2 migrations were quite quick. Is this acceptable? I feel like that 30 minutes of downtime is not great but I am not sure if it can be improved.

One idea that I got - the first migration (adding TF mapping to copr explicitly) could be done separately without the groups, this would split the downtime into 2 periods (10 minutes and 20 minutes).

jpopelka:
Did you say you'd like to merge it this or next week? If next, then we can discuss on the arch meeting.
By "increase limits for moving stable" do you mean moving stable branches? I'm asking because I don't see how that is related to running the migrations.

FrNecas:
Let's try to aim to review this week and merge next Tuesday. Sorry by not being clear - by increasing memory limits I meant for the deployment so that the migration doesn't run out of memory

jpopelka:
I slept on in and now I understand :+1:
Sure we can increase the limit anytime - I will actually redeploy the packit-service because of packit/deployment#433 at some point so I can increase the limit at the same time.

packit_service/worker/handlers/testing_farm.py
```diff
@@ -116,6 +120,44 @@ def get_checkers() -> Tuple[Type[Checker], ...]:
             CanActorRunJob,
         )
 
+    def _get_or_create_group(
+        self,
+        targets: List[str],
+        copr_build: Optional[CoprBuildTargetModel],
+        copr_builds_for_targets: List[CoprBuildTargetModel],
+    ) -> TFTTestRunGroupModel:
```
lbarcziova, line 128: could we please have a docstring here since from the first look it may not be clear what the arguments really are and generally how is the creation of the group done?

packit_service/worker/helpers/build/copr_build.py
```diff
@@ -537,10 +596,62 @@ def submit_copr_build(self, script: Optional[str] = None) -> Tuple[int, str]:
 
         return build.id, self.api.copr_helper.copr_web_build_url(build)
 
-    def handle_build_submit_error(self, ex) -> TaskResults:
+    def run_copr_build(self) -> TaskResults:
```
lbarcziova, line 599: I'm a little confused here; if I am checking correctly, this method was removed already, maybe a mistake when rebasing?

packit_service/worker/helpers/build/copr_build.py
```diff
@@ -442,10 +446,65 @@ def run_copr_build_from_source_script(self) -> TaskResults:
             url=get_srpm_build_info_url(self.srpm_model.id),
         )
 
-        self.handle_rpm_build_start(build_id, web_url, waiting_for_srpm=True)
+        self.handle_rpm_build_start(group, build_id, web_url)
 
         return TaskResults(success=True, details={})
 
+    def _get_or_create_build_group(
+        self, wait_for_srpm: bool = False
```
lbarcziova, line 454: I think that now as we switched to SRPMs in Copr, `wait_for_srpm` should be always `True`; we should have removed this already, so we can do a follow-up PR to not add more work to you

packit_service/worker/handlers/testing_farm.py
```diff
@@ -171,7 +214,17 @@ def run_with_copr_builds(self, targets: List[str], failed: Dict):
...
 
-        for target, copr_build in targets_with_builds.items():
+        if not targets_with_builds:
+            return
+
+        group = self._get_or_create_group(
+            list(targets_with_builds.keys()),
+            copr_build,
+            list(targets_with_builds.values()),
+        )
```
lbarcziova, line 228: it is not clear to me why we have to pass here the lastly assigned `copr_build`, which should be part of the `targets_with_builds` dictionary already; how does this work? :thinking:

lbarcziova (commented):
very nice work with the migrations and improvements (also docs) in general, thanks!!

packit_service/worker/handlers/testing_farm.py
FrNecas: The second argument to `get_or_create_group` (in other words, the latest copr build) is used to get the `PipelineModel` to attach the tests to. I think that with copr groups, this `PipelineModel` is probably shared between all builds in `targets_with_builds` (they are probably part of the same group, i.e. attached to the same pipeline). This would mean that we can pick any build from `targets_with_builds`, I am just not completely sure if there can't be some edge-case where it would result in a different behavior. But it sounds like that the second argument is not necessary :thinking:

packit_service/worker/helpers/build/copr_build.py
FrNecas: I created https://github.com/packit/packit-service/issues/1866 as a follow-up

jpopelka:
I ran the migrations locally and it took 40 minutes and the peak mem consumption was around 3.7Gi so we might raise the current limit of 4Gi  even more just to be on the safe side.

I generated this diagram, which you (anyone) might be interested in:

packit - public-groups

FrNecas:
That diagram is awesome! Could we add it to the CONTRIBUTING docs or somewhere else? I know that it may become outdated with migrations but even then it gives at least a rough idea of the relations

packit_service/worker/handlers/testing_farm.py
```diff
@@ -135,7 +135,7 @@ def _get_or_create_group(
                 trigger_id=self.db_trigger.id,
             )
             if self.testing_farm_job_helper.skip_build or not copr_build
-            else copr_build.runs[-1]
+            else copr_build.group_of_targets.runs[-1]
```
mfocko: I love how Python has a lot of useless methods, yet you cannot do `.last()` like a civilized person…

---

I might've been corrupted by either Kotlin or Rust…

packit_service/worker/helpers/build/copr_build.py
```diff
@@ -537,10 +596,62 @@ def submit_copr_build(self, script: Optional[str] = None) -> Tuple[int, str]:
 
         return build.id, self.api.copr_helper.copr_web_build_url(build)
 
-    def handle_build_submit_error(self, ex) -> TaskResults:
+    def run_copr_build(self) -> TaskResults:
+        """
+        Run copr build using SRPM built by us.
+        """
+        self._report(
```
mfocko: Can't we get rid of it now?

I'd think we can get rid of the whole function :eyes:

mfocko: Resolved by d09fdac9d6c4ea65185670421c937e34f1e570a0

packit_service/worker/handlers/testing_farm.py
```diff
@@ -134,18 +144,18 @@ def _get_or_create_group(
                 type=self.db_trigger.job_trigger_model_type,
                 trigger_id=self.db_trigger.id,
             )
-            if self.testing_farm_job_helper.skip_build or not copr_build
-            else copr_build.group_of_targets.runs[-1]
+            if self.testing_farm_job_helper.skip_build or not builds
+            # All the builds should be in the same copr build group, therefore
+            # connected to the same pipeline, just take the first one
+            else list(builds.values())[0].group_of_targets.runs[-1]
```
mfocko, line 150: Do you really need a list if you take just the first one?

mfocko (commented):
lgtm

packit_service/worker/handlers/testing_farm.py
FrNecas: Oh, `next()` will probably be better, good point

lbarcziova (approved):
:rocket:

TomasTomecek (approved):
ship it to the cloud :sunglasses:

FrNecas:
Okay, now that prod seems to be fixed, let's rock!

PR #1788, opened 2022-11-29
Fix KT-002-PACKIT-SERVICE-4YN

TomasTomecek:
TODO:

Two fixes here: check commit messages :)

Fixes KT-002-PACKIT-SERVICE-4YN

### DIFF
--- a/packit_service/worker/checker/vm_image.py
+++ b/packit_service/worker/checker/vm_image.py
@@ -82,16 +82,25 @@ class IsCoprBuildForChrootOk(Checker, GetVMImageBuildReporterFromJobHelperMixin)
             return False
 
         for build in copr_builds:
-            if build.project_name in self.job_config.packages:
-                job_config = self.job_config.packages[build.project_name]
+            # make sure that the build is done in the configured Copr project
+            if (
+                build.project_name == self.job_config.project
+                and build.owner == self.job_config.owner
+            ):
+                # we trust the user has correctly selected the binary RPM to be installed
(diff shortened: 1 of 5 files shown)


TomasTomecek:
14 minutes later... queued :<

TomasTomecek:
recheck

packit_service/worker/checker/vm_image.py
```diff
@@ -82,15 +82,21 @@ def pre_check(
...
-                    build.target == job_config.copr_chroot
-                    and build.status == BuildStatus.success
-                ):
-                    return True
+            for copr_package in build.built_packages:
+                copr_package_name = copr_package["name"]
+                if copr_package["arch"] == "src":
+                    # we care about built packages, not SRPM
+                    continue
+                if copr_package_name in self.job_config.packages:
```
majamassarini, line 90:
```suggestion
```

 I am testing your PR with my local test and now it does not pass anymore the copr build check (which is awkward but was passing with the old code 😅). 
The problem is this line.
In my project the `copr_project_name` is `python-knx-stack` and the `job_config.packages` is `[knx-stack]`. 
If I remove this line everything seems ok. Do you think we can safely remove it?

TomasTomecek: we have discussed on Matrix

agreed this is not a correct solution

TomasTomecek:
recheck

majamassarini (approved):
Thanks! Now is working :)

PR #1789, opened 2022-11-29
Fix make check-db to run compose down

FrNecas:
I don't really see any reason for `project=` something to be in the Makefile, I believe that it may have been an unfortunate paste from the clipboard :)


lbarcziova (approved):
:+1:

Issue #1747, opened 2022-11-09
Checks for bodhi updates have to give feedback to the user


majamassarini:
Follow-up of https://github.com/packit/packit-service/pull/1729

All classes in [this module](https://github.com/packit/packit-service/blob/1ca6834a8bf83d4c475a2a114461d0843569a1d9/packit_service/worker/checker/bodhi.py) should not only log failures but they should also give feedback to the user in the same way others checker classes do, like [here](https://github.com/packit/packit-service/blob/90d0a9a01d8d45ff3fb0dbb2dcde020a82b2993f/packit_service/worker/checker/koji.py#L53)

majamassarini:
Fixed with #1796

majamassarini:
Fixed with https://github.com/packit/packit-service/pull/1796

PR #1796, opened 2022-12-02
Retrigger koji and bodhi services by issue comment

### LINKED ISSUE #1536: Implement retriggering the downstream jobs via issue comments in issue_repository
Implement retriggering downstream Koji build and Bodhi update jobs via issue comments in issues in `issue_repository` specified by the user:
- [ ] retriggering will be done via dedicated Packit commands - e.g `/packit koji-build` `/packit bodhi-update`
- [ ] the handling should be very similar to `/packit propose-downstream`
- [ ] the user needs to enable Packit app on the `issue_repostitory` so that we get the comment events => update the docs with this requirement
- [ ] user who triggers the jobs should be allowlisted - write access to `issue_repository` is required - update the docs so that users know about this when setting `issue_repository`

majamassarini:
Fixes #1536 #1747 

Merge before packit/packit.dev#567

---

RELEASE NOTES BEGIN
Users can now re-trigger a *Bodhi update* or a *Koji build* writing comments like `/packit create-update` and `/packit koji-update` in a issue (opened by Packit in the repository issue if anything goes wrong during the *propose-downstream*)
RELEASE NOTES END


majamassarini:
recheck

packit_service/worker/checker/bodhi.py
```diff
@@ -65,27 +80,59 @@ class IsKojiBuildCompleteAndBranchConfiguredCheckService(
...
             )
-            return False
+            if not has_write_access:
+                logger.warning(
+                    f"Re-triggering Bodhi update through comment in "
+                    f"repo {self.project.repo} and issue {self.data.issue_id} "
+                    f"is not allowed for the user {self.actor} "
+                    f"which has not write permissions on the project."
+                )
+                return False
```
TomasTomecek, line 103: we should comment in the issue when this happens

lbarcziova (commented):
Maybe I am missing something, but where/how is done the mapping from the issue (repository) to the dist-git repo? (If I remember correctly we were discussing in the past this could be done via parsing the issue title/description and including the dist-git repo URL before)

Could we please also have some integration test for this use case?

packit_service/worker/checker/bodhi.py
majamassarini: Ok! I can also add the feedback for the other checks in the same way and we can resolve the #1747

majamassarini:
> Maybe I am missing something, but where/how is done the mapping from the issue (repository) to the dist-git repo? (If I remember correctly we were discussing in the past this could be done via parsing the issue title/description and including the dist-git repo URL before)
> 
> Could we please also have some integration test for this use case?

Probably I am the one who is missing something here.
I can not remember if the same repository issue could be used by multiple dist-git repos?
If not, since it should have a packit.yaml, could we find the dist-git reference there?

lbarcziova:
Maybe @lachmanfrantisek can correct me, but I think the same issue repository can be configured for multiple dist-git repositories. And the issue repository also doesn't need to have the packit configuration (and we should not look into it there).

majamassarini:
> Maybe @lachmanfrantisek can correct me, but I think the same issue repository can be configured for multiple dist-git repositories. And the issue repository also doesn't need to have the packit configuration (and we should not search for it there).

I may be wrong but I believe that without `packit.yaml` we do not react to events in that repository 🤔.

lbarcziova:
> I may be wrong but I believe that without packit.yaml we do not react to events in that repository thinking.

Currently, we probably do not, therefore we need to get the dist-git repository and get the package config from there.

packit_service/worker/checker/bodhi.py
```diff
@@ -65,27 +82,92 @@ class IsKojiBuildCompleteAndBranchConfiguredCheckService(
...
             )
-            return False
+            if not has_write_access:
+                msg = (
+                    f"Re-triggering Bodhi update through comment in "
+                    f"repo **{self.project_url}** and issue **{self.data.issue_id}** "
+                    f"is not allowed for the user *{self.actor}* "
+                    f"which has not write permissions on the project."
+                )
+                logger.warning(msg)
```
TomasTomecek, line 108: I would make this info, since this case can easily happen or be an accident. Please advise what was your reason to pick warning

packit_service/worker/checker/bodhi.py
```diff
@@ -65,27 +82,92 @@ class IsKojiBuildCompleteAndBranchConfiguredCheckService(
...
+                f"repo {self.project_url} and issue {self.data.issue_id} "
+                f"by {self.actor}."
             )
-            return False
+            if not has_write_access:
+                msg = (
+                    f"Re-triggering Bodhi update through comment in "
+                    f"repo **{self.project_url}** and issue **{self.data.issue_id}** "
+                    f"is not allowed for the user *{self.actor}* "
+                    f"which has not write permissions on the project."
```
TomasTomecek, line 106: I like how descriptive this message is

TomasTomecek (commented):
Looks really good!

packit_service/worker/checker/bodhi.py
majamassarini: It is always hard to decide the level of a log 😅 I just thought that I want to read this message when info messages are turned off. I felt it to be more important than just an info. But probably you are right this could be too much noise. So let it be an info!

lachmanfrantisek:
> Maybe @lachmanfrantisek can correct me, but I think the same issue repository can be configured for multiple dist-git repositories. And the issue repository also doesn't need to have the packit configuration (and we should not search for it there).

:+1: Yes, that's true. It's a valid use case currently.

packit_service/worker/handlers/mixin.py
```diff
@@ -232,6 +232,19 @@ class GetKojiBuildDataFromKojiServiceMultipleBranches(
...
+    def build(self):
+        # call it every time since dist_git_branch reference can change
+        build = self.koji_helper.get_latest_build_in_tag(
+            package=self.project.repo,
+            tag=self._dist_git_branch,
+        )
+        if not build:
+            raise PackitException(
+                f"No build found for package={self.project.repo} and tag={self.dist_git_branch}"
+            )
```
TomasTomecek, line 245: I am not familiar with this code base: does this exception imply there is a bug in our code? meaning, if there is no such build, we should have create it?

packit_service/worker/checker/bodhi.py
```diff
@@ -65,27 +83,92 @@ class IsKojiBuildCompleteAndBranchConfiguredCheckService(
...
+        has_write_access = self.project.has_write_access(user=self.actor)
+        if self.data.event_type in (
+            IssueCommentEvent.__name__,
+            IssueCommentGitlabEvent.__name__,
+        ):
+            logger.debug(
+                f"Re-triggering Bodhi update through comment in "
+                f"repo {self.project_url} and issue {self.data.issue_id} "
+                f"by {self.actor}."
             )
```
TomasTomecek, line 101: very nice, this will totally help identify potential problems

packit_service/worker/handlers/mixin.py
```diff
@@ -82,47 +83,97 @@ def koji_build_helper(self) -> KojiBuildJobHelper:
         return self._koji_build_helper
 
 
-class GetKojiBuildData(Protocol):
+@dataclass
+class KojiBuildData:
```
TomasTomecek, line 87: sweet, dataclasses!

TomasTomecek (approved):

packit_service/worker/handlers/mixin.py
majamassarini: Yes, this mixins are mostly used in the bodhi update handlers/checkers and we should create a koji build before calling them. So probably if we don't find any build we missed something or servers are not yet updated/synced?

packit_service/worker/mixin.py
```diff
@@ -242,6 +266,36 @@ def issue(self):
...
+        | --------------- | ----- |
+        | `f37` | `` |
+
+
+        You can retrigger the update by adding a comment (`/packit propose-downstream`) into this issue.
+        """
+        branches = []
+        issue = self.data.project.get_issue(self.data.issue_id)
+        for line in issue.description.splitlines():
+            if m := re.match(r"\s*\| `(\S+)` \|", line):
```
lbarcziova, line 294: I am a little confused, the comment in the docstring is for propose-downstream, but the `GetBranchesFromIssueMixin` is used for triggering Bodhi updates and Koji builds...and for them, the format of the error messages is different if I am checking it correctly :thinking:

packit_service/worker/jobs.py
```diff
@@ -283,6 +287,17 @@ def report_task_accepted(
             task_accepted_time, handler_kls, number_of_build_targets
         )
 
+    def search_distgit_config_in_issue(self):
+        issue = self.event.project.get_issue(self.event.issue_id)
+        if m := match(r"[\w\s-]+dist-git \((\S+)\):", issue.description):
```
lbarcziova, line 292: could you please add a comment here on what is the format of the description where we are checking the match?

packit_service/worker/mixin.py
lbarcziova: yes, that should work, thanks!

packit_service/worker/handlers/mixin.py
TomasTomecek: oki, understood

just let's be sure that when this happens, we have all the info to diagnose the problem and figure out a solution

majamassarini:
recheck

packit_service/constants.py
```diff
@@ -26,6 +26,17 @@
 TESTING_FARM_INSTALLABILITY_TEST_URL = "https://gitlab.com/testing-farm/tests"
 TESTING_FARM_INSTALLABILITY_TEST_REF = "main"
 
+MSG_PROPOSE_DOWNSTREAM_ERROR_HEADER = (
```
lbarcziova, line 29:
```suggestion
MSG_DOWNSTREAM_JOB_ERROR_HEADER = (
```

lbarcziova (approved):
:rocket:

TomasTomecek (approved):

PR #1800, opened 2022-12-07
change(celery): retry backoff 3 → 7

TomasTomecek:
This also changes retry setup for creation of bodhi updates to 5.
All other tasks are still on 2.
    
When will the retrying happen?
    
    0-7s, 0-14s, 0-28s, 0-48s, 0-96s
    
We have jitter on (celery's default) so it's actual `random(0, value)`.
    
This is done mainly to account for koji build tagging as it can take
tenths of seconds easily. We also wanted to increase the backoff since
it makes more sense to retry later: the initial condition may not be
resolved right away.
    
Brutally related #1615

// meant to be a oneliner

---

RELEASE NOTES BEGIN
We've increased internal task retry backoff time in Packit GitHub app from 3 to 7 seconds. We hope this will increase success for network flakes and *random* infrastructure issues. Creation of bodhi updates should be now more reliable too as Packit will try more times (from 2 to 5).
RELEASE NOTES END

### DIFF
--- a/packit_service/constants.py
+++ b/packit_service/constants.py
@@ -72,11 +72,14 @@ COPR_FAILED_STATE = "failed"
 COPR_API_SUCC_STATE = 1
 COPR_API_FAIL_STATE = 2
 
+# Retry 2 times
 DEFAULT_RETRY_LIMIT = 2
 # Retry more times for outages
 DEFAULT_RETRY_LIMIT_OUTAGE = 5
-# retry in 3s, 6s
-DEFAULT_RETRY_BACKOFF = 3
+# retry in 0-7s, 0-14s, 0-28s, 0-48s, 0-96s
+# because jitter is enabled by default, celery makes these retries random:
+# https://docs.celeryq.dev/en/latest/userguide/tasks.html#Task.retry_jitter
+DEFAULT_RETRY_BACKOFF = 7
 RETRY_INTERVAL_IN_MINUTES_WHEN_USER_ACTION_IS_NEEDED = 10
 BASE_RETRY_INTERVAL_IN_MINUTES_FOR_OUTAGES = 1
 BASE_RETRY_INTERVAL_IN_SECONDS_FOR_INTERNAL_ERRORS = 10
--- a/packit_service/worker/tasks.py
+++ b/packit_service/worker/tasks.py
@@ -100,10 +100,15 @@ def setup_loggers(logger, *args, **kwargs):
 
 class HandlerTaskWithRetry(Task):
     autoretry_for = (Exception,)
-    retry_kwargs = {
-        "max_retries": int(getenv("CELERY_RETRY_LIMIT", DEFAULT_RETRY_LIMIT))
-    }
+    max_retries = int(getenv("CELERY_RETRY_LIMIT", DEFAULT_RETRY_LIMIT))
+    retry_kwargs = {"max_retries": max_retries}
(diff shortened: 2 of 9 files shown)


packit_service/worker/tasks.py
```diff
@@ -100,12 +100,19 @@ def setup_loggers(logger, *args, **kwargs):
 
 class HandlerTaskWithRetry(Task):
     autoretry_for = (Exception,)
+    max_retries = int(getenv("CELERY_RETRY_LIMIT", DEFAULT_RETRY_LIMIT))
     retry_kwargs = {
         "max_retries": int(getenv("CELERY_RETRY_LIMIT", DEFAULT_RETRY_LIMIT))
     }
```
lbarcziova, line 106:
```suggestion
    max_retries = int(getenv("CELERY_RETRY_LIMIT", DEFAULT_RETRY_LIMIT))
    retry_kwargs = {
        "max_retries": max_retries
    }
```

packit_service/worker/tasks.py
```diff
@@ -100,12 +100,19 @@ def setup_loggers(logger, *args, **kwargs):
...
     retry_backoff = int(getenv("CELERY_RETRY_BACKOFF", DEFAULT_RETRY_BACKOFF))
 
 
+class BodhiHandlerTaskWithRetry(HandlerTaskWithRetry):
+    # hardcode for creating bodhi updates to account for the tagging race condition
+    max_retries = 5
+    retry_kwargs = {"max_retries": 5}
```
lbarcziova, line 113:
```suggestion
    max_retries = 5
    retry_kwargs = {"max_retries": max_retries}
```

lbarcziova (approved):
LGTM :+1:

packit_service/worker/tasks.py
TomasTomecek: 🤦

packit_service/constants.py
```diff
@@ -72,11 +72,14 @@
...
+# Retry 2 times
 DEFAULT_RETRY_LIMIT = 2
 # Retry more times for outages
 DEFAULT_RETRY_LIMIT_OUTAGE = 5
-# retry in 3s, 6s
-DEFAULT_RETRY_BACKOFF = 3
+# retry in 0-7s, 0-14s, 0-28s, 0-48s, 0-96s
+# because jitter is enabled by default, celery makes these retries random:
+# https://docs.celeryq.dev/en/latest/userguide/tasks.html#Task.retry_jitter
+DEFAULT_RETRY_BACKOFF = 7
```
lachmanfrantisek, line 82: Because of the jitter, it can still end really soon when we have bad luck, can't it?

(Just wondering if jitter helps here.)

lachmanfrantisek (approved):

packit_service/constants.py
TomasTomecek: yes, that's true, though very unlikely, we'd need to roll 1-5s 6-7 times in a row; and especially once we get to 0-28, 0-48, 0-96 that's becoming 0.001% :D

completely agree with you though if we could specify the jitter function that would be awesome

TomasTomecek: tl;dr celery's reasoning for jitter is not to cause a barrage of requests in a given timeframe and rather spread these, which makes sense

I'm totally up to jitter=false in the next PR to try it out

lachmanfrantisek: You are right, it's really unlikely and if they hit at least one higher value, we are good.

lachmanfrantisek: And we can improve if needed.

PR #1809, opened 2022-12-14
Enforce HTTPS with HSTS

jpopelka:
```
$ mod_wsgi-express-3 start-server --help
--https-only          Flag indicating whether any requests made using a HTTP
                      request over the non secure connection should be
                      redirected automatically to use a HTTPS request over
                      the secure connection.
--hsts-policy PARAMS  Specify the HSTS policy that should be applied when
                      HTTPS only connections are being enforced.
```
https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security
https://cheatsheetseries.owasp.org/cheatsheets/HTTP_Strict_Transport_Security_Cheat_Sheet.html

Enterprise Security Standard (ESSv9) requirement.


TomasTomecek (approved):

jpopelka:
the tests are broken in main as well, merged manually

PR #1812, opened 2022-12-15
Specify f35 as dist_git_branches in bodhi update test

lbarcziova:
CI started failing for this test because the test data is for f35 which is now EOL.

---

RELEASE NOTES BEGIN
n/a
RELEASE NOTES END


FrNecas:
Oh, I just ran into this and was starting to wonder what is wrong. Thanks Laura!

FrNecas (approved):

PR #1817, opened 2022-12-20
Bump the Fedora version in test messages

lachmanfrantisek:

Fixes the reverse-dep tests in Packit.

e.g. https://softwarefactory-project.io/zuul/t/packit-service/build/20372f4534794001abdba097403a1237/console

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -385,8 +385,8 @@ def koji_build_start_rawhide():
 
 
 @pytest.fixture()
-def koji_build_start_f35():
-    with open(DATA_DIR / "fedmsg" / "koji_build_start_f35.json") as outfile:
+def koji_build_start_f36():
+    with open(DATA_DIR / "fedmsg" / "koji_build_start_f36.json") as outfile:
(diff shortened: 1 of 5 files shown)


majamassarini (approved):
Am I wrong or in this way, when the f36 will be no more mantained, we need to fix this again?

lachmanfrantisek:
> Am I wrong or in this way, when the f36 will be no more mantained, we need to fix this again?

Sadly, yes. But I've preferred this so we don't skip any code...

majamassarini:
> > Am I wrong or in this way, when the f36 will be no more mantained, we need to fix this again?
> 
> Sadly, yes. But I've preferred this so we don't skip any code...

I tried to mock the `get_branches` function here (https://github.com/packit/packit-service/blob/ee8f798a792167801f72f0b4ed1a5427033ada22/packit_service/worker/checker/bodhi.py#L41) to return always  `{'f35', 'f36'}`. In this way I think we don't need to skip code and we will not need to modify it in the future.
But I give up because of flexmock... I was not able to do that. Maybe you can? 😅

Issue #1224, opened 2021-09-23
Mismatch between application code and executed tests


psss:
After the recent change in packit/packit/pull/1344 (merge pull request into the target branch before building the srpm) there is a mismatch between the features built in the package and tests executed in the testing farm. This appeared in the recent pull request here:

 * https://github.com/psss/tmt/pull/892

A new feature introduced in the default branch causes tests to unexpectedly fail because test code has been modified in `master` as well. Packit should request Testing Farm to execute tests from the corresponding branch instead.

psss:
@FrNecas, @thrix, fyi.

thrix:
oh well, if the merged to target branch is not exposed in the repo (and connot be), we will need to implement this merging on our side. I do not like it much. Seems we have a mtg today, let's discuss there.

TomasTomecek:
Let us know if we should participate in that discussion.

I'd say this race condition is gonna be hard to solve properly since the merge would need to happen twice: once on our side and the second time on your.

Should we also specify commit SHA of the contributor's branch?

psss:
Using the `refs/pull/892/merge` is not an option here?

mfocko:
> Using the `refs/pull/892/merge` is not an option here?

If you mean the test merge commit that GitHub or GitLab (on request) provide, the problem with them is that they're quite unpredictable (from outdated to having already different revisions that are problematic to track), by local merging when building SRPM we log the heads of branches we merge, so it's easier to both debug and track what the SRPM contains.

lachmanfrantisek:
We just agreed on the following:
* We will add support to make the merge configurable: #1228 
* Testing Farm will implement the support on their side.

mfocko:
We provide `PACKIT_TARGET_SHA` to the testing farm, should be deployed to prod on Tuesday.

TomasTomecek:
shall we close this one?

mfocko:
not done on TF side afaik

thrix:
indeed :( https://issues.redhat.com/browse/TFT-756 :(

thrix:
@TomasTomecek @mfocko did we discuss why for Testing Farm it would be not enough to use `refs/pull/<number>/merge` ? That might would be so much cleaner for us ....

mfocko:
@thrix mainly inconsistency (IIRC cockpit team was trying to use GitHub merges too, see discussion packit/packit-service#245).

[...]

PR #1837, opened 2023-01-11
fix(tf): pass merge SHA to the TF API

mfocko:
This change reflects introduction of the „new“ API parameter of the TF API. Merging is not yet implemented on the Testing Farm' side, just the API support.

Related MR: https://gitlab.com/testing-farm/api/-/merge_requests/30

TODO:

Related to #1224

### DIFF
--- a/packit_service/worker/helpers/testing_farm.py
+++ b/packit_service/worker/helpers/testing_farm.py
@@ -117,6 +117,10 @@ class TestingFarmJobHelper(CoprBuildJobHelper):
         return self.job_config.skip_build
 
     @property
+    def custom_fmf(self) -> bool:
+        return bool(self.job_config.fmf_url)
+
+    @property
     def fmf_url(self) -> str:
         return (
             self.job_config.fmf_url
@@ -131,7 +135,7 @@ class TestingFarmJobHelper(CoprBuildJobHelper):
 
     @property
     def fmf_ref(self) -> str:
-        if self.job_config.fmf_url:
+        if self.custom_fmf:
             return self.job_config.fmf_ref
 
         return self.metadata.commit_sha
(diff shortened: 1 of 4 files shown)


jpopelka (approved):

packit_service/worker/helpers/testing_farm.py
```diff
@@ -290,6 +294,26 @@ def _payload_without_token(payload: Dict) -> Dict:
...
+
+            # We assign a commit hash for merging only if:
+            # • there are no custom fmf tests set
+            # • we merge and have a PR
+            if (
+                not self.custom_fmf
+                and self.job_config.merge_pr_in_ci
+                and self.target_branch_sha
+            ):
+                fmf["merge_sha"] = self.target_branch_sha
```
TomasTomecek, line 310: what are the expectations here?

I thought that we would send 2 hashes to TF so they know what to merge

tests/unit/test_testing_farm.py
```diff
@@ -677,35 +678,46 @@ def test_payload(
...
-        (
+        (  # defaulting to the tests in repo, also merging
             None,
             None,
             "https://github.com/packit/packit",
             "feb41e5",
+            True,
         ),
-        (
+        (  # specifying only ref and merging
```
mfocko, line 697: I've almost forgotten, is this a valid use-case? It basically says that tests are in a different branch/tag/commit, but in the same repository; I think that in this case, it should be considered as a custom tests-scenario and we should not pass the `merge_sha`.

---

Although I cannot imagine use-case for tests from different _ref_ and in the same repository at the same time.

packit_service/worker/helpers/testing_farm.py
mfocko: That's what's happening?

1º you can define `fmf_url` (and `fmf_ref`) → custom tests, no merging should happen, cause it's different repo
2º you don't have that defined, you must have enabled merging and present `target_branch_sha` (means it's not running for commit) → we pass `merge_sha`, so the TF can perform the merge

TomasTomecek: ahaaa, so it's confusing naming then: I thought that merge_sha is the hash *after we merge*, not what's meant to be merged where

makes sense to me now

tests/unit/test_testing_farm.py
TomasTomecek: +1, such merge wouldn't make sense and could easily screw things up

TomasTomecek: > Although I cannot imagine use-case for tests from different ref and in the same repository at the same time.

don't underestimate, people are very creative :)

TomasTomecek (approved):

tests/unit/test_testing_farm.py
mfocko: OK, I'm going to update the `custom_fmf` meaning then… To make sure no merging happens with custom ref **∨**  url for fmf

mfocko: Looks like no change is required, when `fmf_url` is not set, the `fmf_ref` is ignored, both before and after this PR.

PR #1841, opened 2023-01-12
Utilize update_release config option

### LINKED ISSUE #1760: `release_suffix` bad behavior in packit-service
In **packit-service** the behavior should be like the following:

-  for `release_suffix = (not empty str)` (release suffix in `.packit.yaml` is a not empty string) the custom `release_suffix` should be used and the spec file should be **bumped** (that means we should change the spec file with the custom *Release*)
-  for `release_suffix == None` (no release suffix keyword  in `.packit.yaml`) the default packit `release_suffix` schema should be used and the spec file should be **bumped** (that means we should change the spec file with the default packit *Release* schema)
-  for `release_suffix == ""` (empty  string for release suffix keyword in `.packit.yaml`) the release found in the spec file should be used and the spec file should not be changed

The Packit API has two flags the `bump` flag and the `release_suffix` flag.
We need to use the bump flag to control changes in the spec file, we can not count on the `release_suffix` flag, otherwise we will affect the CLI tool.
For this reason when *packit-service* finds a `release_suffix=="" (empty)` then it should force the `bump` flag to `False`!

**But** it seems that *packit-service* is not able to distinguish between a `None` or an `""` (empty) `release_suffix`.  

[This PR](https://github.com/packit/packit-service/pull/1744/files) has solved this [issue](https://github.com/packit/packit/issues/1695#issuecomment-1317258776)  for the *csutils* project  but has introduced a new problem for the *cockpit* project.

Before introducing the above PR we were bumping the Release into the specfile (**and also the Version which seems wrong**):

https://copr-dist-git.fedorainfracloud.org/cgit/@cockpit/cockpit-preview/cockpit.git/commit/?id=b6ded94cb02b2491d570f5ec47e536b813adf285

`Version` was 0 instead of 279 and `Release` seems ok .

After the above PR the *packit-service* did not bump the *Revision* into the specfile

https://copr-dist-git.fedorainfracloud.org/cgit/@cockpit/cockpit-preview/cockpit.git/commit/?id=072531b10de3eeeb952b27f358c4b84525ebb790

The **cockpit** project has no `release_suffix` in its `.packit.yaml`.
The **csutils** has a `release_suffix=""` in its `.packit.yaml`

We solved the problem for **csutils** and introduced a bug for the **cockpit** .
I think we are not able to distinguish between a `None` or empty `release_suffix` in *packit-service*. Maybe this is a serialize/deserialize problem.

Action items:

- [ ] find a reliable way, for the *packit-service*, to tell when the **bump** flag for the *Packit API* has to be `True` or `False`
Possible solutions:
    - instead of carry two different meanings into the `release_suffix` flag we could introduce the `bump` flag into the `.packit.yaml` to be more explicit (like we are with the CLI tool)
    - solve the serialization/deserialization problem (if any)
    - introduce another custom syntax for the `release_suffix` keyword which means -> do not bump

- [ ] be sure we are not doing anything wrong with the Version value inside the spec file when bumping

jpopelka:
The `update_release: false` in a package config now works the same way as `release_suffix: ""`.
The `self.job_config.release_suffix == ''` condition can be removed once users move to `update_release: false`.

TODO:

- [x] packit/packit.dev#587

Fixes #1760 

Merge after packit/packit#1827

---

RELEASE NOTES BEGIN
￼
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/helpers/build/babysit.py
+++ b/packit_service/worker/helpers/build/babysit.py
@@ -200,10 +200,10 @@ def update_copr_builds(build_id: int, builds: Iterable["CoprBuildTargetModel"]) 
         return True
 
     if not build_copr.ended_on and not build_copr.started_on:
-        logger.info("The copr build has not started yet.")
+        logger.info(f"The copr build {build_id} has not started yet.")
         return False
 
-    logger.info(f"The status is {build_copr.state!r}.")
+    logger.info(f"The status of {build_id} is {build_copr.state!r}.")
 
     current_time = datetime.now(timezone.utc)
     for build in builds:
@@ -218,7 +218,7 @@ def update_copr_builds(build_id: int, builds: Iterable["CoprBuildTargetModel"]) 
             continue
         if build.status not in (BuildStatus.pending, BuildStatus.waiting_for_srpm):
             logger.info(
-                f"DB state says {build.status!r}, "
+                f"DB state of {build_id} says {build.status!r}, "
                 "things were taken care of already, skipping."
             )
             continue
--- a/packit_service/worker/helpers/build/build_helper.py
+++ b/packit_service/worker/helpers/build/build_helper.py
@@ -431,17 +431,19 @@ class BaseBuildJobHelper(BaseJobHelper):
         )
         self._srpm_model.set_start_time(datetime.datetime.utcnow())
 
-        bump_version = (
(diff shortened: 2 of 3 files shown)


TomasTomecek (approved):

jpopelka:
recheck

Issue #1692, opened 2022-10-12
Make sure all the handlers are idempotent


lachmanfrantisek:
* [ ] Make sure every handler can be re-triggered in any state of the execution.
  * [ ] Fix the easy problems.
  * [ ] Create issues for the more complicated problems.

lbarcziova:
Some notes regarding what actions are done in the handlers code and potential improvements:

Firstly we have the handlers that report the state of the task to the user and only update the state of existing models in DB (+ in some, we are doing GET requests to get e.g. end time of the Copr build). As for the reporting, as discussed, creating the commit statuses/checks is almost idempotent (there is only different timestamp on the check) and as for commenting, we have methods that check existing duplicate comments.

1. KojiTaskReportHandler
2. KojiBuildReportHandler
3. TestingFarmResultsHandler
4. VMImageBuildResultHandler
5. CoprBuildStartHandler
6. CoprBuildEndHandler - here is an extra step: creating TF handler task(s)
-----------------------------------
these 2 are just a little bit different, they are **creating** or updating the DB entries
7. GithubAppInstallationHandler
8. GithubFasVerificationHandler


- in these, the code could be updated so that in the beginning, we always check whether the state in DB is different than the one about to report and update the state in DB at the end of the handler code; the CoprBuildEndHandler could be problematic because of the creation of TF handler task(s).

Then, there are tasks where we are making POST requests to external services, creating new models in the DB, ... I think with these, we cannot make many improvements:

1. DownstreamKojiBuildHandler - submitting build in Koji
2. CoprBuildHandler - submitting build in Copr
3. TestingFarmHandler - submitting TF run
4. BodhiUpdateHandler - submitting updates in Bodhi
5. KojiBuildHandler - submitting build in Koji
6. VMImageBuildHandler - submitting the image build
------------------------------------------------
9.  AbstractSyncReleaseHandler - this one is a little bit different; we are uploading new sources (if not present) and then doing changes on Git repo (using a deterministic branch) and, in the end, force pushing.

TomasTomecek:
The question to me seems what do we do when a network error happens while submitting a build: do we submit another one? do we check if there is a pending one a reuse it?

lbarcziova:
As we were discussing during the last week's architecture that, we could decide for some handlers to not ensure the idempotency (and when implementing the mentioned Celery `acks_late` do not mark these). 

[...]

PR #1856, opened 2023-01-19
Improvements for idempotency

lbarcziova:
TODO:

Related to #1692 

---

RELEASE NOTES BEGIN
N/A
RELEASE NOTES END

### DIFF
--- a/packit_service/config.py
+++ b/packit_service/config.py
@@ -276,20 +276,24 @@ class PackageConfigGetter:
         title: str,
         message: str,
         comment_to_existing: Optional[str] = None,
+        add_packit_prefix: Optional[bool] = True,
     ) -> Optional[Issue]:
         # TODO: Improve filtering
         issues = project.get_issue_list()
         packit_title = f"[packit] {title}"
 
         for issue in issues:
-            if packit_title in issue.title or title in issue.title:
+            if title in issue.title:
+                logger.debug(f"Title of issue {issue.id} matches.")
                 if comment_to_existing:
                     issue.comment(body=comment_to_existing)
                     logger.debug(f"Issue #{issue.id} updated: {issue.url}")
                 return None
 
         # TODO: store in DB
-        issue = project.create_issue(title=packit_title, body=message)
+        issue = project.create_issue(
+            title=packit_title if add_packit_prefix else title, body=message
+        )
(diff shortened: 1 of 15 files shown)


packit_service/worker/handlers/testing_farm.py
```diff
@@ -310,7 +310,14 @@ def run(self) -> TaskResults:
...
 
-        test_run_model.set_status(self.result, created=self.created)
+        if test_run_model.status == self.result:
+            logger.debug(
+                "Testing farm results already processed "
+                "(state in the DB is the same as the one about to report)."
+            )
+            return TaskResults(
+                success=True, details={"msg": "Testing farm results already processed"}
+            )
```
lachmanfrantisek, line 320: Can't we lose commit statuses with this approach?

(E.g. Handler will update the database and is killed right after that.)

I would either don't skip too much to not lose anything or do the same thing twice (can't hurt the UX).

lbarcziova: I moved the setting of the status in DB after reporting the status to the user to prevent that, WDYT?

lachmanfrantisek: :+1:

lachmanfrantisek (approved):

TomasTomecek (approved):

packit_service/config.py
```diff
@@ -276,20 +276,24 @@ def create_issue_if_needed(
...
         comment_to_existing: Optional[str] = None,
+        add_packit_prefix: Optional[bool] = True,
     ) -> Optional[Issue]:
         # TODO: Improve filtering
         issues = project.get_issue_list()
         packit_title = f"[packit] {title}"
 
         for issue in issues:
             if packit_title in issue.title or title in issue.title:
+                logger.debug(f"Title of issue {issue.id} matches.")
```
jpopelka, line 287: Just wondering about the condition

Since `packit_title = f"[packit] {title}"`, then `packit_title in issue.title` is a subset of  `title in issue.title` and  thus can be removed from the OR, i.e. it can be changed to just

```python
if title in issue.title:
```

can't it?

jpopelka (approved):

lbarcziova:
recheck

PR #1859, opened 2023-01-23
Run API tests during make check-db

FrNecas:
SSIA. I thought about creating a separate target but DB and API tests are very related and I feel like the goal of the target is to reproduce tests_openshift from zuul as much as possible


lachmanfrantisek (approved):
Maybe `make check-with-db`? But `make check-db` is fine...

lbarcziova (approved):

PR #1860, opened 2023-01-23
Get the actual latest Koji build when retriggering Bodhi update

nforro:
Koji tags named equally to dist-git branches are what stable builds are tagged into, but when creating a Bodhi update, we are interested in the latest non-stable build that is tagged into a candidate tag.

Get candidate tag corresponding to the dist-git branch being processed and use that to get the latest build from Koji.

---

This is a problem when retriggering both from a dist-git PR and from an issue repository, see for example https://github.com/packit/specfile/issues/194.


lbarcziova (approved):
thanks!

PR #1863, opened 2023-01-24
Get branches also from comments in issue repository

nforro:
If a Bodhi update fails and there is an existing issue in the configured issue repository, packit adds a comment to it mentioning the corresponding dist-git branch(es) rather than updating the description.

Collect a set of branches from both the issue description and its comments.

See for example https://github.com/packit/specfile/issues/194.

### DIFF
--- a/packit_service/worker/mixin.py
+++ b/packit_service/worker/mixin.py
@@ -288,12 +288,17 @@ class GetBranchesFromIssueMixin(Config, GetBranches):
 
         You can retrigger the update by adding a comment (`/packit propose-downstream`) into this issue.
         """
-        branches = []
+        branches = set()
+        branch_regex = re.compile(r"\s*\| `(\S+)` \|")
         issue = self.data.project.get_issue(self.data.issue_id)
         for line in issue.description.splitlines():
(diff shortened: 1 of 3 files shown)


packit_service/worker/mixin.py
```diff
@@ -289,10 +289,17 @@ def branches(self) -> List[str]:
...
         for line in issue.description.splitlines():
-            if m := re.match(r"\s*\| `(\S+)` \|", line):
-                branches.append(m[1])
+            if m := branch_regex.match(line):
+                if m[1] not in branches:
+                    branches.append(m[1])
+        for comment in issue.get_comments():
+            for line in comment.body.splitlines():
+                if m := branch_regex.match(line):
+                    if m[1] not in branches:
```
jpopelka, line 301: if `branches` was a set you wouldn't need to check

jpopelka (approved):

packit_service/worker/mixin.py
nforro: I wasn't sure if the order wasn't important, but I suppose it doesn't matter, I'll use a set.

Issue #1721, opened 2022-10-20
Improve the permission schema for Bodhi updates


lachmanfrantisek:
Currently, Packit needs to ask for commit rights to dist-git to be able to create Bodhi updates.

This issue is to track the discussions/communication with Fedora infra and the Bodhi team.

These are the relevant tickets elsewhere:
* https://github.com/fedora-infra/bodhi/issues/4671
* https://issues.redhat.com/browse/PACKIT-2143
* https://pagure.io/fedora-infrastructure/issue/10763 (general request)
* https://pagure.io/fedora-infrastructure/issue/10959 (ask for a new FAS group to be created)

lachmanfrantisek:
We were given [global Bodhi rights](https://accounts.fedoraproject.org/group/bodhi_update_bots/) so we don't need dist-git commit rights anymore. So, it's on us now to:

* [ ] Check that it really works both for `packit` and `packit-stg`.
* [ ] Improve the code to not ask for that (=remove the unneeded code).
* [ ] Either automatically, or manually get rid of the dist-git rights we've already been given.

mfocko:
It seems that the created group was not added to the Bodhi deployment playbooks

PR with fix: https://pagure.io/fedora-infra/ansible/pull-request/1285
Follow-up: https://pagure.io/fedora-infrastructure/issue/10959#comment-833133

PR #1868, opened 2023-01-30
fix(bodhi): auth failure is not expected anymore

mfocko:
After we were given the global rights to run Bodhi updates, we don't need to ask for ‹commit› privileges on dist-git, therefore this kind of failure is now unexpected and should be reported to Sentry.

Also we shouldn't suggest adding ‹commit› access to the dist-git as it is no longer needed.

TODO:

Related to #1721

### DIFF
--- a/packit_service/worker/handlers/bodhi.py
+++ b/packit_service/worker/handlers/bodhi.py
@@ -9,7 +9,6 @@ import abc
 from typing import Tuple, Type
 
 from celery import Task
-from fedora.client import AuthError
 
 from packit.config import JobConfig, JobType, PackageConfig
 from packit.exceptions import PackitException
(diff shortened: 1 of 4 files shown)


packit_service/worker/handlers/bodhi.py
```diff
@@ -159,37 +150,6 @@ def report_in_issue_repository(
...
-
-        # Notify user on each task run and set a more generous retry interval
-        # to let the user fix this issue in the meantime.
-        if not self.celery_task.is_last_try():
-            body += (
-                f" Task will be retried in "
-                f"{RETRY_INTERVAL_IN_MINUTES_WHEN_USER_ACTION_IS_NEEDED} minutes.*"
-            )
-            self.celery_task.retry(
-                delay=RETRY_INTERVAL_IN_MINUTES_WHEN_USER_ACTION_IS_NEEDED * 60,
```
lbarcziova, line 185: can we remove also the constant or is it used somewhere else?

lbarcziova (approved):
:+1:

packit_service/worker/handlers/bodhi.py
mfocko: removed; thought it's used in other places where we retry

mfocko:
recheck

jpopelka (approved):
love such diffs :smiling_imp:

TomasTomecek:
I thought this fixes https://github.com/packit/packit/issues/1638 but we will drop even more code!!

PR #1870, opened 2023-01-30
LocalProjectMixin before PackitAPIWithUpstreamMixin

### LINKED ISSUE #1861: packit_api.up.local_project.git_repo is None in AbstractSyncReleaseHandler
Sentry Issue: [PCKT-002-PACKIT-SERVICE-549](https://sentry.io/organizations/red-hat-it/issues/3893427876/?referrer=github_integration)

```
AttributeError: 'NoneType' object has no attribute 'is_dirty'
  File "packit_service/worker/handlers/distgit.py", line 179, in sync_branch
    downstream_pr = self.packit_api.sync_release(
  File "packit/api.py", line 766, in sync_release
    if not force and self.up.is_dirty() and not use_local_content:
  File "packit/base_git.py", line 444, in is_dirty
    return self.local_project.git_repo.is_dirty()

AttributeError: 'NoneType' object has no attribute 'head'
  File "packit_service/worker/handlers/distgit.py", line 280, in run
    downstream_pr = self.sync_branch(
  File "packit_service/worker/handlers/distgit.py", line 212, in sync_branch
    self.packit_api.up.local_project.git_repo.head.reset(
```

creator: @lachmanfrantisek
notes:
* Looks like the `git_repo` is not filled during the API initialisation.
* ~~Problem should be caused by this (very nicely done!) refactor: https://github.com/packit/packit-service/commit/bf1b977f37b6f21bb083983db42884034dbf9dd1~~ (too old)  (Also, a few GitLab-related fixes has been included to the last release.)
* This problem cause `propose-downstream` to fail. (e.g. [here](https://github.com/rhinstaller/anaconda/runs/10840558704))

majamassarini:
Since PackitAPIWithUpstreamMixin uses the LocalProjectMixin we should define it first (probably)

fix #1861

### DIFF
--- a/packit_service/worker/handlers/abstract.py
+++ b/packit_service/worker/handlers/abstract.py
@@ -33,7 +33,7 @@ from packit_service.worker.checker.abstract import Checker
 
 from packit_service.worker.mixin import (
     Config,
-    PackitAPIWithDownstreamMixin,
+    PackitAPIProtocol,
 )
 
 logger = logging.getLogger(__name__)
@@ -219,7 +219,7 @@ class TaskName(str, enum.Enum):
     pull_from_upstream = "pull_from_upstream"
 
 
-class Handler(PackitAPIWithDownstreamMixin, Config):
+class Handler(PackitAPIProtocol, Config):
     def run(self) -> TaskResults:
         raise NotImplementedError("This should have been implemented.")
 
--- a/packit_service/worker/handlers/copr.py
+++ b/packit_service/worker/handlers/copr.py
@@ -59,6 +59,7 @@ from packit_service.worker.handlers.abstract import (
     run_for_check_rerun,
     RetriableJobHandler,
 )
+from packit_service.worker.mixin import PackitAPIWithDownstreamMixin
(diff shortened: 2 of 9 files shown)


lbarcziova (approved):
this should fix #1861 , right?

majamassarini:
> this should fix #1861 , right?

Yes, it should

FrNecas (approved):
LGTM, hopefully the tests pass now :)

TomasTomecek (approved):
Aside from those 2 new tests cases, what can we do to prevent this in future?

majamassarini:
> Aside from those 2 new tests cases, what can we do to prevent this in future?

I don't really know. But when I did the refactoring, I did not create new tests for the new Mixins classes, I just re-used the old tests, but probably a more broader coverage would have been useful.

FrNecas:
I think that after this fix and change in inheritance structure, we are less prone to such bugs in the future. Overriding mixins is complicated and tricky due to MRO so inheriting from mixins at the lowest level possible is IMO a way to prevent this.

PR #1873, opened 2023-01-31
Add broken test for monorepo refactoring

majamassarini:
DO NOT MERGE.

This is just a SPIKE PR to propose a solution on how to refactor the packit_api object to support monorepos.

related to https://github.com/packit/packit/issues/1618
related to https://github.com/packit/packit/pull/1838
related to https://github.com/packit/packit/pull/1840

Here there is a broken unit test that when fixed we could probably say we support the monorepo configuration in `packit-service`.
In [get_jobs_matching_event](https://github.com/majamassarini/packit-service/blob/2e8881d347ce4553cd6f4d48cd919291511c5187/packit_service/worker/jobs.py#L555) we return a `JobConfig` which is not *usable*. It is defined in the unit test [here](https://github.com/majamassarini/packit-service/blob/multiple_distgit_packit_api/tests/unit/test_jobs.py#L3174-L3182) and its properties can not be accessed, like the [`downstream_package_name`](https://github.com/majamassarini/packit-service/blob/2e8881d347ce4553cd6f4d48cd919291511c5187/tests/unit/test_jobs.py#L3228), because this `JobConfig` references multiple packages.

There are many possible solutions; here below I describe my favourite one.

Our handlers are being designed to work with one package at a time and I think they should keep doing that. For this reason if a user specifies multiple downstream packages for the same old job, like the `propose_downstream` job, then I can image that the call `self.event.package_config.jobs` should probably be substituted with another call able to return two different jobs, let me call them `JobConfigView` or something like that, for the same `JobConfig` which has multiple packages...

I think we should try to keep our handlers as  simple as we can and not manage multiple packages in the same handler instance run.
But if we are forced to do that, then we can write new handlers able to deal with a `JobConfig` holding multiple packages. Probably we could create a new class for a `JobConfig` holding multiple packages  and pass instances of this class to the new handlers.

Let me know if you see any problem with this solution.

### DIFF
--- a/tests/unit/test_jobs.py
+++ b/tests/unit/test_jobs.py
@@ -3,11 +3,19 @@
 
 from typing import Type
 
+import copy
 import celery
 import pytest
 from flexmock import flexmock
 
-from packit.config import CommonPackageConfig, JobConfig, JobConfigTriggerType, JobType
+from packit.config import (
+    CommonPackageConfig,
+    JobConfig,
+    JobConfigTriggerType,
+    JobType,
+    PackageConfig,
+)
+
 from packit_service.config import ServiceConfig
 from packit_service.constants import COMMENT_REACTION
 from packit_service.worker.events import (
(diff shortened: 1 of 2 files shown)


tests/unit/test_jobs.py
```diff
@@ -3121,3 +3129,106 @@ def actor(self):
...
+        original_ref = 0
+        double_ref = 0
+        for job_config in job_configs:
+            if job_config.downstream_package_name == "a double":
+                double_ref += 1
+            else:
+                original_ref += 1
+
+        assert original_ref == 2
+        assert double_ref == 2
```
TomasTomecek, line 3234: I'd say we should also follow here and check if all jobs for all packages were processed correctly, something like:
```python
for package in packages_config.packages:
    for job in package.jobs:
        assert something(handlers) == other_thing(job)
```

lachmanfrantisek:
> Our handlers are being designed to work with one package at a time and I think they should keep doing that.

 :+1: 

> I think we should try to keep our handlers as simple as we can and not manage multiple packages in the same handler instance run.

:+1:

We just need to somehow resolve the check status names... Maybe somehow reuse identifiers? So the identifier will not be `None` by default but `{package_name}` if more packages are defined?
But that's a technical detail and will not be relevant for downstream jobs at all..

majamassarini:
> We just need to somehow resolve the check status names... Maybe somehow reuse identifiers? So the identifier will not be `None` by default but `{package_name}` if more packages are defined? 

Sure, I think that this can be done easily when building a `JobConfigView` from the PackageConfig. We can definitely customize it based on the package for which the *view* has been built.

majamassarini:
I close this PR, it was just an example for supporting the SPIKE on how to handle packit refactoring for monorepo.

Related research could be found here: https://github.com/packit/research/tree/main/monorepo-support/refactoring

PR #1887, opened 2023-02-07
Fix group-related bugs

FrNecas:
I will try to add a new test case for the TF bug in the afternoon (testing the first typo/bug doesn't make much sense since it's hard to assert over IDs assigned by postgres and such test would be quite fragile).

### DIFF
--- a/packit_service/worker/handlers/testing_farm.py
+++ b/packit_service/worker/handlers/testing_farm.py
@@ -102,7 +102,7 @@ class TestingFarmHandler(
         job_config: JobConfig,
         event: dict,
         celery_task: Task,
-        testing_farm_group_id: Optional[int] = None,
+        testing_farm_target_id: Optional[int] = None,
         build_id: Optional[int] = None,
     ):
         super().__init__(
@@ -112,7 +112,7 @@ class TestingFarmHandler(
             celery_task=celery_task,
         )
         self.build_id = build_id
-        self._testing_farm_group_id = testing_farm_group_id
+        self._testing_farm_target_id = testing_farm_target_id
         self._testing_farm_job_helper: Optional[TestingFarmJobHelper] = None
 
     @staticmethod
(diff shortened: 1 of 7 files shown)


packit_service/worker/handlers/testing_farm.py
```diff
@@ -135,11 +135,13 @@ def _get_or_create_group(
...
 
         Returns:
-            The existing or created test run group.
+            The existing or created test run group and the test targets
+            to run the tests for.
 
         """
         if self._testing_farm_group_id is not None:
-            return TFTTestRunGroupModel.get_by_id(self._testing_farm_group_id)
+            group = TFTTestRunGroupModel.get_by_id(self._testing_farm_group_id)
```
lbarcziova, line 143: so what about the scenario when the tests are triggered after Copr build (=> for each target separately), and there is an exception for the last ran target, and the test handler is retried...in that case, here we get all the targets in the DB again..can this happen?

FrNecas: I was thinking about this and thought that wouldn't occur but your scenario sounds like that this may be the case. It sounds like that to solve this, we need to save "target ID" not "group ID" in the handler during retries. From the code it looks like that the handler will be triggered for each failed test. Does this sound reasonable?

lbarcziova: maybe I am speculating too much, but if we would forward `target ID` when retrying, in case the tests were triggered by a comment (=> tests for all targets are submitted in one handler task) and e.g. we fail to submit tests for the first processed target, the handler retry task should still process all targets...wouldn't it make more sense to act depending on the configuration + targets overrides? :thinking:

FrNecas: As discussed on chat, the task does not stop if one submission fails, therefore we can retry just the single target. Code adjusted, test case added.

lbarcziova (approved):
LGTM :+1:

PR #1902, opened 2023-02-10
Fix/packit service/1879

### LINKED ISSUE #1879: pull-from-upstream requires Packit-as-a-Service to be enabled in upstream GitHub project
I [set up](https://src.fedoraproject.org/rpms/gpxsee/c/802c0f4a03230caba437a3f951e7cbb9122cae3f?branch=rawhide) `pull-from-upstream` for a project where upstream doesn't know anything about packit (I think that will be a common use case), using a minimalistic config:

```yaml
specfile_path: gpxsee.spec
 
upstream_package_name: GPXSee
downstream_package_name: gpxsee
 
upstream_project_url: https://github.com/tumic0/GPXSee
 
jobs:
  - job: pull_from_upstream
    trigger: release
    dist_git_branches:
      - fedora-all
```

But the job failed with: `Failed to retrieve a token: App is not installed on tumic0/GPXSee`.
I think installing the app kind of defeats the purpose of `pull-from-upstream`.

Here is coresponding Sentry issue: https://sentry.io/organizations/red-hat-it/issues/3901515095/?project=1767823

majamassarini:
Customize authentication method for `PullFromUpstream`: use token instead of tokman.
Token should be added to the service configuration.
I tested it with a personal token and it seems to be working, we could use a token created for packit or a bot?

- [x] Write new unit tests

Fixes #1879

Merge after packit/ogr#767

### DIFF
--- a/packit_service/worker/helpers/sync_release/sync_release.py
+++ b/packit_service/worker/helpers/sync_release/sync_release.py
@@ -57,7 +57,9 @@ class SyncReleaseHelper(BaseJobHelper):
         Return all valid branches from config.
         """
         branches = get_branches(
-            *self.job.dist_git_branches, default=self.default_dg_branch
+            *self.job.dist_git_branches,
+            default_dg_branch=self.default_dg_branch,
+            default=self.default_dg_branch,
         )
         if self.branches_override:
             logger.debug(f"Branches override: {self.branches_override}")
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -50,9 +50,15 @@ You can also run only certain pieces of packit-service for local development
 (e.g. worker, database or service/httpd).
 You also need to populate `secrets/packit/dev/` manually, for instructions
 see [deployment repo](https://github.com/packit/deployment/tree/main/secrets).
+Be sure to copy the content you need from `extra-vars.yml` to `packit-service.yaml`
 
 When you are running service/httpd and making requests to it,
 make sure that `server_name` configuration file in `packit-service.yaml` is set.
+
+### tokman
+
+To make tokman work with docker-compose create a `config.py` file in `./secrets/packit/dev/tokman-files` using this [template](https://github.com/packit/tokman/blob/main/config.py.example).
+In `.secrets/packit/dev/packit-service.yaml` fix url to tokman from `http://tokman` to `http://tokman:8000`
 
 ### binding on localhost
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,6 +1,20 @@
 version: "2"
 
 services:
+  tokman:
+    image: quay.io/packit/tokman
+    container_name: tokman
+    ports:
+      - 8000:8000
+    environment:
+      LOG_LEVEL: debug
+      TOKMAN_CONFIG: /config/config.py
+      BIND_ADDR: 0.0.0.0:8000
+    volumes:
+      - ./secrets/packit/dev/tokman-files:/config:ro,z
+      - ./secrets/packit/dev/tokman-files:/access_tokens:ro,z
+      - ./secrets/packit/dev/:/secrets:ro,z
+
   redis:
     image: quay.io/sclorg/redis-6-c9s
     container_name: redis
@@ -53,6 +67,7 @@ services:
     image: quay.io/packit/packit-worker:dev
     command: /usr/bin/run_worker.sh
     depends_on:
+      - tokman
       - redis
       - postgres
     environment:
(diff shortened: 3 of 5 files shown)


packit_service/worker/helpers/sync_release/sync_release.py
```diff
@@ -57,7 +57,7 @@ def branches(self) -> Set[str]:
         Return all valid branches from config.
         """
         branches = get_branches(
-            *self.job.dist_git_branches, default=self.default_dg_branch
```
majamassarini, line 60: @lbarcziova to make things work in my field test I had to change this initialization parameter, but I am not totally sure this is right. Just asking. Thanks!

files/run_worker.sh
```diff
@@ -13,8 +13,10 @@ source /usr/bin/setup_env_in_openshift.sh
 mkdir -p "${PACKIT_HOME}/.ssh"
 chmod 0700 "${PACKIT_HOME}/.ssh"
 pushd "${PACKIT_HOME}/.ssh" || exit
-install -m 0400 /packit-ssh/id_rsa .
-install -m 0400 /packit-ssh/id_rsa.pub .
+# with rootless podman, if id_rsa* files have 0400 mode,
+# in the host, then we can not access them -> fail here
+install -m 0400 /packit-ssh/id_rsa . || exit
```
majamassarini, line 18: @jpopelka Do you think this `exit`s are ok even in opeshift?

I realized that using rootless podman when my id_rsa* files inside the host have a 0400 mask docker-compose build is not able to load them into the container and it tooks me a while to realize this file were missing.
I think that if we can fail here, debugging is easier.

jpopelka: Yes, I think it's OK.

I actually already have a work-in-progress commit that adds `set -e` to the top of the script so that it exits when **any** command fails. I just need to test it.

packit_service/worker/helpers/sync_release/sync_release.py
lbarcziova: the `default` here should be correct, what was the issue when testing with this?

lbarcziova (commented):
code LGTM :ok_hand:

packit_service/worker/helpers/sync_release/sync_release.py
majamassarini: This is my [`.packit.yaml`](https://src.fedoraproject.org/rpms/python-teamcity-messages/blob/rawhide/f/.packit.yaml) file in the `rawhide` branch in dist-git.
The default branch is `rawhide`.
If I do not change the above call, I can not work on dist-git. The code will look for a dist-git `default` like `main` or `master` (I don't remember) instead of `rawhide` and fails.

lbarcziova: ah, interesting. This code was not introduced by the `pull-from-upstream` changes; I think it was like that all the time. 

[Here](
https://github.com/packit/packit/blob/281a5acbdbc8358405bc75e72f568cb7fd7a8a84/packit/config/aliases.py#L152-L154) is the related code, but I am not that familiar with it. @lachmanfrantisek @jpopelka , from the git blame, do you think we should pass here also the `default_dg_branch` from API and not rely on the default `main` as e.g. repo linked by Maja doesn't include that branch?

majamassarini: I saw that a lot of tests are failing now with my change, so I reverted it.
I am wondering if this [code](https://github.com/packit/packit/blob/281a5acbdbc8358405bc75e72f568cb7fd7a8a84/packit/config/aliases.py#L155-L156) could be related with my problem.
If I [use here](https://github.com/packit/packit-service/blob/3bea9d46865314c6e6a62a4cadcc4e0e808575d3/packit_service/worker/helpers/sync_release/sync_release.py#L59) the `with_aliases=True` flag then branch `rawhide` is used and everything goes smoothly.
I can also see that with the `with_aliases` flag less unit tests fail...

files/run_worker.sh
jpopelka: #1912

majamassarini: > #1912

I have tested it, and for me it works, it fails if id_rsa* files were not copied. So I have removed my commit. Thanks!

majamassarini:
recheck

packit_service/worker/helpers/sync_release/sync_release.py
lbarcziova: from looking into the code, it would make sense to me to pass the `default_dg_branch` and not enable `with_aliases`, since `with_aliases` will result in getting both `main` and `rawhide`, but getting `main` doesn't make sense if it doesn't exist (your case)...but I would like someone to confirm this will be correct

lachmanfrantisek: I would try to use the default branch of the Pagure repository so we are safe to use repository of any structure (even if the `main` branch misses because of an issue or for a good reason).

lbarcziova (approved):

packit_service/worker/helpers/sync_release/sync_release.py
majamassarini: > I would try to use the default branch of the Pagure repository so we are safe to use repository of any structure (even if the `main` branch misses because of an issue or for a good reason).

I tested my last commit on my weird repo config and it is working and no tests are failing. So probably it is safe to add it.

PR #1904, opened 2023-02-13
Don't sync default files in pull-from-upstream

lbarcziova:
In propose-downstream, we by default sync the specfile and Packit config, this should not be done in pull-from-upstream that doesn't contain these files.

---

RELEASE NOTES BEGIN
N/A
RELEASE NOTES END


lbarcziova:
recheck

Issue #1908, opened 2023-02-14
Speed-up queries after introduction of groups


FrNecas:
After https://github.com/packit/packit-service/pull/1787 was merged to prod, the dashboard takes longer to load (e.g. showing copr builds), some queries are probably not as efficient as they used to be. They may be missing indexes that cause this, however further investigation is needed

jpopelka:
From what I remember, there were quite some indexes added in #1787. But it's definitely worth [analyzing](https://www.postgresql.org/docs/current/using-explain.html#USING-EXPLAIN-ANALYZE) because last time (#1639) the new indexes helped a lot.

I also noticed that the PostgreSQL CPU consumption increased quite substantially (about 10x I'd say).

jpopelka:
Use [SQLALCHEMY_ECHO env. var.](https://github.com/packit/packit-service/pull/1913) to see the performed SQL statements.

majamassarini:
Looking at our sentry I found this which could be somehow related: https://red-hat-it.sentry.io/issues/3980287229/?query=is%3Aunresolved&referrer=issue-stream

jpopelka:
If I took this card I would:

- Download prod database backup from AWS and import it into locally running `postgres`
- Start `service` (API) with [SQLALCHEMY_ECHO env. var.](https://github.com/packit/packit-service/pull/1913) set.
- `http --verify=no https://service.localhost:8443/api/runs?page=1&per_page=10`
- In the `service` logs see how the performed SQL queries look like
- Try those queries with [EXPLAIN ANALYZE](https://www.postgresql.org/docs/current/using-explain.html#USING-EXPLAIN-ANALYZE) in `psql -d packit` in the `postgres`
- Try to find the bottlenecks - what takes the most time? where does it use [Seq scan](https://pganalyze.com/docs/explain/scan-nodes/sequential-scan) instead of [[Bitmap] Index scan](https://pganalyze.com/docs/explain/scan-nodes/index-scan)? 
- If there are some sequential scans, add index(es) (`CREATE INDEX`) and try those queries again

See also commits and commit messages in #1639

TomasTomecek:
When I worked in django in my previous life, I used [django-debug-toolbar](https://github.com/jazzband/django-debug-toolbar#django-debug-toolbar-)

It has a mindblowing feature of showing all queries done to display a view (see the screenshot) with detailed ANALYZE, as Jirka explains above.

I'm sure that sentry or some additional tool has to have something similar.

[...]

PR #1913, opened 2023-02-15
Allow to set echo arg of sqlalchemy.create_engine() via env. var.

jpopelka:
`SQLALCHEMY_ECHO`

[sqlalchemy.create_engine()](https://docs.sqlalchemy.org/en/20/core/engines.html#sqlalchemy.create_engine.params.echo)

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -7,7 +7,6 @@ Data layer on top of PSQL using sqlalch
 
 import enum
 import logging
-import os
 from collections import Counter
 from contextlib import contextmanager
 from datetime import datetime, timedelta, timezone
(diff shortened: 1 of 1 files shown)


FrNecas (approved):

TomasTomecek (approved):
I assume that you want to run those queries yourself with [`EXPLAIN`](https://www.postgresql.org/docs/current/sql-explain.html), right?

jpopelka:
The one working on #1908 will run those queries with [EXPLAIN ANALYZE](https://www.postgresql.org/docs/current/using-explain.html#USING-EXPLAIN-ANALYZE), but first they need to know what queries to run, hence this PR.

PR #1921, opened 2023-02-20
Adjust package config path discovery

csomh:
Due to packit/packit#1846 some mocks need to be updated to match the new behaviour.

Add a new configuration field for the service in order to allow overriding the default package config path in Hardly.


jpopelka (approved):

csomh:
recheck

PR #1935, opened 2023-02-23
fix(tasks): HandlerTaskWithRetry.acks_late = True

### LINKED ISSUE #1927: Utilise acks_late for Celery tasks where possible
If the handler is restarted in the middle of processing a task, currently, the task is not executed again, so we lose any progress.
Celery documentation [suggests](https://docs.celeryq.dev/en/stable/faq.html#should-i-use-retry-or-acks-late) that if the tasks are idempotent, using the `acks_late` option should help with this and we have agreed to utilise this option in one of our architecture meetings previously.

TODO:
- [x] look into the comments in #1692 that explain which tasks should be now idempotent
- [x] for these tasks, set [acks_late](https://docs.celeryq.dev/en/stable/userguide/tasks.html#Task.acks_late) to True
- [ ] check in staging this works as expected

Followup of #1692.

TomasTomecek:
since we retry failed tasks that use HandlerTaskWithRetry, let's also do
acks_late for these to make sure they are rescheduled if a worker is
shut down in the middle of processing

TODO:

- [x] ~~Write new tests or update the old ones to cover new functionality.~~ I will verify on stg, I can't see how we could write a test case for this :) tear down a worker and see if the tasks is restarted

Fixes #1927

---

RELEASE NOTES BEGIN
Packit will retry tasks that are interrupted by a worker shutdown. This should improve throughput and reduce cases where there is no outcome, i.e. "builds should have been done but nothing happened".
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/tasks.py
+++ b/packit_service/worker/tasks.py
@@ -118,6 +118,9 @@ class HandlerTaskWithRetry(Task):
     max_retries = int(getenv("CELERY_RETRY_LIMIT", DEFAULT_RETRY_LIMIT))
     retry_kwargs = {"max_retries": max_retries}
     retry_backoff = int(getenv("CELERY_RETRY_BACKOFF", DEFAULT_RETRY_BACKOFF))
+    # https://docs.celeryq.dev/en/stable/userguide/tasks.html#Task.acks_late
+    # retry if worker gets obliterated during execution
+    acks_late = True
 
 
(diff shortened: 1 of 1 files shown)


TomasTomecek:
We just spoke with @lbarcziova and since we all have off on Fri (tomorrow), we will only have Monday to test this on stg which is not enough for such a big change here.

Let's merge this on Tue to get 5 full days to verify on stg.

lbarcziova (approved):
let's try this on stage

PR #1936, opened 2023-02-23
Remove leftovers from aef1b57dc

jpopelka:
@csomh I was checking the  aef1b57dc (because of hardly) and this looks suspicious.

### DIFF
--- a/tests/integration/test_check_rerun.py
+++ b/tests/integration/test_check_rerun.py
@@ -249,7 +249,6 @@ def test_check_rerun_pr_testing_farm_handler(
     flexmock(TestingFarmJobHelper).should_receive("run_testing_farm").once().and_return(
         TaskResults(success=True, details={})
     )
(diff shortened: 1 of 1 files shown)


jpopelka:
recheck

csomh:
Why are these leftovers? AFAIK, `get_files()` is called two times in `get_package_config_from_repo()`: once when looking for packit.yaml, the second time when looking for the spec file (when no `specfile_path` is defined). Are the tests failing with these expectations?

jpopelka:
From our discussion:

It's unclear how the flexmock knows which `and_return()` to use.
Given that the tests pass with my change I think the first one is simply ignored, hence I suggested removing it.

A better solution would be if there were a different `with_args()` in each.

I'm closing this as there are actually no tests failing with the current code.
I just wanted to bring this to your attention and created a PR as a way of asking.

csomh:
Thanks! :bow:

PR #1939, opened 2023-02-28
Add API endpoint for pull-from-upstream runs

lbarcziova:
Related to packit/dashboard#199

TODO:

---

RELEASE NOTES BEGIN
￼
RELEASE NOTES END

### DIFF
--- a/packit_service/service/api/pull_from_upstream.py
+++ b/packit_service/service/api/pull_from_upstream.py
@@ -14,55 +14,54 @@ from packit_service.models import (
 from packit_service.service.api.parsers import indices, pagination_arguments
 from packit_service.service.api.utils import (
     response_maker,
-    get_sync_release_info,
     get_sync_release_target_info,
+    get_sync_release_info,
 )
 
 logger = getLogger("packit_service")
 
-ns = Namespace("propose-downstream", description="Propose Downstream")
+ns = Namespace("pull-from-upstream", description="Pull from upstream")
 
 
 @ns.route("")
-class ProposeDownstreamList(Resource):
+class PullFromUpstreamList(Resource):
     @ns.expect(pagination_arguments)
-    @ns.response(HTTPStatus.PARTIAL_CONTENT.value, "Propose Downstreams results follow")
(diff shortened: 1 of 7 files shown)


packit_service/service/api/propose_downstream.py
```diff
@@ -74,23 +54,16 @@ class ProposeResult(Resource):
...
-        if not dowstream_pr:
+        if (
+            not sync_release_target_model
+            or sync_release_target_model.sync_release.job_type
+            != SyncReleaseJobType.propose_downstream
+        ):
             return response_maker(
                 {"error": "No info about propose downstream target stored in DB"},
                 status=HTTPStatus.NOT_FOUND,
             )
```
lachmanfrantisek, line 67:
```suggestion
            return response_maker(
                {"error": "No info about target release sync stored in DB"},
                status=HTTPStatus.NOT_FOUND,
            )
```

packit_service/service/api/pull_from_upstream.py
```diff
@@ -0,0 +1,68 @@
...
+from packit_service.service.api.parsers import indices, pagination_arguments
+from packit_service.service.api.utils import (
+    response_maker,
+    get_sync_release_target_info,
+    get_sync_release_info,
+)
+
+logger = getLogger("packit_service")
+
+ns = Namespace("pull-from-upstream", description="Pull from upstream")
```
lachmanfrantisek, line 23: Do you think we should differentiate between `pull-from-upstream` and  `propose-downstream` even if we use the same model for those?

But I agree that they will probably diverge even more in the future...

(= Alternatively, we can create a new `sync-release` endpoint, but I am really not sure here..;)

lachmanfrantisek (approved):
:+1:

jpopelka (approved):
:+1:

packit_service/service/api/pull_from_upstream.py
lbarcziova: > But I agree that they will probably diverge even more in the future...

Yes, that's the reason why I think it makes sense to have separate endpoints.

PR #1946, opened 2023-03-02
fix: do not request access to GitLab forks

### LINKED ISSUE #1892: Do not ask for fork permission on GitLab by default
On GitLab, to be able to show pipelines on merge requests, Packit needs to have access to the source (=usually fork) project.

Currently, if we don't have the rights, we request access and use comments instead. The comment has a footer whit the info about the issue and process (related to #1875).

Since we can't provide any context to the user, it's better to:
* [ ] not ask for access automatically
* [ ] improve the wording in the footer to:
  * [ ] clearly state how this logic works
  * [ ] that it's just about reporting, not about functional change
  * [ ] suggest to users how they can give Packit access (ideally, prepare a specific link)

Benefit?
* One-time contributors are not spammed.
* Regular contributors perform a one-time action to resolve it.

This is a quick and easy solution before we implement a better way (#1893).

---
*Part of packit/packit-service#1821 epic.*

mfocko:
When creating commit statuses, we need to have access to the repository having the commits, in case of one-time contributors automatic requesting of access can be an unnecessary noise and spam.

Do not request the access and just post a note about the possibility to add the GitLab user to the said repository.

Fixes #1892

TODO:

Fixes #1892

---

RELEASE NOTES BEGIN
Packit will no longer automatically request access to the forks on GitLab. This will prevent us from spamming one-time contributors with requesting the access, whereas the regular contributors can add Packit following the instructions from the comments on the MRs, so they can receive the results of Packit pipeline right from the commit statuses.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/helpers/job_helper.py
+++ b/packit_service/worker/helpers/job_helper.py
@@ -164,14 +164,14 @@ class BaseJobHelper:
         so we can extend it in subclasses easily.
         """
         if self.is_gitlab_instance and not self.is_reporting_allowed:
-            self.request_project_access()
             description = (
-                f"{description}\n\n---\n*Packit-User does not have access to the "
+                f"{description}\n\n---\nPackit-User does not have access to the "
                 "source project (=usually author's fork of the project). "
-                "We have requested access to be able to set "
-                "commit statuses / pipelines instead of the comments. "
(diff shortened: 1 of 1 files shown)


jpopelka (approved):

mfocko:
recheck

PR #1951, opened 2023-03-06
Use correct dashboard URL for pull-from-upstream

lbarcziova:
When reporting pull-from-upstream issue, reference the dashborad result view correctly.
Related to packit/dashboard#219

Merge afte  packit/dashboard#219

---

RELEASE NOTES BEGIN
￼
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/handlers/distgit.py
+++ b/packit_service/worker/handlers/distgit.py
@@ -8,7 +8,7 @@ import logging
 import shutil
 import abc
 from datetime import datetime
-from typing import Optional, Tuple, Type, List
+from typing import Optional, Tuple, Type, List, Callable
 
 from celery import Task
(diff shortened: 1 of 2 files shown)


packit_service/worker/handlers/distgit.py
```diff
@@ -167,6 +170,7 @@ class AbstractSyncReleaseHandler(
     helper_kls: type[SyncReleaseHelper]
     sync_release_job_type: SyncReleaseJobType
     job_name_for_reporting: str
+    get_dashboard_url: staticmethod
```
nforro, line 173: Wouldn't it be better to type the attribute as `Callable[[int], str]` rather than `staticmethod`?

nforro (approved):
LGTM

PR #1963, opened 2023-03-15
Get rid of the out-dated task list in README

lachmanfrantisek:
Do you have a better idea of what to put there instead of the old task list?

Or, do we want to add new items instead?


TomasTomecek (approved):
much better

PR #1967, opened 2023-03-20
Add dashboard and postgres configuration

majamassarini:
- Echoing of SQL statements (useful for debugging, disabled by default)
- Add `dashboard` service to the network
- Add `postgres` limits for performance testing purposes
- Add configuration file for `postgres` (commented by default)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -56,6 +56,17 @@ services:
       POSTGRESQL_DATABASE: packit
     ports:
       - 5432:5432
+    deploy:
+      resources:
+        limits:
+          memory: 4GB
+          #cpus: "0.3"
+        reservations:
+          memory: 2G
+    # volumes:
+    # a file with packit.conf content from
+    # https://github.com/packit/deployment/blob/main/openshift/postgres.yml.j2#L84
+    # - ./secrets/packit/dev/postgres-config:/opt/app-root/src/postgresql-cfg/packit.conf:ro,Z
 
   worker:
     container_name: worker
@@ -84,6 +95,7 @@ services:
       AWS_ACCESS_KEY_ID: ""
       AWS_SECRET_ACCESS_KEY: ""
       GIT_SSH_COMMAND: "ssh -F /home/packit/.ssh/config"
+      #SQLALCHEMY_ECHO: 1
     volumes:
       - ../packit/packit:/usr/local/lib/python3.11/site-packages/packit:ro,z
       - ./packit_service:/usr/local/lib/python3.11/site-packages/packit_service:ro,z
@@ -123,7 +135,11 @@ services:
       - redis
       - postgres
     ports:
-      - 8443:8443
+      # Port 443 is needed here because the certificate we have is bundled with it
+      # otherwise dashboard.localhost can not reach service.localhost/api
+      # docker-compose needs access to ports lower than 1024:
+      # echo "net.ipv4.ip_unprivileged_port_start=443" > /etc/sysctl.d/docker-compose.conf; sysctl --system
+      - 443:8443
     environment:
       DEPLOYMENT: dev
       REDIS_SERVICE_HOST: redis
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -123,14 +138,15 @@ services:
       - redis
       - postgres
     ports:
-      - 8443:8443
+      - 443:8443
```
jpopelka, line 141: I must have something configured differently because this doesn't work for me.
With this change, the container never starts for me locally, it keeps waiting at `Attaching to service`.

docker-compose.yml
```diff
@@ -84,6 +98,7 @@ services:
       AWS_ACCESS_KEY_ID: ""
       AWS_SECRET_ACCESS_KEY: ""
       GIT_SSH_COMMAND: "ssh -F /home/packit/.ssh/config"
+      SQLALCHEMY_ECHO: 1
```
jpopelka, line 101: The log is quite verbose with it, but it doesn't hurt in the dev environment, so :+1:

docker-compose.yml
```diff
@@ -143,6 +159,28 @@ services:
       - ./files/run_httpd.sh:/usr/bin/run_httpd.sh:ro,z
     user: "1024"
 
+  dashboard:
+    image: quay.io/packit/dashboard:stg # built using local service API -> service.localhost (server name in packit-service.yaml)
```
jpopelka, line 163:
```suggestion
  dashboard:
    container_name: dashboard
    image: quay.io/packit/dashboard:stg # built using local service API -> service.localhost (server name in packit-service.yaml)
```

docker-compose.yml
majamassarini: I changed my `secrets/packit/dev/packit-service.yaml`
```yaml
#server_name: stg.packit.dev
#server_name: service.localhost:8443
server_name: service.localhost
#dashboard_server_name: dashboard.localhost
#dashboard_url: https://dashboard.stg.packit.dev
dashboard_url: https://dashboard.localhost
```

I tried to use port 443 in dashboard but I can not remember why I did not manage it. But with the above changes I can query both the API and the dashboard.

majamassarini: receck

majamassarini:
recheck

docker-compose.yml
jpopelka: hmm, still the same, i.e. waiting  for `Attaching to service` :crying_cat_face:

majamassarini: Here the logs from my docker-compose run: https://gist.github.com/majamassarini/8a089630519153dcd7a019a1ec295501

maybe you can spot some difference?

jpopelka: OK, tried again and this time podman gave me a reason why it doesn't work

```
Attaching to service
Error response from daemon: rootlessport cannot expose privileged port 443,
you can add 'net.ipv4.ip_unprivileged_port_start=443' to /etc/sysctl.conf (currently 1024),
or choose a larger port number (>= 1024): listen tcp 0.0.0.0:443: bind: permission denied
```

and
`echo "net.ipv4.ip_unprivileged_port_start=443" > /etc/sysctl.d/docker-compose.conf; sysctl --system`
makes it work, indeed, but I'm not sure that's a way I want to follow.

Could we use a port higher than 1024 (e.g. 1443 or 9443) for either the service or (preferably) the dashboard?

majamassarini: Oh you are right, in the past I had changed the `net.ipv4.ip_unprivileged_port_start` value in my laptop,so I did not realize this configuration was a problem!
Yes, I can try to use port 9443 for dashboard.

docker-compose.yml
```diff
@@ -143,6 +158,29 @@ services:
       - ./files/run_httpd.sh:/usr/bin/run_httpd.sh:ro,z
     user: "1024"
 
+  dashboard:
+    container_name: dashboard
+    image: quay.io/packit/dashboard:stg # built using local service API -> service.localhost (server name in packit-service.yaml)
```
jpopelka, line 163: Not sure I understand the comment.

jpopelka (approved):

Issue #1864, opened 2023-01-25
Database improvements: revisit the approach (spike)


lachmanfrantisek:
Let's spend some time going through [the original research](https://github.com/packit/research/pull/128) once a few steps have been implemented and make sure that the original approach is still valid and we can commit to finishing that.

The task of this issue is:
* [x] go through  [the original research](https://github.com/packit/research/pull/128)
* [x] take a look at the current state of the database schema
* [x] if needed, suggest any improvements to the plan (change/create/close issues; you can also update/enhance the original research)
  * #1985 
  * #1986 
  * updated #1773 
* [ ] probably do it as a hands-on session (meeting on Thursday)

mfocko:
Based on this article, I can imagine pipelines being implemented in a similar fashion: https://dev.to/tommydreamer57/how-functions-and-composite-types-changed-the-way-i-use-postgresql-7a2
And sqlalchemy has support for these if I'm not mistaken.

suggested by @csomh

PR #1984, opened 2023-04-03
feat: add DB diagram and makefile target

mfocko:
Related to #1864


jpopelka:
Nice!
We already have one [diagram in CONTRIBUTING.md](https://github.com/packit/packit-service/blob/main/CONTRIBUTING.md#er-diagram).
Can we perhaps move the old one to the alembic/ as well and show them both in the alembic/README?
And the CONTRIBUTING.md would just point to it.

mfocko:
> Nice! We already have one [diagram in CONTRIBUTING.md](https://github.com/packit/packit-service/blob/main/CONTRIBUTING.md#er-diagram). Can we perhaps move the old one to the alembic/ as well and show them both in the alembic/README? And the CONTRIBUTING.md would just point to it.

I would probably go with both diagrams in `alembic/` and point to them from `CONTRIBUTING.md`, if you want to move that part out of the `CONTRIBUTING.md`, I would probably go with a `docs/` directory structure as you did with the deployment repo :thinking: WDYT

jpopelka:
Please do, I'm not sure how it's different from what I suggested :)

lachmanfrantisek (approved):

PR #2002, opened 2023-04-12
Propagate exception message to propose_downstream logs

### LINKED ISSUE #1953: Always show the reason of failure in sync release jobs' logs
When the `propose-downstream` or `pull-from-upstream` fail, it should be easy for user to understand why. Depending where in the code the syncing fails, sometimes the actual error/exception is not present in the final logs shown in the dashboard. For propose-downstream, we at least show this in the check runs, but for pull-from-upstream the dashboard logs are the main source of information for user.
It looks like the problem is that the logs are collected only from `packit` ([related code](https://github.com/packit/packit-service/blob/2879822f3ea5372c07550341989eada19514ec2c/packit_service/worker/handlers/distgit.py#L341)), so if we catch the exception in `packit_service` and log it in the packit_service code, it is not also shown in dashboard.

Examples:
- https://dashboard.packit.dev/results/pull-from-upstream/1856 (the reason here was not existing tag)
- https://dashboard.packit.dev/results/propose-downstream/1834

nforro:
There is another case of failed `propose_downstream` with no error message in the logs:
https://dashboard.packit.dev/results/propose-downstream/2152

This ensures that any exception raised during `sync_release()` is logged.

Fixes #1953.

I'm not sure if I should also revert https://github.com/packit/packit/commit/1ed4f97292ccf316db037b8cc795c8e8efb3a5fe (to prevent the error from being logged twice).

### DIFF
--- a/packit_service/worker/handlers/distgit.py
+++ b/packit_service/worker/handlers/distgit.py
@@ -325,6 +325,8 @@ class AbstractSyncReleaseHandler(
             raise
         except Exception as ex:
             logger.debug(f"{self.sync_release_job_type} failed: {ex}")
+            # make sure exception message is propagated to the logs
+            logging.getLogger("packit").error(str(ex))
             # eat the exception and continue with the execution
             self.sync_release_helper.report_status_for_branch(
                 branch=branch,


majamassarini (approved):
Thanks!

lbarcziova (approved):
this should fix #1953 , right?

as for logging twice, is that the only place where this could happen?

nforro:
> this should fix #1953 , right?

Yes, it seems so, I'll link it.

> as for logging twice, is that the only place where this could happen?

I'm not sure. I'm only aware of that one instance because it was me who introduced it, but there could be more instances of exceptions being raised only after logging the error.

lbarcziova:
But since for Packit CLI users, it can be still useful, I would probably leave it as it is; having the message there twice is still a better UX than not having it at all. Thanks for taking care of this!

PR #2015, opened 2023-04-19
Unify checking of the trigger branch name in build and test handlers

### LINKED ISSUE #1981: Bug: test jobs are not filtered by trigger/branch
# Why is this a bug?

Lets consider the following configuration:
```yaml
jobs:
  - job: tests
    trigger: pull_request
  - job: copr_build
    trigger: pull_request
    project: CI
  - job: tests
    trigger: commit
    branch: main
  - job: copr_build
    trigger: commit
    branch: main
    project: nightly
  - job: copr_build
    trigger: release
    project: releases
```

Here `copr_build.commit` is only triggered on branch `main` which is only intended for actual nightly releases. There you also want to check if the nightly release is functional, e.g. after a PR is merged. So far so good. Then you also have a PR workflow that pushes to a separate project so that it doesn't interfere with the nightly builds, so you also include a test job for PR, and that works all fine.

The issue appears when you combine them and you are working on the same repository and not a fork. Because `tests.commit` is not filtered by branch (or trigger apparently), it always requests a test job on pushes to e.g. `fix/issue23` (and waits indefinitely). It pollutes the commit status, but that is not the main issue. The issue is when you then do a PR. Because the `test.commit` jobs are waiting for a `copr_build.commit` which is never executed, the CI checks never finish, and if the branch is protected, you cannot merge without force.

lbarcziova:
When the trigger type of the configured job is a commit, for builds and tests we additionally need to check whether the branch name from the trigger matches. Before the change, this was done only in Copr and Koji build handler pre-checks: unify the method that checks the match and use it also in TF handler pre-check.

Fixes #1981

TODO:

---

RELEASE NOTES BEGIN
When reacting to branch pushes, Packit now correctly checks whether the branch name matches the configuration for the test jobs with configured `commit` trigger.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/checker/copr.py
+++ b/packit_service/worker/checker/copr.py
@@ -2,27 +2,31 @@
 # SPDX-License-Identifier: MIT
 
 import logging
-import re
 
-from packit_service.worker.checker.abstract import ActorChecker, Checker
-from packit_service.worker.events.enums import GitlabEventAction
+from packit_service.constants import (
+    INTERNAL_TF_BUILDS_AND_TESTS_NOT_ALLOWED,
+)
+from packit_service.worker.checker.abstract import (
+    ActorChecker,
+    Checker,
+)
 from packit_service.worker.events import (
     MergeRequestGitlabEvent,
-    PushGitHubEvent,
-    PushGitlabEvent,
-    PushPagureEvent,
 )
+from packit_service.worker.events.enums import GitlabEventAction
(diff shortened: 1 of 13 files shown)


lbarcziova:
@lachmanfrantisek  it is a good point and initially, I wanted to do this when obtaining the matching handlers... but then I realised that we also have the `commit` trigger for downstream Koji builds and Bodhi updates and there we do not check the `branch`, therefore I left it handler-specific

PR #2024, opened 2023-04-27
fix: mention proper permission level and login for GitLab

mfocko:
When creating commit statuses to GitLab, Packit needs to have Developer permissions to create them. Also add a proper login instead of the generic “author of the comment”.

### DIFF
--- a/packit_service/worker/helpers/job_helper.py
+++ b/packit_service/worker/helpers/job_helper.py
@@ -181,14 +181,15 @@ class BaseJobHelper:
         so we can extend it in subclasses easily.
         """
         if self.is_gitlab_instance and not self.is_reporting_allowed:
+            login = self.project.service.user.get_username()
             description = (
-                f"{description}\n\n---\nPackit-User does not have access to the "
+                f"{description}\n\n---\nPackit does not have access to the "
                 "source project (=usually author's fork of the project). "
                 "(This is only about the representation of the results. "
                 "Packit is still able to do its job without having the permissions.)\n\n"
                 "*In case you wish to receive commit statuses instead of comments, please "
-                "add login of the author of this comment to your fork with a role "
-                "`Reporter`.*"
+                f"add {login} to your fork with a role `Developer`.*"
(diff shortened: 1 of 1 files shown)


packit_service/worker/helpers/job_helper.py
```diff
@@ -165,14 +165,14 @@ def _report(
...
             description = (
-                f"{description}\n\n---\nPackit-User does not have access to the "
+                f"{description}\n\n---\nPackit does not have access to the "
                 "source project (=usually author's fork of the project). "
                 "(This is only about the representation of the results. "
                 "Packit is still able to do its job without having the permissions.)\n\n"
                 "*In case you wish to receive commit statuses instead of comments, please "
-                "add login of the author of this comment to your fork with a role "
-                "`Reporter`.*"
+                f"add {login} to your fork with a role `Developer`.*"
```
lbarcziova, line 175: would it make sense to link https://docs.gitlab.com/ee/user/permissions.html here and in the issue as well?

mfocko: I think it would be ideal to link our own docs… The fact that there are:

* Developer
* Reporter
* Maintainer

present basically means that we're not capable of keeping it in sync during “smallish” changes :D And it's definitely not an issue of the TODO list in PRs, cause I had problems “grepping” through the service code to find this hunk on which we're discussing right now :D I knew that we add this footer to the commit comments, but no keywords found it for me

lbarcziova: ok, then we should fix our docs and link them here, right?

lbarcziova: just going through the opened PRs and noticed this got forgotten, so what should we do here?

mfocko: I would avoid as much details as possible and just link the docs, which I think you have already updated?

lbarcziova: no, I did not touch the docs related to this PR

lbarcziova (approved):

PR #2037, opened 2023-05-05
Sentry: Set-up Transactions/Performance

jpopelka:
Setting the traces sampler seems to be all we need to see the Transactions/Performance.

* https://docs.sentry.io/platforms/python/performance
* https://docs.sentry.io/platforms/python/performance/instrumentation/automatic-instrumentation

From [docs](https://docs.sentry.io/platforms/python/configuration/sampling/#sampling-transaction-events):
_Choose a sampling rate with the goal of finding a balance between performance and volume concerns with data accuracy. You don't want to collect too much data, but you want to collect sufficient data from which to draw meaningful conclusions. If you’re not sure what rate to choose, start with a low value and gradually increase it as you learn more about your traffic patterns and volume._

### DIFF
--- a/packit_service/sentry_integration.py
+++ b/packit_service/sentry_integration.py
@@ -4,7 +4,7 @@
 import logging
 from contextlib import contextmanager
 from os import getenv
-from typing import List
+from typing import List, Dict
 
 from sentry_sdk.integrations import Integration
 from sentry_sdk.integrations.logging import LoggingIntegration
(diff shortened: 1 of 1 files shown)


packit_service/sentry_integration.py
```diff
@@ -66,6 +83,7 @@ def configure_sentry(
         secret_key,
         integrations=integrations,
         environment=getenv("DEPLOYMENT"),
+        traces_sampler=traces_sampler,
```
mfocko, line 86: I've briefly read through the docs, this does not affect the exceptions, right? Because the docs suggest so :D

jpopelka: Can you be more specific? What docs, where? I couldn't find it.

mfocko: from the comment https://docs.sentry.io/platforms/python/configuration/sampling/

oh… i see… I got confused cause docs mention sampling of »the errors« as the first thing, but this is just for traces.

mfocko (approved):
:ship: it

PR #2047, opened 2023-05-11
Don't react to pull request comments by Packit

lbarcziova:
This was already done for GitHub, mirror the behaviour for GitLab and Pagure.

Related to #2044

---

RELEASE NOTES BEGIN
TBD
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/parser.py
+++ b/packit_service/worker/parser.py
@@ -702,6 +702,10 @@ class Parser:
             logger.warning("No Gitlab username from event.")
             return None
 
+        if actor in {"packit-as-a-service", "packit-as-a-service-stg"}:
+            logger.debug("Our own comment.")
+            return None
+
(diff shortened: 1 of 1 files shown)


packit_service/worker/parser.py
```diff
@@ -702,6 +702,10 @@ def parse_merge_request_comment_event(
             logger.warning("No Gitlab username from event.")
             return None
 
+        if actor in {"packit-as-a-service", "packit-as-a-service-stg"}:
```
mfocko, line 705: Please add a TODO to make it part of the git forge config.

mfocko (approved):

packit_service/worker/parser.py
lbarcziova: discussed in person, the problem here is we want to ignore comments by both instances, not only for the one that is processing the message

PR #2048, opened 2023-05-11
Do not load Packit config for non-packit comment commands

lbarcziova:
If the event we are reacting to is a PR comment, check whether there is a packit command included. If not, do not try to load the config, end directly.

Related to #2044 

---

RELEASE NOTES BEGIN
￼
RELEASE NOTES END


mfocko (approved):
lgtm

PR #2063, opened 2023-05-24
tests: Mock 'get_repo_download_url()' calls

csomh:
So that they don't work with real Copr repositories.


mfocko (approved):

majamassarini (approved):
Much better now, thanks!!!

PR #2065, opened 2023-05-24
Fix/packit service/1328

### LINKED ISSUE #1328: Database: project reference model
Follow-up for: packit/research#128

The current situation with project/trigger/commit-related models:

* Project for trigger not effective to get for multiple entries.
* `JobTriggerModel` is more a git reference than a trigger.
* Commit is spread across the models (in build/test models and release).

Let's transform `JobTriggerModel` into something better-named and better-connected. We would like to use it as a project/git reference.

* [x] @lachmanfrantisek promise to explain this task and showcase it via the diagram
* [x] Pick a better name for `JobTriggerModel` (something like `ProjectReference` or `CommitModel`).
* [x] Add a new column for commit hash (e.g. `commit`/`commit_hash`/ `commit_sha`).
* [ ] Connect the model directly to the  `ProjectModel`.  Each commit occurrence (e.g. pr x branch push) will have a different model -- so, no change in the behaviour...;)
  * [ ] With this, we can simplify a lot of queries that try to get a project model throught the pr/branch/release/issue model. Either fix it (if you think it's simple) or create a follow-up card to do it. 
* [x] Create migration steps that add the newly introduced connections/values also for the already present data.
* [x] Regenerate the database diagram

majamassarini:
Fixes #1328 

---

RELEASE NOTES BEGIN
RELEASE NOTES END


packit_service/models.py
```diff
@@ -1084,88 +1096,88 @@ def __repr__(self):
...
-    * Each PipelineModel has to be connected to exactly one JobTriggerModel.
-    * There can be multiple PipelineModels for one JobTriggerModel.
-      (e.g. For each push to PR, there will be new PipelineModel, but same JobTriggerModel.)
+    * Each PipelineModel has to be connected to exactly one ProjectEventModel.
+    * There can be multiple PipelineModels for one ProjectEventModel.
+      (e.g. For each push to PR, there will be new PipelineModel, but same ProjectEventModel.)
     """
 
-    __tablename__ = "job_triggers"
+    __tablename__ = "project_events"
```
lachmanfrantisek, line 1151: This will require manual intervention during the migration to not loose data... (Alembic will remove the old one and create a new one...)

lachmanfrantisek: You can also leave the table as is and rename it manually as a follow-up.

alembic/versions/4033221ea50c_rename_jobtriggermodel.py
```diff
@@ -0,0 +1,58 @@
...
+    op.alter_column("project_events", "trigger_id", new_column_name="event_id")
+
+    op.execute("ALTER SEQUENCE job_triggers_id_seq RENAME TO project_events_id_seq")
+    op.execute("ALTER INDEX job_triggers_pkey RENAME TO project_events_pkey")
+    op.execute(
+        "ALTER INDEX ix_job_triggers_trigger_id RENAME TO ix_project_events_event_id"
+    )
+
+    op.alter_column("pipelines", "job_trigger_id", new_column_name="project_event_id")
+    op.drop_constraint("runs_job_trigger_id_fkey", "pipelines", type_="foreignkey")
```
lachmanfrantisek, line 29: Why we are dropping the constraint?

lachmanfrantisek (commented):
Looks good to me!

packit_service/models.py
lachmanfrantisek: Aha, ok. You've already handled the renaming..;)

docker-compose.yml
```diff
@@ -123,14 +135,15 @@ services:
       - redis
       - postgres
     ports:
-      - 8443:8443
+      - 443:8443
```
jpopelka, line 138: oh, I completely forgot about #1967 
I'll try to take another look.

alembic/versions/4033221ea50c_rename_jobtriggermodel.py
majamassarini: Because I didn't find a way to rename it. So the only solution for me it was to drop it and re-create it with the new name.

lachmanfrantisek (approved):

PR #2066, opened 2023-05-24
GitHub: Consider only the events we are interested in

csomh:
This extends the interested() method in the GitHub webhook, so that checks on the event type and action are done in the API. This will help avoid creating Celery tasks for these events.

This change makes similar checks in Parser() redundant, but I've decided not to remove those, yet, mainly b/c it's hard to tell how that will influence parsing and telling which event is handled.

TODO:


csomh:
We should find out if this really makes a difference by looking at this diagram on our metrics dashboard, after `stable` is moved nex week. The expectation is that the percentage of not handled events should go significantly down.

image

PR #2067, opened 2023-05-25
GitHub: Pass the source and the event type to the worker

csomh:
And use this information to be explicit about which parser to use to create the Event object.

### DIFF
--- a/packit_service/service/api/webhooks.py
+++ b/packit_service/service/api/webhooks.py
@@ -96,7 +96,11 @@ class GithubWebhook(Resource):
 
         celery_app.send_task(
             name=getenv("CELERY_MAIN_TASK_NAME") or CELERY_DEFAULT_MAIN_TASK_NAME,
-            kwargs={"event": msg},
+            kwargs={
+                "event": msg,
+                "source": "github",
+                "event_type": request.headers.get("X-GitHub-Event"),
+            },
         )
         github_webhook_calls.labels(result="accepted", process_id=os.getpid()).inc()
 
@@ -117,11 +121,10 @@ class GithubWebhook(Resource):
 
         interests = {
             "check_run": action == "rerequested",
+            "issue_comment": action in {"created", "edited"},
             "pull_request": action in {"opened", "reopened", "synchronize"},
-            "pull_request_review_comment": action in {"created", "edited"},
-            "issue_comment": action == "created",
-            "release": action == "published",
             "push": not deleted,
+            "release": action == "published",
             "installation": action == "created",
         }
         _interested = interests.get(event, False)
--- a/packit_service/worker/jobs.py
+++ b/packit_service/worker/jobs.py
@@ -16,6 +16,7 @@ import celery
 from ogr.exceptions import GithubAppNotInstalledError
 from packit.config import JobConfig, JobType, JobConfigTriggerType
 from packit.config.job_config import DEPRECATED_JOB_TYPES
+from packit.utils import nested_get
 from packit_service.config import PackageConfig, PackageConfigGetter, ServiceConfig
 from packit_service.constants import (
     DOCS_CONFIGURATION_URL,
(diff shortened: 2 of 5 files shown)


csomh:
@TomasTomecek I ment something like this when talking about preserving the information about the type of event received: to be explicit about which parser to use to get the arguments for the Event object from the event data, and don't let `Parser.parse_event()` second-guess the service and event type again.

This PR re-factors only the GitHub parts, follow up work is required to cover the other parsers, too. 

Note, that this change should have no performance impact (#2066 is the one which should reduce the number of Celery tasks which ultimately end up not doing anything), but should make the code more explicit.

packit_service/service/api/webhooks.py
```diff
@@ -96,7 +96,11 @@ def post(self):
 
         celery_app.send_task(
             name=getenv("CELERY_MAIN_TASK_NAME") or CELERY_DEFAULT_MAIN_TASK_NAME,
-            kwargs={"event": msg},
+            kwargs={
+                "event": msg,
+                "service": "github",
```
mfocko, line 101: When you mentioned `service`, I thought we're talking about something like `packit`, `stream`, `fedora-source-git`… Maybe I would reconsider and use something like `source`?

packit_service/worker/jobs.py
```diff
@@ -126,17 +127,27 @@ def service_config(self) -> ServiceConfig:
...
         Args:
             event: Dict with webhook/fed-msg payload.
+            service: Name of the service.
+            event_type: Type of the event.
 
         Returns:
             List of results of the processing tasks.
         """
-        event_object: Optional[Event] = Parser.parse_event(event)
+        parser = nested_get(
```
mfocko, line 147: :+1:

packit_service/worker/parser.py
```diff
@@ -1544,3 +1544,15 @@ def parse_vm_image_build_result_event(event) -> Optional[VMImageBuildResultEvent
             message,
             created_at,
         )
+
+    MAPPING = {
```
mfocko, line 1548: I wonder if we could incorporate this with the decorators.

mfocko (commented):
:+1:

packit_service/worker/parser.py
mfocko: I'm not sure if I can persuade you, cause I thought about that issue when writing it down :smile:

lbarcziova (commented):
nice changes!

packit_service/worker/parser.py
```diff
@@ -1544,3 +1544,15 @@ def parse_vm_image_build_result_event(event) -> Optional[VMImageBuildResultEvent
...
+        "github": {
+            "check_run": parse_check_rerun_event,
+            "pull_request": parse_pr_event,
+            "pull_request_review_comment": parse_pull_request_comment_event,
+            "issue_comment": parse_issue_comment_event,
```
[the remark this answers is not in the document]
csomh: Good catch! Yes, it can, I completely missed it in the [documentation](https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads#issue_comment).

The issue in the code above is that I mistakenly assumed, that we are listening to `pull_request_review_comment` events, but we are not! We are not even subscribed to those! :smile:

Instead, `issue_comment` events are parsed by `parse_issue_comment_event()` if they don't have a `issue.pull_request` field, and by `parse_pull_request_comment_event()` if they have.

packit_service/service/api/webhooks.py
```diff
@@ -96,7 +96,11 @@ def post(self):
 
         celery_app.send_task(
             name=getenv("CELERY_MAIN_TASK_NAME") or CELERY_DEFAULT_MAIN_TASK_NAME,
-            kwargs={"event": msg},
+            kwargs={
+                "event": msg,
+                "source": "github",
+                "event_type": request.headers.get("X-GitHub-Event"),
```
TomasTomecek, line 102: sweet! you're right it's amazing we know this up front and can decide without spawning the celery tasks

TomasTomecek (commented):
:clap: :clap: :clap:

packit_service/worker/parser.py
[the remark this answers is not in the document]
csomh: Fixed this by introducing `parse_github_comment_event()`. PTAL.

packit_service/service/api/webhooks.py
jpopelka: #2066 is about spawning fewer celery tasks. This one is about not calling all the [Parser.parse_xyz_event()](https://github.com/packit/packit-service/blob/main/packit_service/worker/parser.py#L132) to realize what type of event we're parsing when we can get the info from outside and then call only one of the `Parser.parse_xyz_event()` to parse it. But definitely a good improvement.

jpopelka (approved):

PR #2073, opened 2023-05-30
Cover the rest of the parsers

jpopelka:
Follow-up of #2067

@majamassarini I was hoping you could help me understand where we send a celery task with an event later parsed in [here](https://github.com/packit/packit-service/blob/main/packit_service/worker/parser.py#L1532). I was looking for another `celery_app.send_task()` but found none.

### DIFF
--- a/packit_service/service/api/testing_farm.py
+++ b/packit_service/service/api/testing_farm.py
@@ -59,12 +59,14 @@ class TestingFarmResults(Resource):
             logger.info(f"/testing-farm/results {exc}")
             return str(exc), HTTPStatus.UNAUTHORIZED
 
-        # There's only one key in the msg,
-        # so make sure we don't confuse this with something else
-        msg["source"] = "testing-farm"
+        msg["source"] = "testing-farm"  # TODO: remove me
         celery_app.send_task(
             name=getenv("CELERY_MAIN_TASK_NAME") or CELERY_DEFAULT_MAIN_TASK_NAME,
-            kwargs={"event": msg},
+            kwargs={
+                "event": msg,
+                "source": "testing-farm",
+                "event_type": "results",
+            },
         )
 
         return "Test results accepted", HTTPStatus.OK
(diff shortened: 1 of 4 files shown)


majamassarini:
> Follow-up of #2067
> 
> @majamassarini I was hoping you could help me understand where we send a celery task with an event later parsed in [here](https://github.com/packit/packit-service/blob/main/packit_service/worker/parser.py#L1532). I was looking for another `celery_app.send_task()` but found none.

I would say that there is no celery task involved here. This is an Event created (internally) by a babysit task which is later used (internally) to find the right handler to deal with the completed vm build. The babysit task is calling the `run_job` for the `handler` which holds this "new" manually crafted Event. I don't know if this answers your question.

jpopelka:
> I don't know if this answers your question.

A bit :-)
I can see all you wrote in the [update_vm_image_build()](https://github.com/packit/packit-service/blob/65fab286bff5df48bf7a03ec2cc3b1789d2263d0/packit_service/worker/helpers/build/babysit.py#L349) called by the babysit task.

But I can't see how/when the [parse_vm_image_build_result_event()](https://github.com/packit/packit-service/blob/65fab286bff5df48bf7a03ec2cc3b1789d2263d0/packit_service/worker/parser.py#L1534) is actually utilized.
 
If the `Event` (VMImageBuildResultEvent) is just internally created and used only in the `update_vm_image_build()`
and it (as a dictionary to be parsed into an `Event`) doesn't come from any outer source; is it possible that the `parse_vm_image_build_result_event()` is actually never utilized?

majamassarini:
> If the `Event` (VMImageBuildResultEvent) is just internally created and used only in the `update_vm_image_build()` and it doesn't come (as a dictionary to be parsed into an `Event`) from any outer source; is it possible that the `parse_vm_image_build_result_event()` is actually never utilized?

Yes I think you are right! That code is not useful.

jpopelka:
recheck

lachmanfrantisek (approved):

majamassarini (approved):
Thanks for cleaning the unuseful code!

majamassarini (approved):
Thanks for cleaning the unuseful code!

lbarcziova (approved):
thanks!

jpopelka:
regate

PR #2075, opened 2023-06-01
Fix results events reporting from packit-service

Frawless:
Fixes issue introduced in https://github.com/packit/packit-service/pull/2069. When `manual_trigger=True` the results event are filtered and the results are not propagated properly back to je github/gitlab.

### DIFF
--- a/packit_service/worker/events/event.py
+++ b/packit_service/worker/events/event.py
@@ -492,3 +492,12 @@ class AbstractForgeIndependentEvent(Event):
             models=CoprBuildTargetModel.get_all_by_commit(commit_sha=self.commit_sha),
             statuses_to_filter_with=statuses_to_filter_with,
         )
+
+
+class AbstractResultEvent(AbstractForgeIndependentEvent):
+    """
+    This class is used only as an Abstract for result events to
+    allow Steve properly filter jobs with manual trigger.
+    """
+
+    pass
--- a/packit_service/worker/jobs.py
+++ b/packit_service/worker/jobs.py
@@ -38,6 +38,7 @@ from packit_service.worker.events.comment import (
     AbstractCommentEvent,
     AbstractIssueCommentEvent,
 )
+from packit_service.worker.events.event import AbstractResultEvent
 from packit_service.worker.handlers import (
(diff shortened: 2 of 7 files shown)


packit_service/worker/jobs.py
```diff
@@ -71,7 +72,7 @@
 logger = logging.getLogger(__name__)
 
 
-MANUAL_EVENTS = [AbstractCommentEvent, CheckRerunEvent]
+MANUAL_EVENTS = [AbstractCommentEvent, AbstractResultEvent, CheckRerunEvent]
```
lbarcziova, line 75: just for the sake of clarity of the code, I would either rename the list to something like `MANUAL_OR_RESULT_EVENTS` or add one more subcondition to the condition you have introduced in the changes before to check whether the event is AbstractResultEvent

lbarcziova (commented):
nice work!

packit_service/worker/jobs.py
Frawless: Makes sense. Pushing changes.

packit_service/worker/events/event.py
```diff
@@ -492,3 +492,21 @@ def get_all_build_targets_by_status(
...
+    def __init__(
+        self,
+        created_at: Union[int, float, str] = None,
+        project_url=None,
+        pr_id: Optional[int] = None,
+        actor: Optional[str] = None,
+    ):
+        super().__init__(
+            created_at=created_at, project_url=project_url, pr_id=pr_id, actor=actor
+        )
```
lachmanfrantisek, line 512:
```suggestion
    pass
```

This should work fine if we don't do anything there.

lachmanfrantisek (approved):
Nice!

Just one small suggestion. Thanks for adding the test cases!

lbarcziova (approved):
:100:

packit_service/worker/events/event.py
lbarcziova: and when changing this, could you please squash your commits?

Frawless: good, didn't know about that :) Pushed

PR #2078, opened 2023-06-01
fix(api): add CORS

### LINKED ISSUE #1883: Bad request returns from API endpoint are missing CORS
CORS is required when APIs are involved to ensure that clients using it can get the data within their browser. If you screw up the request params or something goes wrong it currently does not include `Access-Control-Allow-Origin`

For example see the response headers for this 
https://stg.packit.dev/api/copr-builds?page=1&per_page=a

image

mfocko:
In the issues of flask-cors there are some mentions that in development/debugging environment flask behaves differently than in production one.

It definitely works for correct requests, even after removing our own “hack” with manual addition of the CORS. However I am not able to get the correct headers via `podman-compose` even with the CORS wrapper. It just doesn't return it, not sure if it is related to the note above regarding the environment, or not.

Fixes #1883

### DIFF
--- a/files/install-deps.yaml
+++ b/files/install-deps.yaml
@@ -24,6 +24,7 @@
           - python3-redis # celery[redis]
           - python3-lazy-object-proxy
           - python3-flask-restx
+          - python3-flask-cors
           - python3-flexmock # because of the hack during the alembic upgrade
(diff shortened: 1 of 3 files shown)


mfocko:
recheck

PR #2079, opened 2023-06-05
Add --identifier to packit comment command

Frawless:
TODO:

- [ ] ‹fill in›

Fixes

Related to

Merge before/after

---

RELEASE NOTES BEGIN
￼
RELEASE NOTES END


packit_service/worker/handlers/testing_farm.py
```diff
@@ -163,21 +163,26 @@ def _get_or_create_group(
...
-                    web_url=None,
-                    test_run_group=group,
-                    copr_build_targets=test_builds,
-                    # In _payload() we ask TF to test commit_sha of fork (PR's source).
-                    # Store original url. If this proves to work, make it a separate column.
-                    data={"base_project_url": self.project.get_web_url()},
+            if (
+                not self.testing_farm_job_helper.comment_arguments.identifier
+                or self.testing_farm_job_helper.comment_arguments.identifier
+                == self.job_config.identifier
```
lbarcziova, line 169: I would probably move this check into the [`checkers` for TF](https://github.com/packit/packit-service/blob/main/packit_service/worker/checker/testing_farm.py)  (probably create a new one, they are then listed [here](https://github.com/packit/packit-service/blob/d708d00beed1738c824f794c0e633a94d49e81f7/packit_service/worker/handlers/testing_farm.py#L119)) , if we checked it here, all the tests would already have the `The task was accepted..` status. If we create a check for this in checkers, the tasks for non-relevant job configurations will not be created

packit_service/worker/helpers/testing_farm.py
```diff
@@ -51,6 +52,41 @@
 logger = logging.getLogger(__name__)
 
 
+class CommentArguments:
```
lbarcziova, line 55: I like the extraction into a new class!

packit_service/worker/checker/testing_farm.py
```diff
@@ -125,3 +125,21 @@ def pre_check(self) -> bool:
...
+    """
+
+    def pre_check(self) -> bool:
+        if (
+            not self.testing_farm_job_helper.comment_arguments.identifier
+            or self.testing_farm_job_helper.comment_arguments.identifier
+            == self.job_config.identifier
+        ):
+            return True
+
```
lbarcziova, line 144: can we add logging here, something like "Skipping running tests for the job, identifiers don't match" and log the values as well?

packit_service/worker/helpers/testing_farm.py
```diff
@@ -229,16 +259,21 @@ def is_comment_event(self) -> bool:
...
 
     def is_test_comment_event(self) -> bool:
-        return self.is_comment_event() and self.comment_command_parts[0] == "test"
+        return (
+            self.is_comment_event() and self.comment_arguments.packit_command == "test"
+        )
 
     def is_test_comment_pr_argument_present(self):
-        return self.is_test_comment_event() and len(self.comment_command_parts) == 2
+        return self.is_test_comment_event() and len(self.comment_arguments.pr_argument)
```
lbarcziova, line 273:
```suggestion
        return self.is_test_comment_event() and self.comment_arguments.pr_argument
```

packit_service/worker/helpers/testing_farm.py
```diff
@@ -229,16 +259,21 @@ def is_comment_event(self) -> bool:
...
+
+    def is_test_comment_identifier_present(self):
+        return self.is_test_comment_event() and len(self.comment_arguments.identifier)
```
lbarcziova, line 276:
```suggestion
        return self.is_test_comment_event() and self.comment_arguments.identifier
```

packit_service/worker/helpers/testing_farm.py
```diff
@@ -1146,15 +1196,12 @@ def _parse_comment_pr_argument(self) -> Optional[Tuple[str, str, str]]:
...
             logger.debug(
                 "Unexpected format of the test argument: "
-                f"not able to split the test argument {pr_argument} with '/'."
+                f"not able to split the test argument "
+                f"{self.comment_arguments.pr_argument} with '/'."
             )
             return None
         namespace, repo = namespace_repo
 
-        logger.debug(
```
lbarcziova, line 1154: why removing this? :thinking:

packit_service/worker/helpers/testing_farm.py
```diff
@@ -554,6 +589,21 @@ def check_comment_pr_argument_and_report(self) -> bool:
 
         return True
 
+    def check_comment_identifier_and_report(self) -> bool:
```
lbarcziova, line 592: maybe I am missing something, but is this method used somewhere?

tests/unit/test_checkers.py
```diff
@@ -422,3 +427,62 @@ def test_koji_branch_merge_queue():
...
+            id="No identifier specified",
+        ),
+        pytest.param(
+            "/packit-dev test --identifier my-id-2",
+            False,
+            id="Non-matching identifier specified",
+        ),
+    ),
+)
+def test_tf_comment_identifier(comment, result):
```
lbarcziova, line 452: :+1:

tests/unit/test_jobs.py
```diff
@@ -2871,6 +2871,47 @@ def __init__(self):
             ],
             {},
         ),
+        # TODO
```
lbarcziova, line 2874: I think these tests should not be extended as the check happens in the pre_check

lbarcziova (commented):
another nice contribution, thanks!

just few small notes

packit_service/worker/helpers/testing_farm.py
Frawless: Guess I removed it because I wanted to move it to a different place. I will put it back.

packit_service/worker/checker/testing_farm.py
```diff
@@ -125,3 +125,26 @@ def pre_check(self) -> bool:
...
+        if (
+            not self.testing_farm_job_helper.comment_arguments.identifier
+            or self.testing_farm_job_helper.comment_arguments.identifier
+            == self.job_config.identifier
+        ):
+            logger.info(
+                f"Skipping running tests for the job, identifiers doesn't match "
+                f"(job:{self.job_config.identifier} "
+                f"!= comment:${self.testing_farm_job_helper.comment_arguments.identifier})"
+            )
```
lbarcziova, line 147: this should be moved to the "`else`" branch

packit_service/worker/helpers/testing_farm.py
lbarcziova: I am thinking about what would be the best UX here, probably rather only adding a comment with the same description instead of setting checks for all test jobs :thinking: WDYT?

But it's a little bit more tricky where to place this check. I can take a look later

lbarcziova (commented):
otherwise LGTM :+1:

packit_service/worker/helpers/testing_farm.py
Frawless: Ye that sounds reasonable to me. So I will remove the method from the PR for now.

lbarcziova (approved):
thank you! :rocket:

PR #2083, opened 2023-06-06
Use GetVMImageDataMixin to access data

lbarcziova:
Add copr_build property to the mixin to get the corresponding CoprBuildTargetModel (also add check for trigger match when getting it) and use the properties of the mixin both in pre-check and in handler itself.
Should fix https://red-hat-it.sentry.io/issues/4231239940/events/d76b2c9df3984b10943195d8c38ded67/

---

RELEASE NOTES BEGIN
N/A
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/mixin.py
+++ b/packit_service/worker/mixin.py
@@ -20,6 +20,7 @@ from packit.vm_image_build import ImageBuilder
 from ogr.abstract import GitProject, PullRequest, PRStatus
 
 from packit_service.config import ServiceConfig
+from packit_service.models import CoprBuildTargetModel, BuildStatus
 from packit_service.worker.reporting import BaseCommitStatus
 from packit_service.worker.events import EventData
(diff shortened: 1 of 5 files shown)


packit_service/worker/mixin.py
```diff
@@ -375,17 +377,15 @@ def identifier(self) -> str:
 
     @property
     def owner(self) -> str:
-        return self.job_config.owner
+        return self.job_config.owner or self.copr_build.owner
```
majamassarini, line 380: Since the `copr_build` can be `None` and also the `job_config.owner` should we manage it in a different way? Or do you think it will not happen that both of them are `None`?

majamassarini (approved):

packit_service/worker/mixin.py
lbarcziova: good point! fixed

Issue #1829, opened 2023-01-05
Implement retriggering of pull from upstream


lbarcziova:
To be consistent with other jobs, people should be able to retrigger `pull_from_upstream` job as well.

- [ ] discuss and decide where this should be possible (e.g. via issues in dist-git, issue_repository, see #1536)
- [ ] implement it
- [ ] document it

lbarcziova:
Ideas from the architecture on where to trigger the pull from upstream:
- issue in dist-git repo
  - enabling the issue tracker via web UI doesn't seem to work
- comment in (any) dist-git PR (we already react to these)

Outcome: do it via dist-git comments.

ptoscano:
> Ideas from the architecture on where to trigger the pull from upstream:

A possible another way: running `packit pull-from-upstream` while being in a dist-git clone.

nforro:
> A possible another way: running `packit pull-from-upstream` while being in a dist-git clone.

That wouldn't be so easy to implement, packit CLI is not a client for Packit Service, I guess we would have to implement an (authenticated?) API endpoint in Packit Service for that.

lbarcziova:
@ptoscano but if you would be interested in a CLI command, there is the [packit propose-downstream](https://packit.dev/docs/cli/propose-downstream/) that can be run in the upstream clone (but it requires user's credentials)

PR #2087, opened 2023-06-07
Implement retriggering of pull-from-upstream via comment in dist-git PR

nforro:
I'll be glad for any feedback. I'm particularly not happy with this:
https://github.com/packit/packit-service/blob/b797adbab1ef2086ad1a5d6f0376ab5bbfa186ce/packit_service/worker/helpers/sync_release/sync_release.py#L82-L87

I thought about using a "hack" similar to this in `PullRequestModel`:
https://github.com/packit/packit-service/blob/cfb5e7f31a0e524bf9bae49c2f67d77db1da8853/packit_service/models.py#L1001-L1002

That works for this case but breaks some other jobs.

Fixes: #1829.

RELEASE NOTES BEGIN
`pull_from_upstream` jobs can now be retriggered with a comment `/packit pull-from-upstream` in a dist-git pull request.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/checker/distgit.py
+++ b/packit_service/worker/checker/distgit.py
@@ -14,6 +14,7 @@ from packit_service.worker.events import (
     IssueCommentGitlabEvent,
 )
 from packit_service.worker.events.pagure import PullRequestCommentPagureEvent
+from packit_service.worker.events.new_hotness import NewHotnessUpdateEvent
 from packit_service.worker.handlers.mixin import GetProjectToSyncMixin
 from packit_service.worker.mixin import (
     GetPagurePullRequestMixin,
@@ -131,7 +132,7 @@ class IsProjectOk(Checker, GetProjectToSyncMixin):
         return self.project_to_sync is not None
 
 
-class ValidInformationForPullFromUpstream(Checker):
+class ValidInformationForPullFromUpstream(Checker, GetPagurePullRequestMixin):
     """
     Check that package config (with upstream_project_url set) is present
     and that we were able to parse repo namespace, name and the tag name.
@@ -141,6 +142,10 @@ class ValidInformationForPullFromUpstream(Checker):
     def pre_check(self) -> bool:
         valid = True
         msg_to_report = None
+        issue_title = (
+            "Pull from upstream could not be run for update "
+            f"{self.data.event_dict.get('version')}"
+        )
 
         if not self.package_config.upstream_project_url:
             msg_to_report = (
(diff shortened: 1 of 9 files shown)


jpopelka (approved):
I can give you ACK but won't help you with your questions, probably worth waiting for a review from Laura/Franta.

packit_service/worker/checker/distgit.py
```diff
@@ -131,48 +132,77 @@ def pre_check(self) -> bool:
...
+                report_in_issue_repository(
+                    issue_repository=self.job_config.issue_repository,
+                    service_config=self.service_config,
+                    title=f"Pull from upstream could not be run for update "
+                    f"{self.data.event_dict.get('version')}",
+                    message=msg_to_report,
+                    comment_to_existing=msg_to_report,
+                )
+
+            return valid
```
lbarcziova, line 180: I would say these checks are relevant also when retriggering, WDYT?

therefore, I think what needs to be implemented is getting the last tag since from the PR we do not have any information about the upstream tag

nforro: Yes, I suppose it makes sense. Where should I implement that though? It can't be in the handler because it has to run before pre-check, right?

lbarcziova: you're right, it's a little bit tricky as it also doesn't make sense for the PR event itself...so maybe we can allow no `tag_name` in the pre_check if we are handling PR event, and add the getting of the tag in the handler if it is None?

lbarcziova:
@nforro that check checks whether we know what the upstream repo is, so I would say it is quite important

nforro:
> @nforro that check checks whether we know what the upstream repo is (=> we were able to parse the name and namespace), so I would say it is quite important

Ok, so I have to modify `PullRequestCommentPagureEvent` and add those properties (similarly to `NewHotnessUpdateEvent`, get it from `upstream_project_url`), right?

lbarcziova:
@nforro yes, I would say so...they do not necessarily need to have the same name as the semantics for the events may be different, but in that case, the handler would need to be adjusted so it could work with the correct arguments

nforro:
Ok, I've added the properties, but it makes `test_pull_request_retrigger_bodhi_update_no_koji_data` and `test_pull_request_retrigger_bodhi_update_with_koji_data` fail because of an invalid `project_url` (but that's on purpose I assume), not sure what I can do about that.

Also, the actual `repo_name` and `repo_namespace` represent the dist-git project, not upstream project, because of this:
https://github.com/packit/packit-service/blob/c136c769aa2181bbe0ee3382bee00463b6c2ab6f/packit_service/worker/events/event.py#L455-L462

That's also why I had to use `job_config` rather than `package_config` here:
https://github.com/packit/packit-service/blob/c136c769aa2181bbe0ee3382bee00463b6c2ab6f/packit_service/worker/handlers/distgit.py#L517-L518

lbarcziova:
so would it be possible to have `upstream_repo_name` and `upstream_repo_namespace` in the PagurePullRequestEvent and then, in pre-check and handler unify the access to these properties?

nforro:
Thanks, let me try that.

nforro:
Actually, renaming the properties doesn't change anything, because they still have to be present in the event dict. But the failed tests turned out to be an easy fix.

I'm still curious about this piece of code:
https://github.com/packit/packit-service/blob/c136c769aa2181bbe0ee3382bee00463b6c2ab6f/packit_service/worker/events/event.py#L455-L462

I don't understand from the comment if it is wrong and should be removed or vice versa. But it doesn't seem right since it replaces `upstream_project_url` with dist-git project URL, at least in this (retriggering pull-from-upstream) case. That being said, I'm still not able to find any usage of `repo_name` and `repo_namespace` in the handler (that's why I was wondering if the check is actually necessary).

lbarcziova:
> But it doesn't seem right since it replaces upstream_project_url with dist-git project URL, at least in this (retriggering pull-from-upstream) case.

my comment meant that I would also like to remove the whole code block and not set any values in package config

as for `repo_name` and `repo_namespace`, they are definitely used in the `ProjectReleaseModel` that is the trigger for pull from upstream, but if in the retriggering use case the trigger will be the PR and if there are no other usages, then  it may be not necessary

nforro:
> my comment meant that I would also like to remove the whole code block and not set any values in package config

Ok, I removed it. All tests pass, hopefully nothing else will break.

> but if in the retriggering use case the trigger will be the PR and if there are no other usages, then it may be not necessary

This seems to be the case, but I don't think the check hurts anything, so I left it there.

lbarcziova (approved):
thanks!! let's ship it and test is

PR #2091, opened 2023-06-15
Make the event-took-more-than-Xs-to-process error message consistent

jpopelka:
With different response times, the error message looks different each time and it generates a new Sentry issue.
With this change, we'll reuse an existing Sentry issue for those events.

### DIFF
--- a/packit_service/worker/jobs.py
+++ b/packit_service/worker/jobs.py
@@ -818,7 +818,7 @@ class SteveJobs:
             # we need more info why this has happened
             logger.debug(f"Event dict: {self.event}.")
             logger.error(
-                f"Event {self.event.__class__.__name__} took ({response_time}s) to process."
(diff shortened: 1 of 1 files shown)


packit_service/worker/jobs.py
```diff
@@ -818,7 +818,7 @@ def push_statuses_metrics(
             # we need more info why this has happened
             logger.debug(f"Event dict: {self.event}.")
             logger.error(
-                f"Event {self.event.__class__.__name__} took ({response_time}s) to process."
+                f"Event {self.event.__class__.__name__} took more than 15s to process."
```
mfocko, line 821: ~~Is there any way we could get the response times in?~~ Though it looks like we have `pushgateway.last_initial_status_time` below, so it should be possible to run some stats on it.

mfocko (approved):

Issue #2058, opened 2023-05-18
Status overrides for shared Copr build


lachmanfrantisek:
When the user configures a monorepo to use a shared Copr project, the build results override all the statuses.

[Example](https://github.com/fedora-copr/copr/pull/2720/commits/39c5e2c2aadee32a73e7da5058cc2072e30f41f0):

```
jobs:
  - job: copr_build
    packages:
      - python-copr
      - copr-cli
      - copr-frontend
    trigger: pull_request
    targets: fedora-37
    owner: packit
    project: copr
```

See that all statuses show the same content:

![image](https://github.com/packit/packit-service/assets/20214043/f6bda7aa-b3b4-4166-9152-78dea62eec23)
![image](https://github.com/packit/packit-service/assets/20214043/581b3481-d37e-4649-bd6e-3531c2548acf)

(Usecase with identifiers might be affected as well.)

When reacting to Copr results, we need to check (at least for the Monorepo) that the package we are working with is affected.
Probably here:
https://github.com/packit/packit-service/blob/96ef508a2a954e8dc390087eb5826e2e8b32a37b/packit_service/worker/events/copr.py#L66

PR #2097, opened 2023-06-26
Share Copr project for monorepo Copr builds

### LINKED ISSUE #2055: Allow sharing temporary Copr project by Monorepo packages
By now, when specifying multiple packages for a Copr build job, Packit creates a Copr project for each.
This might not be what people expect and might cause issues when packages are dependent.

This issue is not about resolving the dependencies. For that, we have a separate epic #1993 

Implementation options:
* A new config option to overwrite Copr project identifier.
    * Statuses will be different, Copr project name will be the same.
    * Looks like a hack from the UX perspective. This should be the default behaviour.
    * Alternatively, have a flag config option instead (to either share Copr project or ignore the identifier).
* Improve the config -> job logic to share the Copr project between the packages by default.
    * This might be tricky since the default Copr project name is resolved later.
* When resolving the default Copr project name, check the full config and realise if we are part of the same job.
    * Goes against the separation of per-package jobs and auto-identifier forwarding.

lbarcziova:
Fixes #2055 #2058

---

RELEASE NOTES BEGIN
Copr builds configured as a monorepo job will now be built in one Copr project together. The bug with overwriting the statuses for monorepo jobs was fixed as well.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/checker/copr.py
+++ b/packit_service/worker/checker/copr.py
@@ -18,6 +18,7 @@ from packit_service.worker.handlers.mixin import (
     GetCoprBuildJobHelperForIdMixin,
     GetCoprBuildJobHelperMixin,
     GetCoprSRPMBuildMixin,
+    GetCoprBuildEventMixin,
 )
 from packit_service.worker.reporting import BaseCommitStatus
 
@@ -70,6 +71,28 @@ class AreOwnerAndProjectMatchingJob(Checker, GetCoprBuildJobHelperForIdMixin):
             f"The Copr project {self.copr_event.owner}/{self.copr_event.project_name} "
             f"does not match the configuration "
             f"({self.copr_build_helper.job_owner}/{self.copr_build_helper.job_project} expected)."
+        )
+        return False
+
+
+class IsPackageMatchingJobView(Checker, GetCoprBuildEventMixin):
+    """
+    When running builds for multiple packages (in monorepo) in one job
+    config, we need to check whether the package that we are handling matches
+    the job configuration.
+    """
+
+    def pre_check(self) -> bool:
+        if (
+            not self.job_config.package
+            or self.copr_event.pkg == self.job_config.package
+        ):
+            return True
+
+        logger.debug(
+            f"The Copr build {self.copr_event.build_id} (pkg={self.copr_event.pkg}) "
+            f"does not match the package from the configuration "
+            f"({self.job_config.package})."
         )
         return False
 
--- a/packit_service/worker/handlers/copr.py
+++ b/packit_service/worker/handlers/copr.py
@@ -36,6 +36,7 @@ from packit_service.worker.checker.copr import (
     IsGitForgeProjectAndEventOk,
     BuildNotAlreadyStarted,
     IsJobConfigTriggerMatching,
+    IsPackageMatchingJobView,
 )
 from packit_service.worker.events import (
     CoprBuildEndEvent,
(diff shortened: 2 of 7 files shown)


packit_service/worker/checker/copr.py
```diff
@@ -74,6 +75,28 @@ def pre_check(self) -> bool:
...
+    """
+    When running builds for multiple packages (in monorepo) in one job
+    config, we need to check whether the package that we are handling matches
+    the job configuration.
+    """
+
+    def pre_check(self) -> bool:
+        if (
+            not self.job_config.package
+            or self.copr_event.pkg == self.job_config.package
```
lbarcziova, line 88: is this the correct way to check this @lachmanfrantisek @majamassarini ? does the configuration key need to always be equal to the package built in Copr?

I was also thinking about storing the value in DB for the builds and comparing it there

lachmanfrantisek: Don't we already save this to the database?
https://github.com/packit/packit-service/blob/716a2a3952dadabd2961756f41e633b8e61817ae/packit_service/models.py#L1401

lbarcziova: but these are stored after a successful build, and we need to compare the values when reporting the state of the build

majamassarini: > is this the correct way to check this @lachmanfrantisek @majamassarini ? does the configuration key need to always be equal to the package built in Copr?

If I am following the idea this check should be needed to solve #2058. 
But I am not really sure this implementation will solve #2058, this check is ok, but we don't use anymore the identifiers (to solve #2055) and in this way even though we have this check we will end up updating the status for the same job as many times as are the packages in the job configuration. 

But I could be wrong, I have just this fear.

lbarcziova: this PR removes the usage of the monorepo identifiers only from the default Copr project name. And for this solution to work, this check needed to be added (not only for #2058), as now the Copr project name will not differ for different packages.

So, do I understand the monorepo support code correctly, for the one job config like [this](https://github.com/packit/packit-service/issues/2055#issuecomment-1552920077), we create for each package a jobconfigview and resolve the jobs for each? If yes, that should work now with the introduced pre-check, as we will for each jobconfigview check whether the package from Copr matches the package in the jobconfigview.

majamassarini: Yes, the monorepo code will create one jobconfigview for every package. And the monorepo code customize the `identifier` value with the name of the package for which the job was built. I don't see a problem with this check but I believe that if we don't use the `identifier` we have 3 different jobs (one copr build job for every package in the list) with just one copr project repo name (and this is ok - is what we want in #2055) and just one status check (and this is not ok - we are making #2058 fail in a new way).

lbarcziova: Thanks for clarification!

> and just one status check 

what do you mean by this? Since the identifier was not removed anywhere else (only in the Copr project name), it will still be used when reporting.

majamassarini: Ok, then I am wrong. I thought that the `identifier` changes should also affect the reporting.

majamassarini (approved):

packit_service/worker/checker/copr.py
lachmanfrantisek, line 88: And this one #2058

lachmanfrantisek (approved):
:+1: 

This looks good to me to solve both Monorepo issues. Thanks!

I would maybe merge tomorrow. What do you think?

lbarcziova:
recheck

PR #2100, opened 2023-06-27
Use Talisman to set HTTP security headers

jpopelka:
See [Talisman docs](https://github.com/wntrblm/flask-talisman#talisman-http-security-headers-for-flask) for what it does by default.

Regarding the content-security-policy:
- I started with the default value, i.e. `default-src: 'self', 'object-src': 'none'`
- added the `image-src` to be able to load icons
- added the `style-src` & `script-src` because of python-restx/flask-restx#252

Related to packit/private#28

### DIFF
--- a/packit_service/service/app.py
+++ b/packit_service/service/app.py
@@ -6,13 +6,14 @@ from os import getenv
 from socket import gaierror
 
 from flask import Flask
-from flask_cors import CORS
 
 # Mypy errors out with Module 'flask' has no attribute '__version__'.
 # Python can find flask's version but mypy cannot.
 # So we use "type: ignore" to cause mypy to ignore that line.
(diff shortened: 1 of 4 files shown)


packit_service/service/app.py
```diff
@@ -78,7 +79,26 @@ def get_flask_application():
 
 
 packit_as_a_service = Proxy(get_flask_application)
-CORS(packit_as_a_service)
+
+cors = CORS(packit_as_a_service)
+
```
lachmanfrantisek, line 84: Is there a reason why we save this into a variable?

lachmanfrantisek (approved):

jpopelka:
recheck

lachmanfrantisek (approved):

PR #2101, opened 2023-06-28
Use token authentication method when getting the last tag

nforro:
This should fix [GithubAppNotInstalledError](https://red-hat-it.sentry.io/issues/4278445161/?project=1767823).

Related to #2087.

### DIFF
--- a/packit_service/worker/handlers/distgit.py
+++ b/packit_service/worker/handlers/distgit.py
@@ -190,7 +190,7 @@ class AbstractSyncReleaseHandler(
         self._project_url = self.data.project_url
         self._sync_release_run_id = sync_release_run_id
         self.helper: Optional[SyncReleaseHelper] = None
-        self.tag = self.data.tag_name
+        self._tag = self.data.tag_name
 
     @property
     def sync_release_helper(self) -> SyncReleaseHelper:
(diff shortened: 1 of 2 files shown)


lachmanfrantisek (approved):
A testcase would be nice..;)

lbarcziova (approved):
thanks for the fix!

I don't like that much that we use the same `with` block twice, what do you think about moving the `tag` to a property that would be lazily evaluated during `run()`?

nforro:
@lbarcziova Does it make sense like this?

@lachmanfrantisek I'm not really sure how to (elegantly) test this, any suggestions?

lbarcziova:
> @lbarcziova Does it make sense like this?

yes, thanks!

lachmanfrantisek (approved):
What about trying requre for the tests?..;-)

OK, I know this one is a bit hard to cover. I was thinking about updating the tests here: https://github.com/packit/packit-service/blob/main/tests/integration/test_new_hotness_update.py

But it's already so flexmocked that it might not give much value. Let's test on stage then.. :rocket:

nforro:
> But it's already so flexmocked that it might not give much value

Added one more flexmock :slightly_smiling_face: (not to `test_new_hotness_update` but to `test_pull_from_upstream_retrigger_via_dist_git_pr_comment`).

PR #2061, opened 2023-05-23
Use CentOS Stream9 as a base image

jpopelka:
[celery](https://bugzilla.redhat.com/show_bug.cgi?id=2032543) & [flexmock](https://bugzilla.redhat.com/show_bug.cgi?id=2120251) are not packaged for CentOS Stream 9 yet, but they can be easily installed from PyPI (we've actually already been using `celery` from PyPI in worker)

Related to packit/deployment#410

PR #2113, opened 2023-07-11
ci(zuul): fix reverse-dependency tests

mfocko:
Related to packit/packit#2008

### DIFF
--- a/files/install-deps.yaml
+++ b/files/install-deps.yaml
@@ -18,16 +18,11 @@
           - python3-mod_wsgi
           - mod_ssl
           - python3-alembic
-          - python3-sqlalchemy+postgresql
           - python3-prometheus_client
-          - python3-celery
           - python3-redis # celery[redis]
           - python3-lazy-object-proxy
           - python3-flask-restx
-          - python3-flask-cors
           - python3-flask-talisman
-          - python3-flexmock # because of the hack during the alembic upgrade
-          # (see d90948124e46_add_tables_for_triggers_koji_and_tests.py )
           - python-jwt
           # v6 = bodhi-client, v5 = python3-bodhi{,-client}
           - bodhi-client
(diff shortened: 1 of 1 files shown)


files/install-deps.yaml
```diff
@@ -40,6 +36,14 @@
         name:
           - sentry-sdk[flask]
           - syslog-rfc5424-formatter
+          - sqlalchemy
+          # postgresql driver that was included in the RPM package
+          - psycopg2
```
nforro, line 41: Shouldn't `python3-sqlalchemy+postgresql` above be removed?

nforro (approved):

files/install-deps.yaml
mfocko: ah, must've missed it, thanks

nforro: I think it's also possible to specify an extra to pip like this: `sqlalchemy[postgresql]`
But the result would be the same.

mfocko:
:+1: to move the openshift tests to TF :smile:

files/install-deps.yaml
jpopelka: Won't don't need to install them with pip, https://github.com/packit/packit-service/pull/2061/commits/271f3622c8e540264313fcf2c609e10814c2ab50 should be sufficient I think.

mfocko: I went through the SQLAlchemy docs just to find that they include the postgresql API by default in SQLAlchemy with the exception of the DB driver which is the psycopg2

The fact I wasn't able to tell the difference between `+something` RPM packages by looking at specfile just frustrated me even more :D

jpopelka:
I'm trying to rebase #2061 which should solve the tests as well.

files/install-deps.yaml
nforro: > The fact I wasn't able to tell the difference between `+something` RPM packages by looking at specfile just frustrated me even more :D

That's the way how to specify extras (it's all automatic based on package metadata, there doesn't have to be anything in a spec except for a `%python_extras_subpkg` macro): https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#Extras

PR #2121, opened 2023-07-19
Install eventlet as well

jpopelka:
One can switch to it by setting `POOL=eventlet` env. var.

Related to #2120


majamassarini (approved):
Oh woow, how did you realize this was the fix?

jpopelka:
> how did you realize this was the fix?

I didn't :-) It's just another concurrency library to try to see if the problem is in the one (or just the version) we're currently using.

jpopelka:
Tried on stage and from https://red-hat-it.sentry.io/issues/4325660438 it seems the `eventlet` doesn't get along well with `sqlalchemy`.

PR #2134, opened 2023-08-03
Add another non copyable object to the list

### LINKED ISSUE #2120: RuntimeError: dictionary changed size during iteration
Sentry Issue: [PCKT-002-PACKIT-SERVICE-631](https://red-hat-it.sentry.io/issues/4323414918/?referrer=github_integration)

```
RuntimeError: dictionary changed size during iteration
(10 additional frame(s) were not displayed)
...
  File "copy.py", line 230, in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
  File "copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "copy.py", line 270, in _reconstruct
    state = deepcopy(state, memo)
  File "copy.py", line 146, in deepcopy
    y = copier(x, memo)
  File "copy.py", line 229, in _deepcopy_dict
    for key, value in x.items():
```

majamassarini:
The build object belongs only to the copr events.

Should fix #2120 which is still failing.

### DIFF
--- a/packit_service/worker/events/copr.py
+++ b/packit_service/worker/events/copr.py
@@ -133,10 +133,12 @@ class AbstractCoprBuildEvent(AbstractResultEvent):
 
         return True
 
+    def get_non_serializable_attributes(self):
+        return super().get_non_serializable_attributes() + ["build"]
+
     def get_dict(self, default_dict: Optional[Dict] = None) -> dict:
         result = super().get_dict()
         result["topic"] = result["topic"].value
-        result.pop("build")
         return result
 
     def get_copr_build_url(self) -> str:
--- a/packit_service/worker/events/event.py
+++ b/packit_service/worker/events/event.py
@@ -267,12 +267,18 @@ class Event:
         """
         return {k: copy.deepcopy(v) for k, v in d.items() if k not in skip}
 
+    def get_non_serializable_attributes(self):
+        return [
+            "_db_project_event",
+            "_project",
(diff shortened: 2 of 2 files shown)


packit_service/worker/events/copr.py
```diff
@@ -133,10 +133,14 @@ def pre_check(self):
 
         return True
 
+    def get_non_serializable_attributes(self):
+        skip = super().get_non_serializable_attributes()
+        skip.append("build")
+        return skip
```
nforro, line 139: How about this?

```suggestion
        return super().get_non_serializable_attributes() + ["build"]
```

nforro (approved):

mfocko (approved):
Could you please reference the #2120 either from the PR description or the commit message?

I guessed based on the changes why are you doing this and then also noticed the branch name.

packit_service/worker/events/copr.py
mfocko: Makes a new copy, not that that it would matter that much.

majamassarini: And probably it is better to work on a new copy. So +1 for copying it. Thanks Nikola.

majamassarini:
> Could you please reference the #2120 either from the PR description or the commit message?
> 
> I guessed based on the changes why are you doing this and then also noticed the branch name.

Yes sure!

PR #2136, opened 2023-08-07
fix(tf): switch to ‹test.tmt›

mfocko:
After talking to @fernflower on Slack about the TF configuration issue, I have noticed that Testing Farm is planning to deprecate the ‹test.fmf› part of the schema in favor of ‹test.tmt›, schemas of both match 1:1 and TF Team suggests preffering the new one, if possible.

TODO:

  - [x] I should probably check whether we're not using `test.fmf` in some of the examples


lbarcziova (approved):
LGTM once this is confirmed by TF team, thanks!

PR #2137, opened 2023-08-07
Do not require a GitProject instance in pull-from-upstream

### LINKED ISSUE #1907: Be able to run `pull-from-upstream` without API calls for upstream repo
Check the operations that require API calls in the current implementation. Don't instantiate the ogr's `GitProject` for the upstream repository and implement the missing functionality using pure git (this may result in diversifying from the original propose-downstream implementation).

Followup of #1879.

nforro:
Fixes #1907.

Merge after https://github.com/packit/packit/pull/2032.

Notes for reviewers:

Not sure if setting namespace and repo name to `None` won't break anything in dashboard for example. Would it be better to try to somehow extract the values from `project_url`?

https://github.com/packit/packit-service/blob/313cb23407e18b54bedb61dcf28b9c669ad9f8a2/packit_service/worker/events/event.py#L203

RELEASE NOTES BEGIN
pull-from-upstream can now be used with upstream repos that are not hosted on a supported git forge.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/events/event.py
+++ b/packit_service/worker/events/event.py
@@ -10,6 +10,7 @@ from logging import getLogger
 from typing import Dict, Optional, Type, Union, Set, List
 
 from ogr.abstract import GitProject
+from ogr.parsing import RepoUrl
 
 from packit.config import JobConfigTriggerType, PackageConfig
 from packit_service.config import PackageConfigGetter, ServiceConfig
@@ -189,7 +190,6 @@ class EventData:
             "ReleaseEvent",
             "ReleaseGitlabEvent",
             "CheckRerunReleaseEvent",
-            "NewHotnessUpdateEvent",
         }:
             (
                 self._db_project_object,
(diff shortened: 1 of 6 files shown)


packit_service/worker/events/event.py
```diff
@@ -194,6 +193,21 @@ def db_project_event(self) -> Optional[AbstractProjectEventDbType]:
...
                     commit_hash=self.commit_sha,
                 )
+            elif self.event_type in {
+                "NewHotnessUpdateEvent",
+            }:
+                if self.project:
+                    namespace = self.project.namespace
+                    repo_name = self.project.repo
+                else:
+                    namespace = repo_name = None
```
lbarcziova, line 203: > Not sure if setting namespace and repo name to None won't break anything in dashboard for example. Would it be better to try to somehow extract the values from project_url?

I think we could utilise `RepoUrl` that is already used in `NewHotnessEvent` implementation to have those values, WDYT?

lbarcziova (commented):
quite straightforward! :+1:

what about accessing `.git_project` in the implementation of release syncing, e.g. [here](https://github.com/packit/packit/blob/6114a82fbd3aca725d58689124c6c4bf13d7e99a/packit/utils/changelog_helper.py#L100), do you plan to include handling those here?

packit_service/worker/events/event.py
nforro: Thanks, done.

nforro:
> what about accessing `.git_project` in the implementation of release syncing, e.g. [here](https://github.com/packit/packit/blob/6114a82fbd3aca725d58689124c6c4bf13d7e99a/packit/utils/changelog_helper.py#L100), do you plan to include handling those here?

I've opened https://github.com/packit/packit/pull/2034 and added: https://github.com/packit/packit-service/blob/10bc544326c3f6c06c52b9a154d55d51043d6f24/packit_service/worker/handlers/distgit.py#L525

Not the most elegant solution, but it should work.

lbarcziova (approved):
nice!

PR #2144, opened 2023-08-11
Fix getting test targets with build target overrides

### LINKED ISSUE #2119: Incorrect reporting for tests when retriggering a build of different target that is not configured for tests
With a configuration like this:
```yaml
    -  job: copr_build
      trigger: pull_request
      targets:
          - fedora-rawhide-x86_64
          - epel-8-x86_64

    - job: tests
      trigger: pull_request
      targets:
        - epel-8-x86_64
```

when one retriggers the `fedora-rawhide-x86_64` build, the status of the SRPM build will be reported for the test job as well, resulting in this case in a `testing-farm:centos-stream-8-x86_64` status pending indefinitely with `SRPM build succeeded. Waiting for RPM build to start...` message. 

This happened in https://github.com/openscanhub/openscanhub/pull/44/.

lbarcziova:
The targets_override can be empty if there is not matching configured test target, therefore check the *_override respectively

Fixes #2119

---

RELEASE NOTES BEGIN
We have fixed the bug resulting in incorrect reporting for tests when retriggering a build of a different target that was not configured for tests.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/helpers/build/build_helper.py
+++ b/packit_service/worker/helpers/build/build_helper.py
@@ -331,7 +331,7 @@ class BaseBuildJobHelper(BaseJobHelper):
 
         return (
             configured_targets & targets_override
-            if targets_override
+            if self.build_targets_override or self.tests_targets_override
             else configured_targets
         )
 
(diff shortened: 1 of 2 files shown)


lbarcziova:
recheck

lachmanfrantisek (approved):

PR #2149, opened 2023-08-17
Move upstream tag check to sync release task

lbarcziova:
It has to be run in the task directly, and not as a pre-check, since pre-check is part of `process_message` task that doesn't run in the long-running worker. Long-running worker is needed here as the Local project is being accessed.

---

RELEASE NOTES BEGIN
N/A
RELEASE NOTES END


lbarcziova:
@nforro actually, I opened #2150 which is probably a better fix than this, PTAL :pray:

PR #2150, opened 2023-08-17
Adjust pre-check for matching tag in sync release

lbarcziova:
If the tag in the event is None (pull-from-upstream triggered by comment), don't check the match with `upstream_tag_include` and `upstream_tag_exclude` as the tags will be filtered out considering these values when obtaining the tag directly in the handler task.

---

RELEASE NOTES BEGIN
N/A
RELEASE NOTES END


nforro (approved):
Makes sense :+1:

PR #2154, opened 2023-08-21
Adjust Copr build epoch handling for None

lbarcziova:
Adjust how we include the epoch in the Testing Farm payload. We get the value from Copr API and the behaviour there recently changed: https://github.com/fedora-copr/copr/pull/2876

---

RELEASE NOTES BEGIN
N/A
RELEASE NOTES END

### DIFF
--- a/packit_service/utils.py
+++ b/packit_service/utils.py
@@ -70,7 +70,8 @@ def get_package_nvrs(built_packages: List[dict]) -> List[str]:
         if package["arch"] == "src":
             continue
 
-        epoch = f"{package['epoch']}:" if package["epoch"] != 0 else ""
+        epoch = f"{package['epoch']}:" if package["epoch"] else ""
+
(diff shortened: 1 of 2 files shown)


packit_service/utils.py
```diff
@@ -70,7 +70,12 @@ def get_package_nvrs(built_packages: List[dict]) -> List[str]:
         if package["arch"] == "src":
             continue
 
-        epoch = f"{package['epoch']}:" if package["epoch"] != 0 else ""
+        epoch = (
+            f"{package['epoch']}:"
+            if package["epoch"] is not None and package["epoch"] != 0
```
mfocko, line 75:
```suggestion
            if package["epoch"]
```

But I would probably add a comment…

mfocko (approved):

lbarcziova:
recheck

PR #2156, opened 2023-08-22
fix(tf): correctly handle tests without labels

### LINKED ISSUE #2155: Fails with labels in TF comment trigger, but without labels in config
Sentry Issue: [PCKT-002-PACKIT-SERVICE-69F](https://red-hat-it.sentry.io/issues/4397832238/?referrer=github_integration)

```
TypeError: 'NoneType' object is not iterable
(4 additional frame(s) were not displayed)
...
  File "packit_service/worker/jobs.py", line 460, in process_jobs
    processing_results.extend(self.create_tasks(job_configs, handler_kls))
  File "packit_service/worker/jobs.py", line 479, in create_tasks
    if self.should_task_be_created_for_job_config_and_handler(
  File "packit_service/worker/jobs.py", line 524, in should_task_be_created_for_job_config_and_handler
    if not handler_kls.pre_check(
  File "packit_service/worker/handlers/abstract.py", line 264, in pre_check
    checks_pass = checks_pass and checker.pre_check()
  File "packit_service/worker/checker/testing_farm.py", line 161, in pre_check
    if not self.testing_farm_job_helper.comment_arguments.labels or any(
```

Job config doesn't have any labels set, but we only check if they were used in the comment that's triggering the TF run which results in iterating over `None` (since there are none in the config).

We should probably provide some reasonable status to the user instead of failing “miserably”.

mfocko:
When retriggering via comment and specifying the labels for the tests, the checker fails because of the missing labels in the job config.

TODO:

Fixes #2155

Related to

Merge before/after

---

RELEASE NOTES BEGIN

RELEASE NOTES END


lachmanfrantisek (approved):
Would it be possible (?as a followup) to provide user feedback when label is specified in a comment and does not exists in a config?

PR #2157, opened 2023-08-23
Fix/packit service/2122

### LINKED ISSUE #2122: Basic test case: monorepos, SRPM builds pending
All SRPM builds submitted from packit/hello-world#1557 are marked as `pending` in [(staging) dashboard](https://dashboard.stg.packit.dev/jobs/srpm-builds) even they succeeded.

Screenshot from 2023-07-19 12-34-37

majamassarini:
Fixes #2122

---

RELEASE NOTES BEGIN
SRPM builds statuses, for monorepos projects, are now updated.
RELEASE NOTES END


packit_service/models.py
```diff
@@ -365,6 +365,9 @@ class ProjectAndEventsConnector:
     def get_project_event_model(self) -> Optional["ProjectEventModel"]:
         return self.runs[0].project_event if self.runs else None
 
+    def get_package_name(self) -> Optional[str]:
+        return self.runs[0].package_name
```
lbarcziova, line 369: I would just change this as the previous method is to be safe
```suggestion
        return self.runs[0].package_name if self.runs else None
```

packit_service/worker/handlers/abstract.py
```diff
@@ -294,6 +294,15 @@ def __init__(
         self._project: Optional[GitProject] = None
         self._clean_workplace()
 
+    def get_package_name(self) -> Optional[str]:
```
lbarcziova, line 297: I'm not sure I follow here, what is this method needed for as for monorepos it returns None?

packit_service/worker/helpers/job_helper.py
```diff
@@ -60,6 +59,12 @@ def __init__(
         self._is_reporting_allowed: Optional[bool] = None
         self._is_gitlab_instance: Optional[bool] = None
 
+    def get_package_name(self):
```
lbarcziova, line 62: would it be too complicated to deduplicate this functionality?

lbarcziova (commented):
nice work! 

just a few comments

packit_service/worker/helpers/job_helper.py
majamassarini: I can create a mixin, shared between helpers and handlers but I was wondering if this isn't too much for just a simple method. At the moment we have no mixins between helpers and handlers... Do we want to create the first one?

packit_service/worker/handlers/abstract.py
lbarcziova: ah okay, I get it now, thanks! Could you maybe add a comment about this to the docstring?

lbarcziova (approved):
thanks!

PR #2159, opened 2023-08-24
Get the build id from the event.

majamassarini:
The SRPMBuildModel, unlike the CoprBuildModel,
 has not a build id associated.

Take the build id  from the event even though the package name is taken from the db.


lbarcziova (approved):
lgtm once the tests are green

PR #2161, opened 2023-08-28
Use project event model ID in reporting external ID

### LINKED ISSUE #2160: Broken retriggering via checks UI
After retriggering a check via Github checks `Re-run` button, nothing was rerun. 

Logs:

```

2023-08-28T08:11:31.439+0000 | Github check run rpm-build:fedora-rawhide-x86_64 rerun event.
2023-08-28T08:11:31.445+0000 | Job project event with ID 72308 not found.

```
We use the external ID of the check run to store the ID of the job project event from the DB so that we can easily use it when retriggering, but it seems that we store ID of some other object?

lbarcziova:
When creating a Github check run, use the db ID of project event model and not project object model so that it can be correctly parsed backed when rerunning.

Fixes #2160

---

RELEASE NOTES BEGIN
We have fixed a bug causing broken retriggering of Github checks.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/helpers/job_helper.py
+++ b/packit_service/worker/helpers/job_helper.py
@@ -159,8 +159,8 @@ class BaseJobHelper:
                 project=self.project,
                 commit_sha=self.metadata.commit_sha,
                 packit_user=self.service_config.get_github_account_name(),
-                project_object_id=self._db_project_object.id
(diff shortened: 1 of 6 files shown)


majamassarini (approved):
Thanks for the fix!

PR #2140, opened 2023-08-10
Use packages_config from default downstream branch

majamassarini:
When retriggering pull-from-upstream

Fixes #2115

---

RELEASE NOTES BEGIN
Retriggering of `pull-from-upstream` via a comment will now use the correct configuration file from the default dist-git branch.
RELEASE NOTES END

PR #2162, opened 2023-08-28
Allow to retrigger pull-from-upstream with config from a PR

nforro:
It makes sense for `/packit pull-from-upstream` to retrigger the job with config taken from the main dist-git branch, but sometimes it could be useful to test config changes in an open dist-git PR, before merging it.

This commit allows that using `/packit pull-from-upstream --with-pr-config`.

Related to #2140.


majamassarini (approved):
I think it could be really useful and its implementation is really straightforward! Nice.

PR #2166, opened 2023-08-29
Move local-project dir inside sandcastle subfolder

### LINKED ISSUE #1956: Be able to access both upstream and downstream repos in actions
If I want to do some custom changes during `pull-from-upstream` job, I want to do them in dist-git (I can't/don't want to touch upstream, that's why I'm using `pull-from-upstream`). But that's not really possible, because all actions have upstream git repo as working directory.

AI: 
- [ ] provide env variables with paths of the upstream and dist-git repo (e.g. `PACKIT_UPSTREAM_REPO`, `PACKIT_DISTGIT_REPO`)
   - people can then define these in their actions: (`cd $UPSTREAM; cd $DOWNSTREAM`)
- [ ] make sure both directories are synced to sandcastle pod and changes are propagated back (sync both directories)

majamassarini:
We need both upstream and dowstream repos in sandcastle shared folder.
For this reason I need to move the upstream repo from the root to a subfolder.

Fixes #1956 

Merge after packit/packit#2054

---

RELEASE NOTES BEGIN
Packit now sets `PACKIT_UPSTREAM_REPO` and `PACKIT_DOWNSTREAM_REPO` environment variables for release syncing actions. The variables represent paths where the respective git repositories are cloned
RELEASE NOTES END

### DIFF
--- a/packit_service/constants.py
+++ b/packit_service/constants.py
@@ -256,3 +256,9 @@ DENIED_MSG = (
     f"You were denied from using Packit Service. If you think this was done by mistake, "
     f"please, [let us know]({CONTACTS_URL})."
 )
+
+# We want to be able to access both
+# upstream and downstream repos through the
+# shared sandcastle dir
+SANDCASTLE_DG_REPO_DIR = "dist-git"
+SANDCASTLE_LOCAL_PROJECT_DIR = "local-project"
--- a/packit_service/worker/handlers/distgit.py
+++ b/packit_service/worker/handlers/distgit.py
@@ -399,8 +399,8 @@ class AbstractSyncReleaseHandler(
             )
         finally:
             # remove temporary dist-git clone after we're done here - context:
-            # 1. the dist-git repo is cloned on worker, not sandbox
-            # 2. it's stored in /tmp, not in the mirrored sandbox PV
+            # 1. the dist-git repo could be cloned on worker, not sandbox
+            # 2. in such case it's stored in /tmp, not in the mirrored sandbox PV
             # 3. it's not being cleaned up and it wastes pod's filesystem space
             shutil.rmtree(self.packit_api.dg.local_project.working_dir)
 
--- a/packit_service/worker/mixin.py
+++ b/packit_service/worker/mixin.py
@@ -4,6 +4,7 @@
 from abc import abstractmethod
 import logging
 import re
+from pathlib import Path
 from typing import Optional, Protocol, Union, List
 
 from fasjson_client import Client
@@ -27,6 +28,8 @@ from packit_service.worker.helpers.job_helper import BaseJobHelper
 
 from packit_service.constants import (
     FASJSON_URL,
+    SANDCASTLE_LOCAL_PROJECT_DIR,
+    SANDCASTLE_DG_REPO_DIR,
 )
 
 logger = logging.getLogger(__name__)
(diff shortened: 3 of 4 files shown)


packit_service/constants.py
```diff
@@ -256,3 +256,9 @@ def from_number(number: int):
     f"You were denied from using Packit Service. If you think this was done by mistake, "
     f"please, [let us know]({CONTACTS_URL})."
 )
+
+# We want to be able to access both
+# upstream and downstream repos trough the
+# shared sandcastle dir
+SANDCASTLE_DG_REPO_DIR = "dg"
+SANDCASTLE_LOCAL_PROJECT_DIR = "local-project"
```
nforro, line 264: I know the names don't matter much, but if local project working dir is in fact the upstream repo, why not use `upstream` and `downstream`/`dist-git`?

packit_service/worker/handlers/distgit.py
```diff
@@ -399,8 +399,8 @@ def run(self) -> TaskResults:
             )
         finally:
             # remove temporary dist-git clone after we're done here - context:
-            # 1. the dist-git repo is cloned on worker, not sandbox
-            # 2. it's stored in /tmp, not in the mirrored sandbox PV
+            # 1. the dist-git repo is cloned on worker, not always on sandbox
+            # 2. it's sometime stored in /tmp, not in the mirrored sandbox PV
```
nforro, line 403:
```suggestion
            # 1. the dist-git repo could be cloned on worker, not sandbox
            # 2. in such case it's stored in /tmp, not in the mirrored sandbox PV
```

packit_service/worker/mixin.py
```diff
@@ -191,6 +194,9 @@ def packit_api(self):
                 self.service_config,
                 self.job_config,
                 upstream_local_project=self.local_project,
+                dist_git_clone_path=Path(
+                    self.service_config.command_handler_work_dir
+                ).joinpath(SANDCASTLE_DG_REPO_DIR),
```
nforro, line 199: Not sure if this is nicer since it doesn't fit in a single line, what do you think?

```suggestion
                dist_git_clone_path=Path(self.service_config.command_handler_work_dir)
                / SANDCASTLE_DG_REPO_DIR,
```

nforro (approved):
LTGM, but I haven't tested it.

nforro:
I'd change release notes to something like this:

Packit now sets `PACKIT_UPSTREAM_REPO` and `PACKIT_DOWNSTREAM_REPO` environment variables for release syncing actions. The variables represent paths where the respective git repositories are cloned.

packit_service/constants.py
majamassarini: The `LocalProjectMixin` sets the suffix for this dir. The local project can be both an upstream or downstream project, it depends on the others mixins in the handler...
I can be sure only about the `dg` suffix, because it is used by the `PackitAPIWithUpstreamMixin` which is setting it.
The `PackitAPIWithUpstreamMixin` "depends" on the `LocalProjectMixin` so it is too late to specify the suffix for the upstream in  `PackitAPIWithUpstreamMixin`.
And I have no other ideas...

I choose `dg` as a name because in the API is called like that, but I can change it in `dist-git` if you think it is better.

majamassarini:
> LGTM, but I haven't tested it.

Me neither, because I can not test locally actions in sandcastle, since I have not openshift.
I can just see that the working dir for `dg` is under the shared sandcastle dir and I hope this will be enough.

packit_service/constants.py
nforro: Yes, I think `dist-git` is better, `dg` makes sense in the code as it is shorter, but I think we can afford to be a bit more explicit in directory names :slightly_smiling_face:

lbarcziova (approved):
thanks! let's test in staging

PR #2168, opened 2023-08-30
Check TF composes using regex matching

lbarcziova:
TF started additionally exposing regexes on top of the exact compose names in the /composes/ endpoints, these changes adjust our code to that.

---

RELEASE NOTES BEGIN
Testing Farm started additionally exposing regexes on top of the exact compose names in the /composes/ endpoints, and we now support this as well when checking the validity of compose.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/helpers/testing_farm.py
+++ b/packit_service/worker/helpers/testing_farm.py
@@ -3,6 +3,7 @@
 
 import logging
 import re
+from re import Pattern
 from typing import Dict, Any, Optional, Set, List, Union, Tuple, Callable
 
 import requests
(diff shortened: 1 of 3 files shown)


packit_service/worker/helpers/testing_farm.py
```diff
@@ -614,6 +614,14 @@ def is_fmf_configured(self) -> bool:
         except FileNotFoundError:
             return False
 
+    @staticmethod
+    def is_compose_matching(compose_to_check, composes) -> bool:
+        """
+        Check whether the compose matches any compose in the list of re-compiled
+        composes.
+        """
+        return any(compose.fullmatch(compose_to_check) for compose in composes)
```
lachmanfrantisek, line 623: Could you please add the type annotations? I think it might be handy to get the info about types right there.

lachmanfrantisek (approved):
Very nice!

PR #2173, opened 2023-09-04
chore: install certifi from PyPI

mfocko:


files/install-deps-worker.yaml
```diff
@@ -54,6 +53,9 @@
           - sentry-sdk
           - syslog-rfc5424-formatter # Logging to Splunk
           - celery==5.3.* # RHBZ#2032543
+          # Mozilla's certificates have been discontinued on C9S, required for Sandcastle
+          # newer version is required because of CVE-2023-37920
+          - certifi==2023.7.22
```
jpopelka, line 58: I don't think we need this, I'd let sandcastle transitively (as a dependency of kubernetes which sandcastle requires) pull it in itself.
The above `python3-certifi` was added in 61b730fb but I don't remember why I actually added other libs than the `(o)auth(lib)` which was causing some problems.
So I'm fine with just removing the `python3-certifi` rpm.

mfocko: I moved the `stable` already, so I'm for commenting it out and testing if we really need it on stage.

jpopelka: > ... testing if we really need it on stage

it'll be installed automatically as a `kubernetes` (package) dependency

> I'm for commenting it out ...

I'm sorry but I'm against commenting it out, I want it to be removed :)
If it stays (commented out) it'll make me think why is it there, and the comment will not help.

jpopelka (approved):

PR #2182, opened 2023-09-12
Implement commenting on failure based on config

### LINKED ISSUE #1911: RFE: Notify on failure + tag github user(s) along with custom failure message
As a packit-service user responsible for monitoring the packit jobs on the containers' team repos, I'd like to be notified whenever there's a packit failure on any PR, preferably via tagging my github id and a customizable message.

The current packit configuration docs only mention notification on build success, nothing for failure.

/cc @evgeni

lbarcziova:
If there is the failure_comment_message defined in the config, notify users via comment on failure using the configured message. Comment only once per commit SHA and job (utilising check of previous identical Packit comment).

Fixes #1911

TODO:

- [x] Update or write new documentation in `packit/packit.dev`. (packit/packit.dev#735)

---

RELEASE NOTES BEGIN
We have introduced a new configuration option `notifications.failure_comment.message` that enables notifying users on failure via a comment using the configured message.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/helpers/build/build_helper.py
+++ b/packit_service/worker/helpers/build/build_helper.py
@@ -31,7 +31,7 @@ from packit_service.trigger_mapping import are_job_types_same
 from packit_service.worker.events import EventData
 from packit_service.worker.helpers.job_helper import BaseJobHelper
 from packit_service.worker.monitoring import Pushgateway
-from packit_service.worker.reporting import BaseCommitStatus
+from packit_service.worker.reporting import BaseCommitStatus, DuplicateCheckMode
 from packit_service.worker.result import TaskResults
 
 logger = logging.getLogger(__name__)
@@ -753,4 +753,22 @@ class BaseBuildJobHelper(BaseJobHelper):
             markdown_content=markdown_content,
             links_to_external_services=links_to_external_services,
             update_feedback_time=update_feedback_time,
+        )
+
+    def notify_about_failure_if_configured(self):
+        """
+        If there is a failure_comment_message configured for the job,
+        post a comment and include the configured message. Do not post
+        the comment if the last comment from the Packit user is identical.
+        """
+        if not (
+            configured_message := self.job_config.notifications.failure_comment.message
+        ):
+            return
+
+        formatted_message = configured_message.format(
+            commit_sha=self.db_project_event.commit_sha
+        )
+        self.status_reporter.comment(
+            formatted_message, duplicate_check=DuplicateCheckMode.check_last_comment
         )
--- a/packit_service/worker/handlers/copr.py
+++ b/packit_service/worker/handlers/copr.py
@@ -311,6 +311,7 @@ class CoprBuildEndHandler(AbstractCoprBuildReportHandler):
                 chroot=self.copr_event.chroot,
             )
             self.measure_time_after_reporting()
+            self.copr_build_helper.notify_about_failure_if_configured()
             self.build.set_status(BuildStatus.failure)
             return TaskResults(success=False, details={"msg": failed_msg})
 
@@ -370,6 +371,7 @@ class CoprBuildEndHandler(AbstractCoprBuildReportHandler):
                 description=failed_msg,
                 url=url,
             )
+            self.copr_build_helper.notify_about_failure_if_configured()
             self.build.set_status(BuildStatus.failure)
             self.copr_build_helper.monitor_not_submitted_copr_builds(
                 len(self.copr_build_helper.build_targets), "srpm_failure"
--- a/packit_service/worker/handlers/koji.py
+++ b/packit_service/worker/handlers/koji.py
@@ -225,6 +225,8 @@ class KojiTaskReportHandler(
                 url=url,
                 chroot=build.target,
             )
+            if self.koji_task_event.state == KojiTaskState.failed:
+                build_job_helper.notify_about_failure_if_configured()
 
         koji_build_logs = KojiTaskEvent.get_koji_build_logs_url(
             rpm_build_task_id=int(build.build_id),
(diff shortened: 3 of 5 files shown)


packit_service/worker/helpers/build/build_helper.py
```diff
@@ -754,3 +755,22 @@ def report_status_to_configured_job(
...
+        the comment if the last comment from the Packit user is identical.
+        """
+        if (
+            configured_message := self.job_config.notifications.failure_comment_message
+        ) is None:
+            return
+
+        full_message = (
+            f":exclamation: **One of the {self.comment_job_name} failed for commit "
+            f"{self.db_project_event.commit_sha}**\n\n---\n\n{configured_message}"
```
lbarcziova, line 772: I was thinking about whether to rather provide these variables so that users can utilise them in the configuration themselves, WDYT?

lachmanfrantisek: I think it would be easier to use/tweak if we users could define the full message. We can provide a good copy-pastable example (e.g. with the content you've just provided) so that every single user does not need to invent that.

packit_service/worker/helpers/build/build_helper.py
```diff
@@ -754,3 +755,22 @@ def report_status_to_configured_job(
...
+        )
+        self.status_reporter.comment(
+            full_message, duplicate_check=DuplicateCheckMode.check_last_comment
```
lachmanfrantisek, line 775: Thinking about that now, I am slightly worried people would like to define the strategy themselves => not sure if it's needed but we need to be prepared with the config structure to support this if needed. (Yes, I know the config PR is already merged..;(

lbarcziova: actually, I was thinking about it too and there is no problem with changing the schema since it is not added to the docs or anywhere yet :) ..I was thinking about changing the schema like:
```yaml
notifications:
  failure_comment:
    message:
    duplicate_check:
```

on the other hand, I am not sure whether we are not overcomplicating this...would we then allow one to one mapping to the [DuplicateCheckMode](https://github.com/packit/packit-service/blob/39bc4f08f13450a6b7abfb537ec4e9d367e8dc57/packit_service/worker/reporting.py#L68)?

packit_service/worker/helpers/build/build_helper.py
lbarcziova: ok, sounds good to me

packit_service/worker/helpers/build/build_helper.py
lachmanfrantisek: > but we can definitely at least change the schema for message to notifications.failure_comment.message so that the configuration for failure_comment can be easily extended if needed

+1

> would we then allow one to one mapping to the [DuplicateCheckMode](https://github.com/packit/packit-service/blob/39bc4f08f13450a6b7abfb537ec4e9d367e8dc57/packit_service/worker/reporting.py#L68)?

This makes sense, but I would first wait for the users to request this.

lbarcziova: ok :+1:

opened https://github.com/packit/packit/pull/2075

packit_service/worker/helpers/build/build_helper.py
```diff
@@ -754,3 +754,21 @@ def report_status_to_configured_job(
...
+        post a comment and include the configured message. Do not post
+        the comment if the last comment from the Packit user is identical.
+        """
+        if (
+            configured_message := self.job_config.notifications.failure_comment_message
+        ) is None:
+            return
+
+        formatted_message = configured_message.format(
+            commit_sha=self.db_project_event.commit_sha
```
lachmanfrantisek, line 770: I quite like the job_name variable, or, maybe some links might be requested. But we can freely stay with this and add on request later on.

lachmanfrantisek (approved):
Very nice!

packit_service/worker/helpers/build/build_helper.py
lbarcziova: I removed it since users can define their custom messages per-job which can replace this, but I am not strongly against

maybe @martinpitt could help us with the feedback here:  would you be interested in Packit providing some other variables besides `commit_sha`  (see [the docs in preparation](https://github.com/packit/packit.dev/blob/6cc46346bbce77ad013594631422f2b57a779fee/docs/configuration/index.md#failure_comment)) that you could utilise in the configuration of the custom failure messages? :)

martinpitt: The big enabler here is to get a configurable list of people to ping, all the details can then be examined on the PR page. 

Personally I liked the first version (i.e. just the first commit) a bit better: it's (mildly) useful to have the job name in the message, and it is IMHO much better to *not* interpolate `configured_message` -- adding variables to the text sounds like a recipe for bugs.

But it's not a strong opinion -- I'd be happy either way (and really looking forward to this!)

lbarcziova: thanks for the feedback! I can see that as @lachmanfrantisek wrote before, some users would like to have this completely customizable and by providing the commit_sha, actually, the duplication of the comment in the PR can be configured as well: when {commit_sha} is not present in the configured message, there will be only one Packit comment per PR and when it will be present, Packit will place a new comment for each failure for different commit

I would probably test it in staging in the current form and we can still easily tweak the content of the comment if needed

martinpitt:
Amazing, thank you! Can't wait to trying it out next week! :100:

lbarcziova:
Tested [here](https://github.com/lbarcziova/hello-world/pull/8#issuecomment-1717098474).

lsm5:
@lbarcziova This is awesome. podman team will be relieved to have this so they won't have to manually tag me going forward :D

PR #2193, opened 2023-09-22
Pass bugs in pull-from-upstream

lbarcziova:
For NewHotnessEvent, take the bug ID from the message directly. For retriggering take the bugs from comment and use them directly.

Fixes packit/packit#1920
Merge after packit/packit#2094

TODO:

- [x] Update or write new documentation in `packit/packit.dev`. (packit/packit.dev#748)

---

RELEASE NOTES BEGIN
Packit now automatically adds bugzilla created by Upstream Release Monitoring as `- Resolves {bugzilla} ` to the changelog (or commit if autochangelog is used) and provides the value in `commit-message` and `changelog-entry` actions as `PACKIT_RESOLVED_BUGS` env variable. When retriggering the `pull-from-upstream` from comment, one can also specify the bug(s) as `/packit pull-from-upstream --resolved-bugs rhbz#123,rhbz#125` and Packit will do the same.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/handlers/distgit.py
+++ b/packit_service/worker/handlers/distgit.py
@@ -36,7 +36,11 @@ from packit_service.service.urls import (
     get_propose_downstream_info_url,
     get_pull_from_upstream_info_url,
 )
-from packit_service.utils import gather_packit_logs_to_buffer, collect_packit_logs
+from packit_service.utils import (
+    gather_packit_logs_to_buffer,
+    collect_packit_logs,
+    get_packit_commands_from_comment,
+)
 from packit_service.worker.checker.abstract import Checker
 from packit_service.worker.checker.distgit import (
     IsProjectOk,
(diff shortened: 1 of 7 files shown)


lbarcziova:
recheck

packit_service/worker/handlers/distgit.py
```diff
@@ -531,6 +543,32 @@ def __init__(
...
+        args = commands[1:] if len(commands) > 1 else ""
+        bugs_keyword = "--resolved-bugs"
+        if bugs_keyword not in args:
+            return []
+
+        bugs = (
+            args[args.index(bugs_keyword) + 1]
+            if args.index(bugs_keyword) < len(args) - 1
+            else None
+        )
```
lachmanfrantisek, line 569: No need to do it now, but we should implement a generic way to work with arguments (maybe as part of #2200).

Maybe, we can really introduce argparse, or some other library as @mfocko has suggested..;)

lachmanfrantisek (approved):
Very nice! This is a huge improvement from the user perspective!

Issue #2216, opened 2023-10-06
Declares running tests as failed


martinpitt:
### What happened? What is the problem?

In https://github.com/cockpit-project/cockpit/pull/19448 and other PRs today I've seen packit tests marked as failure:

![image](https://github.com/packit/packit-service/assets/200109/5161bcd1-df2e-4d50-9c9e-f44d23b6bdb8)

while they are still running. At the time of writing, https://artifacts.dev.testing-farm.io/692a5d85-93a3-4b57-82e9-291a3211d32e/ is running happily, with two plans passed, and the third one still in progress.

### What did you expect to happen?

Yellow "in progress" status.

### Example URL(s)

https://github.com/cockpit-project/cockpit/pull/19448

### What is the impacted category (job)?

Testing Farm tests

### Workaround

- [X] There is an existing workaround that can be used until this issue is fixed.

### Participation

- [ ] I am willing to submit a pull request for this issue. (Packit team is happy to help!)

martinpitt:
And now https://artifacts.dev.testing-farm.io/692a5d85-93a3-4b57-82e9-291a3211d32e/ actually finished, but the PR still shows it as failed. Something similar happened this morning in https://github.com/containers/podman/pull/20286 , where the test was notified for failure, but it actually passed.

martinpitt:
https://github.com/cockpit-project/cockpit/pull/19401 is another case -- https://artifacts.dev.testing-farm.io/c0e0f78d-94f6-4d1f-9a28-04b1f143d7f0/ is currently running:

```
+-----------------------------------------------------+---------+---------+-----------+------------------------------------+--------------------------------------+----------+
| SE                                                  | Stage   | State   | Result    | Environment                        | Guest                                | Runner   |
|-----------------------------------------------------+---------+---------+-----------+------------------------------------+--------------------------------------+----------|
| CentOS-Stream-9-aarch64:aarch64:/plans/all/basic    | RUNNING | OK      | UNDEFINED | aarch64 CentOS-Stream-9-aarch64 S- | aarch64 CentOS-Stream-9-aarch64 S-   | tmt      |
|                                                     |         |         |           |                                    | bb9e7e08-1220-4390-b21c-cf241c7aa7e1 |          |
[...]

PR #2217, opened 2023-10-06
Testing Farm status reporting improvements

lbarcziova:
Related to #2216
See the commit messages.

---

RELEASE NOTES BEGIN
N/A
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/parser.py
+++ b/packit_service/worker/parser.py
@@ -1133,9 +1133,17 @@ class Parser:
             tuple: project_url, ref, result, summary, copr_build_id,
                 copr_chroot, compose, log_url, identifier
         """
-        result: TestingFarmResult = TestingFarmResult(
-            nested_get(event, "result", "overall") or event.get("state") or "unknown"
-        )
+        tf_state = event.get("state")
+        tf_result = nested_get(event, "result", "overall")
+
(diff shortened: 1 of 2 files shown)


packit_service/worker/parser.py
```diff
@@ -1133,9 +1133,19 @@ def parse_data_from_testing_farm(
...
+
+        logger.debug(f"TF payload: state = {tf_state}, result['overall'] = {tf_result}")
+
+        # error and complete are the end states
+        if tf_state not in ("complete", "error"):
+            result = TestingFarmResult(tf_state)
+        else:
+            result = TestingFarmResult(tf_result or tf_state)
+
+        result = result or TestingFarmResult.unknown
```
mfocko, line 1147: `TestingFarmResult.unknown` is unreachable. You either use `tf_state` or `tf_result or tf_state`, constructor will return an object and here the _unknown_ has therefore no effect.

---

Unless you can get `None` out of the constructor which, tbf, wouldn't even surprise me in this language :D

mfocko, line 1147:
```suggestion
            result = TestingFarmResult(tf_result or tf_state or "unknown")
```

I guess you wanted this

lbarcziova: I wanted to use it as default even in the `if` branch

mfocko (approved):

PR #2224, opened 2023-10-11
Do not suffix aarch64 composes for TF

### LINKED ISSUE #2028: aarch64 Testing Farm composes does not need to be suffixed
The `aarch64` composes used to be named differently because of what we have to behave specifically for those:

TBD

This is no longer true and we can use the same compose names for `aarch64` as well. Both names are possible for now but will be deprecated.

TODO:
* [ ] Remove the specific handling for `aarch64`.
* [ ] Enhance tests to cover that change.
* [ ] Let @thrix know we're done

Here is an example provided by @thrix :

```
❯ testing-farm-public request --git-url https://gitlab.com/testing-farm/tests --plan /testing-farm/sanity --compose CentOS-Stream-9 --arch aarch64
📦 repository https://gitlab.com/testing-farm/tests ref main test-type fmf
💻 CentOS-Stream-9 on aarch64 
🔎 api https://api.dev.testing-farm.io/v0.1/requests/a3dc8251-7a2d-40b6-984d-289c9a6ef79f
💡 waiting for request to finish, use ctrl+c to skip
👷 request is queued
🚀 request is running
🚢 artifacts https://artifacts.dev.testing-farm.io/a3dc8251-7a2d-40b6-984d-289c9a6ef79f
✅ tests passed
```

lbarcziova:
This is no longer required from TF.
Fixes #2028

---

RELEASE NOTES BEGIN
N/A
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/helpers/testing_farm.py
+++ b/packit_service/worker/helpers/testing_farm.py
@@ -684,10 +684,6 @@ class TestingFarmJobHelper(CoprBuildJobHelper):
         if compose == "CentOS-Stream":
             compose = "CentOS-Stream-8"
 
-        if arch == "aarch64":
-            # TF has separate composes for aarch64 architecture
-            compose += "-aarch64"
-
         if self.job_config.use_internal_tf:
             if self.is_compose_matching(compose, compiled_composes):
                 return compose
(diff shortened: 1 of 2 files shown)


lachmanfrantisek (approved):
Thanks!

thrix:
LGTM, let's get this in! BTW this was never needed for RH ranch, and it even causes there a bug. We just did not realize, as nobody was testing against `aarch64` :)

PR #2230, opened 2023-10-13
Fix bypassing compose name sanitization

lbarcziova:
distro2compose always accepts target (including arch), therefore we cannot match the whole target as a compose.

See https://redhat-internal.slack.com/archives/C04MU29TES1/p1696945995707489

CC @thrix


lachmanfrantisek (approved):

lbarcziova:
recheck

lachmanfrantisek:
As discussed with Laura, merging manually...

PR #2248, opened 2023-11-07
Storing downstream Koji build fixes

lbarcziova:
Fixes:

https://red-hat-it.sentry.io/issues/4609526388/?project=1767823&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=30d&stream_index=3&utc=true

https://red-hat-it.sentry.io/issues/4609585723/?project=1767823&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=30d&stream_index=8&utc=true


packit_service/worker/handlers/koji.py
```diff
@@ -305,14 +305,20 @@ def run(self):
...
         if (
             new_commit_status
             and self.build.status
-            and self.build.status != KojiBuildState.building
+            and self.build.status
+            in (
+                BaseCommitStatus.failure.value,
+                BaseCommitStatus.error.value,
+                BaseCommitStatus.success.value,
+            )
```
majamassarini, line 317: Thanks it lgtm. I have just one question, that is not related with your refactoring but it is just something that I don't understand... why shouldn't we update the status supposing the `new_commit_status` is different from `KojiBuildState.building`?

majamassarini (approved):

packit_service/worker/handlers/koji.py
lbarcziova: the point was to not overwrite the "final" status if we already have that in the DB, but since we set the value corresponding to BaseCommitStatus and not the KojiBuildState, therefore I changed it here

majamassarini: I was just wondering why don't have the "real" (latest) *final* state. I can't imagine why we shouldn't have the latest one 🤔

lbarcziova: I think this could happen e.g. if we process the events in a different order that they were sent.

majamassarini: Oh wired but thanks a lot for the answer!

PR #2253, opened 2023-11-14
Expose Copr repo from all builds in the env var for TF

lbarcziova:
The PACKIT_COPR_PROJECT env var will now also include the name of the Copr project from an additional build, if there is one provided in the retrigger comment.

CC @inknos

---

RELEASE NOTES BEGIN
`PACKIT_COPR_PROJECT` env var that is exposed to Testing Farm now includes the Copr project of the additional build specified in comment, if present.
RELEASE NOTES END


majamassarini (approved):
🙏🏻 probably a release note can be written for this feature :)

PR #2255, opened 2023-11-21
Get the latest release including pre-releases 

nforro:
For GitHub projects, `GitProject.get_latest_release()` considers only stable releases. If sync_release failed for a pre-release, the retrieved git ref is incorrect. Fix that.

Fixes https://github.com/packit/specfile/issues/325.


lbarcziova (commented):
could we rather use the first release from  `get_releases` (that one should list all the releases including the pre-releases)?

nforro:
recheck

lbarcziova (approved):
LGTM once the tests are green, thanks!

Issue #2271, opened 2023-12-07
Check of allowed_pr_authors for dist-git PRs with multiple commits is not correct


lbarcziova:
For pushes to dist-git, when downstream Koji builds are configured, we check in case the commit is coming from a PR whether the PR author matches the configuration of `allowed_pr_authors`. 

The problem is in getting the PR here in case we don't react to the head commit (there are multiple commits in the PR and the head commit doesn't have specfile change):
https://github.com/packit/packit-service/blob/d222ac18e1614fe057561c9797ceb30cecbd9e89/packit_service/worker/mixin.py#L259-L272

Example:
- PR: https://src.fedoraproject.org/rpms/python-ibm-cloud-sdk-core/pull-request/6
- we reacted to this commit as it has a specfile change: https://apps.fedoraproject.org/datagrepper/v2/id?id=63799bf2-5bcf-417a-ab42-e2eb3f0039c2&is_raw=true&size=extra-large which was not the PR head commit and therefore we were not able to get the PR (author) . Therefore the pre-check failed on this which resulted in Koji build not being triggered.
- 
Logs: 
```
2023-11-16T07:03:31.506+00:00	Getting pull request with head commit 43b577b8642d26db0509c342d3c2dfcf674320a9for repo rpms/python-ibm-cloud-sdk-core
2023-11-16T07:03:32.091+00:00	PR author: None
2023-11-16T07:03:32.092+00:00	Push event rawhide with corresponding PR created by None that is not allowed in project configuration: ['packit'].
```

lbarcziova:
I opened draft PRs for the proposed solution: packit/ogr#826, #2275 and packit/packit-service-fedmsg#99 - in fedmsg, filter the pushes (with the check for specfile changes) only if the commit doesn't come from a PR. If the commit comes for a PR, we should run the Koji build for the head commit of the PR which doesn't necessarily have to have the specfile change, so we will now check whether the whole PR has a specfile change. WDYT about this solution?

nforro:
I think it makes sense, but it doesn't solve the case when someone pushes multiple commits directly (with the HEAD commit not containing a spec file change), right? That could theoretically happen if someone follows this workflow: https://packit.dev/docs/fedora-releases-guide#keeping-dist-git-branches-non-divergent

lbarcziova:
good point, but I have to correct myself, after looking into the code of running the build, we just checkout the corresponding dist-git branch and do not checkout the specific commit, so that should be ok. 

[...]

PR #2275, opened 2023-12-11
Adjust check of dist-git PRs for Koji builds

lbarcziova:
If the dist-git push comes from a PR, check that
the specfile was changed in that PR.

Related to #2271
Needs packit/ogr#826

TODO:

---

RELEASE NOTES BEGIN
We have fixed a bug of not running Koji builds for Packit PRs with multiple commits if the last commit of the PR did not contain the specfile change.
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/checker/distgit.py
+++ b/packit_service/worker/checker/distgit.py
@@ -24,6 +24,22 @@ logger = logging.getLogger(__name__)
 
 
 class PermissionOnDistgit(Checker, GetPagurePullRequestMixin):
+    def contains_specfile_change(self):
+        """
+        Check whether the dist-git commit contains
+        any specfile change (do the check only for pushes from PRs,
+        with direct pushes we do the filtering in fedmsg).
+        """
+        if not self.pull_request:
+            return True
+
+        pr_id = self.pull_request.id
+        diff = self.project.get_pr_files_diff(pr_id) or {}
+        if not any(change.endswith(".spec") for change in diff):
+            logger.info(f"PR {pr_id} does not contain a specfile change.")
+            return False
+        return True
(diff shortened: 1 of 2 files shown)


packit_service/worker/checker/distgit.py
```diff
@@ -40,6 +56,15 @@ def pre_check(self) -> bool:
                 return False
 
             if self.data.event_dict["committer"] == "pagure":
+                if not self.pull_request:
+                    logger.debug(
+                        "Not able to get the pull request "
+                        "(may not be the head commit of the PR)."
+                    )
```
nforro, line 63: Wouldn't it make more sense to return `True` here rather than in `contains_specfile_change()`?

packit_service/worker/checker/distgit.py
```diff
@@ -24,6 +24,22 @@
 
 
 class PermissionOnDistgit(Checker, GetPagurePullRequestMixin):
+    def contains_specfile_change(self):
+        """
+        Check whether the dist-git commit contains
+        any specfile change (do the check only for pushes from PRs,
+        with direct pushes we do the filtering in fedsmg).
```
nforro, line 31:
```suggestion
        with direct pushes we do the filtering in fedmsg).
```

packit_service/worker/checker/distgit.py
lbarcziova: sure, I would swear it was there :D

lachmanfrantisek (commented):
The approach looks good to me.

lbarcziova:
recheck

majamassarini (approved):

PR #2282, opened 2023-12-12
fix: migrate the API from dashboard

mfocko:


lachmanfrantisek (approved):
Thanks!

packit_service/service/api/architecture.py
```diff
@@ -0,0 +1,47 @@
...
+# )
+# from packit_service.service.api.utils import response_maker
+
+# logger = getLogger("packit_service")
+
+# TODO: Migrate properly
+# architecture_ns = Namespace(
+#     "architecture",
+#     description="Provides API endpoints related to architecture of Packit"
+# )
```
lachmanfrantisek, line 33: :+1: We might do it with some Patternfly instead.

Issue #2287, opened 2023-12-19
"PR does not contain a specfile change" but it does


majamassarini:
### What happened? What is the problem?

After merging this PR https://src.fedoraproject.org/rpms/ddnet/pull-request/4 packit does not react creating a koji build and bodhi update.

Looking in sentry I can see:
```
12/19/23
10:56:17.153 AM	
2023-12-19T10:56:17.153909576+00:00 stderr F [2023-12-19 10:56:17,153: ERROR/MainProcess] {'error': 'No statistics could be computed for this PR', 'error_code': 'ENOPRSTATS'}
12/19/23
10:56:17.157 AM	
2023-12-19T10:56:17.157516615+00:00 stderr F [2023-12-19 10:56:17,157: INFO/MainProcess] PR 4 does not contain a specfile change.
```

But the PR has a change in the specfile.

### Steps to reproduce

```bash
1.
2.
3.
```

### What is the impacted category (job)?

Fedora release automation

### Workaround

- [ ] There is an existing workaround that can be used until this issue is fixed.

### Participation

- [ ] I am willing to submit a pull request for this issue. (Packit team is happy to help!)

PR #2288, opened 2023-12-19
Fix race condition causing ENOPRSTATS

nforro:
Fixes https://github.com/packit/packit-service/issues/2287.


nforro:
We decided to implement another solution.

PR #2291, opened 2023-12-21
Increase waiting time and retries in get_files_diff

majamassarini:
When we call this check the PR is still in open state transitioning to a merged state.
Until the transition is done we can not get diff stats.


majamassarini:
recheck

tests/integration/test_dg_commit.py
```diff
@@ -935,7 +935,7 @@ def test_precheck_koji_build_push_pr(
         ]
     )
     flexmock(PagureProject).should_receive("get_pr_files_diff").with_args(
-        5, retries=3, wait_seconds=3
+        5, retries=5, wait_seconds=5
```
nforro, line 938: So we don't have to modify tests every time we change the values:

```suggestion
        5, retries=int, wait_seconds=int
```

nforro (approved):

PR #2302, opened 2024-01-11
Use correct git ref to identify most recent tag

nforro:
Fixes https://github.com/packit/packit/issues/2184.

Merge after https://github.com/packit/packit/pull/2193.

### DIFF
--- a/packit_service/worker/helpers/build/build_helper.py
+++ b/packit_service/worker/helpers/build/build_helper.py
@@ -514,12 +514,28 @@ class BaseBuildJobHelper(BaseJobHelper):
         else:
             update_release = None  # take the value from config (defaults to True)
 
+        # use correct git ref to identify most recent tag
+        if self.job_config.trigger == JobConfigTriggerType.pull_request:
+            merged_ref = self.project.get_pr(
+                self._db_project_object.pr_id
+            ).target_branch
+        elif self.job_config.trigger == JobConfigTriggerType.commit:
+            merged_ref = self._db_project_object.commit_sha
+        elif self.job_config.trigger == JobConfigTriggerType.release:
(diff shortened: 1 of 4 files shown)


packit_service/worker/helpers/build/build_helper.py
```diff
@@ -514,12 +514,25 @@ def _create_srpm(self):
...
+        # use correct git ref to identify most recent tag
+        if self.job_config.trigger == JobConfigTriggerType.pull_request:
+            merged_ref = self.project.get_pr(
+                self._db_project_object.pr_id
+            ).target_branch
+        elif self.job_config.trigger == JobConfigTriggerType.commit:
+            merged_ref = self._db_project_object.commit_sha
+        elif self.job_config.trigger == JobConfigTriggerType.release:
+            merged_ref = self._db_project_object.tag_name
+        else:
```
mfocko, line 526: When does this happen? :eyes:

mfocko (approved):

packit_service/worker/helpers/build/build_helper.py
nforro: It should never happen, it just kind of demonstrates the default/fallback value :slightly_smiling_face: Should I remove it?

lbarcziova (approved):

packit_service/worker/helpers/build/build_helper.py
mfocko: I would probably add a message to the logs, warning maybe? ideally with the trigger, so we know how we got there and that we're using the fallback.

With regards to the `.create_srpm`, I would probably prefer default to be `None` that gets resolved to `HEAD`, so that we don't have the default value both in service and packit, though I'll leave that decision up to you :D

PR #2319, opened 2024-01-24
Try to handle PendingRollbackError

nforro:
I think this could solve the PendingRollbackError issue, however I don't really have a way to test it. Once this is merged, I can try to rescale postgres on stage in order to hopefully trigger the error. What do you think?

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -12,6 +12,7 @@ from contextlib import contextmanager
 from datetime import datetime, timedelta, timezone
 from os import getenv
 from typing import (
+    Any,
     Dict,
     Iterable,
     List,
@@ -47,12 +48,14 @@ from sqlalchemy.dialects.postgresql import array as psql_array
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import (
     Session as SQLASession,
+    Query,
     relationship,
     scoped_session,
     sessionmaker,
 )
 from sqlalchemy.sql.functions import count
 from sqlalchemy.types import ARRAY
+from sqlalchemy.exc import PendingRollbackError
 
 from packit.config import JobConfigTriggerType
 from packit_service.constants import ALLOWLIST_CONSTANTS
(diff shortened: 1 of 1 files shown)


packit_service/models.py
```diff
@@ -128,6 +131,19 @@ def sa_session_transaction() -> SQLASession:
         raise
 
 
+def db_query(session: SQLASession, *entities: Any) -> Query:
+    """
+    Wrapper for SQLASession.query() that tries to access the connection first
+    and deal with potential PendingRollbackError.
+    """
+    try:
```
lbarcziova, line 139: could you please reference some docs/resources that you used?

majamassarini (approved):
Nice idea! Thanks.

lbarcziova (approved):
let's try it out!

packit_service/models.py
nforro: I've added a link to https://stackoverflow.com/a/69687698. Even though in our case the reason is different (I think it's because a rollback, even though issued, can't happen when postgres is unavailable), the solution could work for us.

lbarcziova: ok, thanks!

nforro:
Right, we should probably revert this. I wanted to try using:

```python
while session.registry().in_transaction():
    session.rollback()
```

instead of plain `session.rollback()`, but I'm not so sure it would change anything.

lbarcziova:
@nforro let's at least try it, this has been happening more often and causes instability of the service (the pod was stuck again 10h again and I had to rescale)

if it doesn't help, we can properly refine this issue and work on it further

nforro:
So, the problem is, `Session` has no `registry()` method, but I don't think it matters, accessing `connection()` clearly doesn't throw the exception so it wouldn't be executed anyway. There seems to be a "disbalance" between the connection that session can access and engine connection. Let me try to simulate the problem locally, maybe I'll figure out a way how to detect a broken connection and recover from it.

PR #2320, opened 2024-01-24
Support aliases and groups in allowed_* config options for Koji

lbarcziova:
Introduce aliases `all_admins` and `all_committers` and also support groups in allowed_pr_authors and allowed_committers config options

Fixes packit/packit#2088
Requires packit/ogr#834

TODO:

- [x] Write new tests or update the old ones to cover new functionality (needs packit/ogr#834)
- [x] Update or write new documentation in `packit/packit.dev`. (packit/packit.dev#814)

---

RELEASE NOTES BEGIN
`allowed_pr_authors` and `allowed_committers` now allow specifying groups and also aliases `all_admins` and `all_committers` (corresponding to the access to the repository).
RELEASE NOTES END

### DIFF
--- a/packit_service/constants.py
+++ b/packit_service/constants.py
@@ -289,3 +289,8 @@ CHANGED_LOADING_BEHAVIOUR_IN_DISTGIT_MESSAGE = (
     "of the `pull_from_upstream` job. We believe this adjustment will simplify the onboarding "
     "process and enhance the overall user experience. "
 )
+
+
+class KojiAllowedAccountsAlias(Enum):
+    all_admins = "all_admins"
+    all_committers = "all_committers"
--- a/packit_service/worker/checker/distgit.py
+++ b/packit_service/worker/checker/distgit.py
@@ -4,8 +4,9 @@
 import logging
 import re
 
+from ogr.abstract import AccessLevel
 from packit.config.aliases import get_branches
-from packit_service.constants import MSG_GET_IN_TOUCH
+from packit_service.constants import MSG_GET_IN_TOUCH, KojiAllowedAccountsAlias
 from packit_service.worker.checker.abstract import Checker, ActorChecker
 from packit_service.worker.events import (
     PushPagureEvent,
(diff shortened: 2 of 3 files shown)


packit_service/worker/checker/distgit.py
```diff
@@ -43,6 +44,66 @@ def contains_specfile_change(self):
...
+        for value in accounts_list:
+            if self.is_koji_allowed_accounts_alias(value):
+                all_accounts.update(self.expand_maintainer_alias(value))
+            elif value.startswith("@"):
+                try:
+                    # remove @
+                    group = self.project.service.get_group(value[1:])
+                    all_accounts.update(group.members)
+                except Exception:
+                    pass
```
majamassarini, line 82: Why not logging the exception here?

majamassarini (approved):

packit_service/worker/checker/distgit.py
lbarcziova: good point, I will change that

packit_service/constants.py
```diff
@@ -289,3 +289,8 @@ def from_number(number: int):
     "of the `pull_from_upstream` job. We believe this adjustment will simplify the onboarding "
     "process and enhance the overall user experience. "
 )
+
+
+class KojiAllowedAccountsAlias(Enum):
+    all_admins = "all_admins"
+    all_committers = "all_committers"
```
nforro, line 296: I'm just wondering about the `all_` prefix. An alias is technically a project-specific virtual group, and if we use a `@` symbol for groups, we could use something similar for the aliases.

nforro (approved):

packit_service/constants.py
lbarcziova: personally I like the original proposal, having it like the current implementation, but no strong opinion. @packit/the-packit-team anyone has any preferences?

majamassarini: I don't have strong opinions either.
But I wouldn't use the @ symbol. I agree that we have created a group, but it is a "packit" group and it will not show up in the fedoraproject UI; for this reason I think it is misleading to use @
We can have another symbol but it will not "immediately suggest" to the user the "group idea", so I don't know if it is worth having it?

nforro: I just think that using a symbol is a little more consistent and gives the feeling that it's some special identifier, on the other hand `all_admins` is also pretty self-explanatory:

```yaml
allowed_pr_authors:
    - $admins
    - @packager
    - packit-stg
```

```yaml
allowed_pr_authors:
    - all_admins
    - @packager
    - packit-stg
```

I'm fine with both.

nforro: If we used `$`, we could call it a variable rather than alias: `$admins expands to all users with admin access to the project`

lbarcziova: We already have the aliases for `targets` and these also do not use any special symbols. So if we would like to be consistent, I would go with the original idea.

PR #2322, opened 2024-01-25
Resolve owner and project through CoprBuildHelper

majamassarini:
If no owner and project are given in the vm_image_build job resolve them through the CoprBuildHelper defaults.

When looking for the copr_build job related with a vm_image_build and COPR owner or project are not given then
guess they are being created by the CoprBuildHelper and use it to find them.


lbarcziova (approved):
thanks!

PR #2328, opened 2024-02-01
Increase timeouts and retries to prevent ENOPRSTATS

majamassarini:
ENOPRSTATS are still happening in prod.

Try to mitigate them a bit more.


lbarcziova (approved):
thanks!

PR #2331, opened 2024-02-02
Roll back also after exceptions raised during queries

nforro:
It turns out a SQLAlchemy session [uses transactions even for queries](https://docs.sqlalchemy.org/en/20/orm/session_basics.html), so if an exception happens during a query execution, it is necessary to roll back to be able to use the session further.

Related to https://github.com/packit/packit-service/pull/2319.

---

I've been able to reproduce the behavior locally and verify that adding rollbacks to queries makes the `PendingRollbackError` go away, so I'm quite confident this will work, however I can't be 100% sure :slightly_smiling_face:

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -103,25 +103,25 @@ else:  # service/httpd
     singleton_session = None
 
 
-def sa_session() -> SQLASession:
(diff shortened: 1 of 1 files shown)


lachmanfrantisek (approved):
:tada: Thanks a lot Nikola!!!

PR #2332, opened 2024-02-04
Factor out reporting

mfocko:


mfocko:
recheck

lbarcziova (approved):
thanks, great improvement!

PR #2334, opened 2024-02-07
Fix checking of PR labels

lbarcziova:
The labels are GithubLabel objects and not strings, so we need to check their name attribute.

---

RELEASE NOTES BEGIN
N/A
RELEASE NOTES END

### DIFF
--- a/packit_service/utils.py
+++ b/packit_service/utils.py
@@ -243,7 +243,7 @@ def pr_labels_match_configuration(
         f"(label.present: {configured_labels_present}, label.absent: {configured_labels_absent})"
     )
 
-    pr_labels = pull_request.labels
+    pr_labels = [label.name for label in pull_request.labels]
     logger.info(f"Labels on PR: {pr_labels}")
 
     return (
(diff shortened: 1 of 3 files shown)


packit_service/utils.py
```diff
@@ -243,7 +243,7 @@ def pr_labels_match_configuration(
         f"(label.present: {configured_labels_present}, label.absent: {configured_labels_absent})"
     )
 
-    pr_labels = pull_request.labels
+    pr_labels = [label.name for label in pull_request.labels]
```
lachmanfrantisek, line 246: Just to be sure, won't this break the use on GitLab/Pagure?

lachmanfrantisek (approved):

packit_service/utils.py
lbarcziova: oh, good point :face_exhaling: at least for GitLab these are indeed strings..I need to have a deeper look to fix it

lbarcziova: opened packit/ogr#839

PR #2335, opened 2024-02-07
fix(event): use fallback for accepted time

### LINKED ISSUE #2258: AttributeError: 'NoneType' object has no attribute 'tzinfo'
Sentry Issue: [PCKT-002-PACKIT-SERVICE-6CJ](https://red-hat-it.sentry.io/issues/4455199048/?referrer=github_integration)

```
JSONDecodeError: Expecting value: line 1 column 1 (char 0)
  File "copr/v3/requests.py", line 123, in handle_errors
    response_json = response.json()
  File "requests/models.py", line 900, in json
    return complexjson.loads(self.text, **kwargs)
  File "__init__.py", line 525, in loads
    return _default_decoder.decode(s)
  File "simplejson/decoder.py", line 370, in decode
    obj, end = self.raw_decode(s)
  File "simplejson/decoder.py", line 400, in raw_decode
    return self.scan_once(s, idx=_w(s, idx).end())

CoprTimeoutException: Gateway Timeout
Consider using pagination for large queries.
(3 additional frame(s) were not displayed)
...
  File "copr/v3/proxies/build.py", line 328, in create_from_custom
    return self._create(endpoint, data, buildopts=buildopts)
  File "copr/v3/helpers.py", line 73, in wrapper
    result = func(*args, **kwargs)
  File "copr/v3/proxies/build.py", line 340, in _create
    response = self.request.send(
  File "copr/v3/requests.py", line 46, in send
    handle_errors(response)
  File "copr/v3/requests.py", line 146, in handle_errors
    raise CoprTimeoutException(message, response=response)

AttributeError: 'NoneType' object has no attribute 'tzinfo'
(9 additional frame(s) were not displayed)
...
  File "packit_service/worker/helpers/build/copr_build.py", line 704, in handle_build_submit_error
    self.monitor_not_submitted_copr_builds(
  File "packit_service/worker/helpers/build/copr_build.py", line 397, in monitor_not_submitted_copr_builds
    time = elapsed_seconds(
  File "packit_service/utils.py", line 184, in elapsed_seconds
    begin = get_timezone_aware_datetime(begin)
  File "packit_service/utils.py", line 171, in get_timezone_aware_datetime
    if is_timezone_naive_datetime(datetime_to_update):
  File "packit_service/utils.py", line 156, in is_timezone_naive_datetime
    datetime_to_check.tzinfo is None
```

It looks like `task_accepted_time` can be `None`:
https://github.com/packit/packit-service/blob/ff6c0f5f4ae9e51311fb8e195d2b2df3d325c4ac/packit_service/worker/helpers/build/copr_build.py#L397-L399

mfocko:
When submitting failed Copr builds to the Pushgateway, we're using time of accepting the task, in some cases we don't have time of accepting, so use ‹created_at› as a fallback.

Fixes #2258

Not sure about this change, since for the last occurrence in Sentry the difference between the reported time of failure and `created_at` is 5 minutes. I can the ‹task_accepted_time› only in DB, but this is happening even before the build starts, cause it's coming from an issue caused **by submitting** the build, so I would say that we should use the `datetime.now()`? But OTOH it doesn't feel like it's skipping the Celere queue and being handled _right away_.

### DIFF
--- a/packit_service/worker/events/event.py
+++ b/packit_service/worker/events/event.py
@@ -113,11 +113,12 @@ class EventData:
         commit_sha = event.get("commit_sha")
         identifier = event.get("identifier")
         issue_id = event.get("issue_id")
+
+        time = event.get("task_accepted_time")
         task_accepted_time = (
-            datetime.fromtimestamp(event.get("task_accepted_time"), timezone.utc)
-            if event.get("task_accepted_time")
-            else None
+            datetime.fromtimestamp(time, timezone.utc) if time else None
         )
+
         build_targets_override = event.get("build_targets_override")
         tests_targets_override = event.get("tests_targets_override")
         branches_override = event.get("branches_override")
--- a/packit_service/worker/helpers/build/copr_build.py
+++ b/packit_service/worker/helpers/build/copr_build.py
@@ -394,6 +394,17 @@ class CoprBuildJobHelper(BaseBuildJobHelper):
         Measure the time it took to set the failed status in case of event (e.g. failed SRPM)
         that prevents Copr build to be submitted.
         """
+
+        # NOTE: When there is no ‹task_accepted_time›, we skip the submission to
+        # the metrics, since there is no delay between the submission and
+        # failure. We could probably track those by a separate metric as
+        # suggested by Maja in the PR.
+        if self.metadata.task_accepted_time is None:
+            logger.warning(
+                "No task_accepted_time for failed Copr build with reason: %s", reason
+            )
+            return
+
         time = elapsed_seconds(
             begin=self.metadata.task_accepted_time, end=datetime.now(timezone.utc)
         )


majamassarini (approved):
At first it was wired to see different "kind" of timestamps used to initialize the "task_accepted_time" field. But I think we use this field just to show it to the user, and the difference in this kind of timestamps is so minimal that no one will ever notice it.

So for me it looks good, maybe I would just add a note.  🙏

mfocko:
@majamassarini The time difference is used for the metrics and SLOs, 5 minutes is exactly the range we have as an SLO, the thing is that these errors are occurring during the submission, so there is no “status set after pass/fail of the Copr build”?

lbarcziova:
> I can the ‹task_accepted_time› only in DB, but this is happening even before the build starts, cause it's coming from an issue caused by submitting the build, so I would say that we should use the datetime.now()

Sorry, I am still not following, why do these miss the information about `task_accepted_time`?

majamassarini:
> @majamassarini The time difference is used for the metrics and SLOs, 5 minutes is exactly the range we have as an SLO, the thing is that these errors are occurring during the submission, so there is no “status set after pass/fail of the Copr build”?
> 
> I'm kind of inclined to do `datetime.now()` if there is no `task_accepted_time`

Oh, I have forgot about the SLOs 🙈
And now probably I am following; if any kind of errors occurs that prevents us to update the *status checks* then the `task_accepted_time` remains with the initial value of `None`, and I think this is ok because we never really accepted the task.
I am wondering if we can't deal with the exception of having a `None` value there? We don't have to calculate a difference in this case, it is another metric: we are not simply slow but an error has occurred.

mfocko:
So do we want to count them in with being an immediate failure, or just drop them from metrics? I'm kinda inclined to count them with time difference of `datetime.now() - datetime.now()`

* :one: for dropping
* :two: for counting

Vote on your devices :) :D

PR #2338, opened 2024-02-10
Install epel-rpm-macros package on worker

nforro:
This makes for example spec files using `%gometa -L` parseable.


gotmax23 (changes requested):
go-rpm-macros-epel should _not_ be in the default buildroot. I would depend on epel-rpm-macros which pulls in the correct package, `go-srpm-macros-epel`, and the other EPEL macro backports.

gotmax23 (approved):
Thanks!

lbarcziova (approved):

PR #2333, opened 2024-02-06
Conditions on PR labels

lbarcziova:
TODO:
- [ ] docs

Fixes #2269 #2186

---

RELEASE NOTES BEGIN
We have introduced new configuration options `require.label.present` and `require.label.absent`. By configuring these you can specify labels that need to be present or absent on a pull request for Packit to react on such PR.
RELEASE NOTES END

PR #2344, opened 2024-02-12
Check the labels also when processing copr build end

lbarcziova:
When scheduling follow-ups testing jobs, this needs to be additionally checked here as well. Followup of #2333

---

RELEASE NOTES BEGIN
￼
RELEASE NOTES END


packit_service/worker/handlers/copr.py
```diff
@@ -415,6 +416,18 @@ def handle_testing_farm(self):
                 if (
                     not job_config.skip_build
                     and not job_config.manual_trigger
+                    and (
+                        job_config.trigger != JobConfigTriggerType.pull_request
```
mfocko, line 420: Probably too late to the party, but `job_config.require.label.satisfied(job_config, self.copr_build_helper.pull_request_obejct` would've been nice, I have a feeling that something like this is in the other PR too.

Maybe `NOTE`-worthy? :D

mfocko (approved):
LGTM

Issue #2068, opened 2023-05-25
Support release syncing to CentOS Stream


lbarcziova:
Followup of packit/packit#1979:

- [ ] update the [parsing of the message from Upstream release monitoring](https://github.com/packit/packit-service/blob/38d6158d1b718a079d4b08c35bbd69caf2e7859b/packit_service/worker/parser.py#L1484) if needed
   - [here](https://github.com/packit/packit-service/blob/38d6158d1b718a079d4b08c35bbd69caf2e7859b/packit_service/worker/parser.py#L1501) we construct the URL of the package in Fedora dist-git, this needs to be done for CentOS Stream (the repo for the particular package in [CentOS Stream dist-git on GitLab](https://gitlab.com/redhat/centos-stream/rpms)) as well
   - this URL should be used as well [when looking](https://github.com/packit/packit-service/blob/38d6158d1b718a079d4b08c35bbd69caf2e7859b/packit_service/worker/events/new_hotness.py#L79) for the configuration (of pull-from-upstream)
- [ ] update the [call of sync_release](https://github.com/packit/packit-service/blob/38d6158d1b718a079d4b08c35bbd69caf2e7859b/packit_service/worker/handlers/distgit.py#L214) according to the implementation in packit/packit#1979 => for `dist_git_branches` such as c8s, c9s trigger the release sync in CentOS Stream and not Fedora
---
- [ ] the retriggering should be later possible as well, but it is currently being worked on in #1829 (will update the action items once the implementation is done)

CC @cathay4t

mfocko:
I had a look at it… It is a bit more complicated to resolve:

* parsing is not affected that much, yet
  * we should make sure we're using the correct package name for both Fedora and CentOS
  * we should include the CentOS dist-git
* handling of the CentOS _v._ Fedora should be separated, since they depend on the dist-git included in the API and we can have only one dist-git, duh…
  * having a separate “package” (one for Fedora, one for CentOS) in Packit config might ease this a lot, since there are different `dist_git_branches` and each “package” would have its own downstream counterpart, there would be 2 separate Celery tasks for handling this, thus resolving the `PackitAPI` issue
* given the two points above, it would probably make sense to extend the event to hold dist-git URL**s** instead of just the one

lbarcziova:
The person who will pick this should make sure this is everything that needs to be changed and in case there is more, create follow-up issues.

nforro:
[...]

PR #2348, opened 2024-02-14
Feat/centos anitya

mfocko:
- [x] Add an event class
- [x] Implement a parser
- [ ] Pass around versions in a list (for New Hotness we can have just one version in the list, for Anitya version update fail)
- [ ] Do not generate DB entry from the event
- [ ] **⚠️** Choosing the correct version
  - Outcomes from the arch:
    - Gather more use cases
    - **Require** the explicit setting for version mask (for CentOS at least)
- [ ] **⚠️** Decide whether to support Fedora for Anitya version update too… right now it's prefiltered for the CentOS

Related to #2068


packit_service/worker/parser.py
```diff
@@ -1568,6 +1571,37 @@ def parse_new_hotness_update_event(event) -> Optional[NewHotnessUpdateEvent]:
             release_monitoring_project_id=release_monitoring_project_id,
         )
 
+    @staticmethod
+    def parse_anitya_version_update_event(event) -> Optional[AnityaVersionUpdateEvent]:
+        if "anitya.project.version.update.v2" not in event.get("topic", ""):
+            return None
+
+        # FIXME: Handle Fedora too in case we want to support multiple releases
+        # such as for Go.
```
mfocko, line 1580: There was an RFE for this on the Slack already…

packit_service/worker/events/new_hotness.py
```diff
@@ -148,3 +147,58 @@ def get_dict(self, default_dict: Optional[Dict] = None) -> dict:
...
+        super().__init__(
+            package_name=package_name,
+            distgit_project_url=distgit_project_url,
+            release_monitoring_project_id=release_monitoring_project_id,
+        )
+
+        self._versions = versions
+
+    @property
+    def version(self) -> str:
```
mfocko, line 195: The major issue here is that we “generate” DB entry for the release from the event object. There's also serialization involved (for Celery).

At the same time it feels a bit weird to keep logic of “deciding the next version” in the »event«.

packit_service/worker/parser.py
lbarcziova: +1, we can create a separate issue for this (e.g.this will imply monitoring settings for the package will not be considered anymore, therefore also changes in documentation etc.) 

we could properly test this for CentOS Stream and once it is tested to be working do the switch for Fedora packages to not break anything for already onboarded Fedora packages

packit_service/worker/events/new_hotness.py
```diff
@@ -148,3 +147,58 @@ def get_dict(self, default_dict: Optional[Dict] = None) -> dict:
...
+        # TODO: Handle here or further down the chain? we should be able to get
+        # the package config from dist-git and resolve the next version; how are
+        # we going to choose the version?
+        #   a) latest greatest
+        #   b) next unreleased?
+        # this can be also influenced by the mask…
```
mfocko, line 201: Example of a problem:

* There's a `1.0.0` release in Fedora.
* We've got releases `1.0.1`, `1.1.0` and `2.0.0` from Anitya
* Which release do we sync? The next possible (`1.0.1`) or latest-greatest `2.0.0`? Syncing all of them would create a lot of noise.
* Ideal scenario here would be having the *version mask* defined, so it would be as explicit as possible.

packit_service/worker/events/new_hotness.py
lbarcziova: discussed together with Mato that we want all the new versions, see e.g. https://apps.fedoraproject.org/datagrepper/v2/id?id=efec10ed-9f86-4e03-9e4f-2e626934b3ae&is_raw=true&size=extra-large in here we would be interested in all 3

lbarcziova (commented):
can we also have a testcase please?

packit_service/worker/events/new_hotness.py
lbarcziova: and then it is up to maintainer to define the per-branch jobs and rules (e.g. via version_udpate_mask, upstream_tag_include, upstream_tag_exclude): [example](https://src.fedoraproject.org/rpms/golang/blob/rawhide/f/.packit.yaml)

mfocko:
recheck

mfocko:
@lbarcziova rebased, added test case; left notes in comments (code) and also updated the original description; this should be ready for review and merge

lbarcziova (approved):
LGTM, thanks for the refactoring of the tests!

PR #2303, opened 2024-01-15
Migrate to org.fedoraproject.prod.pagure.git.receive

nforro:

PR #2351, opened 2024-02-16
  Migrate to org.fedoraproject.prod.pagure.git.receive

### LINKED ISSUE #2305: Migrate to org.fedoraproject.prod.pagure.git.receive
Currently used `org.fedoraproject.prod.git.receive` will be removed as it is triggered by the same action.

lbarcziova:
Supersedes #2303

Related to #2305

---

RELEASE NOTES BEGIN
N/A
RELEASE NOTES END

### DIFF
--- a/packit_service/worker/checker/distgit.py
+++ b/packit_service/worker/checker/distgit.py
@@ -41,25 +41,6 @@ class LabelsOnDistgitPR(Checker, GetPagurePullRequestMixin):
 
 
 class PermissionOnDistgit(Checker, GetPagurePullRequestMixin):
-    def contains_specfile_change(self):
-        """
-        Check whether the dist-git commit contains
-        any specfile change (do the check only for pushes from PRs,
-        with direct pushes we do the filtering in fedmsg).
-        """
-        if not self.pull_request:
-            return True
-
-        pr_id = self.pull_request.id
-        logger.debug(f"PR {pr_id} status: {self.pull_request.status}")
-        # Pagure API tends to return ENOPRSTATS error when a pull request is transitioning
-        # from open to merged state, give it some extra time
-        diff = self.project.get_pr_files_diff(pr_id, retries=7, wait_seconds=7) or {}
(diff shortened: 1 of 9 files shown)


packit_service/worker/checker/distgit.py
```diff
@@ -139,17 +120,7 @@ def pre_check(self) -> bool:
...
-                    logger.debug(
-                        "Not able to get the pull request "
-                        "(may not be the head commit of the PR)."
-                    )
-                    return False
-
-                if not self.contains_specfile_change():
-                    return False
-
+            if self.pull_request:
```
lbarcziova, line 123: @nforro as you mentioned, there is no way to tell from the message whether the commit(s) is coming from PR, do you think it is ok to always try to get the PR via `end_commit` or do you have some other suggestions?

nforro: I don't think we have a choice. But I think this has to be adjusted as well:

https://github.com/packit/packit-service/blob/a439cb5d734252ae00cb1f8143cbba301def7571/packit_service/worker/mixin.py#L257

lbarcziova: > I don't think we have a choice. 

there is still option to revisit the 2 config options, but that would require changes in configs for users who use these options

nforro: I was wondering if it still makes sense to differentiate between the two events. But we need to be able to recognize Packit PRs, and the person making the push (agent) is not packit, e.g.: https://apps.fedoraproject.org/datagrepper/v2/id?id=a934e891-c91c-4d7e-b287-3d7ef6771f8f&is_raw=true&size=extra-large.

lbarcziova: Yes I know, therefore I wrote it would require changes in the configs (I meant having only one option, for the committer, but the default could not be `packit` and would be more complicated). I will go with the current implementation and if we notice some problems with this, we can revisit.

nforro (approved):

PR #2352, opened 2024-02-16
Make check-onboarded-projects a daily job

majamassarini:
This should hopefully fix the long prefetch task queue in celery


lbarcziova (approved):
thanks!

PR #2365, opened 2024-03-08
Declare babysit tasks celery queue in a different way

majamassarini:
send_task  should have the same options as apply_async,
nevertheless passing queue as an option is not working.
Probably because the referenced queues have to be defined in task_queues.
Instead of defining task_queues let define explicitly the routes.

### DIFF
--- a/packit_service/celery_config.py
+++ b/packit_service/celery_config.py
@@ -7,6 +7,12 @@ import packit_service.constants
 # https://docs.celeryq.dev/en/stable/userguide/configuration.html#std-setting-task_default_queue
 task_default_queue = packit_service.constants.CELERY_TASK_DEFAULT_QUEUE
 
+
+task_routes = [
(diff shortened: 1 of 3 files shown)


lbarcziova (approved):

packit_service/celery_config.py
```diff
@@ -7,6 +7,12 @@
 # https://docs.celeryq.dev/en/stable/userguide/configuration.html#std-setting-task_default_queue
 task_default_queue = packit_service.constants.CELERY_TASK_DEFAULT_QUEUE
 
+
+task_routes = [
+    ("task.babysit_vm_image_build", {"queue": "short-running"}),
```
mfocko, line 12: Didn't you want `long-running` here?

majamassarini: 🙈

PR #2366, opened 2024-03-08
Change task_routes format

majamassarini:
The previous one was not working:
too many values to unpack (expected 2)

### DIFF
--- a/packit_service/celery_config.py
+++ b/packit_service/celery_config.py
@@ -8,10 +8,10 @@ import packit_service.constants
 task_default_queue = packit_service.constants.CELERY_TASK_DEFAULT_QUEUE
 
 
-task_routes = [
-    ("task.babysit_vm_image_build", {"queue": "short-running"}),
-    ("task.babysit_copr_build", {"queue": "short-running"}),
-]
+task_routes = {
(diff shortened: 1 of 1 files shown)


packit_service/celery_config.py
```diff
@@ -8,10 +8,10 @@
 task_default_queue = packit_service.constants.CELERY_TASK_DEFAULT_QUEUE
 
 
-task_routes = [
-    ("task.babysit_vm_image_build", {"queue": "short-running"}),
-    ("task.babysit_copr_build", {"queue": "short-running"}),
-]
+task_routes = {
```
lbarcziova, line 11: shouldn't the value still be `{"queue": "long-running"}`? (just quickly checked https://docs.celeryq.dev/en/stable/userguide/routing.html#automatic-routing)

lbarcziova: > The error I am getting now from stage is too many values to unpack so I think the {"queue": "long-running"} format is not what it is expecting.

I think the problem there is with the list iteration and not that dict itself. Not sure, haven't tested it locally either, I am ok with checking in stg

lbarcziova (approved):

PR #2373, opened 2024-03-13
Don't comment about logs on just submitted sync PR

lbarcziova:
Fixes packit/packit#2230

RELEASE NOTES BEGIN

We have adjusted commenting behaviour regarding Packit logs on release syncing PRs.

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/handlers/distgit.py
+++ b/packit_service/worker/handlers/distgit.py
@@ -404,8 +404,8 @@ class AbstractSyncReleaseHandler(
             model.set_logs(collect_packit_logs(buffer=buffer, handler=handler))
 
         dashboard_url = self.get_dashboard_url(model.id)
-        downstream_pr.comment(
-            f"Logs and details of the syncing: [Packit dashboard]({dashboard_url})"
+        self.report_dashboard_url(
+            sync_release_pull_request, downstream_pr, dashboard_url
         )
         self.sync_release_helper.report_status_for_branch(
             branch=branch,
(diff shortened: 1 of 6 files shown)


packit_service/worker/handlers/distgit.py
```diff
@@ -488,6 +488,20 @@ def _report_errors_for_each_branch(self, message: str):
...
+        downstream_pr: PullRequest,
+        dashboard_url: str,
+    ):
+        msg = f"Logs and details of the syncing: [Packit dashboard]({dashboard_url})"
+        # this is a retrigger
+        if len(sync_release_pull_request.sync_release_targets) > 1:
+            downstream_pr.comment(msg)
+        else:
+            original_title = downstream_pr.title
+            downstream_pr.title = original_title + "\n---\n" + msg
```
majamassarini, line 503:
```suggestion
            downstream_pr.title = original_title + "\n---\n" + msg
```

I will be surprised if we can do this operation.
I don't think we have the permissions to update our own PR, at least not until I finish my work on the Pagure side and I create a new token.

And I am also wondering if this is not too long as a title? 🤔

majamassarini: But we can try it on stg, to be 100% sure.

majamassarini (approved):

packit_service/worker/handlers/distgit.py
lbarcziova: I am dumb :D it was meant to be `description`. But yet, you are right, I am getting exc because of insufficient ACLs. I will have a look into it more, but I wanted to avoid having this logic directly in Packit as it is service-related. In case I won't find a way to do this nicely, I would wait for your Pagure changes to be deployed for this as well. Thanks for pointing this out!

hroncok:
Thank you.

PR #2375, opened 2024-03-18
Do not show build status in manual test checks

### LINKED ISSUE #2147: [RFE] Do not show queued jobs
### Description

Please do not show queued jobs as checks, when `manual_trigger: true` is in place.
Context:
When using the `manual_trigger:true` option in the Packit config, all jobs defined for the stage (pull_request) get triggered and show up in the checks overview.

In our scenario, we now have multiple tiers defined for the pull request stage, with just some subset of tests being recognized as essential for the PR to be merged.
We plan to introduce even more granularity to the job configuration and with that the check overview would be cluttered by the amount of jobs queued with a possibility to not ever be dispatched to the Testing Farm.

Related to #1754 

### Benefit

For more complex and multiple job definitions, the check overview might be overwhelming with many queued entries.
The pull request will get a :heavy_check_mark: when the submitted (and completed) jobs are green. :)))

### Importance

low

### What is the impacted category (job)?

Testing Farm tests

### Workaround

- [ ] There is an existing workaround that can be used until this feature is implemented.

### Participation

- [ ] I am willing to submit a pull request for this issue. (Packit team is happy to help!)

majamassarini:
Unless there are failures or errors.

This can be improved having a real dependency
between tests and builds.

Showing a neutral state for a successful build
message is not working on gitlab and either on
github is not working always
(my test falls back on commit statuses and
they don't have a neutral state).

Fixes #2147

RELEASE NOTES BEGIN

Packit no longer shows status checks for not yet triggered manual tests.

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/helpers/build/build_helper.py
+++ b/packit_service/worker/helpers/build/build_helper.py
@@ -650,7 +650,7 @@ class BaseBuildJobHelper(BaseJobHelper):
         update_feedback_time: Callable = None,
     ) -> None:
         for test_job in self.job_tests_all:
-            if test_job.skip_build:
+            if test_job.skip_build or test_job.manual_trigger:
                 continue
             check_names = self.test_check_names_for_test_job(test_job)
             self._report(
@@ -696,8 +696,10 @@ class BaseBuildJobHelper(BaseJobHelper):
         update_feedback_time: Callable = None,
     ) -> None:
         for test_job in self.job_tests_all:
-            if not test_job.skip_build and chroot in self.build_targets_for_test_job(
-                test_job
+            if (
+                not test_job.skip_build
+                and not test_job.manual_trigger
+                and chroot in self.build_targets_for_test_job(test_job)
             ):
                 test_targets = self.build_target2test_targets_for_test_job(
                     chroot, test_job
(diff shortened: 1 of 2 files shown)


packit_service/worker/helpers/build/build_helper.py
```diff
@@ -652,6 +652,9 @@ def report_status_to_all_test_jobs(
         for test_job in self.job_tests_all:
             if test_job.skip_build:
                 continue
+            elif test_job.manual_trigger:
+                if state not in (BaseCommitStatus.error, BaseCommitStatus.failure):
```
majamassarini, line 656: I thought that showing errors can make it easier for a user to understand why tests are not running, but then I realized that when the copr build will work the failed manual test check will not be overwritten. And this is misleading. So it is probably better never to show these check statuses at all. WDYT?

lbarcziova: I would rather be consistent and really hide all the statuses

lbarcziova (commented):
otherwise looks good :+1:

lbarcziova (approved):
thanks!

PR #2380, opened 2024-03-22
Do not show taks accepted status for manual tests

majamassarini:
Related with commit ce2a7882cc7087cbd2ec58225535e0b09c3d1070

### DIFF
--- a/packit_service/worker/helpers/testing_farm.py
+++ b/packit_service/worker/helpers/testing_farm.py
@@ -1298,6 +1298,42 @@ class TestingFarmJobHelper(CoprBuildJobHelper):
         """
         return self.test_target2build_target_for_test_job(test_target, self.job_config)
 
+    def is_manual_job_being_triggered_by_label(self) -> bool:
+        """
+        Is this job meant to be triggered manually
+        and is it being triggered by a matching
+          /packit test --labels label
+        command?
+        """
+        if (
+            self.job_config.manual_trigger
+            and self.job_config.labels
+            and self.comment_arguments.labels
+        ):
+            return (
+                True
+                if set(self.job_config.labels).intersection(
+                    set(self.comment_arguments.labels)
+                )
+                else False
+            )
+        return False
+
+    def is_manual_job_being_triggered_by_identifier(self) -> bool:
+        """
+        Is this job meant to be triggered manually
+        and is it being triggered by a matching
+          /packit test --identifier job-identifier
+        command?
+        """
+        if (
+            self.job_config.manual_trigger
+            and self.job_config.identifier
+            and self.comment_arguments.identifier
+        ):
+            return self.job_config.identifier == self.comment_arguments.identifier
+        return False
+
     @property
     def build_targets_for_tests(self) -> Set[str]:
         """
@@ -1316,6 +1352,12 @@ class TestingFarmJobHelper(CoprBuildJobHelper):
         links_to_external_services: Optional[Dict[str, str]] = None,
         update_feedback_time: Callable = None,
     ) -> None:
+        if self.job_config.manual_trigger and not (
+            self.is_manual_job_being_triggered_by_label()
+            or self.is_manual_job_being_triggered_by_identifier()
+        ):
+            return
+
         if chroot in self.build_targets_for_tests:
             test_targets = self.build_target2test_targets_for_test_job(
                 chroot, self.job_config
@@ -1341,6 +1383,12 @@ class TestingFarmJobHelper(CoprBuildJobHelper):
         links_to_external_services: Optional[Dict[str, str]] = None,
         update_feedback_time: Callable = None,
     ) -> None:
+        if self.job_config.manual_trigger and not (
+            self.is_manual_job_being_triggered_by_label()
+            or self.is_manual_job_being_triggered_by_identifier()
+        ):
+            return
+
         if target in self.tests_targets:
             self._report(
                 description=description,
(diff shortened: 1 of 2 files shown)


nforro:
Wouldn't it be better to have the check in `SteveJobs.report_task_accepted()` and not call `job_helper.report_status_to_configured_job()` at all if not wanted, rather than relying on what's in the description?

Also, there is a typo in the commit message :slightly_smiling_face:

majamassarini:
> Wouldn't it be better to have the check in `SteveJobs.report_task_accepted()` and not call `job_helper.report_status_to_configured_job()` at all if not wanted, rather than relying on what's in the description?

I need to skip this only for the test jobs which have the manual trigger set to True. I would need a lot of code to extract this information in `SteveJobs` class. This is a sort of "ugly hack" in the *testing farm helper* only. 
And, I don't want to skip all the tests statuses but just some of them (thus I am using the description to be sure to skip only the wanted statuses updates).

I feel like both this commit and the previous one are just an ugly workaround because we don't have real dependencies between tests and builds.

I am hiding useful information to the user and I have to hide one more, you will see a new commit (and maybe more after I have tested all the retriggering combinations).

I don't really have other solutions, I think. I tried, instead of hiding this statuses, to make them neutral but in my local instance they are falling back with a failure status because there is no neutral state for commit statuses (as I read it in our code) and also @lbarcziova said that the neutral state is not supported in gitlab.

The only solution I see now, until we don't have dependencies between jobs, is to hide all those pending statuses related to a manual triggered test.

> Also, there is a typo in the commit message 🙂
 Thanks!

nforro:
> I need to skip this only for the test jobs which have the manual trigger set to True. I would need a lot of code to extract this information in `SteveJobs` class. This is a sort of "ugly hack" in the _testing farm helper_ only.
> And, I don't want to skip all the tests statuses but just some of them (thus I am using the description to be sure to skip only the wanted statuses updates).

Oh, I see, you would have to differentiate between types of jobs. Let's keep it like this then.

nforro:
> I am hiding useful information to the user and I have to hide one more, you will see a new commit (and maybe more after I have tested all the retriggering combinations).

So is this not yet ready for review or will you do that in separate PRs?

majamassarini:
> > I am hiding useful information to the user and I have to hide one more, you will see a new commit (and maybe more after I have tested all the retriggering combinations).
> 
> So is this not yet ready for review or will you do that in separate PRs?

I already added the new commit. I hope I will no need more exceptions... I will try more retriggering combinations locally to see if I found out something more. In case I need to add more exceptions I will ping you  🙂🙏🏻

packit_service/worker/helpers/testing_farm.py
```diff
@@ -1380,6 +1387,12 @@ def report_status_to_configured_job(
         links_to_external_services: Optional[Dict[str, str]] = None,
         update_feedback_time: Callable = None,
     ):
+        if description == TASK_ACCEPTED and self.job_config.manual_trigger:
```
lbarcziova, line 1390: will this work in case the test is triggered manually (in that case we want to show that status)?

majamassarini: No, this is always hidden. I thought about trying to show it when you really triggered the test manually. But I think this will require some code (as putting this information into the database) and I don't know if it is worth it.

I would prefer to work on adding dependencies to our code instead of trying to improve this workaround. So that we could show all the messages that now I am hiding at the proper moment.

packit_service/worker/helpers/testing_farm.py
```diff
@@ -1298,6 +1298,22 @@ def test_target2build_target(self, test_target: str) -> str:
...
+        command?
+        """
+        if self.job_config.manual_trigger:
+            if self.comment_arguments.labels:
+                return (
+                    True
+                    if self.job_config.identifier in self.comment_arguments.labels
+                    else False
+                )
+        return False
```
lbarcziova, line 1315:
```suggestion
        if self.job_config.manual_trigger and self.comment_arguments.labels:
                return self.job_config.identifier in self.comment_arguments.labels
        return False
```

packit_service/worker/helpers/testing_farm.py
```diff
@@ -1298,6 +1298,22 @@ def test_target2build_target(self, test_target: str) -> str:
         """
         return self.test_target2build_target_for_test_job(test_target, self.job_config)
 
+    def is_manual_job_being_triggered(self) -> bool:
```
lbarcziova, line 1301: could we have some more explicit name of the method mentioning the labels?

packit_service/worker/helpers/testing_farm.py
lbarcziova: I think we should check here both identifier and labels? Something like we have in pre-checks [here](https://github.com/packit/packit-service/blob/9bbb1d743d02166a97070c4bfd9e7cf647c1eefe/packit_service/worker/checker/testing_farm.py#L165) and [here](https://github.com/packit/packit-service/blob/9bbb1d743d02166a97070c4bfd9e7cf647c1eefe/packit_service/worker/checker/testing_farm.py#L132)?

packit_service/worker/helpers/testing_farm.py
```diff
@@ -1298,6 +1298,42 @@ def test_target2build_target(self, test_target: str) -> str:
...
+            and self.job_config.labels
+            and self.comment_arguments.labels
+        ):
+            return (
+                True
+                if set(self.job_config.labels).intersection(
+                    set(self.comment_arguments.labels)
+                )
+                else False
+            )
```
lbarcziova, line 1319:
```suggestion
            return bool(set(self.job_config.labels).intersection(
                    set(self.comment_arguments.labels)
                ))
```

lbarcziova (approved):
lgtm :+1:

PR #2389, opened 2024-04-04
Hide manual test statuses only when build handler is required

### LINKED ISSUE #2386: Missing reporting for manual tests
In this PR https://github.com/rpm-software-management/ci-dnf-stack/pull/1475, tests were triggered by a comment and Packit did not report anything to user, even though the tests were actually submitted.

Configuration: https://github.com/rpm-software-management/ci-dnf-stack/blob/main/.packit.yml

lbarcziova:
We should not check the labels and identifiers, this should be handled by pre-checks and results in no reporting for TF notifications.

Fixes #2386

TODO:
- [ ] testcase

RELEASE NOTES BEGIN

We have fixed reporting of tests with `manual_trigger: True`.

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/handlers/testing_farm.py
+++ b/packit_service/worker/handlers/testing_farm.py
@@ -260,13 +260,14 @@ class TestingFarmHandler(
                     "The latest build has not finished yet, "
                     "waiting until it finishes before running tests for it."
                 )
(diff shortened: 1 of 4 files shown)


majamassarini (approved):
I have the feeling that something is missing but this can be tested on stg.
Just one note about the testing, I found out that many corner cases where related with no having a build yet.

PR #2391, opened 2024-04-04
Fix handling of not finished build and manual test

lbarcziova:
Set neutral status for this case and also adjust the message.

RELEASE NOTES BEGIN

N/A

RELEASE NOTES END


majamassarini (approved):
🙏🏻🤞🏻

PR #2394, opened 2024-04-09
Store package config in DB project event

### LINKED ISSUE #1940: Load the package config once per pipeline run
In order to reduce the number of GitHub API calls made from the service, get the package configuration from GitHub when the pipeline is started, save it to the data base, and access it from there in subsequent pipeline steps.

See packit/research#171 for more details.

lbarcziova:
So that it doesn't have to be obtained via Github API multiple times during the pipeline (e.g. Copr build start, end, Testing farm results)

Fixes #1940

TODO:

- [ ] think about cleanup in DB (periodic/pipeline end)
- [ ] make sure that now that `Event.get_dict()` calls `get_packages_config()`, it doesn't cause obtaining package config in case it is not really needed (e.g. maybe in Event.__str__)
- [ ] test

RELEASE NOTES BEGIN

N/A

RELEASE NOTES END

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -369,6 +369,11 @@ class ProjectAndEventsConnector:
         if project_event_model:
             project_event_model.commit_sha = value
 
+    @property
+    def packages_config(self) -> Optional[dict]:
+        project_event_model = self.get_project_event_model()
+        return project_event_model.packages_config if project_event_model else None
+
     def get_pr_id(self) -> Optional[int]:
         project_event_object = self.get_project_event_object()
         if isinstance(project_event_object, PullRequestModel):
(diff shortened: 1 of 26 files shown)


mfocko:
I haven't had a look at the code yet, but I hope this happens »before« the backward-compatibility “hacks” run… cause the way we “inherit” parts of config makes them rather big (you can see that in the endless configs in logs)

lbarcziova:
I have just realised this might be even more complicated as I noticed the way we dump the config [here](https://github.com/packit/packit-service/blob/101169072e0c6dfc80a43d821b2c767bd53d3392/packit_service/worker/handlers/abstract.py#L357) is different than the one I added, so this is definitely not ready for review yet

PR #2397, opened 2024-04-11
Store package config in DB and use it in result events

### LINKED ISSUE #1940: Load the package config once per pipeline run

lbarcziova:
TODO:
- [x] test after Packit PR gets merged

Related to

Merge after packit/packit#2278
Fixes #1940 

RELEASE NOTES BEGIN

N/A

RELEASE NOTES END


packit_service/worker/events/event.py
```diff
@@ -320,14 +319,25 @@ def get_non_serializable_attributes(self):
...
         # whole dict has to be JSON serializable because of redis
         d = self.make_serializable(d, self.get_non_serializable_attributes())
         d["event_type"] = self.__class__.__name__
 
-        # we are trying to be lazy => don't touch database if it is not needed
-        d["event_id"] = self._db_project_object.id if self._db_project_object else None
+        if store_project_event:
+            package_config_dict = self.packages_config.get_raw_dict_with_defaults()
+            db_project_object = self.db_project_object
+            if self.db_project_event and package_config_dict:
```
majamassarini, line 333: Just not sure what could happen here if `package_config_dict` is None

lbarcziova: the package config will not be stored to DB, which shouldn't be a problem as the "old" solution (getting the config via API) is still a fallback for the result events

majamassarini (approved):

lbarcziova:
I have made few changes, PTAL :pray:

PR #2399, opened 2024-04-15
Don't trigger builds from TestingFarmHandler

### LINKED ISSUE #2381: /packit test in for a missing build scenario triggering builds for each test job in config
Steps to reproduce:
1.  multiple test jobs in config
2. no successful Copr builds 
3. `/packit test` comment

This results in submitting Copr build for each test job separately, resulting in flooding of both Copr and later Testing Farm (once the builds finish and tests are not configured to be run only manually).

This happened for https://github.com/oamg/convert2rhel/pull/1149#issuecomment-2012747884, see multiple [Copr](https://copr.fedorainfracloud.org/coprs/g/oamg/convert2rhel/builds/) builds for that PR+ref and for TF @thrix reported 367 test requests for that PR+ref submitted in a very short period of time.

lbarcziova:
This changes to behaviour of /packit test: in case there is missing build for some target, the build will not be triggered anymore, it will just be reported to the user.

Previous code was prone to the race condition if there were multiple test jobs in configs, as there were scheduled multiple TestingFarmHandler runs where each could trigger Copr build jobs. This potentially meant wasting of lot of resources.

Fixes #2381 

TODO:

- [x] agree on which solution we want to go with (this PR or #2400 )

RELEASE NOTES BEGIN

We have changed the behaviour of `/packit test` comment command: in case there is a missing build for some target, the build will not be triggered anymore, it will just be reported to the user. We needed to make this change as with the increased complexity of the configuration (multiple test jobs), the previous implementation was prone to race conditions leading to wasting of resources of Copr and Testing Farm.

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/handlers/testing_farm.py
+++ b/packit_service/worker/handlers/testing_farm.py
@@ -9,7 +9,6 @@ from datetime import datetime, timezone
 from typing import Optional, Dict, List, Tuple, Type
 
 from celery import Task
-from celery import signature
 
 from packit.config import JobConfig, JobType
 from packit.config.package_config import PackageConfig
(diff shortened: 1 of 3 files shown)


lbarcziova:
@majamassarini totally agreed, thanks for the feedback!

PR #2401, opened 2024-04-15
Refactor calling GitProject.get_pr()

lbarcziova:
Try to use the pull_request_object properties where possible.

RELEASE NOTES BEGIN

N/A

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/helpers/testing_farm.py
+++ b/packit_service/worker/helpers/testing_farm.py
@@ -8,7 +8,7 @@ from typing import Dict, Any, Optional, Set, List, Union, Tuple, Callable
 
 import requests
 
-from ogr.abstract import GitProject, PullRequest
+from ogr.abstract import GitProject
 from ogr.utils import RequestResponse
 from packit.config import JobConfig, PackageConfig
 from packit.exceptions import PackitConfigException, PackitException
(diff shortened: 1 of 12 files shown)


lachmanfrantisek (approved):
Thanks!

packit_service/worker/helpers/testing_farm.py
```diff
@@ -171,12 +171,7 @@ def custom_fmf(self) -> bool:
     def fmf_url(self) -> str:
         return (
             self.job_config.fmf_url
-            or (
-                self.metadata.pr_id
-                and self.project.get_pr(
-                    self.metadata.pr_id
-                ).source_project.get_web_url()
-            )
+            or (self.pr_id and self.pull_request_object.source_project.get_web_url())
```
majamassarini, line 174: This is a bit strange since `self.pr_id` doesn't seem to be used in the second clause anymore. Can we remove it now?

nforro:
Won't this cause issues, for example if the target branch of a PR is changed, will it be reflected in `StatusReporter.pull_request_object.target_branch`?

lbarcziova:
> Won't this cause issues, for example if the target branch of a PR is changed, will it be reflected in StatusReporter.pull_request_object.target_branch?

I don't think it should cause issues, `pull_request_object` will be initialised per an instance of `StatusReporter`, so in each handler run this is obtained via API again. Or am I missing something from your question?

nforro (approved):

PR #2402, opened 2024-04-15
Feat/configurable check name

mfocko:
TODO:

- [ ] [FIXME]: retriggering via GitHub re-run BROKEN

RELEASE NOTES BEGIN

We have introduced a new `status_name_template` option that allows you to configure status name for a Packit job. For further details have a look at our [docs](https://packit.dev/docs/configuration#status_name_template). This feature is still experimental and it is not possible to retry those jobs via GitHub Checks' re-run at the moment.

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/helpers/build/build_helper.py
+++ b/packit_service/worker/helpers/build/build_helper.py
@@ -412,9 +412,16 @@ class BaseBuildJobHelper(BaseJobHelper):
         chroot: str = None,
         project_event_identifier: Optional[str] = None,
         identifier: Optional[str] = None,
+        package: Optional[str] = None,
+        template: Optional[str] = None,
     ):
         return cls.get_check_cls(
-            cls.status_name_build, chroot, project_event_identifier, identifier
+            cls.status_name_build,
+            chroot,
+            project_event_identifier,
+            identifier,
+            package=package,
+            template=template,
         )
 
     @classmethod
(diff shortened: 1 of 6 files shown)


nforro:
I wonder if it wouldn't be better/safer to use [template strings](https://docs.python.org/3/library/string.html#template-strings), what do you think?

packit_service/worker/helpers/build/build_helper.py
```diff
@@ -385,14 +385,24 @@ def get_check_cls(
...
         project_event_identifier: Optional[str] = None,
         identifier: Optional[str] = None,
+        package: Optional[str] = None,
+        template: Optional[str] = None,
     ):
+        if project_event_identifier:
+            project_event_identifier = project_event_identifier.replace(":", "-")
+
+        if template is not None:
+            return template.format(
```
lbarcziova, line 395: should we wrap this to `try-except` for the case of misconfigured template?

lbarcziova (commented):
nicely done!

packit_service/worker/helpers/build/build_helper.py
mfocko: that's a good idea :+1:

packit_service/worker/helpers/build/build_helper.py
```diff
@@ -385,14 +385,24 @@ def get_check_cls(
...
+                job_name=job_name,
+                chroot=chroot,
```
lachmanfrantisek, line 397: Can we also include a compose name for tests?

lachmanfrantisek (approved):
Very nice!

packit_service/worker/helpers/build/build_helper.py
mfocko: I have a feeling that chroot is the compose for testing farm :eyes:

Issue #2195, opened 2023-08-30
Dead URL in dashboard when Packit fails to complete first COPR step


mulkieran:
### What happened? What is the problem?

When a Packit copr build fails during packit's actions, the "Build Logs URL" link in the HTML is not actually a link.

We find an example here: https://dashboard.packit.dev/results/copr-builds/966457 .

Hover over "Build Logs URL" with cursor, and no link shows in your browser window. Press cursor, and nothing happens.

Additional information: As I did not know for a while, the logs are still accessible by clicking on the "Web URL" link and then the "builder-live.log.gz" link . So there is a workaround. But it was confusing and a bit frustrating last week to not be able to see the links to the logs of the build failures (as I thought).

### What did you expect to happen?

I expected a link and to be able to view the logs pointed to by the link.

### Example URL(s)

https://dashboard.packit.dev/results/copr-builds/966457

### Steps to reproduce

```bash
1. Fail a packit action at a particular place.
2. View dashboard for the job.
3. Note that the "Build Logs URL" link text has no target.
```

### Workaround

- [X] There is an existing workaround that can be used until this issue is fixed.

### Participation

- [ ] I am willing to submit a pull request for this issue. (Packit team is happy to help!)

PR #2406, opened 2024-04-22
Report proper status in case RPM build failure is a result of SRPM build failure

nforro:
Related to https://github.com/packit/packit-service/issues/2195.

This fixes two cases:

- babysit task updates build statuses in the DB but also overwrites status checks with a wrong message and link
- `CoprBuildEndEvent` for SRPM is never received and SRPM build status in the DB stays in `pending` forever

RELEASE NOTES BEGIN

We have fixed an issue when `copr_build` job status checks were sometimes wrongly updated with a misleading message after a SRPM build failure.

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/handlers/copr.py
+++ b/packit_service/worker/handlers/copr.py
@@ -321,18 +321,20 @@ class CoprBuildEndHandler(AbstractCoprBuildReportHandler):
         if self.copr_event.status != COPR_API_SUCC_STATE:
             failed_msg = "RPMs failed to be built."
             packit_dashboard_url = get_copr_build_info_url(self.build.id)
-            self.copr_build_helper.report_status_to_all_for_chroot(
-                state=BaseCommitStatus.failure,
-                description=failed_msg,
-                url=packit_dashboard_url,
-                chroot=self.copr_event.chroot,
-            )
-            self.measure_time_after_reporting()
-            self.copr_build_helper.notify_about_failure_if_configured(
-                packit_dashboard_url=packit_dashboard_url,
-                external_dashboard_url=self.build.web_url,
-                logs_url=self.build.build_logs_url,
-            )
+            # if SRPM build failed it has been reported already so skip reporting
+            if self.build.get_srpm_build().status != BuildStatus.failure:
+                self.copr_build_helper.report_status_to_all_for_chroot(
+                    state=BaseCommitStatus.failure,
+                    description=failed_msg,
+                    url=packit_dashboard_url,
+                    chroot=self.copr_event.chroot,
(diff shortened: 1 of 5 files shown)


packit_service/worker/helpers/build/babysit.py
```diff
@@ -222,6 +225,19 @@ def update_copr_builds(build_id: int, builds: Iterable["CoprBuildTargetModel"])
 
     logger.info(f"The status of {build_id} is {build_copr.state!r}.")
 
+    if (
+        srpm_build := SRPMBuildModel.get_by_copr_build_id(build_id)
+    ) and srpm_build.status == BuildStatus.pending:
+        build_copr_srpm = copr_client.build_proxy.get_source_chroot(build_id)
```
lbarcziova, line 231: shouldn't this be wrapped in the try-except as well?

packit_service/worker/handlers/copr.py
```diff
@@ -319,8 +319,15 @@ def run(self):
 
         # https://pagure.io/copr/copr/blob/master/f/common/copr_common/enums.py#_42
         if self.copr_event.status != COPR_API_SUCC_STATE:
-            failed_msg = "RPMs failed to be built."
-            packit_dashboard_url = get_copr_build_info_url(self.build.id)
+            if (
+                srpm_build := self.build.get_srpm_build()
+            ) and srpm_build.status == BuildStatus.failure:
```
lbarcziova, line 324: if this is `True`, hasn't the message and URL been reported to the user previously?

packit_service/worker/helpers/build/babysit.py
nforro: It probably should, but in that case I should do it also here, right?
https://github.com/packit/packit-service/blob/710e6942e811711b081915684a828fd95f98fcf4/packit_service/worker/helpers/build/babysit.py#L242

lbarcziova: I would say so

packit_service/worker/handlers/copr.py
lbarcziova: sorry for not elaborating more, I meant whether we cannot just log it and return here, so that the status is not updated another time

nforro: Good point, let me change it.

packit_service/worker/helpers/build/babysit.py
nforro: Does it make sense like this?

lbarcziova (approved):
thanks for the improvements!

PR #2410, opened 2024-04-29
chore: add first codeowners

mfocko:


lbarcziova (approved):
thanks!

(can we fix the error?)

lbarcziova:
@mfocko can we merge this like this or should we resolve the error?

Venefilyn:
If the group exists it likely is missing write access to this repo

Venefilyn (approved):

PR #2413, opened 2024-04-30
Fix flask restarts

majamassarini:
Make usage function calls cacheable.
Collect usage statistics in a long running celery worker.

### DIFF
--- a/packit_service/celerizer.py
+++ b/packit_service/celerizer.py
@@ -20,10 +20,12 @@ class Celerizer:
             password = getenv("REDIS_PASSWORD", "")
             port = getenv("REDIS_SERVICE_PORT", "6379")
             db = getenv("REDIS_SERVICE_DB", "0")
+            celery_beckend = getenv("REDIS_CELERY_BECKEND", "1")
             broker_url = f"redis://:{password}@{host}:{port}/{db}"
+            backend_url = f"redis://:{password}@{host}:{port}/{celery_beckend}"
 
             # http://docs.celeryq.dev/en/stable/reference/celery.html#celery.Celery
-            self._celery_app = Celery(broker=broker_url)
+            self._celery_app = Celery(backend=backend_url, broker=broker_url)
 
             # https://docs.celeryq.dev/en/stable/getting-started/first-steps-with-celery.html#configuration
             self._celery_app.config_from_object("packit_service.celery_config")
--- a/packit_service/celery_config.py
+++ b/packit_service/celery_config.py
@@ -40,6 +40,11 @@ beat_schedule = {
         "schedule": crontab(minute=0, hour=2),  # nightly at 2AM
         "options": {"queue": "long-running"},
     },
+    "get_usage_statistics": {
+        "task": "packit_service.worker.tasks.get_usage_statistics",
+        "schedule": crontab(minute=1, hour=0),  # nightly at 0AM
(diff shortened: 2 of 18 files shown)


majamassarini:
recheck

packit_service/celerizer.py
```diff
@@ -20,10 +20,12 @@ def celery_app(self):
             password = getenv("REDIS_PASSWORD", "")
             port = getenv("REDIS_SERVICE_PORT", "6379")
             db = getenv("REDIS_SERVICE_DB", "0")
+            celery_beckend = getenv("REDIS_CELERY_BECKEND", "1")
```
lbarcziova, line 23: what is this value exactly? :thinking: I was confused seeing the default is `1` after reading the name

```suggestion
            celery_backend = getenv("REDIS_CELERY_BECKEND", "1")
```

packit_service/celery_config.py
```diff
@@ -40,6 +43,11 @@
         "schedule": crontab(minute=0, hour=2),  # nightly at 2AM
         "options": {"queue": "long-running"},
     },
+    "get_usage_statistics": {
+        "task": "packit_service.worker.tasks.get_usage_statistics",
+        "schedule": crontab(minute=1, hour=0),  # nightly at 0AM
```
lbarcziova, line 48: :+1:

packit_service/service/api/usage.py
lbarcziova, line 1: I just quickly went through the code and am a little confused, how will now the data be exposed?

packit_service/worker/tasks.py
```diff
@@ -75,6 +84,8 @@
 from packit_service.worker.jobs import SteveJobs
 from packit_service.worker.result import TaskResults
 
+_CACHE_MAXSIZE = 100
```
lbarcziova, line 87: I don't see this to be used anywhere

packit_service/celerizer.py
majamassarini: This is a new redis db, to be used as celery backend. The 0 db is  already taken by "Packit", so I am using the next available.

packit_service/service/api/usage.py
majamassarini: I haven't created new API endpoints. 
I have only fixed the caching of the `get_usage_data` calls. 
When someone calls the usage endpoints, now the computation is made in a Celery task. 
Since Celery computes the usage statistics once a day every night, and since the calls are cached, when you ask for the usage endpoint the response is almost immediate. 
If the nightly task is not run yet, for any reason, the first time you call the usage endpoint it will timeout (but in a "graceful way" without crashing Falsk - this can be improved using React, making the client polling on the result page, but I have not done it yet, since I thought this can be a later improvement if we really want to do it, I hope it will not be necessary since the caching should make the calls almost immediate all the day long...)

packit_service/worker/tasks.py
majamassarini: It was an old value, I don't know why `black` didn't complain. I will try to remove it.

packit_service/service/api/usage.py
lbarcziova: ok, sorry, I am blind :D I thought the api/usage.py is completely removed, therefore the confusion, but it was just the GitHub UI not rendering it because of a bigger diff...but still, thanks for the explanation!

lbarcziova (approved):
thanks!

PR #2414, opened 2024-04-30
Handle VM image builds in /api/runs

nforro:


packit_service/service/api/runs.py
```diff
@@ -126,6 +128,24 @@ def process_runs(runs):
...
+                    "packit_id": vm_image_build_id,
+                    "target": vm_image_build.target,
+                    "status": vm_image_build.status,
+                }
+            )
+            if "trigger" not in response_dict:
+                response_dict["time_submitted"] = optional_timestamp(
+                    vm_image_build.build_submitted_time
+                )
+                response_dict["trigger"] = get_project_info_from_build(vm_image_build)
```
lbarcziova, line 147: could we also have this in a separate method?

lbarcziova (approved):
thanks for looking into this!

packit_service/service/api/runs.py
nforro: Ok, let me move it.

PR #2419, opened 2024-05-06
Store results only of usage data tasks

nforro:
Storing results of all tasks is unnecessary and requires all result data to be serializable. Ignore results of all tasks by default (they will still be visible in the logs) and only store them for usage data tasks.


mfocko (approved):
Is that why we ran out of memory in Redis at 4pm? :smile:

nforro:
That would be my guess :slightly_smiling_face:

mfocko:
> That would be my guess 🙂

Well, at least I don't have to open a PR for the resource changes then :D

majamassarini (approved):
Thanks a lot!

PR #2420, opened 2024-05-07
Improvements for showing onboarded projects

majamassarini:
- Increase the timeout when collecting usage statistics
- Cache the result of get_onboarded_projects
- Collect statistics every hour -> should help with caching data in all the pods where long tasks are run and also should help with the pods restarts. But cpu usage should stay low since the caching.

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -972,7 +972,7 @@ class GitProjectModel(Base):
     @classmethod
     def get_known_onboarded_downstream_projects(
         cls,
-    ) -> set["GitProjectModel"]:
+    ) -> Set["GitProjectModel"]:
         """
         List already known onboarded projects.
         An onboarded project is a project with a bodhi update or a koji build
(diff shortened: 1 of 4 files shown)


packit_service/models.py
```diff
@@ -3787,3 +3787,45 @@ def get_usage_data(datetime_from=None, datetime_to=None, top=10) -> dict:
         },
         jobs=jobs,
     )
+
+
+@cached(cache=TTLCache(maxsize=1, ttl=(60 * 60 * 24)))
+def get_onboarded_projects() -> Tuple[dict[int, str], dict[int, str]]:
+    """Returns a tuple with two numbers:
```
lbarcziova, line 3794:
```suggestion
    """Returns a tuple with two dictionaries of project IDs and URLs:
```

lbarcziova (approved):
thanks!

PR #2421, opened 2024-05-07
Consider default job timeout for VM image build babysits

lbarcziova:


majamassarini (approved):
Thanks, I think this is useful. 
But do you think this will solve exceptions in prod? I read `Compose with given id not found` returned by the image builder service api, so I would have expected another kind of "code". Should I give it a look tomorrow?

lbarcziova:
> But do you think this will solve exceptions in prod? I read Compose with given id not found returned by the image builder service api, so I would have expected another kind of "code". Should I give it a look tomorrow?

I checked which builds are causing these errors and there were only 2 which keep triggering this exception and they are both older than week, so I assume it could be really that those are already deleted from image builder or something like that.

PR #2424, opened 2024-05-10
Add downstream package name to Sentry tags

nforro:
With the rise of pull-from-upstream usage it can be useful to have related Sentry events tagged with downstream package name (which can often be different from the upstream repository name).


lbarcziova (approved):
nice!

PR #2442, opened 2024-06-10
fix(news): remove warning about changelog defaults

mfocko:
As some time has already passed since we changed the defaults and there don't seem to be any affected users, revert this change as it is no longer relevant.


mfocko:
pre-commit CI passed… pre-commit in Zuul failed… :sigh:

lbarcziova (approved):
thanks!

mfocko:
looks like a flake of pre-commit in Zuul

PR #2450, opened 2024-06-24
chore: Enable HTTP/2 protocol for packit-service

Venefilyn:
With this change we should now have mod_http2 installed and enabled.

As I haven't figured out how to run packit-service for development purposes this is just a guess on what needs to be done.

RELEASE NOTES BEGIN
Enable HTTP/2 for Packit service
RELEASE NOTES END


lbarcziova (approved):
with docker-compose, the curl output looked ok, let's try on staging

Issue #2378, opened 2024-03-21
Implement `koji_build` trigger for `koji_build` job


nforro:
TODO:

- [x] start listening to `org.fedoraproject.prod.buildsys.tag` events
- [x] implement `koji_build` trigger:
  - check if the event's tag matches a sidetag in the `sidetags` table
  - if it does, get list of packages tagged into the sidetag (`koji list-tagged $SIDETAG`)
  - for every package/project get Packit config and find any jobs with `koji_build` trigger and `sidetag_group` matching the sidetag (if this proves impractical, we can resort to having a global configuration for all packages sharing a sidetag)
  - for every job found check if their `tagged_packages` are a subset of packages tagged into the sidetag
  - if yes, trigger the build (from the HEAD commit of the corresponding dist-git branch)

nforro:
Here is an example `org.fedoraproject.prod.buildsys.tag` payload:

```json
{
  "body": {
    "base_url": "https://koji.fedoraproject.org",
    "build_id": 2424213,
    "instance": "primary",
    "name": "python-smart_open",
    "owner": "music",
    "release": "1.fc40",
    "tag": "f40-build-side-85541",
    "tag_id": 85541,
    "user": "music",
    "version": "7.0.3"
  },
  "headers": {
    "fedora_messaging_rpm_python-smart_open": true,
    "fedora_messaging_schema": "koji_fedoramessaging.tag.TagV1",
    "fedora_messaging_severity": 10,
    "fedora_messaging_user_music": true,
    "priority": 2,
    "sent-at": "2024-03-21T15:17:43+00:00"
  },
  "id": "57d89d96-8eac-4f9f-b787-053cd2532305",
  "queue": null,
  "topic": "org.fedoraproject.prod.buildsys.tag"
}
```

PR #2457, opened 2024-07-08
Implement `koji_build` trigger for `koji_build` job

nforro:
Fixes https://github.com/packit/packit-service/issues/2378.

Merge after https://github.com/packit/packit/pull/2349.
Merge after https://github.com/packit/packit/pull/2350.


alembic/versions/e05e1b04de87_add_koji_build_tags.py
```diff
@@ -0,0 +1,61 @@
...
+        ["koji_tag_name"],
+        unique=False,
+    )
+    op.create_index(
+        op.f("ix_koji_build_tags_project_id"),
+        "koji_build_tags",
+        ["project_id"],
+        unique=False,
+    )
+    op.execute("ALTER TYPE projecteventtype ADD VALUE 'koji_build_tag'")
```
nforro, line 49: There is no easy way to remove a value from an enum type, so there is no corresponding reverse operation in `downgrade()`. Could that be a problem?

nforro:
recheck

nforro:
recheck

nforro:
recheck

nforro:
recheck

alembic/versions/e05e1b04de87_add_koji_build_tags.py
lbarcziova: I don't think so, I think in the past we did the same for other migrations

nforro:
recheck

packit_service/worker/handlers/koji.py
```diff
@@ -354,3 +366,79 @@ def run(self):
...
+        packages_to_trigger = set()
+        for package_name in tagged_packages:
+            distgit_project_url = f"{dg_base_url}rpms/{package_name}"
+            project = self.service_config.get_project(url=distgit_project_url)
+            packages_config = PackageConfigGetter.get_package_config_from_repo(
+                base_project=None,
+                project=project,
+                pr_id=None,
+                reference=None,
+                fail_when_missing=True,
```
lbarcziova, line 407: do we want to have this `True`? maybe we should handle this case in a different way?

lbarcziova (commented):
very nice work!

packit_service/worker/handlers/koji.py
nforro: Right, I didn't realize it's possible for builds of packages without Packit config to be tagged into the sidetag, so in this case skipping the package would be certainly better, same below. Let me change that.

lbarcziova (approved):
LGTM!

PR #2464, opened 2024-07-12
Don't stop babysitting TF runs if there is issue with getting info from API

lbarcziova:
This may be likely an API outage and we should continue babysitting the run (we have the timeouts, so in case this is some other issue, we will eventually stop).

See discussion with @thrix in https://redhat-internal.slack.com/archives/C04MU29TES1/p1720786823958689.

RELEASE NOTES BEGIN

We have improved the babysitting of Testing Farm runs to better handle API outages.

RELEASE NOTES END


lbarcziova:
recheck

lbarcziova:
recheck

nforro (approved):

Issue #2379, opened 2024-03-21
Implement `koji_build` trigger for `bodhi_update` job


nforro:
TODO:

- [x] implement triggerring Bodhi updates in the same way as Koji builds
- [ ] after a successful update, remove the corresponding mapping from the sidetag mapping table ~~and delete the corresponding sidetag in Koji~~ (Bodhi does this automatically)

Issue #2412, opened 2024-04-30
Implement periodic cleanup of unused sidetags


nforro:
In the ideal case, a Packit-created sidetag should be removed by Bodhi after submitting an update from it. However, that doesn't necessarily have to happen for various reasons and we don't want useless sidetags to be piling up and wasting resources.

TODO:

- [ ] in periodic intervals go through all side tags in `sidetags` table
- [ ] if a sidetag doesn't contain any builds (`koji list-tagged`) or if it was created more than N weeks ago, remove it both from Koji (`KojiHelper.remove_sidetag()`) and from the `sidetags` table
- [ ] document this

nforro:
This turns out to be most likely unnecessary. Bodhi removes sidetags updates have been created from, and even stray (created but never actually used) sidetags seem to be auto-cleaned after some time (I haven't found any documentation on this though).

lbarcziova:
Based on Nikola's previous comment, we don't need to implement this.

PR #2475, opened 2024-07-26
Implement `koji_build` trigger for `bodhi_update` job

nforro:
Fixes https://github.com/packit/packit-service/issues/2379.

I purposefully didn't implement removing sidetags from DB after a successful update creation because I want to see first how exactly the Bodhi workflow works - at which point is the sidetag removed from Koji etc. Maybe it would be enough to leave it to the periodic cleanup: https://github.com/packit/packit-service/issues/2412


packit_service/service/api/bodhi_updates.py
```diff
@@ -77,7 +77,7 @@ def get(self, id):
             "status": update.status,
             "branch": update.target,
             "web_url": update.web_url,
-            "koji_nvr": update.koji_nvr,
+            "koji_nvrs": update.koji_nvrs,
```
lbarcziova, line 80: this should be probably updated in dashboard code as well

packit_service/worker/handlers/koji.py
```diff
@@ -436,13 +439,23 @@ def run(self) -> TaskResults:
                 continue
             for job in packages_config.get_job_views():
                 if (
-                    job.type == JobType.koji_build
+                    job.type in [JobType.koji_build, JobType.bodhi_update]
```
lbarcziova, line 442:
```suggestion
                    job.type in (JobType.koji_build, JobType.bodhi_update)
```

lbarcziova (approved):
LGTM, let's test this! 

> Maybe it would be enough to leave it to the periodic cleanup: https://github.com/packit/packit-service/issues/2412

:+1: let's see the behaviour first

nforro:
recheck

PR #2477, opened 2024-07-26
Parse OpenScanHub URL correctly from stdout

lbarcziova:
Stdout may include lot of info e.g. about Kerberos ticket processing before the actual dict response from the osh-client, which should be in the last line of the output.

### DIFF
--- a/packit_service/worker/handlers/copr.py
+++ b/packit_service/worker/handlers/copr.py
@@ -3,6 +3,7 @@
 
 import json
 import logging
+import re
 import tempfile
 from datetime import datetime, timezone
 from os import getenv
@@ -564,17 +565,20 @@ class ScanHelper:
                 logger.debug("Something went wrong, skipping the reporting.")
                 return
 
-            logger.info(f"Scan submitted successfully, output: {output}")
+            logger.info("Scan submitted successfully.")
 
-            response = json.loads(output)
-            if not (url := response.get("url")):
+            response_dict = self.parse_dict_from_output(output)
+
+            logger.debug(f"Parsed dict from output: {response_dict} ")
+
+            if not (url := response_dict.get("url")):
                 logger.debug("It was not possible to get the URL from the response.")
                 return
 
             self.copr_build_helper._report(
                 state=BaseCommitStatus.success,
                 description=(
-                    "Scan in OpenScanHub submitted succesfully. Check the URL for more details."
+                    "Scan in OpenScanHub submitted successfully. Check the URL for more details."
                 ),
                 url=url,
                 check_names=["osh-diff-scan:fedora-rawhide-x86_64"],
@@ -586,6 +590,17 @@ class ScanHelper:
                     f"see [docs]({DOCS_URL}/configuration#osh_diff_scan_after_copr_build)."
                 ),
             )
+
+    @staticmethod
+    def parse_dict_from_output(output: str) -> dict:
+        json_pattern = r"\{.*?\}"
+        matches = re.findall(json_pattern, output, re.DOTALL)
+
+        if not matches:
+            return {}
+
+        json_str = matches[-1]
+        return json.loads(json_str)
 
     def find_base_build_job(self) -> Optional[JobConfig]:
         """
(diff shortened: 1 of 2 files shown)


packit_service/worker/handlers/copr.py
```diff
@@ -587,6 +591,16 @@ def handle_scan(self):
                 ),
             )
 
+    @staticmethod
+    def parse_dict_from_output(output: str):
+        json_pattern = r"\{.*?\}"
```
siteshwar, line 596: Why do you need this? Why it can not be as simple as the [CLI code](https://github.com/packit/packit/blob/2e75e6ff4c0cadb55da1c8daf9315e4b0a69e4a8/packit/cli/scan_in_osh.py#L81-L83)?

lbarcziova: currently it is the same as the CLI code, but it raises a `JSONNDecodeError: ExtraData` ([see](https://red-hat-it.sentry.io/issues/5646624565/?project=1767823&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=30d&stream_index=0&utc=true))

as I write in the description, after looking into logs I see the output actually includes also some other info, mostly looks like Kerberos processing info

packit_service/worker/handlers/copr.py
```diff
@@ -587,6 +591,16 @@ def handle_scan(self):
...
+        matches = re.findall(json_pattern, output, re.DOTALL)
+
+        if matches:
+            json_str = matches[-1]
+
+        return json.loads(json_str)
```
siteshwar, line 602: `json_str` is only defined if there the regular expression matches. This line would throw an exception if `json_str` is not defined.

packit_service/worker/handlers/copr.py
```diff
@@ -564,10 +565,13 @@ def handle_scan(self):
...
+            logger.info("Scan submitted successfully.")
 
-            response = json.loads(output)
-            if not (url := response.get("url")):
+            response_dict = self.parse_dict_from_output(output)
+
+            logger.debug(f"Parsed dict from output: {response_dict} ")
+
+            if not (url := response_dict.get("url")):
                 logger.debug("It was not possible to get the URL from the response.")
```
siteshwar, line 575: It should probably be an `error`.

lbarcziova: we are sending those to Sentry, so it could cause a lot of noise, therefore I prefer keeping it as is

siteshwar (approved):

majamassarini (approved):
lgtm!

packit_service/worker/handlers/copr.py
nforro: Are you sure about that? `init_kerberos_ticket()` should run in a separate process, it can't affect stdout of `osh-cli`.

lbarcziova: yes I know, I think it is not about initialising the Kerberos ticket itself, I was coming from these logs:

```
2024-07-26T10:33:43.051834226+00:00 stderr F [2024-07-26 10:33:43,051: INFO/MainProcess] task.run_copr_build_end_handler[f1824d81-6a07-48e3-9ad1-d9ec6171e7d6] Scan submitted successfully, output: [41] 1721990020.122452: ccselect module realm chose cache DIR::/home/packit/kerberos/tkt with client principal packit-stg@FEDORAPROJECT.ORG for server principal HTTP/openscanhub.fedoraproject.org@FEDORAPROJECT.ORG
2024-07-26T10:33:43.051834226+00:00 stderr F [41] 1721990020.122454: Retrieving packit-stg@FEDORAPROJECT.ORG -> krb5_ccache_conf_data/start_realm@X-CACHECONF: from DIR::/home/packit/kerberos/tkt with result: -1765328243/Matching credential not found (filename: /home/packit/kerberos/tkt)
```
etc

nforro: That's still suspicious to me, I'm not against the workaround for the time being, but I think we should investigate and fix this.

PR #2482, opened 2024-07-29
Adjust markdown content for scans reporting

lbarcziova:
TODO:
- [x] link the discussion


siteshwar:
@kdudka Review?

lbarcziova:
> The Name/Job field is empty. Can we add "Differential Scan" (or something similar) there?

that will probably require some more rework of our reporting, I will have a closer look and either adjust the PR or create a followup issue for that.

kdudka (approved):
Looks good.  Thanks!  Once https://github.com/rpm-software-management/mock/pull/1393 is resolved, we will need to update `csmock` to use the new feature of `mock`.  But that should be easy, so we do no need to bother users with this technical detail.

siteshwar (approved):

majamassarini (approved):

PR #2484, opened 2024-07-29
Fix some issues with sidetag builds and updates

nforro:

### DIFF
--- a/packit_service/worker/handlers/bodhi.py
+++ b/packit_service/worker/handlers/bodhi.py
@@ -185,7 +185,12 @@ class BodhiUpdateHandler(
         self._bodhi_update_group_model_id = bodhi_update_group_model_id
 
     def run(self) -> TaskResults:
-        group = self._get_or_create_bodhi_update_group_model()
+        try:
+            group = self._get_or_create_bodhi_update_group_model()
+        except PackitException as ex:
+            logger.debug(f"Bodhi update failed to be created: {ex}")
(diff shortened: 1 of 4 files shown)


packit_service/worker/handlers/bodhi.py
```diff
@@ -185,7 +185,12 @@ def __init__(
         self._bodhi_update_group_model_id = bodhi_update_group_model_id
 
     def run(self) -> TaskResults:
-        group = self._get_or_create_bodhi_update_group_model()
+        try:
+            group = self._get_or_create_bodhi_update_group_model()
+        except PackitException as ex:
```
lbarcziova, line 190: what can happen here causing an PackitException?

lbarcziova (approved):

packit_service/worker/handlers/bodhi.py
nforro: https://github.com/packit/packit-service/blob/2103beccec95039e6cbfae652c8d100030ad5d71/packit_service/worker/handlers/bodhi.py#L115-L119

https://github.com/packit/packit-service/blob/2103beccec95039e6cbfae652c8d100030ad5d71/packit_service/worker/handlers/bodhi.py#L127-L129

nforro: The first case should never really happen, but the second one can happen pretty often.

PR #2485, opened 2024-07-30
More fixes for sidetag builds and updates

nforro:

### DIFF
--- a/packit_service/worker/checker/distgit.py
+++ b/packit_service/worker/checker/distgit.py
@@ -164,6 +164,24 @@ class IsProjectOk(Checker, GetProjectToSyncMixin):
         return self.project_to_sync is not None
 
 
+class TaggedBuildIsNotABuildOfSelf(Checker):
+    """
+    Make sure to skip the build if the event that triggered it is a reaction
+    to an already completed Koji build of the same package that was tagged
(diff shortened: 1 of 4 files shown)


packit_service/worker/checker/distgit.py
```diff
@@ -164,6 +164,18 @@ def pre_check(self) -> bool:
         return self.project_to_sync is not None
 
 
+class TaggedBuildIsNotABuildOfSelf(Checker):
```
lachmanfrantisek, line 167: Can you please add a short docstring to make it really clear also for not-so-experienced packagers? Thanks!

lachmanfrantisek (approved):

lbarcziova (approved):

packit_service/worker/checker/distgit.py
nforro: I'm not sure this is related to packaging experience, this is to break the cycle - completing the build generates an event that triggred the build in the first place, we don't want to repeat that indefinitely (though in this case it would fail anyway because Koji doesn't allow to build the same NVR twice).

nforro: @lachmanfrantisek WDYT?

PR #2489, opened 2024-08-01
Store target in Koji build tag event

nforro:
This should make trigger links in the dashboard work for jobs triggered by Koji tagging events.


lbarcziova (commented):
would it make sense to store the `target` for `KojiBuildTagModel` directly?

nforro:
> would it make sense to store the `target` for `KojiBuildTagModel` directly?

You mean doing it here?

https://github.com/packit/packit-service/blob/567ccab694aabe2286b98b9db47108ba815e2eea/packit_service/models.py#L1441-L1462

Sure, that would work, but wouldn't there be unnecessary duplicities in the DB?

lbarcziova (approved):
lgtm

PR #2496, opened 2024-08-08
Remove sidetag immediately after successful Bodhi update creation

nforro:
Bodhi removes the sidetag an update has been created from automatically when the update hits stable, but it doesn't allow creating new updates from the same sidetag. In order to allow creating e.g. hotfix updates, remove the sidetag immediately after successful update creation so a new one can be created for any new builds and a new update.

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -3592,10 +3592,14 @@ class SidetagModel(Base):
         with sa_session_transaction() as session:
             return session.query(SidetagModel).filter_by(koji_name=koji_name).first()
 
-    def set_koji_name(self, koji_name: str):
+    def set_koji_name(self, koji_name: str) -> None:
         with sa_session_transaction(commit=True) as session:
             self.koji_name = koji_name
             session.add(self)
+
+    def delete(self) -> None:
+        with sa_session_transaction(commit=True) as session:
+            session.delete(self)
 
 
 @cached(cache=TTLCache(maxsize=2048, ttl=(60 * 60 * 24)))
--- a/packit_service/worker/handlers/bodhi.py
+++ b/packit_service/worker/handlers/bodhi.py
@@ -29,6 +29,7 @@ from packit_service.models import (
     BodhiUpdateTargetModel,
     KojiBuildTargetModel,
     SidetagGroupModel,
+    SidetagModel,
 )
 from packit_service.worker.checker.abstract import Checker
 from packit_service.worker.checker.bodhi import (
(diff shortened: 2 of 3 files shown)


LecrisUT:
Wouldn't that mean that additional builds cannot be added to the update/side-tag manually. E.g. let's say there is a gating on rpmddeplint and installability, then as we naviate through the test results, we can add the dependencies that were broken until it's satisfied

nforro:
> E.g. let's say there is a gating on rpmddeplint and installability, then as we naviate through the test results, we can add the dependencies that were broken until it's satisfied

And how do you deal with this situation in case of a regular (no sidetag involved) Packit-created update? As far as I know, you are not able to edit the update.

LecrisUT:
With normal updates, no easy method that I can think of, just untag and re-tag in a side-tag. But what I want to say is that if the bodhi update is already a side-tag, then we can patch the side-tag to add builds to the update.

For the issue in the description, ideally the deletion is only done prior to adding the conflicting update, i.e. check if there are any nvr already present, if so and it is controlled by packit than cancel the previous one and delete the side-tag. Afaiu you can still add multiple side-tags to a build so it should not be blocking koji in that respect right?

mfocko (approved):

nforro:
> But what I want to say is that if the bodhi update is already a side-tag, then we can patch the side-tag to add builds to the update.

That's not how it works though. Once the update is created, you have to edit it to add/remove builds.



> For the issue in the description, ideally the deletion is only done prior to adding the conflicting update, i.e. check if there are any nvr already present, if so and it is controlled by packit than cancel the previous one and delete the side-tag. Afaiu you can still add multiple side-tags to a build so it should not be blocking koji in that respect right?

Sure, there are other ways how to deal with this problem, we chose this one because it's the simplest -  if a sidetag doesn't exist, a new one is created, if it exists, all configured builds are being tagged into it, once all required packages are there, update is created and the sidetag removed.

LecrisUT:
> > But what I want to say is that if the bodhi update is already a side-tag, then we can patch the side-tag to add builds to the update.
> 
> That's not how it works though. Once the update is created, you have to edit it to add/remove builds.

Weird, I will check that again next time I have the opportunity. I thought you can just edit the side-tag as you go. The [documentation](https://docs.fedoraproject.org/en-US/package-maintainers/Package_Update_Guide/#bodhi_update_for_builds_in_a_side_tag) indicates that you can simply refresh the side-tag?

nforro:
> to fix problems found during testing, send the new build to your side tag, then edit the update and refresh the list of builds using the 🔃 (refresh) button

It explicitly mentions **edit**. But I've never done it myself so I could be wrong. And it would be really nice if Packit could do this automatically, it shouldn't be hard to implement, but as you are saying removing the sidetag would break that possibility. But for now I would merge this PR, we can easily revert it in the future.

nforro:
Looking at the code, the refresh button just refreshes (client-side) the list of builds when editing an update, so it's the same as if you added/removed build(s) manually - you have to edit the update and submit the changes. Anyway, I've opened https://github.com/packit/packit-service/issues/2497.

Issue #2468, opened 2024-07-16
Implement a command to tag builds of a package into a sidetag group


nforro:
Users need a way to get a build of a package into a Packit-managed sidetag without a rebuild (a rebuild couldn't even be possible if there is no NVR change).

Let's say that package C depends on packages A and B. Packages B and C get updated and their sync_release PRs get merged, but package A isn't updated and its last builds need to be tagged to the configured sidetag group in order for the pipeline to continue.

Implement a command, perhaps something like `/packit koji-tag-latest`, that will tag the latest builds (for each configured dist-git branch) of a package to a configured sidetag group.

PR #2507, opened 2024-08-22
Handle `/packit koji-tag` dist-git PR comment command

nforro:
Fixes https://github.com/packit/packit-service/issues/2468.

### DIFF
--- a/packit_service/worker/checker/distgit.py
+++ b/packit_service/worker/checker/distgit.py
@@ -6,7 +6,10 @@ import re
 
 from packit.config.aliases import get_branches
 from packit_service.constants import MSG_GET_IN_TOUCH
-from packit_service.utils import pr_labels_match_configuration
+from packit_service.utils import (
+    pr_labels_match_configuration,
+    get_packit_commands_from_comment,
+)
 from packit_service.worker.checker.abstract import Checker, ActorChecker
 from packit_service.worker.checker.helper import DistgitAccountsChecker
 from packit_service.worker.events import (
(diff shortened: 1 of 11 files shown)


nforro:
recheck

packit_service/worker/handlers/distgit.py
```diff
@@ -1015,3 +1016,83 @@ def get_trigger_type_description(self) -> str:
...
+            if (
+                job.type == JobType.koji_build
+                and job.sidetag_group
+                and job.dist_git_branches
+            ):
+                configured_branches = aliases.get_branches(
+                    *job.dist_git_branches, default_dg_branch="rawhide"
+                )
+                if "--all-branches" in args:
+                    branches = configured_branches
```
majamassarini, line 1087: I may be wrong, but if there are multiple branches in `dist_git_branches` I expect we tag them multiple times. I don't know if this could be a problem? Probably not...

packit_service/worker/events/pagure.py
```diff
@@ -173,11 +173,15 @@ def get_packages_config(self) -> Optional[PackageConfig]:
...
-                self.packages_config.upstream_project_url
-                if self.packages_config
-                else None
-            )
+            try:
+                upstream_project_url = (
+                    self.packages_config.upstream_project_url
+                    if self.packages_config
+                    else None
+                )
```
majamassarini, line 181: I am not sure I understand this.
Why should `upstream_project_url` be ambiguous? Because there are multiple packages? In that case shouldn't it happen if we access the data through the `PackageView`, probably I am missing something else?

packit_service/worker/handlers/distgit.py
nforro: Sorry, I don't understand. Do you mean I should get rid of the `--all-branches` argument?

packit_service/worker/events/pagure.py
nforro: Well, we even have [an issue](https://github.com/packit/packit-service/issues/2172) about it, though that's elsewhere in the code.

packit_service/worker/handlers/distgit.py
majamassarini: No, I think it is handy. But I am wondering if there could be problems if we run it multiple times for the same branch, probably not? You probably can tell it better than me.

packit_service/worker/events/pagure.py
nforro: Perhaps there is a better solution for this problem, but in this particular case `upstream_project_url` is not important (I'm not even sure why it's being accessed, we don't need upstream repo here, perhaps for `propose-downstream` retriggering?).

majamassarini (approved):

packit_service/worker/checker/distgit.py
```diff
@@ -94,22 +94,24 @@ def pre_check(self) -> bool:
                     return False
         elif self.data.event_type in (PullRequestCommentPagureEvent.__name__,):
             commenter = self.data.actor
+            tagging = "koji-tag" in self.data.event_dict.get("comment", "")
```
mfocko, line 97: Can't I do `/packit koji-tag koji-build`?

packit_service/worker/handlers/distgit.py
mfocko: Oh, I think I know what Maja means… Multiple job configs for the same branch, not sure if there is a use case for such thing :thinking:

mfocko (approved):

packit_service/worker/handlers/distgit.py
nforro: I guess in theory such a config could exist, but like I said it shouldn't do any harm. You can have multiple `koji_build` jobs with the same sidetag group and the same dist-git branches in different packages (monorepo config), but then the packages will most likely have different `downstream_package_name`, so Packit will be tagging a different build.

packit_service/worker/events/pagure.py
nforro: Fixed in https://github.com/packit/packit/pull/2394.

packit_service/worker/checker/distgit.py
nforro: I suppose you can, but you will get the same result as with `/packit koji-tag`. But I get your point, `/packit koji-build koji-tag` would be problematic. Let me fix that.

mfocko: thanks :+1:

PR #2509, opened 2024-08-26
Implicitly include self in dependents of a koji_build job

nforro:
While it is possible to omit declaring self as a dependency in `bodhi_update` job, it is still necessary to explicitly declare self as a dependent for the `bodhi_update` job to be triggered:

```yaml
downstream_package_name: foo

jobs:
- job: koji_build
  trigger: commit
  sidetag_group: test
  dependents:
    - foo

- job: bodhi_update
  trigger: koji_build
  sidetag_group: test
  # dependencies:
  #   - foo
```

Fix that.


majamassarini (approved):
LGTM

PR #2511, opened 2024-08-30
Increase celery hard time limit for get_usage_statistics

majamassarini:
Since last deployment the get_usage_statistics task is failing because of exception:

Hard time limit (900s) exceeded for packit_service.worker.tasks.get_usage_statistics


lbarcziova (approved):
thanks!

lachmanfrantisek (approved):
Thanks for taking a look!

Issue #2495, opened 2024-08-07
Fix searching for base build when triggering OpenScanHub scan


nforro:
When triggering the scan in OpenScanHub, we are looking for a base build that was triggered from the head commit of the target branch of a PR. The problem is, if the target branch was updated recently, the corresponding build can still be in progress, or it could be an older build that didn't succeed, either way, the scan is not triggered.

Issue #2432, opened 2024-05-23
Previously built Copr PR packages used for buildroot initialization


jan-kolarik:
In dnf5, we use Packit to run CI tests on pull requests in GitHub.

Once the dnf5 package is successfully built, it becomes part of the Copr repository associated with the GitHub pull request. Any subsequent changes to the pull request retrigger the CI tests and the RPM package build in the same Copr repository containing previously built packages.

Since dnf5 is used to build packages, including itself, it is installed during buildroot initialization using the previously built version from the Copr repository. If this version has a runtime issue, it can affect the buildability of subsequent packages in the pull request.

The only workaround is to create a new pull request, which generates a new Copr repository.

hroncok:
Copr chroots can be configured to use mock bootstrap images. That should prevent this issue.

lachmanfrantisek:
Thanks @hroncok for suggesting this! Looks like this is [also configurable via API](https://python-copr.readthedocs.io/en/latest/client_v3/proxies.html#copr.v3.proxies.project.ProjectProxy.edit). This would be really handy!

In the meantime, I've discussed this with the Copr team and they've suggested taking a look at "Copr project directories" (~subproject). It's not documented much and could change a bit but still worth considering -- the directories should not influence each other.

jamacku:
We have hit the same issue in chkconfig/alternatives:
- https://github.com/fedora-sysv/chkconfig/pull/128

The initial PR caused alternatives to segfault. Alternatives are in minimal buildroot so any follow-up commits ended with the Packit copr build failing.

We went the same route as @jan-kolarik, and Packit passed since it installed the package from Fedora instead of from copr.

- https://github.com/fedora-sysv/chkconfig/pull/132

But for some reason, it also rewrote the status checks on the original PR. :confused: So PR 128 has status checks from PR 132.

jamacku:
@lachmanfrantisek, Do you have any updates regarding this issue? It's very inconvenient to have to open new PRs every time we encounter this problem.

Thank you

e.g. https://github.com/fedora-sysv/chkconfig/pull/135

lachmanfrantisek:
@jamacku thanks for the ping. Will prioritise this one -- adding a config option to set a mock bootstrap image sounds like being more easier to do compared to swapping to Copr directories. So, if this works, we will go that way.

nforro:
[...]

PR #2518, opened 2024-09-06
Fix searching for base build when triggering OpenScanHub scan

nforro:
Fixes https://github.com/packit/packit-service/issues/2495.

Two notes:
- I'm not sure what are the performance implications of enabling `require_git_repo_in_local_project`
- In case there are no successful builds but there are a lot of commits it could take a while before it fails, however I'm not sure how to detect such case

### DIFF
--- a/packit_service/worker/handlers/copr.py
+++ b/packit_service/worker/handlers/copr.py
@@ -646,26 +646,38 @@ class ScanHelper:
         base_build_owner = self.copr_build_helper.job_owner_for_job_config(
             base_build_job
         )
+
+        def get_srpm_build(commit_sha):
+            logger.debug(
+                f"Searching for base build for {target_branch_commit} commit "
+                f"in {base_build_owner}/{base_build_project_name} Copr project in our DB. "
+            )
+
+            builds = CoprBuildTargetModel.get_all_by(
+                commit_sha=commit_sha,
+                project_name=base_build_project_name,
+                owner=base_build_owner,
+                target="fedora-rawhide-x86_64",
+                status=BuildStatus.success,
+            )
+            try:
+                return next(iter(builds)).get_srpm_build()
+            except StopIteration:
+                return None
+
         target_branch_commit = (
             self.copr_build_helper.pull_request_object.target_branch_head_commit
         )
 
-        logger.debug(
-            f"Searching for base build for {target_branch_commit} commit "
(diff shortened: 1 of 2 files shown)


lbarcziova:
> I'm not sure what are the performance implications of enabling require_git_repo_in_local_project

it means now for the copr build handler tasks, the git repo would be cloned, which we tried to avoid previously :/ do you think we could solve this in another way?

nforro:
> it means now for the copr build handler tasks, the git repo would be cloned, which we tried to avoid previously :/ do you think we could solve this in another way?

I suppose it would be possible (though a bit hacky) to only enable `require_git_repo_in_local_project` per instance and rebuild `LocalProject`, would that make sense?

lbarcziova:
I am still not sure we should be cloning the whole repo to get few commit hashes, what about getting it via API?

mfocko:
Wouldn't it be easier to get last successful build? From the practical PoV I don't really see a difference between:
* checking commit by commit for a build
* taking last successful build

Or is the issue with different packages in the Copr repo? Can't we just filter by the package name in that case?

nforro:
I think the problem is we need to get only builds triggered from the target branch of a PR. Although it would be weird to build in the same Copr project from different branches, it is possible.

nforro:
I have opened https://github.com/packit/ogr/pull/857.

nforro:
I decided to leave `target_branch_head_commit` there and use it first, since it should work in most cases and `get_commits()` could take some time, so it makes sense to call it only if necessary.

lachmanfrantisek:
> I think the problem is we need to get only builds triggered from the target branch of a PR. Although it would be weird to build in the same Copr project from different branches, it is possible.

We can use Build directories for this purpose. But this would require quite a lot of work.

nforro:
> We can use Build directories for this purpose. But this would require quite a lot of work.

Are you referring to [this](https://github.com/fedora-copr/copr/blob/main/doc/user_documentation.rst#subprojects)? I like the idea of having commit/release triggered builds in a project and PR builds in its subdirectories (I'm just starting to work on https://github.com/packit/packit-service/issues/2432), but I'm not sure if we should force users do configure that if they want to use OSH. Or maybe I just don't get your comment :slightly_smiling_face:

lbarcziova (approved):
lgtm :+1:

PR #2524, opened 2024-09-11
Create model for multiple Anitya versions

### LINKED ISSUE #2440: Do not generate database entry from the event for Anitya / New Hotness
Part of #2106

Currently we generate a database entry from the event:
https://github.com/packit/packit-service/blob/ca15dc13fde8d5b644c8de5b3c87ef176054066a/packit_service/worker/events/new_hotness.py#L77-L83

This is a problematic state, cause when we start reacting to Anitya, we can receive multiple versions (not just the latest one as for the Rawhide) from which we then choose one based on to-be implemented rules or version mask defined by the user.

Move this functionality out of the event, as the tight-coupling of `event × DB entry` doesn't make sense in this case.

Be careful about the consequences of this action, i.e., database being populated by calling the `_add_*` on the events.

> [!NOTE]
> Alternative approach could be populating the tag name in DB with a dummy value that would be overwritten in the handler in later stages, instead of creating the database entry later by the handler.

lbarcziova:
Fixes #2440

Discussed with @mfocko that we can store the multiple versions in the DB.

RELEASE NOTES BEGIN

N/A

RELEASE NOTES END


packit_service/worker/events/new_hotness.py
```diff
@@ -215,13 +215,17 @@ def __init__(
...
-        # the package config from dist-git and resolve the next version; how are
-        # we going to choose the version?
-        #   a) latest greatest
-        #   b) next unreleased?
-        # this can be also influenced by the mask…
-        #
-        # It would be ideal to handle here because of the serialization…
-        raise NotImplementedError()
+    def version(self) -> Optional[str]:
+        return None
```
majamassarini, line 219: I am not sure I understand why we return `None` here, can we add a comment? 🙏

packit_service/models.py
```diff
@@ -493,6 +497,53 @@ def get_or_create(
...
+
+    @classmethod
+    def get_by_id(cls, id_: int) -> Optional["AnityaMultipleVersionsModel"]:
+        with sa_session_transaction() as session:
+            return session.query(AnityaVersionModel).filter_by(id=id_).first()
+
+    def __repr__(self):
+        return (
+            f"AnityaMultipleVersionsModel("
+            f"version={self.version}, "
```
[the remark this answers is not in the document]
lbarcziova: yes, copy-pasted, good catch!

packit_service/worker/events/new_hotness.py
lbarcziova: sure, will add

lbarcziova: added

majamassarini (approved):
🙏🏻

Issue #2526, opened 2024-09-12
ACTION REQUIRED: Sentry Project Transactions/Errors Sample Rates


majamassarini:
As stated in the email received on date September 11th.

We are in `the top 6 projects that are impacting the Transactions quota significantly and these project should take quick action to reduce their Transactions consumption`

Packit transactions in 30 day: 248k
Packit errors in 30 days: 37k

Red Hat overall quota for transactions: 2 million transactions
Red Hat overall quota for errors: 8 million errors

Recommended Actions:
  - [x] Reduce Transaction Sample Rate (traceSampleRate):
         Lowering the transaction sample rate can significantly reduce the volume of data being sent to Sentry.
         [We should already have a low transaction sample rate](https://github.com/packit/packit-service/blob/8e545435bee15d60da6be2c7d9bda851d31715c1/packit_service/sentry_integration.py#L17-L31)

  - [ ] Reduce Error Sample Rate (sampleRate):
         Similar to transaction rates, reducing the error sample rate will help decrease data volume.
         
  - [x] Filter Out Non-Essential Errors and Transactions:
         Utilize Sentry’s ignoreErrors and beforeSend features to filter out less critical data before it’s sent. This ensures that only the most relevant issues are tracked, helping to manage our quota.
         Suppress errors that aren't application errors (e.g linting, typing, etc)
         [We are sending to sentry not just errors but also debug data, but I am not sure this will help with the transaction data](https://github.com/packit/packit-service/blob/8e545435bee15d60da6be2c7d9bda851d31715c1/packit_service/sentry_integration.py#L76-L79)

[According to this overview, now that I check it, there were 7 transactions collected in the last minute](https://red-hat-it.sentry.io/performance/?environment=prod&project=1767823&statsPeriod=24h). This data can cope with the ~250K transactions in a month. Now that I am looking at it, we have:
- 5 transactions/minute for [api.webhooks_github_webhook](https://red-hat-it.sentry.io/performance/summary/?environment=prod&project=1767823&query=http.method%3APOST&referrer=performance-transaction-summary&statsPeriod=24h&transaction=api.webhooks_github_webhook&unselectedSeries=p100%28%29&unselectedSeries=avg%28%29)
[...]

PR #2541, opened 2024-09-24
docs: use Google-style docstrings for task results

mfocko:


packit_service/worker/result.py
```diff
@@ -17,14 +17,13 @@ class TaskResults(dict):
...
-                          True - we processed the event
-                          False - there was an error while processing it -
-                                  usually an exception
-        :param details: more info from job handler
-                        (optional) 'msg' key contains a message
-                        more keys to be defined
+        Args:
+            success: Represents the resulting state of the job handler.
+                `True`, if we processed the event; `False` an error occurred
+                while processing it (usually an exception)
```
mfocko, line 23: @lbarcziova related to your #2526… Going through the Sentry we have a lot of “errors” that we can't do much about, e.g., failed SRPM build, failed RPM build; for both of the examples, there's rather small chance it is our fault, in case of SRPMs it could be (we run `packit srpm` in Copr), but it would have rather high impact for the Sentry error to be **the one** to raise the suspicion…

Since `success=False` implies `logger.error(some message)`, this causes a lot of entries in the Sentry. In the ideal world this should be an enumeration of `ok, nok, err` (or `success, fail, error`) where the `nok`/`fail` represents not successful result, but not directly failing on our side (e.g., failed (S)RPM build above), and `err`/`error` issue on our end (e.g. caught exception).

Additionally we could retry certain `nok`/`fail` (like we do right now for `sync-release`), but it doesn't make sense to retry an `error` which would be an internal error on our end.

---

Maybe open for a discussion on Thursday's arch.

PR #2554, opened 2024-10-01
Avoid race condition when creating sidetag in Koji

### LINKED ISSUE #2545: Duplicate sidetags
The `podman-releases` sidetag group is shared between `podman`, `buildah`, `skopeo` and `containers-common` packages. When releasing `podman` and `buildah` while skipping the other two packages using `/packit koji-tag --all-branches` to tag them, somehow it happened that two sidetags were created for `f41` and `f40`. One set of them as a result of the command and the other set most likely during a build of `buildah`. This then messed up resolving dependencies which prevented Bodhi update for `f40` from being triggered and would have caused issues in the future if the duplicate sidetags weren't manually removed.

TODO:
- [ ] add unique constraint on (`sidetag_group_id`, `target`) pair in the `sidetags` table so that there can be only one sidetag per group and dist-git branch
- [ ] figure out how it could have happened in the first place and fix it

nforro:
There is a big time window between reading Koji name of a sidetag from the DB and updating it with a new value returned after a new sidetag in Koji is created, making race conditions possible.

Prevent this by doing everything in a single DB transaction.

Fixes #2545.

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -13,6 +13,7 @@ from datetime import datetime, timedelta, timezone
 from os import getenv
 from typing import (
     Dict,
+    Generator,
     Iterable,
     List,
     Optional,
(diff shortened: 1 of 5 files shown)


lbarcziova (approved):
lgtm :+1:

PR #2559, opened 2024-10-04
Skip running Koji builds if they are triggered already

### LINKED ISSUE #2503: Downstream KojiBuild state not handled correctly
### What happened? What is the problem?

I discovered that sometimes BodhiUpdate is not filed by Packit, despite being correctly configured.

I did and upstream release of fedora-upgrade and submitted it to Fedora using `tito release fedora-git`.  And I forgot to add `--no-build`. This uploaded src.rpm to all branches in dist-git **and** triggered koji build. The builds were submitted by both Packit and Tito. As you can see in https://koji.fedoraproject.org/koji/packageinfo?packageID=15136
Packit was first for F40 and Tito everywhere else. Who is second is denied to proceed. As you can see in F39 build https://koji.fedoraproject.org/koji/taskinfo?taskID=122080964 submitted by Packit. Although it failed Packit reports in dashboard that the build is still pending https://dashboard.packit.dev/results/koji-builds/6850

And what is important is that F39 bodhi update was not filled. Despite the condition was meant - there is successful build https://koji.fedoraproject.org/koji/buildinfo?buildID=2531586

### What did you expect to happen?

I expected that Bodhi update for F39 is created.

### Example URL(s)

_No response_

### Steps to reproduce

```bash
1. fedpkg import foo.src.rpm
2. fedpkg build (be quicker than Packit)
3. Bodhi update is not created.
```

### What is the impacted category (job)?

Fedora release automation

### Workaround

- [X] There is an existing workaround that can be used until this issue is fixed.

### Participation

- [ ] I am willing to submit a pull request for this issue. (Packit team is happy to help!)

lbarcziova:
Fixes #2503
Requires packit/packit#2435

TODO:

- [x] test it

RELEASE NOTES BEGIN

Before triggering the non-scratch Koji builds, we now check whether there is not already a build in progress or completed for the same NVR.

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/handlers/distgit.py
+++ b/packit_service/worker/handlers/distgit.py
@@ -25,6 +25,7 @@ from packit.exceptions import (
     PackitDownloadFailedException,
     ReleaseSkippedPackitException,
 )
+from packit.utils.koji_helper import KojiHelper
 from packit_service import sentry_integration
 from packit_service.config import PackageConfigGetter, ServiceConfig
 from packit_service.constants import (
@@ -35,6 +36,7 @@ from packit_service.constants import (
     DEFAULT_RETRY_BACKOFF,
     MSG_RETRIGGER_DISTGIT,
     RETRY_LIMIT_RELEASE_ARCHIVE_DOWNLOAD_ERROR,
+    KojiBuildState,
 )
 from packit_service.models import (
     SyncReleasePullRequestModel,
@@ -737,6 +739,8 @@ class AbstractDownstreamKojiBuildHandler(
     topic = "org.fedoraproject.prod.pagure.git.receive"
     task_name = TaskName.downstream_koji_build
 
+    _koji_helper: Optional[KojiHelper] = None
+
     def __init__(
         self,
         package_config: PackageConfig,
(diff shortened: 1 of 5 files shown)


packit_service/worker/handlers/distgit.py
```diff
@@ -806,6 +829,14 @@ def run(self) -> TaskResults:
                 )
                 continue
 
+            if self.is_already_triggered(branch):
```
nforro, line 832: I think we should skip this if `self.job_config.scratch` it set (scratch builds can have the same NVR as existing production builds and `KojiHelper.get_build_info()` ignores scratch builds).

packit_service/worker/handlers/distgit.py
```diff
@@ -791,6 +793,27 @@ def _get_or_create_koji_group_model(self) -> KojiBuildGroupModel:
     def get_branches(self) -> List[str]:
         """Get a list of branch (names) to be built in koji"""
 
+    def is_already_triggered(self, branch: str) -> bool:
+        """
+        Check if the build was already triggered
+        (building or completed state).
+        """
+        nvr = self.packit_api.dg.get_nvr(branch)
+        existing_build = KojiHelper().get_build_info(nvr)
```
nforro, line 802: I'm not sure how expensive is instantiation of `koji.ClientSession` but it's probably worth keeping a single session open rather than creating a new one for every branch per build. We could use the same approach as in the past:

https://github.com/packit/packit-service/blob/6f028d40eb02100e5f6ae1d5c12123ddb5688c3e/packit_service/worker/handlers/distgit.py#L719-L744

nforro (approved):

PR #2561, opened 2024-10-04
docs(ib): add note to the handler about Copr

mfocko:
As we discussed with @majamassarini, we cannot find any use cases for supporting the VM image builds via Packit Service without any Copr repository set. Therefore adding a note about the fact that it's not possible as opposed to CLI support of this through ‹packit build in-image-builder›.

Related to packit/packit#1966, packit/packit#2434

### DIFF
--- a/packit_service/worker/handlers/vm_image.py
+++ b/packit_service/worker/handlers/vm_image.py
@@ -62,6 +62,9 @@ class VMImageBuildHandler(
     def get_checkers() -> Tuple[Type[Checker], ...]:
         return (
             HasAuthorWriteAccess,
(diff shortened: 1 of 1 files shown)


majamassarini (approved):
Thank you! Probably would be helpful to have a note also in our documentation in packit.dev, it could be puzzling this behaviour difference between CLI and service. But maybe before doing it we could discuss with the others?

mfocko:
> Thank you! Probably would be helpful to have a note also in our documentation in packit.dev, it could be puzzling this behaviour difference between CLI and service. But maybe before doing it we could discuss with the others?

I'll add arch topic for the next week

PR #2580, opened 2024-10-11
Receive and save OpenScanHub task finish event

### LINKED ISSUE #2543: Implement OpenScanHub initial reporting
- [ ] in fedmsg repo, process the messages about OSH task states (see https://github.com/openscanhub/openscanhub/pull/299 ) 
- [ ] create new event class for that and implement parsing of these 
- [ ] create a model for a scan
- [ ] store the task in the DB [somewhere here](https://github.com/packit/packit-service/blob/4507b83b9272bebd30c584b00dcdf716ab34e2e8/packit_service/worker/handlers/copr.py#L570) and instead of `successful` status there, report `in-progress`
- [ ] create a handler to report the results for the new event
   - obtain the task in our DB, update the state and report it to user
   - for now, just forward the URLs from the message (=> no additional processing)
   - take into consideration #2515 
   - by default don't fail the CI, have a look also into `action_required` conclusion from [here](https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#create-a-check-run), whether that could be used, or use neutral/successful
   - report neutral status for failed, cancelled or interrupted builds

---

Part of #2516

majamassarini:
Fixes #2543

Merge after:
- https://github.com/packit/packit-service-fedmsg/pull/113
- https://github.com/packit/deployment/pull/603


siteshwar:
> Fixes #2543

Thank you for working on this! But it does not handle the case when the tasks may end up in `failed`, `cancelled` or `interrupted` state. So, the reporting is incomplete.

majamassarini:
> > Fixes #2543
> 
> Thank you for working on this! But it does not handle the case when the tasks may end up in `failed`, `cancelled` or `interrupted` state. So, the reporting is incomplete.

I am still working on it 🙂 I will push new commits soon (I hope).

packit_service/models.py
```diff
@@ -3824,6 +3834,52 @@ def get_by_koji_name(cls, koji_name: str) -> Optional["SidetagModel"]:
             return session.query(SidetagModel).filter_by(koji_name=koji_name).first()
 
 
+class ScanModel(Base):
```
lbarcziova, line 3837:
```suggestion
class OSHScanModel(Base):
```

so that we are explicit and consistent (e.g. TFT models, Copr,..)

packit_service/models.py
```diff
@@ -3824,6 +3834,52 @@ def get_by_koji_name(cls, koji_name: str) -> Optional["SidetagModel"]:
...
+    __tablename__ = "scans"
+    id = Column(Integer, primary_key=True)
```
lbarcziova, line 3839: I think it would make sense to store `status` as well

packit_service/worker/handlers/open_scan_hub.py
```diff
@@ -0,0 +1,125 @@
...
+                        "No job configuration found for "
+                        f"openscanhub_task_finish in {self.project.repo}"
+                    )
+                },
+            )
+
+        branches = aliases.get_build_targets(
+            *self.job_config.targets,
+        )
+        if "fedora-rawhide-x86_64" not in branches:
```
lbarcziova, line 76: I'm wondering if this shouldn't be rather in `pre-check`

packit_service/worker/handlers/open_scan_hub.py
```diff
@@ -0,0 +1,125 @@
...
+            return TaskResults(
+                success=False,
```
lbarcziova, line 78:
```suggestion
                success=True,
```

it makes sense to me to have this marked as successful

packit_service/worker/handlers/open_scan_hub.py
```diff
@@ -0,0 +1,125 @@
...
+
+        scan_helper = ScanHelper(copr_build_helper=build_helper, build=event.build)
+
+        external_links = {
+            "Added issues": event.issues_added_url,
+            "Fixed issues": event.issues_fixed_url,
+            "Scan results": event.scan_results_url,
+        }
+
+        # TODO: probably we need a babysit task for when the build is not finished yet
```
lbarcziova, line 102: what is this scenario of receiving the message but scan not being  finished?

packit_service/worker/handlers/open_scan_hub.py
```diff
@@ -0,0 +1,125 @@
...
+    @staticmethod
+    def get_checkers() -> Tuple[Type[Checker], ...]:
+        return ()
+
+    def run(self) -> TaskResults:
+        task_id = self.data.event_dict["task_id"]
+        event = self.data.to_event()
+        if not event.scan or not event.build:
+            return TaskResults(
+                success=False,
```
lbarcziova, line 57:
```suggestion
                success=True,
```

packit_service/worker/handlers/open_scan_hub.py
```diff
@@ -0,0 +1,125 @@
...
+                details={
+                    "msg": f"Scan {task_id} not found or not associated with a Copr build"
+                },
+            )
+        elif not self.job_config:
+            return TaskResults(
+                success=False,
```
lbarcziova, line 64:
```suggestion
                success=True,
```

packit_service/models.py
lbarcziova: so that we know and can show on dashboard pending vs completed scan

packit_service/models.py
```diff
@@ -1970,6 +1970,8 @@ class CoprBuildTargetModel(GroupAndTargetModelConnector, Base):
         "CoprBuildGroupModel", back_populates="copr_build_targets"
     )
 
+    scan = relationship("ScanModel", back_populates="copr_build_target")
```
lbarcziova, line 1973: I was wondering whether this should not be connected on the `PipelineModel` level. Have you considered this as well?

packit_service/worker/handlers/open_scan_hub.py
```diff
@@ -0,0 +1,125 @@
...
+        scan_helper = ScanHelper(copr_build_helper=build_helper, build=event.build)
+
+        external_links = {
+            "Added issues": event.issues_added_url,
+            "Fixed issues": event.issues_fixed_url,
+            "Scan results": event.scan_results_url,
+        }
+
+        # TODO: probably we need a babysit task for when the build is not finished yet
+        if event.build.status == "success":
```
lbarcziova, line 103: shouldn't we be checking the status of the scan and not build? We submit the scans just for successful build, so there should be no need for this

lbarcziova (commented):
good job Maja, this will be great!

packit_service/worker/handlers/open_scan_hub.py
majamassarini: No, I was really thinking about the build... I mean the copr build. I thought we wanted the scan result somehow coupled with the build result. This is also related to what I wrote below... But I understand it now and it makes no sense :)

packit_service/worker/handlers/open_scan_hub.py
```diff
@@ -0,0 +1,125 @@
...
+    LocalProjectMixin,
+    PackitAPIWithUpstreamMixin,
+)
+
+from packit_service.worker.reporting import BaseCommitStatus
+
+logger = logging.getLogger(__name__)
+
+
+@configured_as(job_type=JobType.copr_build)
```
mfocko, line 38: I don't even recall what we use this for :eyes:

if it's used for matching the job to the received event, I have a feeling it could break with multiple Copr builds being specified :thinking:

packit_service/worker/handlers/open_scan_hub.py
```diff
@@ -0,0 +1,125 @@
...
+        if event.build.status == "success":
+            state = BaseCommitStatus.success
+            description = (
+                "Scan in OpenScanHub is finished. Check the URL for more details."
+            )
+        else:
+            state = BaseCommitStatus.neutral
+            description = (
+                "Scan in OpenScanHub is finished but the build did not "
+                "finish yet or did not succeed."
```
mfocko, line 112: We're talking about RPM build in Copr, right? Cause from the wording I'm a bit confused :D SRPM build triggers OSH. Now that I think about it… Does it matter whether the RPM build succeeds or fails? :thinking: OSH does static analysis, the results should be relevant regardless

packit_service/worker/handlers/open_scan_hub.py
lbarcziova: good point, @siteshwar could you please provide more details how the message looks when the build is failed/cancelled/interrupted vs when it finishes successfully? Is it detectable by the result files (like added, fixed) not being present?

packit_service/worker/handlers/open_scan_hub.py
lbarcziova: we trigger OSH just when the RPM build finishes successfully, so that we do not overload OSH. But there is mock build being done again after the submission in OSH and scan is being run when the mock build finishes successfully, am I right @siteshwar ?

packit_service/worker/handlers/open_scan_hub.py
siteshwar: > good point, @siteshwar could you please provide more details how the message looks when the build is failed/cancelled/interrupted vs when it finishes successfully? Is it detectable by the result files (like added, fixed) not being present?

These are the topics for tasks that do not finish successfully:
```
org.fedoraproject.stg.openscanhub.task.cancel
org.fedoraproject.stg.openscanhub.task.fail
org.fedoraproject.stg.openscanhub.task.interrupt
```

And the body contains the build id.

Other than that, you may also want to process:
```
org.fedoraproject.stg.openscanhub.task.open
```

which means a worker has started running the task.

packit_service/worker/handlers/open_scan_hub.py
siteshwar: > we trigger OSH just when the RPM build finishes successfully, so that we do not overload OSH. But there is mock build being done again after the submission in OSH and scan is being run when the mock build finishes successfully, am I right @siteshwar ?

OpenScanHub uses a wrapper over `mock` which is known as `csmock` that is used to capture and process the output of a RPM build, so basically you rebuild the entire package again. As an example, you can see the [stdout.log](https://openscanhub.fedoraproject.org/task/15319/log/stdout.log) for the curl package.

packit_service/worker/handlers/open_scan_hub.py
siteshwar: I have summarized the steps again in https://github.com/packit/packit-service/issues/2543#issuecomment-2417233326

packit_service/worker/handlers/open_scan_hub.py
majamassarini: Yes exactly, and thus I am checking that in the targets for the copr_build job there is  `fedora-rawhide-x86-64`.

packit_service/worker/handlers/open_scan_hub.py
majamassarini: Thanks a lot @siteshwar now it is far more clear to me. I think it will not be a big deal to handle the other messages as well. Still I want to ask you if there is a reason why you created different messages? I would have expected a `started` and a `finished` task message but not a `cancel or fail or interrupt` message. I see the *cancel or fail or interrupt* more as states for the *finished* action. For example for the [copr.build.end](https://apps.fedoraproject.org/datagrepper/v2/id?id=353d4c59-eafa-4043-aa7e-6386e7bb595b&is_raw=true&size=extra-large) message there is a `status` field and there are just the `copr.build.start` and `copr.build.end` topics/messages. As I wrote, we can deal with more topics/messages but maybe you also prefer having just one finish message?

majamassarini: One more thing, if we will have all this messages I suppose I will never receive a `finish` message together with an `interrupt or cancel or fail` message for the same `task_id`, right? Otherwise, if the order of the messages is not preserved in our workers, I could override a failed scan as successful.

packit_service/worker/handlers/open_scan_hub.py
majamassarini: Sorry for the noise, and yes I was talking about the copr build. I didn't know where to take the result state for the scan and I thought that as a **first step** implementation we wanted just to mark the check as finished and mirroring the state of the "copr build"...
Now that I know where to take the info I will handle the state of the scan and change the check description!

packit_service/worker/handlers/open_scan_hub.py
```diff
@@ -0,0 +1,125 @@
...
+        # TODO: probably we need a babysit task for when the build is not finished yet
+        if event.build.status == "success":
+            state = BaseCommitStatus.success
+            description = (
+                "Scan in OpenScanHub is finished. Check the URL for more details."
+            )
+        else:
+            state = BaseCommitStatus.neutral
+            description = (
+                "Scan in OpenScanHub is finished but the build did not "
```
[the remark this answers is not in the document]
majamassarini: See the above comment :)

packit_service/models.py
majamassarini: We agreed on keeping the actual design.

packit_service/worker/handlers/open_scan_hub.py
siteshwar: > As I wrote, we can deal with more topics/messages but maybe you also prefer having just one finish message?

I am just reflecting the different states of tasks in the messages as Kobo.

@kdudka Shall we change the message topics as Copr or keep them as it is?

kdudka: Yep, the proposed change sounds reasonable.  Let's implement the messaging such that it is better aligned with other services that Packit listens to.

packit_service/worker/handlers/open_scan_hub.py
```diff
@@ -0,0 +1,170 @@
...
+            self.event.scan.set_issues_added_url(self.event.issues_added_url)
+            self.event.scan.set_issues_fixed_url(self.event.issues_fixed_url)
+            self.event.scan.set_scan_results_url(self.event.scan_results_url)
+        else:
+            state = BaseCommitStatus.neutral
+            description = (
+                f"Scan in OpenScanHub is finished in a {self.event.status} state."
+            )
+            external_links = {}
+            if self.event.status == OpenScanHubTaskFinishedEvent.Status.cancel:
```
lbarcziova, line 125: what about the `interrupt` state here?

lbarcziova (approved):
just small notes, otherwise LGTM, great job!

packit_service/worker/handlers/open_scan_hub.py
majamassarini: I thought that from a Packit point of view the `interrupted` and `canceled` states can have the same "view". I mapped them both to a `canceled` state in our db. 
But the details are available in the Open Scan Hub logs and also the message we give back has the detail if this was canceled or interrupted...
But if we want to map all the states in Packit too, I can add the missing one.

lbarcziova: ok, that makes sense, thanks for the explanation!

Issue #2200, opened 2023-08-30
Testing Farm on-demand customized execution


obabec:
### Description

Currently, Packit does not support adding tf_extra_params from the trigger command. This means that users cannot in any way be more specific about test execution.

Workaround using Packit does not exist. You would need to modify `.packit.yaml` and then change it back or use testing farm CLI directly.

### Benefit

The benefit of this feature is definitely noticeable. Users will be able to specify extra environment variables to test environments. 

For example, the project I am working on (Debezium - https://github.com/debezium/debezium/blob/main/.packit.yaml) is using Packit to trigger testing farm. We have quite a few jobs with 600+ tests running 1h+. But if I broke just one test with my PR and I want confirmation that I have fixed this issue I could save time and just add environment variable `ORACLE_ARG=-Dit=io.debezium.connector.oracle.OracleConnectorIT` (in our case) and our test script will run only one class. 

That results as the test run is not 100 minutes but 15.

### Importance

I don't think that it is a critical issue but it would be a very nice addition to a lot of projects.

### Workaround

- [ ] There is an existing workaround that can be used until this feature is implemented.

### Participation

- [X] I am willing to submit a pull request for this issue. (Packit team is happy to help!)

lbarcziova:
We have discussed this with the team, here are the notes:
- let's start with reusing the same status name for the test with modified parameters from comment
- we should consider `argparse` to ease the way of processing the comment arguments
- forwarding the env variables should be quite straightforward and be a good start

lbarcziova:
@obabec would you be interested in the implementation of this? Could we somehow help?

obabec:
@lbarcziova Sure I am interested, I will look into that next week hopefully.

lbarcziova:
@obabec great, let me know if you will need some assistance! :)

obabec:
@lbarcziova Just FYI, I have finally managed to get some time. I will start with implementation this :)

lbarcziova:
@Frawless mentioned on Slack that he would like to implement this.

PR #2583, opened 2024-10-14
Add --env to comment trigger phrase

Frawless:
Checklist:

This PR adds possibility to specify `--env` parameters in manual trigger command. 

User can specify multiple env variables in format like `--env MY_ENV_VAR=value --env MY_ENV_VAR2=value2`. Every env var has to have it's own `--env` keyword. 

Env variables specified with this option has the highest priority when building TF payload.

Fixes https://github.com/packit/packit-service/issues/2200

RELEASE NOTES BEGIN

Packit now supports `--env` parameters in manual trigger command.

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/helpers/testing_farm.py
+++ b/packit_service/worker/helpers/testing_farm.py
@@ -1,9 +1,10 @@
 # Copyright Contributors to the Packit project.
 # SPDX-License-Identifier: MIT
-
+import argparse
 import logging
 import re
 from re import Pattern
+import shlex
 from typing import Dict, Any, Optional, Set, List, Union, Tuple, Callable
 
 import requests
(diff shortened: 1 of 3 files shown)


packit_service/worker/helpers/testing_farm.py
```diff
@@ -523,6 +556,16 @@ def _payload(
         env_variables = self.job_config.env if hasattr(self.job_config, "env") else {}
         predefined_environment.update(env_variables)
 
+        # User-defined variables from comments have priority
+        if self.is_comment_event:
+            if self.comment_arguments.envs is not None:
```
lbarcziova, line 561:
```suggestion
        if self.is_comment_event and self.comment_arguments.envs is not None:
```

packit_service/worker/helpers/testing_farm.py
```diff
@@ -523,6 +556,16 @@ def _payload(
...
+                env_variables = {
+                    k: v
+                    for k, v in self.comment_arguments.envs.items()
+                    if v is not None
+                }
```
lbarcziova, line 566: what about a use-case of "unsetting" env var via comment arg?

packit_service/worker/helpers/testing_farm.py
```diff
@@ -60,40 +61,72 @@ class CommentArguments:
...
-        match = re.search(r"--labels[\s=](?P<labels>\S+)", comment)
-        if match:
-            self.labels = match.group("labels").split(",")
-            logger.debug(f"Parsed test argument -> labels: {self.labels}")
+        # Set up argparse
+        parser = argparse.ArgumentParser()
+        parser.add_argument("packit_command")
+        parser.add_argument("--identifier", "--id", "-i")
+        parser.add_argument("--labels")
+        parser.add_argument("--env", action="append")  # Allows multiple --env arguments
```
lbarcziova, line 96: maybe we could create property out of this, or at least separate the creation of the parser

packit_service/worker/helpers/testing_farm.py
```diff
@@ -60,40 +61,72 @@ class CommentArguments:
...
+        logger.debug(f"Parsed identifier: {self.identifier}")
+
+        if args.labels:
+            self.labels = args.labels.split(",")
+            logger.debug(f"Parsed labels: {self.labels}")
+
+        if args.env:
+            # Parse envs into dictionary
+            self.envs = dict(env.split("=") for env in args.env)
+            logger.debug(f"Parsed envs: {self.envs}")
```
lbarcziova, line 121: I think here we should add some error handling for bad input (not `key=value` format)

packit_service/worker/helpers/testing_farm.py
```diff
@@ -60,40 +61,72 @@ class CommentArguments:
     Parse arguments from trigger comment and provide the attributes to Testing Farm helper.
     """
 
-    packit_command: str = None
-    identifier: str = None
-    labels: List[str] = None
-    pr_argument: str = None
-
     def __init__(self, command_prefix: str, comment: str):
```
lbarcziova, line 64: the `__init__` method grew quite a lot with the new code, do you think it could be split up a bit?

lbarcziova (commented):
really nice work, thanks for looking into it so quickly!! 

Just few comments, and could you please fix the commits history (there is one in the PR that's already in main)?

packit_service/worker/helpers/testing_farm.py
Frawless: I didn't think about this one. You mean something like `--env MY_ENV_VAR=""` will unset it or do you have something else how should look like? It shouldn't be a problem to implement it.

packit_service/worker/helpers/testing_farm.py
Frawless: Sure, I will try to split it

packit_service/worker/helpers/testing_farm.py
Frawless: I did some changes, please take a look :)

packit_service/worker/helpers/testing_farm.py
Frawless: I did some changes, please take a look :)

packit_service/worker/helpers/testing_farm.py
Frawless: I am not python expert so maybe my changes are not python compliant, but I think it is better now

packit_service/worker/helpers/testing_farm.py
```diff
@@ -60,40 +61,94 @@ class CommentArguments:
...
+        self.labels: List[str] = None
+        self.pr_argument: str = None
+        self.envs: Dict[str, str] = None
+
         if comment is None:
             return
 
-        # Try to parse identifier argument from comment
+        # Init argparse
+        self.parser = self.init_arg_parse
```
nforro, line 76: Shouldn't the method be actually invoked?

```suggestion
        self.parser = self.init_arg_parse()
```

Also, `init_arg_parser()` would make a bit more sense to me.

packit_service/worker/helpers/testing_farm.py
```diff
@@ -60,40 +61,94 @@ class CommentArguments:
...
+        self.parse_known_arguments(args)
+        self.parse_unknown_arguments(unknown_args)
+
+    @property
+    def init_arg_parse(self):
+        # Set up argparse
+        parser = argparse.ArgumentParser()
+        parser.add_argument("packit_command")
+        parser.add_argument("--identifier", "--id", "-i")
+        parser.add_argument("--labels")
```
nforro, line 113:
```suggestion
        parser.add_argument("--labels", type=lambda s: s.split(","))
```

This should make `args.labels` below to already be a list.

packit_service/worker/helpers/testing_farm.py
Frawless: > Shouldn't the method be actually invoked?
That's because of `@property` annotation on the method

> Also, init_arg_parser() would make a bit more sense to me.
I will update it

nforro: > That's because of `@property` annotation on the method

Oh, sorry, I missed that. But if it's a property then it should be called `arg_parser` or perhaps just `parser`, the verb doesn't belong there.

nforro: See e.g. here how such a property should be implemented:
https://github.com/packit/packit-service/blob/c376983ead260911b07001f083608f2e5b1246e1/packit_service/worker/handlers/mixin.py#L462-L478

Frawless: Oh I see now, I updated the code to follow this implementation.

packit_service/worker/helpers/testing_farm.py
```diff
@@ -60,40 +61,92 @@ class CommentArguments:
...
+            self.labels = args.labels
+            logger.debug(f"Parsed labels: {self.labels}")
+
+        if args.env:
+            self.envs = {}
+            for env in args.env:
+                if "=" in env:
+                    key, value = env.split("=", 1)
+                    self.envs[key] = value
+                    logger.debug(f"Added env variable: {key}={value}")
```
lbarcziova, line 135:
```suggestion
                    logger.debug(f"Parsed env variable: {key}={value}")
```

just for consistency

packit_service/worker/helpers/testing_farm.py
```diff
@@ -60,40 +61,92 @@ class CommentArguments:
...
 
-        match = re.search(r"(?P<pr_arg>[^/\s]+/[^#]+#\d+)", comment)
-        if match:
-            self.pr_argument = match.group("pr_arg")
-            logger.debug(f"Parsed test argument -> pr_argument: {self.pr_argument}")
+        self.parse_known_arguments(args)
+        self.parse_unknown_arguments(unknown_args)
+
+    @property
+    def parser(self):
```
nforro, line 105:
```suggestion
    def parser(self) -> argparse.ArgumentParser:
```

packit_service/worker/helpers/testing_farm.py
lbarcziova: yes, practically both, like if someone passes `--env ENV_VAR_TO_UNSET=  --env OTHER_ENV_VAR=value`, which would lead to the value being None

lbarcziova (commented):
thanks for the improvements! As for me, I would just agree on what to do with the unsetting the env var, and then after squashing the commits this could be merged! :rocket:

Frawless:
@lbarcziova I updated the code, please take a look if the unset functionality is fine.

lbarcziova (approved):
LGTM, thanks a lot!

let's test in staging and adjust in followup PRs if needed

PR #2586, opened 2024-10-17
Draft: GitLab events / parsing refactoring

### LINKED ISSUE #2471: Prepare and refactor events/parsers for CentOS Stream dist-git comments
### TODO

- [ ] Decide whether it makes sense to have duplicitous “events” for GitLab (downstream and upstream), or handle them in a similar way as @majamassarini did with the mixins (common GitLab part, common downstream/upstream part), although I'm not sure how feasible it is to use mixins here :thinking: 
- [ ] Refactor parsing of GitLab events (since there will be _upstream_ events directly via GitLab webhook and _downstream_ events provided via Fedora Messaging)
- [ ] Parse the dist-git GitLab events ([example](https://apps.fedoraproject.org/datagrepper/v2/id?id=8cae2ffc-5bab-4a6b-81f7-2ac9f93e66c4&is_raw=true&size=extra-large) of such event)
- [ ] Revamp the hierarchy for dist-git events, if necessary (afterwards the events will be coming from both GitLab and Pagure)

### ACs

- [ ] There is a **one** parser for each GitLab event, e.g., parsing of a push to either upstream on GitLab or downstream on GitLab should be parsed by one function only, there should be no duplication present that would differentiate upstream from downstream on the parsing level.
- [ ] Downstream events should have common ancestors that allow handlers to be handling **just** one ancestor for downstream.
   i.e. there should be a notion of a dist-git event in general (such as `AbstractDownstreamEvent`) and more specific ones (such as `AbstractDownstreamPushEvent`)
- [ ] Current downstream workflows **must not** be broken.
- [ ] Current upstream workflows on GitLab **must not** be broken.

### Links

Part of #2191

mfocko:
TODO:

- [ ] ‹fill in›

Fixes #2471

Related to

Merge before/after

RELEASE NOTES BEGIN

Packit now supports automatic ordering of ☕ after all checks pass.

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/jobs.py
+++ b/packit_service/worker/jobs.py
@@ -640,7 +640,16 @@ class SteveJobs:
         Returns:
             List of job configs.
         """
-        matching_jobs = []
+
+        def compare_jobs_without_triggers(a, b):
+            # check if two jobs are the same or differ only in trigger
+            ad = dict(a.__dict__)
+            ad.pop("trigger")
+            bd = dict(b.__dict__)
+            bd.pop("trigger")
+            return ad == bd
+
+        matching_jobs: List[JobConfig] = []
         if isinstance(self.event, PullRequestCommentPagureEvent):
             for job in self.event.packages_config.get_job_views():
                 if (
@@ -650,9 +659,15 @@ class SteveJobs:
                     and self.event.job_config_trigger_type
                     == JobConfigTriggerType.pull_request
                 ):
-                    # A koji_build or bodhi_update job with comment or koji_build trigger
-                    # can be re-triggered by a Pagure comment in a PR
-                    matching_jobs.append(job)
+                    # avoid having duplicate koji_build jobs
+                    if job.type != JobType.koji_build or not any(
+                        j
+                        for j in matching_jobs
+                        if compare_jobs_without_triggers(job, j)
+                    ):
+                        # A koji_build or bodhi_update job with comment or koji_build trigger
+                        # can be re-triggered by a Pagure comment in a PR
+                        matching_jobs.append(job)
                 elif (
                     job.type == JobType.pull_from_upstream
                     and job.trigger == JobConfigTriggerType.release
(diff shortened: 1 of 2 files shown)


majamassarini (dismissed):
Thank you 🙏🏻 
Looking at this I think I would like to have more namespaces in our code. I would like to have `github.PullRequestEvent` or even better `event.github.PullRequest` instead of `GithubPullRequestEvent`. I don't know what others think about it and I don't know if this could be the right moment for such a big refactoring... but still I think it would be nicer 🙂

mfocko:
> Thank you 🙏🏻 Looking at this I think I would like to have more namespaces in our code. I would like to have `github.PullRequestEvent` or even better `event.github.PullRequest` instead of `GithubPullRequestEvent`. I don't know what others think about it and I don't know if this could be the right moment for such a big refactoring... but still I think it would be nicer 🙂

I'm thinking about it too :D and I will need to do some changes around anyways, so… the biggest issue is fixing the imports afterwards :D

mfocko:
> I'm thinking about it too :D and I will need to do some changes around anyways, so… the biggest issue is fixing the imports afterwards :D

I think it will break everything, so… let's do it \o/

mfocko:
How about…
```
packit_service.worker.events
|- abstract
|  |- AbstractForgeIndependentEvent
|  `- AbstractResultEvent
|- github
|  |- abstract
|  |  `- AbstractGithubEvent
|  |- pull_request
|  |  |- Comment
|  |  `- Update # ← not sure here, right now it's just PullRequestGithubEvent, covers create and push
…
```

@lbarcziova @majamassarini @packit/the-packit-team

The bigger issue how to fit the downstream events in there… Do they even count as downstream-only if they come from both GitLab and Pagure? It almost feels like there should be two parsers (parsing Fedora Messaging event) returning the same type (downstream event) regardless of whether it's from GitLab or Pagure. We shouldn't even care about this…

I don't want to add another layer of `events.upstream.github.…` :/

lbarcziova:
>It almost feels like there should be two parsers (parsing Fedora Messaging event) returning the same type (downstream event) regardless of whether it's from GitLab or Pagure. 

you are right, this is probably how it should be done the most correctly

but otherwise the proposal looks good to me! Is it manageable to have it like that considering your note about upstream vs downstream? @mfocko

mfocko:
> you are right, this is probably how it should be done the most correctly
> 
> but otherwise the proposal looks good to me! Is it manageable to have it like that considering your note about upstream vs downstream? @mfocko

maybe
```
• packit_service.worker.events
  • copr
  • openscanhub
  • downstream
  • github
  • gitlab
```
?

I think that for the beginning I could just rename Pagure to downstream (hide the fact it's pagure) and try to make it more abstract for supporting GitLab (or potentially Forgejo too, if and when Fedora migrates)

mfocko:
GitHub PLEASE /o

Issue #2527, opened 2024-09-12
The number of Koji build jobs triggered when using sidetags can be overwhelming


nforro:
For example, this is from the last release of Packit:

![sidetags](https://github.com/user-attachments/assets/548dead7-ab22-457f-884b-a76ec25ccbf1)

There are two issues here:

- The builds without a task ID are in `skipped` state because of missing dependencies, but perhaps it could be possible to check for the  dependencies earlier and not record the builds in the DB at all.
- The purple builds with a task ID are in `pending` state, but they shouldn't be there at all, each of them complements a successful build that was for some reason triggered for a second time. Another issue is that their state should be `error` rather than `pending` (the tasks failed with `GenericError: Build already in progress`).

PR #2588, opened 2024-10-20
Reduce the number of Koji build jobs related to a sidetag(s) release

nforro:
Fixes https://github.com/packit/packit-service/issues/2527.


lbarcziova (approved):
lgtm

nforro:
@lbarcziova I realized that it would make sense to still check Koji (after checking the DB) for an already processed NVR to keep considering builds triggered outside Packit (even though it probably won't help in most cases as the time between running `koji build` and the corresponding build becoming visible in Koji API is pretty long). Let me fix that.

lbarcziova (approved):

PR #2589, opened 2024-10-21
Implement non-PR retriggering

### LINKED ISSUE #1062: Be able to retrigger the non-PR jobs
Currently, we can retrigger the jobs only in those two ways:

* PR comment for PR jobs => This needs to stay the same.
* Issue comments for `propose-downstream` => There is some hardcoding done in the code to match the release object.

We have at least two ways how we can allow retriggers or reactions with Packit:

* Create an issue with the known format. Some options (can be combined):
    * title == `/packit propose-downstream 0.3.0` / `/packit propose-downstream` / `/packit build master` / ....
        * Reopening of the issue triggers the action again.
    * Any title, but react on commands in description or comment (same syntax as requested in #540 or #970)
 * Create a commit comment with the command
     * How can we differentiate between branches and releases?
 * Implement elsewhere -- e.g. via dashboard

Some more ideas? How other projects do that?

lbarcziova:
TODO:

- [x] implement reacting to commit comments (eyes emoji) in ogr

Fixes #1062

RELEASE NOTES BEGIN

We have added support for retriggering jobs that are configured for `commit` and `release` trigger. Retriggering can be done via commit comments on the relevant commit, using the usual comments, such as `/packit build` or `/packit test`, but the branch or tag need to be specified like this (without the arguments, we will default to `commit` trigger for the default branch of the repository): 

```
/packit build --commit <branch-name>
```
or
```
/packit build --release <tag-name>
```

RELEASE NOTES END


lbarcziova:
recheck

packit_service/worker/events/comment.py
```diff
@@ -183,3 +190,125 @@ def get_dict(self, default_dict: Optional[dict] = None) -> dict:
...
+    def tag_name(self) -> Optional[str]:
+        if not self._tag_name and "--release" in self.comment:
+            release_match = re.search(r"--release[\s=](?P<release>\S+)", self.comment)
+            if release_match:
+                self._tag_name = release_match.group("release")
+        return self._tag_name
+
+    @property
+    def git_ref(self) -> Optional[str]:
+        if not self._git_ref and "--release" not in self.comment:
```
majamassarini, line 231: Just wondering if we don't need the `self._git_ref` attribute, somewhere in the code, when we already have the `self._tag_name` attribute. I would probably populate the `self._git_ref` when there is `--release` in the comment just to be safe. But if you think it is not necessary I am fine with it.

majamassarini (approved):
🚀

packit_service/worker/events/comment.py
lbarcziova: good points! I will rather change it

PR #2590, opened 2024-10-21
GitLab events / parsing refactoring

mfocko:
### TODO

- [ ] Open up follow-up issues.

- [x] We create instances of **abstract** classes in tests… I have implemented `.event_type()` for those with an _assertion_ to check that the tests are actually running…

### Comments / questions from the commits

- [x] Manual caching of the `db_project_object`, `db_project_event` and `project` (currently unimplemented method in superclasses, property that caches to an attribute)
  Comment from the review:
  ```
  Are these supposed to be abstract methods?
  Also, if they're abstract (and cached, based on the commits from blame),
  why don't we just use cachedproperty after abstractmethod? Having these
  return ‹None› by default »can« result in unimplemented override somewhere
  down the chain.
  ```
  - [x] Leaving as is for now, both `project_object` and `project_event` are constructed at the same time via one helper function, therefore the caching can't be _simply_ wrapped with a `cached_property`
  - [ ] Ideally drop explicit getters and implement cached abstract properties (even though I recall some issues with type checking that, or it's not very bulletproof…)
- [x] `Event.pre_check` returns `True` by default…
  - [x] after a discussion with @lbarcziova just adding a note why we have chose the specific default there
  - [ ] log a warning
  - [ ] default to `False`

### Reconsider structuring

- [x] `forge.push.Push` (respectively `gitlab.push.TagPush`) → `forge.push.Commit` or `forge.push.Tag`
- [x] `anitya.base` → `anitya.events` (not used directly though, events are exposed via `__init__.py` as it doesn't make much sense to structure them further)
- [x] `forge.pr.Synchronize` doesn't make sense, it is “some kind of an action” on the PR/MR, but Synchronize has been inspired by the GitHub itself (and it's an example of an action on the PR), but I don't have a better name… maybe even `forge.pr.Change` / `forge.pr.Action`?
- [x] Enumerations: there are some enums in `events.enums` for PR/MRs, comments and FedMsg…
  - [x] Combination of both… as for the PR and comment actions they are shared by GitHub and Pagure, therefore I kept those shared actions in the `events.enums`, while moving the GitLab enum that's also shared by GitLab-only events to `events.gitlab.enums` and as for the OpenScanHub, the least disruptive change (since the enum is tied only to the task) to `events.openscanhub.task.Status`
  - [ ] Follow OpenScanHub implementation by Maja: _nests_ the enumerations into the event type itself, which is a possible way, but it looks a bit weird used in code
  - [ ] Have a separate module (e.g. `openscanhub.enum`) for all of them consistently
  - [ ] Going from the current change:
    ```diff
    -        if self.event.status == OpenScanHubTaskFinishedEvent.Status.success:
    +        if self.event.status == openscanhub.task.Finished.Status.success:
    ```
    
    I'd probably prefer `openscanhub.task.Status.success` :eyes:

Fixes

Supersedes #2586 (cause GitHub is a :flatbread:)

Merge before/after


mfocko:
@majamassarini @lbarcziova WDYT? At the moment it's broken everywhere possible :D it looks like we have also some pretty intrusive mocking in place, so this reshuffle opened the can of worms…

I still haven't made up my mind about the usage of the events, cause they can be imported as
```py
from packit_service.worker.events.github.pr import Synchronize
```

but at the same time I have a feeling that it won't look very good once it gets mixed up with the other forges, for now I replaced the imports with `as ‹original_name›`, so that it breaks as least stuff as possible, but going forward it doesn't feel very feasible… it probably needs a better hierarchy :thinking:

mfocko:
> [!TIP]
> I had a feeling something will crawl out, when I make `Event` an abstract class
> ```
> packit_service/worker/events/github/installation.py:41: error: Cannot instantiate abstract class "Installation" with abstract attributes "base_project" and "get_packages_config"  [abstract]
> ```

lbarcziova (commented):
the structure looks good to me :rocket:

> for now I replaced the imports with as ‹original_name›, so that it breaks as least stuff as possible, but going forward it doesn't feel very feasible
 
agreed it may be a good middle step

majamassarini:
> @majamassarini @lbarcziova WDYT? At the moment it's broken everywhere possible :D it looks like we have also some pretty intrusive mocking in place, so this reshuffle opened the can of worms…
> 
> I still haven't made up my mind about the usage of the events, cause they can be imported as
> 
> ```python
> from packit_service.worker.events.github.pr import Synchronize
> ```
> 
> but at the same time I have a feeling that it won't look very good once it gets mixed up with the other forges, for now I replaced the imports with `as ‹original_name›`, so that it breaks as least stuff as possible, but going forward it doesn't feel very feasible… it probably needs a better hierarchy 🤔

Thanks for starting this big refactoring, to me it looks good!
If I understand your concern, I think in future we should avoid `import` like this
```python
from packit_service.worker.events.github.pr import Synchronize
```
and use instead the namespaces we have created
```python
from packit_service.worker.events import github, gitlab, pagure
[...]
github.pr.Synchronize
[...]
gitlab.pr.Synchronize
```
So I think the hierarchy is fine 🤔

mfocko:
> Thanks for starting this big refactoring, to me it looks good! If I understand your concern, I think in future we should avoid `import` like this
> 
> ```python
> from packit_service.worker.events.github.pr import Synchronize
> ```
> 
> and use instead the namespaces we have created
> 
> ```python
> from packit_service.worker.events import github, gitlab, pagure
> [...]
> github.pr.Synchronize
> [...]
> gitlab.pr.Synchronize
> ```
> 
> So I think the hierarchy is fine 🤔

oh i see, so you **really** wanna use it like that :grin: I was a bit worried how _explicit_ it will be, but I guess it's better

lbarcziova (commented):
I really like how you broke this up into smaller pieces, this is so much more readable, good job!

packit_service/worker/events/event.py
```diff
@@ -392,6 +392,14 @@ def __init__(self, created_at: Optional[Union[int, float, str]] = None):
         self._db_project_object: Optional[AbstractProjectObjectDbType] = None
         self._db_project_event: Optional[ProjectEventModel] = None
 
+    @classmethod
+    @abstractmethod
+    def event_type(cls) -> str:
+        """
+        ‹TODO›
+        """
+        ...
```
majamassarini, line 409: Have you tried to build it dynamically? I think something like the following should work?
```
cls.__module__ + "." + cls.__name__
```

packit_service/worker/events/copr.py
```diff
@@ -67,6 +68,11 @@ def __init__(
         self.pkg = pkg
         self.timestamp = timestamp
 
+    @classmethod
+    def event_type(cls) -> str:
+        assert os.environ.get("PYTEST_VERSION"), "Should be initialized only during tests"
```
majamassarini, line 73: Unfortunately in Visual Studio Code this trick does not work, it would be hard to use concrete events instead of abstract ones in tests?

majamassarini (commented):
This is really an huge refactoring and code now looks much better! Thanks a lot for the great work!

majamassarini:
> Event.pre_check returns True by default…

I agree that this isn't safe, we should try to turn it to `False` but I fear we will need to adjust something later. 

> forge.pr.Synchronize doesn't make sense, it is “some kind of an action” on the PR/MR, but Synchronize has been inspired > by the GitHub itself (and it's an example of an action on the PR), but I don't have a better name… maybe even forge.pr.Change / forge.pr.Action?

For my (bad) english `Synchronize`, `Change`, `Action` are almost the same, I would stick with `Synchronize` probably.

 > Enumerations: there are some enums in events.enums for PR/MRs, comments and FedMsg…
 > Follow OpenScanHub implementation by Maja: nests the enumerations into the event type itself, which is a possible way, but it looks a bit weird used in code

> Have a separate module (e.g. openscanhub.enum) for all of them consistently

Now that we have the `openscanhub` namespace I totally agree that we can move the enums there. 

> Going from the current change:
> -        if self.event.status == OpenScanHubTaskFinishedEvent.Status.success:
> +        if self.event.status == openscanhub.task.Finished.Status.success:
> I'd probably prefer openscanhub.task.Status.success 👀

Sure, I agree.

packit_service/worker/events/copr.py
mfocko: Just out of interest, what version of pytest are you running? I did unpin the version on Friday :eyes:

packit_service/worker/events/event.py
mfocko: I've tried just now, I have 2 failing tests and I'm not really sure what's the cause of the issue

mfocko: Oh I see… that includes »whole« path… I'm not sure if it's worth it :thinking:

majamassarini: We can always remove the unwanted prefix (or probably we need to use the new one everywhere?), so that we don't have to define the `event_type` for every subclass.

packit_service/worker/events/copr.py
majamassarini: Nevermind, I thought that using pytest directly I had the `PYTEST_VERSION` env variable already initialized (and not when calling tests from VS code) but I see now that it isn't initialized even when calling pytest from command line. So now we should always define this variable first? right?

mfocko: It's defined since pytest 8.2.0; we had pinned an older version till Friday when I unpinned and switched to the pytest from the PyPI… I really hate the fact we have different environments (for packit we run TF **and** Zuul), and `Makefile` also allows running tests from the container (they can easily diverge in the versions… I switched the pytest to PyPI exactly because of the different version in Fedora and EPEL).

lbarcziova:
> I agree that this isn't safe, we should try to turn it to False but I fear we will need to adjust something later.

as for my understanding, this would mean that we would require for each subclass of Event to explicitly add override returning True, if there are no actual pre-checks to be run. And since in events, we mostly don't even use the pre-check (as opposed to handlers), therefore I don't think we need to spend time on doing the switch here.

> forge.pr.Synchronize doesn't make sense, it is “some kind of an action” on the PR/MR, but Synchronize has been inspired by the GitHub itself (and it's an example of an action on the PR), but I don't have a better name… maybe even forge.pr.Change / forge.pr.Action?

I prefer Action, as even in your description you used this word, sounds the most fitting to me

> I'd probably prefer openscanhub.task.Status.success 👀

 :+1:
 
 > forge.push.Push (respectively gitlab.push.TagPush) → forge.push.Commit or forge.push.Tag
 
 +1, sounds more precise

majamassarini (approved):
🚀

lbarcziova (approved):
thanks a lot! Let's get this to staging 🚢

PR #2598, opened 2024-10-24
ci(black): set line length to 100

### LINKED ISSUE #2597: black and ruff are fighting with each other
black has a default line length limit of 88 characters, but in case it's not possible to achieve it, it seems to give up and try to minimize the number of lines such chunk of code spans over. But ruff with its line length limit set to 100 doesn't accept that and formats the chunk so it fits, or at least complains about line being too long.

Here is the code:

black variant:

```python
                    existing_models = BodhiUpdateTargetModel.get_all_successful_or_in_progress_by_nvrs(
                            target_model.koji_nvrs,
                    )
```

ruff variant:

```python
                    existing_models = (
                        BodhiUpdateTargetModel.get_all_successful_or_in_progress_by_nvrs(
                            target_model.koji_nvrs,
                        )
                    )
```

See also here: https://github.com/packit/packit-service/pull/2596

mfocko:
* Also run ruff before formatting, still **doesn't** resolve deps though

Fixes #2597

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,10 +3,18 @@
 # pre-commit install -t pre-commit -t pre-push
 
 repos:
-  - repo: https://github.com/psf/black
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    # Ruff version.
+    rev: v0.0.291
+    hooks:
+      - id: ruff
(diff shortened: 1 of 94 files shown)


mfocko:
Closed in favor of #2599

PR #2599, opened 2024-10-24
ci(ruff): enable ruff-format

### LINKED ISSUE #2597: black and ruff are fighting with each other

mfocko:
* Enable ruff-format
* Disable black

Fixes #2597

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,10 +3,12 @@
 # pre-commit install -t pre-commit -t pre-push
 
 repos:
-  - repo: https://github.com/psf/black
-    rev: 24.8.0
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.0.291
     hooks:
-      - id: black
+      - id: ruff
+        args: [--fix, --exit-non-zero-on-fix]
+      - id: ruff-format
   - repo: https://github.com/pre-commit/mirrors-prettier
     rev: v4.0.0-alpha.8
     hooks:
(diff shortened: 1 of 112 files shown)


lbarcziova (commented):
+1 for this solution, thanks!

mfocko:
> +1 from me too, what's up with the failed test though?
> 
> Ah, it seems to convert `list(x)[0]` to `next(x)`, where `x` is a set. I would say the conversion is fine, but getting the first element of a set seems a bit nonsensical 🙂

Oh… so that's why it wasn't autofix :D yet it let me do that :DD

nforro:
I suppose it's like that because it is known that the set contains only one element. In that case it would probably be better to change it from `y = list(x)[0]` to `[y] = x`.

nforro (approved):

PR #2607, opened 2024-10-29
Do not submit scan multiple times for the same build

majamassarini:
Should fix: #2604

The fix is described in more detail [here...](https://github.com/packit/packit-service/issues/2604#issuecomment-2444321483)

### DIFF
--- a/packit_service/worker/helpers/open_scan_hub.py
+++ b/packit_service/worker/helpers/open_scan_hub.py
@@ -65,11 +65,19 @@ class OpenScanHubHelper:
             logger.debug("No base build job needed for diff scan found in the config.")
             return
 
-        logger.info("Preparing to trigger scan in OpenScanHub...")
-
-        if not (base_srpm_model := self.get_base_srpm_model(base_build_job)):
+        if self.build.scan:
+            # see comment https://github.com/packit/packit-service/issues/2604#issuecomment-2444321483
+            logger.debug(
+                f"Scan for build {self.build.id} already submitted, "
+                "scan task_id {self.build.scan.task_id}."
+            )
             return
 
+        if not (base_srpm_model := self.get_base_srpm_model(base_build_job)):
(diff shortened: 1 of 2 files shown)


packit_service/worker/helpers/open_scan_hub.py
```diff
@@ -78,6 +78,13 @@ def handle_scan(self):
 
             build_dashboard_url = get_copr_build_info_url(self.build.id)
 
+            if self.build.scan:
+                # see comment https://github.com/packit/packit-service/issues/2604#issuecomment-2444321483
+                logger.debug(
+                    f"Scan for build {self.build.id} already submitted, "
+                    "scan task_id {self.build.scan.task_id}."
+                )
+                return
```
mfocko, line 87: Why here? How about

```py
        if not (base_build_job := self.find_base_build_job()):
            logger.debug("No base build job needed for diff scan found in the config.")
            return

        if self.build.scan:
            # see comment https://github.com/packit/packit-service/issues/2604#issuecomment-2444321483
            logger.debug(
                f"Scan for build {self.build.id} already submitted, "
                "scan task_id {self.build.scan.task_id}."
            )
            return

        if not (base_srpm_model := self.get_base_srpm_model(base_build_job)):
            logger.debug("Successful base SRPM build has not been found.")
            return

        logger.info("Preparing to trigger scan in OpenScanHub...")
        srpm_model = self.build.get_srpm_build()

        with tempfile.TemporaryDirectory() as directory:
```

I would say that we already have the `.build.scan` populated, even before we try to find base SRPM model **and** download the SRPMs for the scan :eyes: Or am I missing something there?

majamassarini: No no, you are right, let's me move it on the top.

lbarcziova (approved):

majamassarini:
recheck

PR #2596, opened 2024-10-23
Bodhi update fixes

nforro:
Fixes https://github.com/packit/packit-service/issues/2573.

Issue #2426, opened 2024-05-15
Failed Koji build and Bodhi update jobs with no relevant error message


lbarcziova:
When a Koji build fails to be triggered, user is not really informed about the cause of the failure:
- dashboard [example](https://dashboard.stg.packit.dev/results/koji-builds/463)
- issue [example](https://github.com/packit/ogr/issues/851)

majamassarini:
For the context, I had triggered the Packit exception with this PR https://github.com/packit/ogr/pull/852
I have created the PR to play with the command line and `build in-koji` a scratch build (this was successful). I was testing scratch builds as a way to "trigger" Koschei and I have broken the `ogr` code on purpose to get a `reverse dependency build failure on Koschei`.

lbarcziova:
I didn't mean to close this yet, the service part is done, and I will create followup PR for dashboard.

PR #2609, opened 2024-10-29
Make sure `bodhi_update` job runs don't get stuck in `queued` state

nforro:
When an unhandled exception occurs, a Sentry issue is created but the job run stays in `queued` state, which breaks the condition that is supposed to prevent submitting the same update more than once, as it considers that job run to be in progress. Make sure the state is updated when that happens.

Question is, in such case should the exception and traceback be stored and issue with it created in the issue repository or do we consider that internal?

Related to https://github.com/packit/packit-service/pull/2596.

### DIFF
--- a/packit_service/worker/handlers/bodhi.py
+++ b/packit_service/worker/handlers/bodhi.py
@@ -192,6 +192,18 @@ class BodhiUpdateHandler(
                 target_model.set_status("error")
                 target_model.set_data({"error": error})
 
+            except Exception as ex:
+                if self.celery_task and not self.celery_task.is_last_try():
+                    target_model.set_status("retry")
+                    raise
+
+                error = f"Internal error, please contact us: {ex}"
+                errors[target_model.target] = error
+
+                target_model.set_status("error")
+                target_model.set_data({"error": error})
+                raise
+
         if errors:
             self.report_in_issue_repository(errors=errors)
 


mfocko (approved):

packit_service/worker/handlers/bodhi.py
```diff
@@ -192,6 +192,10 @@ def run(self) -> TaskResults:
                 target_model.set_status("error")
                 target_model.set_data({"error": error})
 
+            except Exception:
+                target_model.set_status("error")
```
lbarcziova, line 196: could we also do the `target_model.set_data({"error": error})`?

lbarcziova (approved):

packit_service/worker/handlers/bodhi.py
nforro: `error` being the exception and traceback? So only storing it in the DB, not creating a user-facing issue?

lbarcziova: yes, I meant the exception and traceback. But if it is the case of our unexpected bug, maybe we could also mark that in the `error`, something like `Internal error, please contact us`, and we could also prioritise #2426 for that.

lbarcziova: and as for creating user-facing issue, I would do the same, like a message about `Internal error`, but we need to account for the retries the same way as in the `PackitException` branch to avoid duplicated issues/comments

nforro:
@lbarcziova Do you think something like this could work? Or do we want to suppress the exception until the last try?

lbarcziova (approved):
thanks, this looks good to me!

PR #2617, opened 2024-11-05
Try to minimize concurrency problems

majamassarini:
Related to #2604

We have concurrency issues. 
This change does not solve them but could minimize them.

Here an example log of the concurrency issue before my change:

**worker1**
```
11/5/24 11:47:31.688 AM	2024-11-05T11:47:31.688406049+00:00 stderr F [2024-11-05 11:47:31,688: DEBUG/MainProcess] task.run_copr_build_end_handler[8d4413e0-ba73-4f78-a14f-487b824a1b0f] Searching for base build for 7c08e96170470eb6e3cb99fbf8b705862a1e0a58 commit in @teemtee/tmt Copr project in our DB.
[...]
11/5/24 11:47:32.592 AM	2024-11-05T11:47:32.592982751+00:00 stderr F [2024-11-05 11:47:32,592: DEBUG/MainProcess] task.run_copr_build_end_handler[8d4413e0-ba73-4f78-a14f-487b824a1b0f] Command: osh-cli version-diff-build --srpm=/tmp/tmp27a0u3z_/tmt-1.39.dev888-1.20241105114529529022.pr3342.18.g3fc4607c.src.rpm --base-srpm=/tmp/tmp27a0u3z_/tmt-1.39.dev17+g7c08e961-main.src.rpm --config=fedora-rawhide-x86_64 --nowait --json --comment=Submitted via Packit Service for https://dashboard.packit.dev/results/copr-builds/1994869
11/5/24 11:47:39.070 AM	2024-11-05T11:47:39.070934307+00:00 stderr F [2024-11-05 11:47:39,070: INFO/MainProcess] task.run_copr_build_end_handler[8d4413e0-ba73-4f78-a14f-487b824a1b0f] Scan submitted successfully.
```

**worker2**
```
11/5/24 11:47:30.159 AM	2024-11-05T11:47:30.159087194+00:00 stderr F [2024-11-05 11:47:30,158: DEBUG/ForkPoolWorker-1] task.babysit_copr_build[fd022210-82ed-43ff-9494-18705b393847] Searching for base build for 7c08e96170470eb6e3cb99fbf8b705862a1e0a58 commit in @teemtee/tmt Copr project in our DB.
[...]
11/5/24 11:47:30.795 AM	2024-11-05T11:47:30.795838802+00:00 stderr F [2024-11-05 11:47:30,795: DEBUG/ForkPoolWorker-1] task.babysit_copr_build[fd022210-82ed-43ff-9494-18705b393847] Command: osh-cli version-diff-build --srpm=/tmp/tmpg70py27t/tmt-1.39.dev888-1.20241105114529529022.pr3342.18.g3fc4607c.src.rpm --base-srpm=/tmp/tmpg70py27t/tmt-1.39.dev17+g7c08e961-main.src.rpm --config=fedora-rawhide-x86_64 --nowait --json --comment=Submitted via Packit Service for https://dashboard.packit.dev/results/copr-builds/1994869
11/5/24 11:47:32.774 AM	2024-11-05T11:47:32.774600167+00:00 stderr F [2024-11-05 11:47:32,774: INFO/ForkPoolWorker-1] task.babysit_copr_build[fd022210-82ed-43ff-9494-18705b393847] Scan submitted successfully.
```

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -43,7 +43,6 @@ from sqlalchemy import (
     null,
 )
 from sqlalchemy.dialects.postgresql import array as psql_array
-from sqlalchemy.exc import IntegrityError
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import (
     Session as SQLASession,
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -23,6 +23,7 @@ from urllib.parse import urlparse
 
 from cachetools import TTLCache, cached
 from cachetools.func import ttl_cache
+from packit.config import JobConfigTriggerType
 from sqlalchemy import (
     JSON,
     Boolean,
@@ -42,11 +43,11 @@ from sqlalchemy import (
     null,
 )
 from sqlalchemy.dialects.postgresql import array as psql_array
+from sqlalchemy.exc import IntegrityError
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import (
     Session as SQLASession,
 )
-from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import (
     relationship,
     scoped_session,
@@ -55,7 +56,6 @@ from sqlalchemy.orm import (
 from sqlalchemy.sql.functions import count
 from sqlalchemy.types import ARRAY
 
-from packit.config import JobConfigTriggerType
 from packit_service.constants import ALLOWLIST_CONSTANTS
 
 logger = logging.getLogger(__name__)
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -23,7 +23,6 @@ from urllib.parse import urlparse
 
 from cachetools import TTLCache, cached
 from cachetools.func import ttl_cache
-from packit.config import JobConfigTriggerType
 from sqlalchemy import (
     JSON,
     Boolean,
@@ -47,6 +46,7 @@ from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import (
     Session as SQLASession,
 )
+from sqlalchemy.exc import IntegrityError
 from sqlalchemy.orm import (
     relationship,
     scoped_session,
@@ -55,6 +55,7 @@ from sqlalchemy.orm import (
 from sqlalchemy.sql.functions import count
 from sqlalchemy.types import ARRAY
 
+from packit.config import JobConfigTriggerType
 from packit_service.constants import ALLOWLIST_CONSTANTS
 
 logger = logging.getLogger(__name__)
@@ -2328,6 +2329,46 @@ class CoprBuildTargetModel(GroupAndTargetModelConnector, Base):
             scan.copr_build_target = self
             session.add(scan)
             return scan
+
+    @contextmanager
+    def add_scan_transaction(self) -> Generator["OSHScanModel"]:
+        """
+        Context manager that creates a ScanModel upon entering the context,
+        provides a corresponding instance of `ScanModel` to be updated within the context
+        and commits the changes upon exiting the context, all within a single transaction.
+
+        This locking mechanism is working on the assumption that just a single scan model
+        for build can exist.
+
+        raise: IntegrityError if the scan model already exists
+        """
+        session = singleton_session or Session()
+        session.begin()
+
+        try:
+            scan = OSHScanModel()
+            scan.copr_build_target = self
+            session.add(scan)
+            session.commit()
+        except Exception as ex:
+            logger.warning(f"Exception while working with database: {ex!r}")
+            session.rollback()
+            raise
+
+        try:
+            yield scan
+        except Exception as ex:
+            logger.warning(f"{ex!r}")
+            session.rollback()
+            raise
+
+        try:
+            session.add(scan)
+            session.commit()
+        except Exception as ex:
+            logger.warning(f"Exception while working with database: {ex!r}")
+            session.rollback()
+            raise
 
 
 class KojiBuildGroupModel(ProjectAndEventsConnector, GroupModel, Base):
--- a/packit_service/worker/helpers/open_scan_hub.py
+++ b/packit_service/worker/helpers/open_scan_hub.py
@@ -9,12 +9,14 @@ from os import getenv
 from os.path import basename
 from pathlib import Path
 from typing import Optional
+from sqlalchemy.exc import IntegrityError
 
 from packit.config import (
     JobConfig,
     JobConfigTriggerType,
     JobType,
 )
+from packit.exceptions import PackitException
 
 from packit_service.constants import (
     OPEN_SCAN_HUB_FEATURE_DESCRIPTION,
@@ -22,6 +24,7 @@ from packit_service.constants import (
 from packit_service.models import (
     BuildStatus,
     CoprBuildTargetModel,
+    OSHScanStatus,
     SRPMBuildModel,
 )
 from packit_service.service.urls import get_copr_build_info_url
@@ -32,6 +35,10 @@ from packit_service.worker.helpers.build import CoprBuildJobHelper
 from packit_service.worker.reporting import BaseCommitStatus
 
 logger = logging.getLogger(__name__)
+
+
+class OSHNoFeedback(PackitException):
+    pass
 
 
 class OpenScanHubHelper:
@@ -65,14 +72,6 @@ class OpenScanHubHelper:
             logger.debug("No base build job needed for diff scan found in the config.")
             return
 
-        if self.build.scan:
-            # see comment https://github.com/packit/packit-service/issues/2604#issuecomment-2444321483
-            logger.debug(
-                f"Scan for build {self.build.id} already submitted, "
-                "scan task_id {self.build.scan.task_id}."
-            )
-            return
-
         if not (base_srpm_model := self.get_base_srpm_model(base_build_job)):
             logger.debug("Successful base SRPM build has not been found.")
             return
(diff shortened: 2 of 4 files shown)


mfocko (commented):
Do you think it would make sense to add an entry to the database **before** submitting the OSH build? Something like a placeholder that would ensure that the other worker/thread cannot submit it too?

mfocko:
Maybe a progression like `pending → { error, in progress, done }`?

majamassarini:
I am not sure I following you, from my point what you are suggesting will move the critical section on other lines.

The problem is to be sure that between checking the db and updating it there will be no another execution of the same code in another worker (that does the checking just before the first one has updated it).

The data in the database is already present: the scan entry for the build and it has already the status.

The existence of a scan entry for the build is checked before submitting and adding the scan to the db, but in this case between the checking and the update of the entry there is a too long operation (searching for the base rpm), so in the meantime the second worker has checked the entry (not yet updated).

I would say this PR should minimize occurrences of this issue reducing the time between checking the db and updating it.
To really solve this I think we should use something like https://github.com/brolewis/celery_mutex or https://github.com/vaidik/sherlock but honestly I would not introduce locks if we can make this happen rarely...

mfocko:
Yes, there are 2 workers racing to submit the build. The issue is that each worker:
- gets the base SRPM
- gets the PR SRPM
- submits the build
- creates DB entry

The issue here is that there's no possible way to tell whether some worker is already executing the submission of the scan. If we swap the order to

- creates DB entry
- gets the base SRPM
- gets the PR SRPM
- submits the build
- updates DB entry

the second worker will fail on the constraint, but at the same time won't submit the scan, handling of this issue is a bit questionable, but the condition could probably stay before you download any of the SRPMs (and handle this case instead of catching an exception from the SQLAlchemy)

Adding entry to the DB uses transaction, so it should create a lock on the DB, therefore they cannot race, **if** you create the DB entry beforehand, one worker would error out anyways and not submit the scan.

packit_service/models.py
```diff
@@ -2329,6 +2329,33 @@ def add_scan(self, task_id: int) -> "OSHScanModel":
             session.add(scan)
             return scan
 
+    @contextmanager
+    def add_scan_transaction(self, cls) -> Generator["OSHScanModel"]:
```
nforro, line 2333: What is `cls`?

packit_service/worker/helpers/open_scan_hub.py
```diff
@@ -78,42 +80,41 @@ def handle_scan(self):
...
+                        logger.debug("Something went wrong, skipping the reporting.")
+                        return
+
+                    logger.info("Scan submitted successfully.")
+
+                    response_dict = self.parse_dict_from_output(output)
+
+                    logger.debug(f"Parsed dict from output: {response_dict} ")
+
+                    scan = None
```
nforro, line 101: What is this supposed to achieve?

packit_service/models.py
majamassarini: Copy and paste error 😅

packit_service/models.py
```diff
@@ -2329,6 +2329,33 @@ def add_scan(self, task_id: int) -> "OSHScanModel":
...
+        """
+        Context manager that creates a ScanModel upon entering the context,
+        provides a corresponding instance of `ScanModel` to be updated within the context
+        and commits the changes upon exiting the context, all within a single transaction.
+
+        This locking mechanism is working on the assumption that just a single scan model
+        for build can exist and that we can commit the transaction after having collected
+        inside the context manager the task_id.
```
nforro, line 2341: Is this supposed to roll back if the scan already exists, raising an exception? I would expect a check for the existence of a scan.

packit_service/worker/helpers/open_scan_hub.py
majamassarini: You mean try/catching the `IntegrityException`?

packit_service/models.py
nforro: I would expect the context manager to behave somewhat like this, yielding a newly created scan model for updating or `None` if it already exists (the scan has been submitted):

```python
        session = singleton_session or Session()
        session.begin()

        submit_scan = False

        try:
            if not self.scan:
                submit_scan = True
                scan = OSHScanModel()
                scan.copr_build_target = self
                session.add(scan)
        except Exception as ex:
            logger.warning(f"Exception while working with database: {ex!r}")
            session.rollback()
            raise

        try:
            yield scan if submit_scan else None
        except Exception:
            session.rollback()
            raise

        try:
            if submit_scan:
                session.add(scan)
                session.commit()
        except Exception as ex:
            logger.warning(f"Exception while working with database: {ex!r}")
            session.rollback()
            raise
```

packit_service/worker/helpers/open_scan_hub.py
nforro: I mean why are you assigning `None` to target variable of the context. But I don't understand the `IntegrityException` either.

packit_service/models.py
majamassarini: If we doesn't commit before yielding the scan, we might be doing a "dirty read" and thus we may submit multiple scans. So I don't think I can rely upon a read. From my point of view this works only if we update the db before running the scan... The only thing I see that can ensure us that we will not submit multiple scans is using the db constraints that will raise an exception when updating the db.

As I said I don't like this solution, is tightly coupled with the db design, instead of being just a generic locking mechanism that we could use in a more explicit way.

I would prefer be able to make atomic just the two lines of code I need, but since we are doing it in this way I agree to be consistent:
```
with some_lock:
    if not self.build.scan:
        run_ohs_build()
```

nforro: > we might be doing a "dirty read"

Where? How? You are in the middle of a transaction, all other workers are waiting to enter their transactions.

majamassarini: Oh, my bad. I didn't imagine the transaction as blocking 😅
I was thinking we were using something like https://en.wikipedia.org/wiki/Optimistic_concurrency_control...

In this case I can make the read. Still I don't like it 🙂 Blocking the db for such a long time I don't think is a good idea.

majamassarini: I don't think we are overriding the db defaults via SqlAlchemy, so it should be as you said.
So you are right, we are using a level of isolation that does not allow to have inconsistent readings.
I will keep your code 🙂

From https://www.postgresql.org/docs/current/sql-set-transaction.html the defaults is the following.
```
READ COMMITTED
A statement can only see rows committed before it began. This is the default.
```

nforro: > ```
> READ COMMITTED
> A statement can only see rows committed before it began. This is the default.
> ```

Right, but unfortunately that doesn't prevent race conditions, because it will allow all workers to continue and commit the changes one after the other, if I understand it correctly. No, I thought SQLAlchemy session works as a locking mechanism, but I'm not so sure anymore. Let me do a test.

nforro: Ok, I tested it and it indeed doesn't work, we would have to introduce explicit locking with something like `LOCK TABLE sidetags IN ACCESS EXCLUSIVE MODE`, which is far from being elegant.
However I found out one can lock an individual record for updating, with [sqlalchemy.orm.Query.with_for_update()](https://docs.sqlalchemy.org/en/20/orm/queryguide/query.html#sqlalchemy.orm.Query.with_for_update), I tested it and it indeed makes all workers block until the one updating the record is finished and commits or rollbacks the transaction. I think that's perfect for our use-case. I suppose your context manager could look something like this:

```python
@contextmanager
def add_scan_for_updating(self):
    with sa_session_transaction(commit=True) as session:
        existing_scan = (
            session.query(OSHScanModel)
            .filter_by(copr_build_target_id=self.id)
            .with_for_update()
            .first()
        )
        if existing_scan:
            new_scan = None
        else:
            new_scan = OSHScanModel()
            new_scan.copr_build_target = self
            session.add(new_scan)
        try:
            yield new_scan
        except Exception:
            session.rollback()
            raise
        else:
            if new_scan:
                session.add(new_scan)
```

packit_service/models.py
```diff
@@ -2329,6 +2329,37 @@ def add_scan(self, task_id: int) -> "OSHScanModel":
...
+            except Exception as ex:
+                logger.warning(f"Exception while working with database: {ex!r}")
+                session.rollback()
+                raise
+
+            try:
+                if new_scan:
+                    yield new_scan
+                    session.add(new_scan)
+                    session.commit()
```
nforro, line 2358: `sa_session_transaction(commit=True)` does the commit for you. As well as the rollback above and below. The only issue is that a potential exception will be prefixed with a DB exception text even if it's not DB related, not sure how to elegantly avoid that.

packit_service/models.py
```diff
@@ -2329,6 +2329,37 @@ def add_scan(self, task_id: int) -> "OSHScanModel":
...
+                    new_scan.copr_build_target = self
+                    session.add(new_scan)
+            except Exception as ex:
+                logger.warning(f"Exception while working with database: {ex!r}")
+                session.rollback()
+                raise
+
+            try:
+                if new_scan:
+                    yield new_scan
```
nforro, line 2356: You can't yield conditionally, otherwise you'll get a "generator didn't yield" exception.

packit_service/models.py
```diff
@@ -2329,6 +2329,37 @@ def add_scan(self, task_id: int) -> "OSHScanModel":
             session.add(scan)
             return scan
 
+    @contextmanager
+    def add_scan_transaction(self) -> Generator["OSHScanModel"]:
+        """
+        Context manager that creates a ScanModel upon entering the context,
+        provides a corresponding instance of `ScanModel` to be updated within the context
+        and commits the changes upon exiting the context, all within a single transaction.
```
nforro, line 2337:
```suggestion
        Context manager that, upon entering the context, creates a `OSHScanModel` if it doesn't
        exist yet, provides the instance as the target variable to be updated within the context
        and commits the changes done to it upon exiting the context. If a scan model already
        exists, the target variable is set to None and nothing is committed.
```

packit_service/models.py
majamassarini: Oh ok, then I really prefer the older solution that relied on committing and raising an Exception if the scan was already there. Instead of returning a `None` value and check in the caller is safe to proceed.

nforro: I don't follow. How is that going to prevent submitting the scan multiple times? We already concluded that transactions are run in parallel, there won't be any exception on commit, but even if there was, it's too late, the scan had already been submitted.

packit_service/models.py
```diff
@@ -2329,6 +2330,49 @@ def add_scan(self, task_id: int) -> "OSHScanModel":
...
+        session.begin()
+
+        try:
+            if not self.scan:
+                scan = OSHScanModel()
+                scan.copr_build_target = self
+                session.add(scan)
+                session.commit()
+            else:
+                raise IntegrityError("Scan model already exists")
```
nforro, line 2355:
```suggestion
            scan = OSHScanModel()
            scan.copr_build_target = self
            session.add(scan)
            session.commit()
```

The condition is redundant, if the scan already exists, `IntegrityError` will be raised either way.

majamassarini: 🤔 if there is already the scan, I don't `commit` anything and no exception will be raised before me yielding a scan. Since I don't want to submit a new scan or check for a `None` scan in the caller, I need to raise the exception here.

nforro (approved):
While I'm not a huge fan of this approach, I believe it does fix the concurrency issues.

PR #2623, opened 2024-11-08
Don't include OSH link in check title

lbarcziova:
In check title, markdown is not supported (and the URL is still present in the URL table).

RELEASE NOTES BEGIN

N/A

RELEASE NOTES END


siteshwar (approved):
LGTM. Thanks for the quick fix!

majamassarini (approved):

PR #2637, opened 2024-11-12
Run job_config handlers in new tasks

majamassarini:
There is a correlation between [this new exception in sentry](https://red-hat-it.sentry.io/issues/6057156460/)
and some [hard time limit errors](https://red-hat-it.sentry.io/issues/4323136874/), look at the graphs.

In splunk, the above celery task id, had more than 25,528 events associated and ran for a really long time.

If we run sequentially all the handlers for all the job_configs, for every entry in the db that needs babysitting we can reach the hard time limit for the task.
If we fail to babysit the "entries in the db", the tasks will sum up leading to more "hard time limit" exceptions.

### DIFF
--- a/packit_service/worker/helpers/build/babysit.py
+++ b/packit_service/worker/helpers/build/babysit.py
@@ -8,8 +8,10 @@ from datetime import datetime, timezone
 from enum import Enum
 from typing import Any
 
+import celery
 import copr.v3
 import requests
+from celery.canvas import Signature
 from copr.v3 import Client as CoprClient
 from requests import HTTPError
 
@@ -53,6 +55,13 @@ from packit_service.worker.mixin import ConfigFromUrlMixin
 from packit_service.worker.parser import Parser
 
 logger = logging.getLogger(__name__)
+
+
+def celery_run_async(signatures: list[Signature]) -> None:
+    logger.debug("Signatures are going to be sent to Celery (from babysit task).")
+    # https://docs.celeryq.dev/en/stable/userguide/canvas.html#groups
+    celery.group(signatures).apply_async()
+    logger.debug("Signatures were sent to Celery.")
 
 
 def check_pending_testing_farm_runs() -> None:
@@ -142,6 +151,7 @@ def update_testing_farm_run(event: TestingFarmResultsEvent, run: TFTTestRunTarge
     )
 
     event_dict = event.get_dict()
+    signatures = []
     for job_config in job_configs:
         package_config = (
             event.packages_config.get_package_config_for(job_config)
@@ -155,7 +165,9 @@ def update_testing_farm_run(event: TestingFarmResultsEvent, run: TFTTestRunTarge
         )
         # check for identifiers equality
         if handler.pre_check(package_config, job_config, event_dict):
-            handler.run_job()
+            signatures.append(handler.get_signature(event=event, job=job_config))
+
+    celery_run_async(signatures=signatures)
 
 
 def check_pending_copr_builds() -> None:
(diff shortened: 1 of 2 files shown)


packit_service/worker/helpers/build/babysit.py
```diff
@@ -155,7 +157,12 @@ def update_testing_farm_run(event: TestingFarmResultsEvent, run: TFTTestRunTarge
         )
         # check for identifiers equality
         if handler.pre_check(package_config, job_config, event_dict):
-            handler.run_job()
+            signatures.append(TestingFarmResultsHandler.get_signature(event=event, job=job_config))
```
nforro, line 160:
```suggestion
            signatures.append(handler.get_signature(event=event, job=job_config))
```

packit_service/worker/helpers/build/babysit.py
```diff
@@ -155,7 +157,12 @@ def update_testing_farm_run(event: TestingFarmResultsEvent, run: TFTTestRunTarge
...
+
+        logger.debug("Signatures are going to be sent to Celery (from update_testing_farm_run).")
+        # https://docs.celeryq.dev/en/stable/userguide/canvas.html#groups
+        celery.group(signatures).apply_async()
+        logger.debug("Signatures were sent to Celery.")
```
nforro, line 165: Shouldn't this be outside of the loop?

packit_service/worker/helpers/build/babysit.py
majamassarini: No, it should be a method of the class object and not of its instance.

packit_service/worker/helpers/build/babysit.py
```diff
@@ -55,6 +57,13 @@
 logger = logging.getLogger(__name__)
 
 
+def celery_run_async(signatures: list[signature]):
```
nforro, line 60:
```suggestion
def celery_run_async(signatures: list[Signature]) -> None:
```

packit_service/worker/helpers/build/babysit.py
```diff
@@ -8,8 +8,10 @@
 from enum import Enum
 from typing import Any
 
+import celery
 import copr.v3
 import requests
+from celery import signature
```
nforro, line 14:
```suggestion
from celery.canvas import Signature
```

packit_service/worker/helpers/build/babysit.py
```diff
@@ -55,6 +57,13 @@
 logger = logging.getLogger(__name__)
 
 
+def celery_run_async(signatures: list[signature]) -> None:
```
nforro, line 60:
```suggestion
def celery_run_async(signatures: list[Signature]) -> None:
```

[`signature`](https://github.com/celery/celery/blob/dfe241ecb2bb8f536ddab83b6e18a806029a227d/celery/canvas.py#L2365) is a function, you are not passing a list of functions.

majamassarini: I wasn't sure, but I was expecting mypy to complain 🤔...

nforro: Yes, it's weird that it didn't.

packit_service/worker/helpers/build/babysit.py
```diff
@@ -55,6 +57,13 @@
 logger = logging.getLogger(__name__)
 
 
+def celery_run_async(signatures: list[Signature]) -> None:
+    logger.debug("Signatures are going to be sent to Celery (from update_copr_build_state).")
```
lbarcziova, line 61: this can be called also from `update_testing_farm_run`, so the log should be adjusted

lbarcziova (approved):
this is a great improvement, thanks Maja!

PR #2639, opened 2024-11-13
Add footer about OSH integration to news footers

lbarcziova:


packit_service/worker/reporting/news.py
```diff
@@ -21,6 +21,8 @@ class News:
         "and be one of the first to know about all the news!",
         "Interested in the Packit team plans and priorities? "
         "Check [our epic board](https://github.com/orgs/packit/projects/7/views/29).",
+        "Want to catch security issues early? [Read more](https://packit.dev/posts/openscanhub-prototype)"
```
nforro, line 24: I would argue that static analysis can identify a lot more than just security issues, but I suppose `security issues` sounds more appealing than just `issues`.

nforro (approved):

packit_service/worker/reporting/news.py
lbarcziova: agreed, but for this purpose, this sounds more fitting to me

Issue #2602, opened 2024-10-24
Dist-git PR scratch build


lachmanfrantisek:
Implement a new handler for triggering dist-git PR scratch builds.

Agree withing a team what to use as an enabler for this feature. (This is just a prototype we can show to people and let them play a bit with it.)

lachmanfrantisek:
And a related one: https://pagure.io/fedora-ci/general/issue/496

PR #2650, opened 2024-11-21
Fedora CI scratch builds

lbarcziova:
Related to #2602

RELEASE NOTES BEGIN

N/A

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/handlers/abstract.py
+++ b/packit_service/worker/handlers/abstract.py
@@ -217,6 +217,7 @@ class TaskName(str, enum.Enum):
     tag_into_sidetag = "task.tag_into_sidetag"
     openscanhub_task_finished = "task.openscanhub_task_finished"
     openscanhub_task_started = "task.openscanhub_task_started"
+    downstream_koji_scratch_build = "task.run_downstream_koji_scratch_build_handler"
 
 
 class Handler(PackitAPIProtocol, Config):
--- a/packit_service/worker/handlers/abstract.py
+++ b/packit_service/worker/handlers/abstract.py
@@ -218,6 +218,7 @@ class TaskName(str, enum.Enum):
     openscanhub_task_finished = "task.openscanhub_task_finished"
     openscanhub_task_started = "task.openscanhub_task_started"
     downstream_koji_scratch_build = "task.run_downstream_koji_scratch_build_handler"
+    downstream_koji_scratch_build_report = "task.run_downstream_koji_scratch_build_report_handler"
 
 
 class Handler(PackitAPIProtocol, Config):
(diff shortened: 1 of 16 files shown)


packit_service/worker/helpers/fedora_ci.py
```diff
@@ -0,0 +1,45 @@
...
+
+from packit_service.worker.events import EventData
+from packit_service.worker.reporting import BaseCommitStatus
+from packit_service.worker.reporting.reporters.base import StatusReporter
+
+logger = logging.getLogger(__name__)
+
+
+class FedoraCIHelper:
+    status_name: str = "Packit - scratch build"
```
majamassarini, line 16: I am not sure I understand the need for a new helper. 

The ` KojiBuildJobHelper` has some more methods that could be useful when building koji builds that are missing in the generic `FedoraCIHelper` (if I am not wrong).
If we need the FedoraCIHelper because of the reporting (that is different in Pagure) then I am wondering if we could factor it out in the `StatusReporter` or somewhere else related with the reporting. And when creating the helpers (that we already have) we can choose which kind of reporter inject?

I just fear that we will need to duplicate all our helpers, creating a new "base".

lbarcziova: yes, at the moment, it really mostly wraps up the status reporting, in my initial implementation it contained some other logic common for the 2 new handlers. So that was the main idea, to have the common logic in here. Other thing is that I didn't want to adjust too much of existing code since this is more of a prototype and might be later changed/moved, therefore I wanted to keep it in one separate place. But having it as it is now, I am ok to just repeat the status reporting logic for the 2 handlers.

packit_service/worker/handlers/abstract.py
```diff
@@ -104,6 +107,30 @@ def _add_to_mapping(kls: type["JobHandler"]):
     return _add_to_mapping
 
 
+def reacts_to_as_fedora_ci(event: type["Event"]):
+    """
+    [class decorator]
+    Specify an event for which we want to use this handler as a Fedora CI.
+    Matching is done via isinstance so you can use some abstract class as well.
```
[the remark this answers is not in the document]
lbarcziova: :+1: , changed to markdown

packit_service/worker/handlers/distgit.py
```diff
@@ -736,6 +742,140 @@ def run(self) -> TaskResults:
...
+            str output
+        """
+        cmd = [
+            "koji",
+            "build",
+            "--scratch",
+            "--nowait",
+            self.dist_git_branch,
+            self.repo_url,
+        ]
```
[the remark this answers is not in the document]
lbarcziova: I picked this solution for staying consistent on how we do the other builds and for its simplicity and familiarity. But I can try out the API method as well and implement the needed wrapper in packit if you prefer it.

majamassarini (approved):

packit_service/worker/helpers/fedora_ci.py
lachmanfrantisek: It does not need to be done now, but I agree with Maja that we should be cautious about adding more helpers. (TBH, I don't think this was a good idea when introducing them... yes, I am probably the one to blame..;)

lachmanfrantisek (commented):
I know I am a bit late for the party but I like the implementation and don't have any concerns.. Thanks, Laura! I am really looking forward to all this!

Issue #1886, opened 2023-02-07
retest-failed triggers also successful test runs


TomasTomecek:
![image](https://user-images.githubusercontent.com/1662493/217184136-a6f41f32-dbd7-4a2d-89f3-07b95eaf06ef.png)

only 37 was failed, the comment triggered testing for 36 and rawhide as well

packit.yaml
```
specfile_path: ansible-bender.spec
actions:
  create-archive:
    - "python3 setup.py sdist --dist-dir ."
    - "sh -c 'echo ansible-bender-$(python3 setup.py --version).tar.gz'"
  get-current-version:
    - "python3 setup.py --version"
jobs:
- job: copr_build
  targets:
  - fedora-all
  trigger: pull_request
- job: tests
  trigger: pull_request
  targets:
  - fedora-all
- job: propose_downstream
  trigger: release
  dist_git_branches: [rawhide]
srpm_build_deps:
  - python3-pip # "python3 setup.py --version" needs it
  - python3-setuptools_scm
```

bocekm:
FYI, there's a known workaround - go to the Checks tab of the PR and re-run the failed jobs manually one by one.

lbarcziova:
This especially doesn't work when using identifiers, as we filter only targets and utilise the `*_targets_override` and do not consider the identifiers at all (encountered in https://github.com/neoave/mrack/pull/262)

PR #2658, opened 2024-11-26
Consider identifiers when retriggering

majamassarini:
Related to #1886


lbarcziova (commented):
looks good!

packit_service/worker/helpers/build/copr_build.py
```diff
@@ -80,8 +80,8 @@ def __init__(
         metadata: EventData,
         db_project_event: ProjectEventModel,
         job_config: JobConfig,
-        build_targets_override: Optional[set[str]] = None,
-        tests_targets_override: Optional[set[str]] = None,
+        build_targets_override: Optional[set[tuple[str, str]]] = None,
+        tests_targets_override: Optional[set[tuple[str, str]]] = None,
```
mfocko, line 84: I have a feeling you're changing the meaning of this attribute, so the name no longer holds very well, but… I guess that's a different rabbit hole to go down…

packit_service/worker/helpers/build/build_helper.py
```diff
@@ -314,14 +326,22 @@ def tests_targets_for_test_job(self, test_job_config: JobConfig) -> set[str]:
 
         if self.build_targets_override:
             logger.debug(f"Build targets override: {self.build_targets_override}")
-            for target in self.build_targets_override:
-                targets_override.update(
-                    self.build_target2test_targets_for_test_job(target, test_job_config),
-                )
+            for target, identifier in self.build_targets_override:
+                if identifier == (test_job_config.identifier if test_job_config.identifier else ""):
```
mfocko, line 330: Why do you need the ternary there? :eyes:

packit_service/models.py
```diff
@@ -2102,6 +2104,8 @@ class CoprBuildTargetModel(GroupAndTargetModelConnector, Base):
 
     scan = relationship("OSHScanModel", back_populates="copr_build_target")
 
+    identifier = Column(String, default="")
```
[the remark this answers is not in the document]
majamassarini: Ok is not a big issue not having the default here, I can remove it. In both cases this could be wrong. I will not create the logic for populate this field correctly because I think it is not worth the time we will spend on it.

packit_service/worker/helpers/build/build_helper.py
majamassarini: `job_config.identifier` could be `None`, so yes I think we need it. A non specified `identifier` is the default identifier "".
We can also say that the default identifier is always `None` but in this case I am a bit afraid of the serialization, if we go through a serialization/deserialization process of a `None` string value I think we will end up with an empty str ""? 
So I feel this comparison more safe.

packit_service/worker/helpers/build/copr_build.py
majamassarini: I would say they are still targets override but a bit more specific. They are targets for a specific identifier.

The most efficient structure here is probably a dictionary and not a `set` anymore, however I would start with this simpler enhancement.

packit_service/worker/events/event.py
```diff
@@ -119,8 +119,16 @@ def from_event_dict(cls, event: dict):
         time = event.get("task_accepted_time")
         task_accepted_time = datetime.fromtimestamp(time, timezone.utc) if time else None
 
-        build_targets_override = event.get("build_targets_override")
-        tests_targets_override = event.get("tests_targets_override")
+        build_targets_override = (
+            {(target, identifier_) for [target, identifier_] in event.get("build_targets_override")}
+            if event.get("build_targets_override")
+            else set()
```
lbarcziova, line 125: just to be sure, I also remember some discussions about the empty set vs None for *targets_override, is it ok to have empty set here instead of None?

packit_service/worker/helpers/build/build_helper.py
```diff
@@ -270,13 +270,25 @@ def build_targets_for_test_job(self, test_job_config: JobConfig) -> set[str]:
 
         if self.build_targets_override:
             logger.debug(f"Build targets override: {self.build_targets_override}")
-            targets_override.update(self.build_targets_override)
+            targets_override.update(
+                [
+                    target
+                    for (target, identifier) in self.build_targets_override
+                    if identifier
+                    == (test_job_config.identifier if test_job_config.identifier else "")
```
lbarcziova, line 278:
```suggestion
                    == (test_job_config.identifier or "")
```

lbarcziova (approved):
thanks!

packit_service/worker/events/event.py
majamassarini: I think it is, however I want to test this a bit deeper in the staging instance.

PR #2666, opened 2024-11-29
Set UID when setting Pagure PR flag

lbarcziova:
Generate it from the check_name. This prevents creating multiple separate flags (for each state). This was already introduced in 0d4bc3653d61e7d10310823995aa97a8071f93fe .

RELEASE NOTES BEGIN

N/A

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/reporting/reporters/pagure.py
+++ b/packit_service/worker/reporting/reporters/pagure.py
@@ -1,6 +1,6 @@
 # Copyright Contributors to the Packit project.
 # SPDX-License-Identifier: MIT
-
+import hashlib
 import logging
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


nforro:
> This prevents not creating multiple separate flags (for each state).

Should `not` be in this sentence? To me it seems to contradict the code change.

packit_service/worker/reporting/reporters/pagure.py
```diff
@@ -47,8 +47,16 @@ def set_status(
             url = CONTACTS_URL
 
         if self.pull_request_object:
+            # generate a custom uid from the check_name,
+            # so that we can update flags we set previously,
+            # instead of creating new ones (Pagure specific behaviour)
+            uid = hashlib.md5(check_name.encode()).hexdigest()
```
nforro, line 53: I would recommend staying away from MD5 (or SHA1), it could potentially disappear in the future.

```suggestion
            uid = hashlib.sha256(check_name.encode()).hexdigest()
```

lbarcziova:
>    Should `not` be in this sentence? To me it seems to contradict the code change.

Fixed.

nforro (approved):

PR #2672, opened 2024-12-09
fix(usage): raise time limit and lower frequency

mfocko:
As discussed with @majamassarini in the DMs, there are numerous occurencies of the hourly usage statistics caching not meeting the Celery deadline. Therefore adjusting both timeout and frequency.

• Timeout: 1800s → 3600s
           30min → 60min

  To make sure it doesn't spam Sentry with unnecessary exceptions.

• Frequency: 3600s → 10800s
               1hr →    3hr

  To avoid parallel runs of the same task and also to avoid congestion
  with other tasks (e.g., regular jobs).


lbarcziova (approved):
:+1:

PR #2674, opened 2024-12-10
Consider identifiers for OpenScanHub

### LINKED ISSUE #2653: Consider identifiers for OpenScanHub reporting
For projects with multiple Copr build jobs configured (e.g. monorepositories), we submit multiple scans, but when reporting, use just one check name, which results in overwriting the statuses and informing the user just about one scan.

TODO:
- [ ] include the identifier (of the Copr build job) in the check name for the OpenScanHub job
- [ ] implement pre-check for the OSH results handler to check for matching identifiers (we might need to store the identifier in the DB)

lbarcziova:
Fixes #2653

RELEASE NOTES BEGIN

We have fixed the reporting of OpenScanHub jobs when identifiers for Copr build jobs are configured.

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/checker/open_scan_hub.py
+++ b/packit_service/worker/checker/open_scan_hub.py
@@ -25,3 +25,15 @@ class RawhideX86Target(
             )
             return False
         return True
+
+
+class IsEventForJob(Checker):
+    def pre_check(self) -> bool:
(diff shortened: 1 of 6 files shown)


majamassarini (approved):
LGTM, thanks!

PR #2677, opened 2024-12-10
Fix the image builds

nforro:


nforro:
Ok, this fixes the matrix: https://github.com/packit/packit-service/actions/runs/12253212121

mfocko (approved):
I guess the question is whether we need alternative arches for service :eyes: but I guess it would be cool if someone ran it on Raspberry :D

nforro:
It looks promising: https://github.com/packit/packit-service/actions/runs/12278848800

PR #2679, opened 2024-12-11
fix(vm): unify separators in statuses

mfocko:
For all other jobs we're using ‹:› as a separator between “parts” of the status names.

When bumping the target in packit/hello-world#2024 I noticed that we use dashes for separating the parts of the status names in the VM Image Builder jobs, therefore switch to colons to have a unified schema across all statuses. Also helps with potential machine parsing, if someone wishes to implement that.

TODO:


lbarcziova (approved):
thanks!

PR #2680, opened 2024-12-11
Install mod_wsgi from updates-testing in service

nforro:
See https://github.com/GrahamDumpleton/mod_wsgi/issues/912. Can be reverted when https://bodhi.fedoraproject.org/updates/FEDORA-2024-235cc1f0b3 reaches stable.


mfocko (approved):
And here we go…

PR #2708, opened 2025-02-07
Trace memory usage with logging

majamassarini:
Sadly `py-spy` was only a time profiler and I wasn't able to find something that does not need changes in the code.

I followed mainly [this guide](https://chase-seibert.github.io/blog/2013/08/03/diagnosing-memory-leaks-python.html) and [this celery issue suggestions](https://github.com/celery/celery/issues/3339).

Deep profiling will be done just in stg, quicker resource check also in prod.


files/install-deps-worker.yaml
```diff
@@ -48,6 +49,11 @@
         name:
           - git+https://github.com/packit/sandcastle.git@{{ source_branch }}
           - eventlet
+    - name: Install guppy3 for memory profiling in non production environments
+      ansible.builtin.pip:
+        name:
+          - guppy3
+      when: deployment in ['stg', 'dev']
```
mfocko, line 56:
```suggestion
      when: deployment != 'prod'
```

mfocko (approved):

packit_service/worker/handlers/abstract.py
```diff
@@ -237,16 +240,34 @@ def get_tag_info(self) -> dict:
             tags.update({"package_name": self.data.event_dict["package_name"]})
         return tags
 
+    def log_memory_stats(self) -> None:
+        gc.collect()
+        memory_usage = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
+        logger.error(f"Memory usage: {memory_usage} (kb)")
```
lbarcziova, line 246: do we need this as error level? It will end up in Sentry iirc

lbarcziova (approved):
otherwise looks good!

packit_service/worker/handlers/abstract.py
majamassarini: No, I forgot to change it...

majamassarini:
recheck

Issue #2430, opened 2024-05-16
Fix Packit metrics "Copr builds (queued, started, finished)" chart


majamassarini:
The *Copr builds (queued, started, finished)* in http://metrics.osci.redhat.com/d/iuAkAaWMk/packit?orgId=1 is not working as expected.

Fix it.

This card is an outcome from the *share debugging Packit tips and tricks*. Do we need something else, apart from fixing the chart, for making this charts useful when debugging Packit? 
If yes, please create a follow up card.

See also https://github.com/packit/packit-service/issues/2519 for follow-ups.

lbarcziova:
After #2722 and #2723, the graphs look now ok:

<img width="1030" alt="Image" src="https://github.com/user-attachments/assets/c7d719cf-aca6-4268-a87e-cf52cd1034e3" />

PR #2722, opened 2025-02-15
Push metrics once per process_message task

lbarcziova:
Previously, metrics could be pushed multiple times within a single process_message task. This could potentially lead to increased CPU usage, redundant network requests, and  metric inconsistencies. Related to #2430 #2697


mfocko (approved):
What should happen in the event of an exception? :eyes: I think that would mean we're not pushing any metrics, but I guess we can handle that in a follow-up, it could be a more generic problem.

majamassarini (approved):

lbarcziova:
@mfocko that's right, in case of exception, the metrics would not be recorded. But as you said, I don't see this as that big of an issue, i.e. can be handled later, and now could potentially help us with the debugging.

PR #2723, opened 2025-02-17
Initialise Pushgateway per instance of SteveJobs

lbarcziova:
Previously, `cls.pushgateway` was used in `process_message`, possibly causing race conditions in multithreaded Celery tasks, leading to inaccurate metric reporting. Related to #2430 #2697

### DIFF
--- a/packit_service/worker/jobs.py
+++ b/packit_service/worker/jobs.py
@@ -130,10 +130,9 @@ class SteveJobs:
     Steve makes sure all the jobs are done with precision.
     """
 
-    pushgateway = Pushgateway()
-
     def __init__(self, event: Optional[Event] = None) -> None:
         self.event = event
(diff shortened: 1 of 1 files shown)


mfocko (approved):
Is that enough to change just these occurrences? From the other review I think there were multiple occurrences of the `pushgateway` :eyes:

lbarcziova:
@mfocko the other occurrences are from `JobHandler` ([see](https://github.com/packit/packit-service/blob/478d94b1a9f81623673075bc2de4d5809cc8e366/packit_service/worker/handlers/abstract.py#L358)), i.e. from the handler tasks and those use instance attribute already (that might explain the inaccuracy in ratio when you compare the graphs for Copr builds queued vs started/finished).

PR #2745, opened 2025-03-10
Introduce FastAPI

### LINKED ISSUE #2715: Introduce FastAPI alongside Flask
* set up FastAPI within the existing project structure
* ensure it runs alongside Flask without breaking existing functionality
* make sure OpenAPI documentation is generated
* add a basic health check endpoint in FastAPI to verify functionality (see [healthz.py](https://github.com/packit/packit-service/blob/main/packit_service/service/api/healthz.py))
* migrate [system.py](https://github.com/packit/packit-service/blob/main/packit_service/service/api/system.py)

lbarcziova:
Merge after #2749 (needed for tests to pass)

Fixes #2715 

To discuss:
- which ASGI server to use: currently I played with this with uvicorn, which is simple and lightweight, it doesn’t have http2 support though (for that we could use e.g. hypercorn), you can see some info on the alternatives [here](https://medium.com/@ramanbazhanau/preparing-fastapi-for-production-a-comprehensive-guide-d167e693aa2b)
  - after reading few articles, I went with uvicorn for local development and gunicorn + uvicorn workers for prod/stg, as it is supposed to be more suitable for prod envs (scalable, robust)
- Flask runs under /api and is hardcoded as WSGI, so I couldn’t make it work so that new endpoints are under /api/v1, for now I put them to /v1, wdyt? (redoc and swagger are also on /v1/redoc and /v1/docs)

TODO:
- [x] tests
- [x] rename of run_httpd.sh and occurrences
- [ ] polish the code, FastAPI settings, look into `security_and_https_middleware` more

RELEASE NOTES BEGIN

N/A

RELEASE NOTES END


tests/unit/test_views.py
```diff
@@ -118,7 +118,7 @@ def test_get_srpm_logs(client):
 
 
 def test_system_api(client):
-    response = client.get("/api/system")
+    response = client.get("/system")
```
Venefilyn, line 121: Does this work?

Venefilyn (commented):
Looks like a good start! I think going with `/v1` etc. instead of `/api/v1` is fine. Would API documentation go to `/v1/docs` or `/docs`?

tests/unit/test_views.py
lbarcziova: yes, since we set the `/api/` prefix on the FastAPI level, therefore I needed to change the test

lbarcziova:
> Would API documentation go to /v1/docs or /docs?

Currently it goes under `/docs`, but this can be easily changed.

Venefilyn:
> > Would API documentation go to /v1/docs or /docs?
> 
> Currently it goes under `/docs`, but this can be easily changed.

Ah okay. Yeah I'd say put it under the specific api version due to using versioning

CONTRIBUTING.md
```diff
@@ -47,11 +47,11 @@ This repository contains [docker-compose.yml](./docker-compose.yml) for
 (can be also [used with podman](https://fedoramagazine.org/use-docker-compose-with-podman-to-orchestrate-containers-on-fedora)).
 Before you run it, we suggest that you open the file and read all the comments.
 You can also run only certain pieces of packit-service for local development
-(e.g. worker, database or service/httpd).
+(e.g. worker, database or service).
```
majamassarini, line 50:
```suggestion
(e.g. worker, database or http service).
```
Now I am used to know that the service is an http service, but in the beginning this was confusing, service could be any kind of service. I would keep a way to say this is an http service.

nforro: :+1: though I'd rather use HTTP.

packit_service/service/app.py
```diff
@@ -62,46 +62,98 @@ def get_flask_application():
...
-    "'unsafe-inline'",
-    "'self'",
-]
+CORS(flask_app)
+INLINE = ["'unsafe-inline'", "'self'"]
 Talisman(
-    packit_as_a_service,
-    # https://github.com/wntrblm/flask-talisman#options
-    # https://infosec.mozilla.org/guidelines/web_security#implementation-notes
+    flask_app,
```
majamassarini, line 105:
```suggestion
    # https://github.com/wntrblm/flask-talisman#options
    # https://infosec.mozilla.org/guidelines/web_security#implementation-notes
    flask_app,
```

Should we keep this links? I think they are still relevant.

CONTRIBUTING.md
lbarcziova: makes sense, will change this 👍

packit_service/service/app.py
lbarcziova: yes, I firstly removed this because I thought it is not necessary anymore with setting it on FastAPI level, but tests revealed this is still needed, good catch!

majamassarini (approved):
I think we should land this on stg soon, because now our stg service instance is down.

Thanks!

lbarcziova:
recheck

lbarcziova:
I fixed the staging in #2750 so I would wait with merging this until Monday afternoon (prod deployment), so that this can be observed a bit more in staging.

lbarcziova:
Unfortunately, with the recent capacity reduction of our team, we needed to deprioritise few of the epics so that we are able to deliver the high-prio ones (see [board](https://github.com/orgs/packit/projects/7/views/29)). I won't merge this until we have the necessary capacity to continue with the epic, as changing the framework may introduce additional work.

PR #2747, opened 2025-03-11
Generate uid for pagure set_flag using target_branch

### LINKED ISSUE #2736: packit-fedora-ci: Scratch builds on a PR to multiple branches overwrite each other
### What happened? What is the problem?

Let's say there is a PR to update `rawhide` which triggers a packit-scratch-build. Then if we have another PR to `f42` using the same commit (e.g. merging `rawhide` to `f42` later on), the latter f42 scratch-build will overwrite the rawhide one.

This is what happened here, where you see only 1 packit-scratch-build job, but fedora-ci and zuul are correctly doubled:
- https://src.fedoraproject.org/rpms/python-scikit-build-core/pull-request/75
- https://src.fedoraproject.org/rpms/python-scikit-build-core/c/50f910da98b7516499f4ae8b4a71ed5cbb6818a8

majamassarini:
Fixes #2736

Not yet tested on a real packit-service instance (I will test it on stg instace later)

RELEASE NOTES BEGIN

We have fixed a bug for Fedora CI, scratch Koji builds statuses in Pagure are no more overwritten when the same commit is shared between different target branches. 

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/helpers/fedora_ci.py
+++ b/packit_service/worker/helpers/fedora_ci.py
@@ -26,6 +26,10 @@ class FedoraCIHelper:
         self._status_reporter = None
 
     @property
+    def target_branch(self) -> str:
+        return self.metadata.event_dict.get("target_branch")
+
+    @property
     def status_reporter(self) -> StatusReporter:
         if not self._status_reporter:
             self._status_reporter = StatusReporter.get_instance(
@@ -42,4 +46,5 @@ class FedoraCIHelper:
             description=description,
             url=url,
             check_name=self.status_name,
+            target_branch=self.target_branch,
         )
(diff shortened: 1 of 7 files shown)


lbarcziova (approved):
LGTM

mfocko (approved):

nforro:
A bit of feedback, this is great, especially compared to Fedora CI statuses that are indistinguishable, however I'm a bit puzzled why there is a third one without a branch name?

ci

nforro:
Aha, so the one without brnach name links to the same job as the epel10.0 one, and the ones with branch name seem not getting updated (both jobs finished successfully):

ci2

majamassarini:
[This](https://github.com/packit/packit-service/pull/2760) should fix the reporting.

Issue #2701, opened 2025-01-20
`version_update_mask` not working as expected


majamassarini:
I released packit 1.0.0 with `version_update_mask` set as '\d+\.'

Looking at the dashboard everything seems ok, packit seems to be released just for `rawhide` (and epel):

![Image](https://github.com/user-attachments/assets/b58eea42-69e5-446b-bc57-6ca180263b9b)

But then in dist-git do exist PRs even for the other branches (f40, f41...):

![Image](https://github.com/user-attachments/assets/06420954-f1b3-451c-ab67-2f7bfb34a394)

nforro:
Isn't that because `fast_forward_merge_into` doesn't care about version?

majamassarini:
Yes, it could be. But I don't think it is the right behaviour, or do you think the `fast_forward_merge_into` option should win over the `version_update_mask`? Now we can merge PRs by mistake.

majamassarini:
There is still another misbehaviour: there was once a skipped icon in dashboard for branches that did not match the pattern. Now there is no icon at all and this could be puzzling for the users.

nforro:
We should fix this ASAP, it affects our own releases as we can't release packit >= 1.0 in stable Fedoras and EPEL, yet we keep getting PRs for them.

PR #2751, opened 2025-03-13
Store PRs created for fast_forward_merge_into branches in DB

lbarcziova:
Adjust to the change of return value in PackitAPI.sync_release, store the returned additional PRs in DB and also expose them via API.

Related to #2701

Merge after packit/packit#2551

TODO:
- [x] create new testcase
- [ ] dashboard followups

RELEASE NOTES BEGIN

N/A

RELEASE NOTES END


packit_service/worker/handlers/distgit.py
```diff
@@ -405,12 +405,13 @@ def run_for_target(
...
 
         try:
-            downstream_pr = self.sync_branch(
+            downstream_pr, additional_prs = self.sync_branch(
                 branch=branch,
                 model=sync_release_run_model,
             )
             logger.debug("Downstream PR created successfully.")
             model.set_downstream_pr_url(downstream_pr_url=downstream_pr.url)
+            model.set_fast_forward_prs({branch: pr.url for (branch, pr) in additional_prs.items()})
```
majamassarini, line 414: I know this is quick. But I am wondering if we should instead save the info in the db as we do for the others prs and mark them as fast forwarded prs. We could have a column or even better a separate table to create the relationship with the parent pr. In this way we could later find all created prs and their relationships easily? 

We can also do this later if necessity arise.

lbarcziova: I was thinking about this, but wasn't sure if the additional DB complexity is worth introducing this. Let me try it out and we can decide then.

lbarcziova: pushed a commit with how this would look like, let me know wdyt 🙏

nforro (approved):
LGTM

PR #2760, opened 2025-03-20
Fix reporting for scratch build CI

majamassarini:
Should fix reporting for https://github.com/packit/packit-service/issues/2736

### DIFF
--- a/packit_service/worker/helpers/fedora_ci.py
+++ b/packit_service/worker/helpers/fedora_ci.py
@@ -27,7 +27,10 @@ class FedoraCIHelper:
 
     @property
     def target_branch(self) -> str:
-        return self.metadata.event_dict.get("target_branch")
+        return (
+            self.metadata.event_dict.get("target_branch")  # for pagure.pr.Action
+            or self.metadata.event_dict.get("target")  # for koji.result.Task
+        )
 
     @property
     def status_reporter(self) -> StatusReporter:


packit_service/worker/parser.py
```diff
@@ -1424,13 +1424,17 @@ def parse_koji_task_event(event) -> Optional[koji.result.Task]:
             if children.get("method") == "buildArch":
                 rpm_build_task_ids[children.get("arch")] = children.get("id")
 
+        request = nested_get(event, "info", "request", default=[])
+        target_branch = request[1] if len(request) >= 2 else None
```
nforro, line 1428: I'm not having any luck finding documentation or source code for the `request` list, but inspecting a few messages it seems the second item is in fact a Koji build target name (e.g. `f41-candidate`, `f43`), not a branch name.

We do have [`KojiHelper.get_build_target()`](https://github.com/packit/packit/blob/9be956e56fccff773eaf1db0d60c12c739e211ed/packit/utils/koji_helper.py#L401) that converts branch name to build target so we could perhaps somehow use it to do the opposite and find a branch from a target, but I wonder if there is a more reliable way to get the matching branch.

majamassarini: Oh right, I missed it 🙈 [here](https://koji.fedoraproject.org/koji/taskinfo?taskID=130730129) it is named *build target* (in the parameters list). So I think you are right. And I wasn't able to find anything else, in the event, to tell which was the target branch. If we can't reliably guess the branch from the target build (I don't know if we can?) then maybe we can save it in the database when submitting the build?

nforro: I've opened https://github.com/packit/packit/pull/2571.

nforro: > maybe we can save it in the database when submitting the build?

That's also an option. Not sure what's better.

packit_service/events/koji/result.py
```diff
@@ -219,13 +222,17 @@ def identifier(self) -> str:
...
+        target_branch = request[1] if len(request) >= 2 else None
+
         return Task(
             task_id=event.get("task_id"),
             state=KojiTaskState(event.get("state")) if event.get("state") else None,
             old_state=(KojiTaskState(event.get("old_state")) if event.get("old_state") else None),
             rpm_build_task_ids=event.get("rpm_build_task_ids"),
             start_time=event.get("start_time"),
             completion_time=event.get("completion_time"),
+            target_branch=target_branch,
```
nforro, line 235: I don't think the `event` dict contains `request`, but it should contain `target_branch`:

```suggestion
            target_branch=event.get("target_branch"),
```

packit_service/worker/parser.py
nforro: Ok, so we trigger the scratch builds with literal branch name as a build target, so in case of Packit-triggerd builds the build target in `request` should always be equal to a branch name. That means we don't have to do any conversion, however as the event object is generic, the attribute shouldn't be called `target_branch` but `build_target`.

nforro:
I'm confused now. We do store the target branch to the DB https://github.com/packit/packit-service/blob/48862afde4b8641b8ed55972f1c2fabbf1cc372e/packit_service/worker/handlers/distgit.py#L807-L816
and `koji.result.Task.target` should be automatically populated (from the DB) when the build completes.
Why do we need this PR?

nforro (approved):
Just remove the first commit please :slightly_smiling_face:

Issue #2765, opened 2025-03-26
propose-downstream sometimes fails with: ERROR  The repository is dirty, will not discard the changes.


martinpitt:
### What happened? What is the problem?

I was doing a Cockpit release this morning, and [propose-downstream failed](https://github.com/cockpit-project/cockpit/issues/21771) for [Fedora 41](https://dashboard.packit.dev/jobs/propose-downstream/15937) and [Fedora 42](https://dashboard.packit.dev/jobs/propose-downstream/15936), but [succeeded for rawhide](https://dashboard.packit.dev/jobs/propose-downstream/15935), resulting in a [valid dist-git PR](https://src.fedoraproject.org/rpms/cockpit/pull-request/330#).

After a `/packit propose-downstream`, [F42 worked](https://dashboard.packit.dev/jobs/propose-downstream/15938) but [F41 failed again](https://dashboard.packit.dev/jobs/propose-downstream/15939). 

In the failing case, this error appears:

> distgit.py        ERROR  The repository is dirty, will not discard the changes. Use --force to bypass.

in the passing runs this does not appear, so it feels significant.

### Steps to reproduce

```bash

```

### What is the impacted category (job)?

Fedora release automation

### Workaround

- [x] There is an existing workaround that can be used until this issue is fixed.

### Participation

- [ ] I am willing to submit a pull request for this issue. (Packit team is happy to help!)

martinpitt:
Two more instances from today: https://dashboard.packit.dev/jobs/propose-downstream/16046 and https://dashboard.packit.dev/jobs/propose-downstream/16045 (see https://github.com/cockpit-project/cockpit/issues/21800)

martinpitt:
This somehow is getting worse.. The third retry still failed.

martinpitt:
After three further retries there is a consistent pattern: rawhide always succeeds (it keeps updating https://src.fedoraproject.org/rpms/cockpit/pull-request/332) but F41/42 always fail. I also tried `/packit propose-downstream f42`, hoping that this would only attempt a single release, but it still did all three.

nforro:
> After three further retries there is a consistent pattern: rawhide always succeeds (it keeps updating https://src.fedoraproject.org/rpms/cockpit/pull-request/332) but F41/42 always fail. I also tried `/packit propose-downstream f42`, hoping that this would only attempt a single release, but it still did all three.

[...]

PR #2768, opened 2025-03-28
Reset repo submodules before the next sync_release run

nforro:
Fixes https://github.com/packit/packit-service/issues/2765.

RELEASE NOTES BEGIN

We have fixed an issue that could cause subsequent runs of `propose_downstream`/`pull_from_upstream` to fail if upstream git repo contains submodules that are manipulated with in actions.

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/handlers/distgit.py
+++ b/packit_service/worker/handlers/distgit.py
@@ -330,6 +330,9 @@ class AbstractSyncReleaseHandler(
                     index=True,
                     working_tree=True,
                 )
+                # reset also submodules
+                for submodule in self.packit_api.up.local_project.git_repo.submodules:
+                    submodule.update(init=True, recursive=True, force=True)
 
         return downstream_pr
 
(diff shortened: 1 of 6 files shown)


majamassarini (approved):
wow submodules 😅 Thanks for fixing this!

I will wait for this before creating the new release!

mfocko:
> wow submodules 😅 Thanks for fixing this!
> 
> I will wait for this before creating the new release, should we add a changelog message?

@majamassarini don't think you need to, it affects mainly service, there won't be many people using `sync-release` from the CLI :)

mfocko (approved):

nforro:
> I will wait for this before creating the new release, should we add a changelog message?

As Maťo said, this is in service. I don't think this affects many users, but I will add a release note.

martinpitt:
Thanks @nforro for the super-fast fix! I wasn't aware that all releases are done from the same prepared tree, but of course that makes sense.

Issue #2738, opened 2025-03-03
Packit as a Fedora dist-git CI: retriggering


nforro:
### Description

Fedora CI reacts to `[citest]` comments and retriggers the scratch build and other tests, however Packit does not.

### Importance

I would say not having retriggering is a blocker.

### Workaround

- [ ] There is an existing workaround that can be used until this is implemented.

LecrisUT:
Should it also react to `[citest]` or `/packit` instead? I don't think `[citest]` is that widely documented that we would need exact mirroring, and it would be nice to have finer control of which job to re-trigger.

nforro:
Let's stick with `/packit` for now.

xsuchy:
Please add documentation entry in https://packit.dev/docs

nforro:
> Please add documentation entry in https://packit.dev/docs

The naming is not final, we are going to discuss and most likely change it. Also, we don't have any docs for Packit as Fedora CI yet, I think we should open a new issue for that.

LecrisUT:
I think it would be fine to continue with this issue so that people subscribed to it can find the relevant changes in the mention history here.

On the naming, something like `/packit (re)check` seems reasonable and it would make it extensible to re-trigger parts of it like with a `/packit (re)check scratch-build`

nforro:
> I think it would be fine to continue with this issue so that people subscribed to it can find the relevant changes in the mention history here.

This issue is about retriggering, I think we want to document more than that.

> On the naming, something like `/packit (re)check` seems reasonable and it would make it extensible to re-trigger parts of it like with a `/packit (re)check scratch-build`

I would think it is necessary to distinguish between _Packit Service_ commands and _Packit as Fedora CI_ commands, a lot of users will not come in touch with the former, while everyone should eventually be affected by the latter.

LecrisUT:
> I would think it is necessary to distinguish between _Packit Service_ commands and _Packit as Fedora CI_ commands, a lot of users will not come in touch with the former, while everyone should eventually be affected by the latter.

I should clarify, `/packit (re)check scratch-build` would be in addition to `/packit (re)check`, aimed at the power-users who would want more control. E.g. when working on a tmt test, you would probably want to `[skip-ci]` all commits while debugging and instead manually trigger `/packit (re)check tmt`. Normal users on the other hand would be using `/packit (re)check` interface and trigger everything as a whole.

lbarcziova:
[...]

PR #2770, opened 2025-04-02
Implement Fedora CI retriggering

nforro:
Fixes https://github.com/packit/packit-service/issues/2738.

---

Notes:
I think it makes sense to allow only packagers to retrigger CI.
Feel free to suggest better command names than `/packit ci`.

---

RELEASE NOTES BEGIN

It is now possible to retrigger dist-git CI scratch builds with `/packit ci`.

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/checker/distgit.py
+++ b/packit_service/worker/checker/distgit.py
@@ -128,6 +128,33 @@ class PermissionOnDistgit(Checker, GetPagurePullRequestMixin):
         return True
 
 
+class PermissionOnDistgitForFedoraCI(Checker, GetPagurePullRequestMixin):
+    def pre_check(self) -> bool:
+        if self.data.event_type in (pagure.pr.Comment.event_type(),):
+            comment = self.data.event_dict.get("comment", "")
+            commands = get_packit_commands_from_comment(
+                comment,
+                self.service_config.comment_command_prefix,
+            )
+            command = commands[0] if commands else ""
+            commenter = self.data.actor
+            logger.debug(f'Triggering CI through "{command}" comment by: {commenter}')
+            if not self.is_packager(commenter):
+                msg = (
+                    "CI retriggering "
+                    f"through comment on PR identifier {self.data.pr_id} "
+                    f"and project {self.data.project_url} "
+                    f"done by {commenter} who is not a packager."
+                )
+                logger.info(msg)
+                self.pull_request.comment(
+                    f"{commenter} is not a packager, retriggering is not allowed."
+                )
+                return False
+
+        return True
+
+
 class HasIssueCommenterRetriggeringPermissions(ActorChecker):
     """To be able to retrigger a koji-build the issue commenter should
     have write permission on the project.
--- a/packit_service/worker/handlers/distgit.py
+++ b/packit_service/worker/handlers/distgit.py
@@ -81,6 +81,7 @@ from packit_service.worker.checker.distgit import (
     IsUpstreamTagMatchingConfig,
     LabelsOnDistgitPR,
     PermissionOnDistgit,
+    PermissionOnDistgitForFedoraCI,
     TaggedBuildIsNotABuildOfSelf,
     ValidInformationForPullFromUpstream,
 )
@@ -760,7 +761,9 @@ class PullFromUpstreamHandler(AbstractSyncReleaseHandler):
             return super().run()
 
 
+@run_for_comment(command="scratch-build")
 @reacts_to_as_fedora_ci(event=pagure.pr.Action)
+@reacts_to_as_fedora_ci(event=pagure.pr.Comment)
 class DownstreamKojiScratchBuildHandler(
     RetriableJobHandler, ConfigFromUrlMixin, LocalProjectMixin, PackitAPIWithDownstreamMixin
 ):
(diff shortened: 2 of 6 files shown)


packit_service/worker/checker/distgit.py
```diff
@@ -128,6 +128,30 @@ def pre_check(self) -> bool:
...
+            commands = get_packit_commands_from_comment(
+                comment,
+                self.service_config.comment_command_prefix,
+            )
+            command = commands[0] if commands else ""
+            commenter = self.data.actor
+            logger.debug(f'Triggering CI through "{command}" comment by: {commenter}')
+            if not self.is_packager(commenter):
+                msg = (
+                    "CI retriggering "
```
lbarcziova, line 144: not sure how we do it for downstream automation (for consistency), but it would be beneficial to comment this also in the PR, WDYT?

packit_service/worker/handlers/distgit.py
```diff
@@ -757,7 +758,9 @@ def run(self) -> TaskResults:
             return super().run()
 
 
+@run_for_comment(command="ci")
```
lbarcziova, line 761: I would go with more granular commands, like in the service itself, i.e. `scratch-build` for example. And to preserve what current Fedora CI has, we could have `ci` that would trigger all the jobs, WDYT?

packit_service/worker/helpers/fedora_ci.py
```diff
@@ -28,8 +29,13 @@ def __init__(
     @property
     def target_branch(self) -> str:
         return (
-            self.metadata.event_dict.get("target_branch")  # for pagure.pr.Action
-            or self.metadata.event_dict.get("target")  # for koji.result.Task
+            # for pagure.pr.Comment
+            self.project.get_pr(self.metadata.pr_id).target_branch
```
lbarcziova, line 33: I see some overlap here with `dist_git_branch` property in the handler, could this be extracted?

packit_service/worker/checker/distgit.py
nforro: For downstream automation we do the "task was accepted" comments, we can do that here as well, though I have a feeling certain users would consider that unnecessary spam :slightly_smiling_face: But I suppose we can make that configurable later.

packit_service/worker/handlers/distgit.py
nforro: I was thinking about `scratch-build` but then I thought it would be better to have the CI commands more distinguishable to users. Since we want to eventually allow users to retrigger individual checks/parts of the CI pipeline, how about something like `/packit ci[scratch-build]` or `/packit ci scratch-build`?

packit_service/worker/helpers/fedora_ci.py
nforro: Let me check.

packit_service/worker/checker/distgit.py
lbarcziova: sorry for not being specific, I meant mostly error-related comments

but yes, the `Task was accepted` could be quite spammy, so not sure whether to do this by default, the reactions UX would be much better here, so that could be used once the Forgejo switch happens

packit_service/worker/handlers/distgit.py
lbarcziova: That's a good point. I remember originally discussing maybe we will use a different command prefix for Fedora CI, so we could end up with `/packit-ci scratch-build`, WDYT? Maybe worth discussing with the rest of the team.

packit_service/worker/checker/distgit.py
nforro: > sorry for not being specific, I meant mostly error-related comments

Oh, right, for downstream automation errors are reported as issues in `issue_repository` if configured. For CI I think it makes sense to reply with a comment. Let me do that.

> but yes, the `Task was accepted` could be quite spammy, so not sure whether to do this by default, the reactions UX would be much better here, so that could be used once the Forgejo switch happens

There is a TODO for it, perhaps I can implement it and we can disable it later?
https://github.com/packit/packit-service/blob/615f072d16e8f4e57eb3ae3f9b6191262159ceea/packit_service/worker/jobs.py#L456

packit_service/worker/handlers/distgit.py
nforro: If we change the prefix then `scratch-build` makes sense. What about the instances though? Let me change the command to `scratch-build` for now and we can discuss this on arch next week and change it.

packit_service/worker/checker/distgit.py
lbarcziova: > For CI I think it makes sense to reply with a comment. Let me do that.

👍 , thanks!

> There is a TODO for it, perhaps I can implement it and we can disable it later?

actually, thinking about it now, for Fedora CI, operating on PRs, it makes sense to do regular commit statuses exactly like in upstream

nforro: > actually, thinking about it now, for Fedora CI, operating on PRs, it makes sense to do regular commit statuses exactly like in upstream

Do you mean the flags (as Pagure calls them) that are already there so no extra reporting needed or did I misunderstood?

lbarcziova: yes I mean the flags, but for Fedora CI we haven't been doing them for `Task was accepted` yet iiuc

nforro: Oh, so you mean another status before `RPM build was submitted ...`?

lbarcziova: yes

nforro: Ok, makes sense, let me to that.

lbarcziova (approved):
LGTM!

Issue #2006, opened 2023-04-14
When using monorepo, react only on changes done to configured paths


lachmanfrantisek:
### Description

When using [monorepo](https://github.com/packit/packit/issues/1543) (=more packages in one git repository, defined using the `packages` key) with many packages, triggering actions for all the packages is not needed.

Let's check the affected paths first and react only if anything has been done to the package-related path.

### Benefit

* Safe of resources. (=Only the important/relevant jobs are done.)
* Cleaner UX. (=Just important/relevant results are shown.)

### Importance

This has been asked multiple times, but lately by @LecrisUT who wants to use a monorepo with ~500 packages.

### What is the impacted category (job)?

General

### Workaround

- [ ] There is an existing workaround that can be used until this feature is implemented.

### Participation

- [ ] I am willing to submit a pull request for this issue. (Packit team is happy to help!)

lachmanfrantisek:
Option 1:

Add the info to the Event classes and use this when creating job tasks:

* [ ] Add a new property to all the relevant event classes:
https://github.com/packit/packit-service/blob/a8dddb3d4be8ca29110f93e671a54a8a3fd6be7a/packit_service/worker/events/github.py#L86
* [ ] Fill in the info when parsing:
https://github.com/packit/packit-service/blob/a8dddb3d4be8ca29110f93e671a54a8a3fd6be7a/packit_service/worker/parser.py#L250
  * [ ] Check if we can actually get that info: https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads?actionType=opened#pull_request (If I've checked correctly, it's not in the event payload and we need to use GitHub API (using [this OGR method](https://packit.github.io/ogr/abstract.html#ogr.abstract.GitProject.get_files) in our case) to get that info.
* [ ] Use the info when filtering the new tasks to be added to our job queue:
https://github.com/packit/packit-service/blob/a8dddb3d4be8ca29110f93e671a54a8a3fd6be7a/packit_service/worker/jobs.py#L446

LecrisUT:
I would say to have it for both PRs and branch commits initially. Later this should be handled by the dependencies logic with `rebuild`/`sync-version` logic [from here](https://github.com/packit/packit/issues/1903). To bridge the gap between them, maybe the user could run the `packit build in-copr` for the select packages or use a `\packit build` in the PR specifying the packages there (later just `\packit build dependents`).

lachmanfrantisek:
Option 2:

[...]

PR #2780, opened 2025-04-16
[WIP] feat: Filter `copr_build` jobs based on `paths` and files changed

LecrisUT:
TODO:

- [ ] Skip the check if `/packit` command is used
- [ ] Actually check the files changed
- [ ] ‹fill in›

Fixes https://github.com/packit/packit/issues/1997
Fixes https://github.com/packit/packit-service/issues/2006

RELEASE NOTES BEGIN

`copr_build` jobs are triggered only if there are changed files under the `paths` field

RELEASE NOTES END


packit_service/worker/checker/copr.py
```diff
@@ -138,3 +142,51 @@ def _pre_check(self) -> bool:
...
+        elif self.job_config.trigger == JobConfigTriggerType.release:
+            # For releases we don't do any checks
+            return None
+
+    def get_files_changed(self, previous_git_ref: str) -> list[Path]:
+        """
+        Get the list of files changed between the current git ref and the previous one.
+        """
+        # FIXME: Implement the relevant git diff
+        return []
```
LecrisUT, line 173: Currently stuck on this part. Isn't the git repo cloned somewhere when the `package_config` was evaluated? I was hoping on using that to do the `git diff`, but I haven't navigated to such a part. Should that be an interface in `ogr` instead?

nforro: `LocalProjectMixin` should give you `self.local_project.git_repo` which is `git.Repo`.
https://github.com/packit/packit-service/blob/11d4722765650d350cab89ba8ec80fcfd0736cfa/packit_service/worker/mixin.py#L212-L246

lbarcziova: on our arch meeting we were discussing we could get this rather [via API](https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#get-a-commit), this would mean implementing the method for getting a commit in [ogr](https://github.com/packit/ogr)

packit_service/worker/checker/copr.py
```diff
@@ -138,3 +141,44 @@ def _pre_check(self) -> bool:
...
+    def get_files_changed(self) -> list[Path]:
+        """
+        Get the list of files changed in the current commit or the current pullrequest
+        """
+        # Get the changes object
+        if self.job_config.trigger == JobConfigTriggerType.pull_request:
+            changes = self.copr_build_helper.pull_request_object.changes
+        elif self.job_config.trigger == JobConfigTriggerType.commit:
+            branch = self.job_config.branch or self.project.default_branch
+            changes = self.copr_build_helper.project.get_commit(branch)
```
LecrisUT, line 161: Something to note here. If a PR is merged as a rebase, only the top commit will be checked for changes

packit_service/worker/checker/copr.py
```diff
@@ -138,3 +141,44 @@ def _pre_check(self) -> bool:
...
+        # Early check if the git root was included, in which case we don't need to
+        # check the files changed
+        # TODO: refine this check when gitignore-like patterns are supported
+        if "./" in paths:
+            return True
+        for changed_file in self.get_files_changed():
+            # Check if any of the files changed are under the paths that are being tracked
+            if any(changed_file.is_relative_to(p) for p in paths):
+                return True
+        return False
```
LecrisUT, line 184: Actually, what do we do if we have an empty commit, like the Fedora rebuild commits? Should it be a special case to skip this check?

packit_service/worker/checker/copr.py
lbarcziova: that's a good point. Could we get and check all the commits?

LecrisUT: :thinking: Maybe it's possible. At least with Pagure we know that there is a link between commit and PR, and maybe there is one for Github as well. Is there a PR merged as rebase (didn't find better wording) in https://github.com/packit/hello-world ?

lbarcziova: I think we could use the `commits` directly from the webhook payload [here](https://docs.github.com/en/webhooks/webhook-events-and-payloads#push), wdyt? It would require changing the [parsing](https://github.com/packit/packit-service/blob/cdce0c7cc1b33c28b07fd87c494d6956ff471328/packit_service/worker/parser.py#L521)  and extending the [class](https://github.com/packit/packit-service/blob/main/packit_service/events/github/push.py). Then in the checker we would check this based on the event type.

lbarcziova: same can also be done for [Gitlab](https://docs.gitlab.com/user/project/integrations/webhook_events/#push-events). Another benefit is having the information directly, without additional API calls.

> We do have a hook we can use there:

is this deterministic though? (e.g. one commit associated to multiple PRs)

lbarcziova:
I briefly looked into the code and I am thinking if it wouldn't be more fitting to have the logic for checking this in [this method](https://github.com/packit/packit-service/blob/e24583035cf6abb96559ee7586bd0127351aaba2/packit_service/worker/jobs.py#L849), which would mean it would work generally, not just for Copr build jobs.

LecrisUT:
> I briefly looked into the code and I am thinking if it wouldn't be more fitting to have the logic for checking this in [this method](https://github.com/packit/packit-service/blob/e24583035cf6abb96559ee7586bd0127351aaba2/packit_service/worker/jobs.py#L849), which would mean it would work generally, not just for Copr build jobs.

Probably, but it's not as extendable as with the Checker, needing another place to be aware of that check are being done. I see that there already are checks like `pr_labels_match_configuration`. What about extending the `Checker` design (or rather reimplement a similar/simpler class) or at least aggregate  the checks in a list of lambdas?

lbarcziova:
@LecrisUT can you elaborate more on what you are suggesting? I agree the code I linked could be refactored, but it is something different than the Checker's logic, as that is tight to the handlers, while the checks I linked are meant for actually getting the relevant job configurations to the particular event.

LecrisUT:
> @LecrisUT can you elaborate more on what you are suggesting? I agree the code I linked could be refactored, but it is something different than the Checker's logic, as that is tight to the handlers, while the checks I linked are meant for actually getting the relevant job configurations to the particular event.

Indeed, not using the same `Checker`, but mainly not having all the checks implemented in the function, i.e. something like
```python
    def get_jobs_matching_event(self) -> list[JobConfig]:
        jobs_matching_trigger = []
        checkers = JobChecker.get_all_checkers()
        for job in self.event.packages_config.get_job_views():
            if all(c.check(job) for c in checkers):
                jobs_matching_trigger.append(job)
```

The `Checker` design that I like is being able to just create a class and *boom* a new check is introduced, but I would design it a bit simpler, like
```python
class JobChecker:
    all_checkers: Final[ClassVar[list[type[JobChecker]]]] = []

    def __init_subclass__(cls):
        # Register all checkers when the class is created
        cls.all_checkers.append(cls)

    @classmethod
    def get_all_checkers(cls) -> list[JobChecker]
        return  [checker_cls() for checker_cls in cls.all_checkers]

    def check(job: JobView) -> bool: ...
```
(Of course with a better interface, constructor, etc.)

lachmanfrantisek:
Hi @LecrisUT , thanks for all the work so far. Do you think it's real to finish this? (Asking for the Copr team that would hugely benefit from this..) Anything we can help with?

LecrisUT:
I can try to make some progress this week.

One question to be answered is https://github.com/packit/packit-service/pull/2780#discussion_r2111302592. There are  a few ways to approach this:
- Use `ogr` and navigate to the PR and get the files changed from there
- Use the webhook metadata directly
- Teach `ogr` to give compare patches [^1] and use the before-after that are already in the webhook payload

I am inclined to go with the 3rd option, just because it will come in handy for the future.

[^1]: https://github.com/PyGithub/PyGithub/blob/2d4785dd08374147a8eab866eeb704508558336a/github/Repository.py#L1342

betulependule (commented):
Nice work. It looks good to me so far. Do you think you'll be able to make progress with this soon?

packit_service/worker/checker/copr.py
```diff
@@ -138,3 +141,44 @@ def _pre_check(self) -> bool:
...
+            raise AssertionError(f"Trigger not supported: {self.job_config.trigger}")
+        # The paths here are relative to the git repo root
+        return [Path(file) for file in changes.files]
+
+    def pre_check(self) -> bool:
+        if self.job_config.trigger == JobConfigTriggerType.release:
+            # For releases we don't do any checks
+            return True
+        # FIXME: This is probably unnecessary
+        package_config = self.package_config.get_package_config_for(self.job_config)
```
betulependule: Why do you think this might be unnecessary?

betulependule: Hm, it seems that the `get_package_config_for` is used to get the config from a config for multiple packages (usually named `packages_config`) based on a specific job config. In other existing code, I only see `packages_config.get_package_config_for` being used, but never `package_config.get_package_config_for`. I think using `self.package_config` directly might be the right way to do it, but I'm not exactly sure. Tests should hopefully reveal what it right.

LecrisUT: Indeed, worst case this is just a no-op, but if anyone with more experience could chyme in and confirm it quickly then we can resolve this one way or the other. Let's keep this open for a bit.

packit_service/worker/checker/copr.py
```diff
@@ -138,3 +141,44 @@ def _pre_check(self) -> bool:
...
+    def get_files_changed(self) -> list[Path]:
+        """
+        Get the list of files changed in the current commit or the current pullrequest
+        """
+        # Get the changes object
+        if self.job_config.trigger == JobConfigTriggerType.pull_request:
+            changes = self.copr_build_helper.pull_request_object.changes
+        elif self.job_config.trigger == JobConfigTriggerType.commit:
+            # TODO: Get changed files from webhook push
+            raise NotImplementedError()
```
betulependule: This should probably be simple to implement. Both GitLab and GitHub webhook push events specify the `commits` field, where each item contains the `added`, `modified` and `removed` fields describing what files have been changed.

https://docs.github.com/en/webhooks/webhook-events-and-payloads#push
https://docs.gitlab.com/user/project/integrations/webhook_events/#push-events

However, I am a bit worried about this note from GitHub's docs: `A maximum of 3000 changed files will be reported per commit.` I am concerned that if a user's commit includes more than 3000 changed files, then the webhook will not report all and the checker might ignore some file changes, resulting in an obscure bug... Maybe there should be a check whether the amount of changes matches 3000, because if it does, then that means that the webhook may have under-reported what files were changes and we should find the rest of changed files using a different method.

betulependule: Or for simplicity, if 3000 changes are reported in one of the fields `added`, `modified` or `removed`, return the root directory `./` to make sure the checker doesn't ignore any files not reported by the webhook push event.

LecrisUT: That comment is the same issue as https://github.com/packit/packit-service/pull/2780#issuecomment-3414461407. I will try to join you on figuring out how to implement this. If you have any ideas from diving in the code, do let me know

> However, I am a bit worried about this note from GitHub's docs: `A maximum of 3000 changed files will be reported per commit.`

My gut instinct is if it ends up with that many change files, we might be having other issues on the side. If they do that big of a change, then might as well either trigger the build for everything or let the users manually trigger it and we just report " :eyes: > 3000 files being changed!? Help me :pray: "

packit_service/worker/checker/copr.py
```diff
@@ -138,3 +141,44 @@ def _pre_check(self) -> bool:
...
+    If not, then just skip the current copr build job.
+    """
+
+    def get_files_changed(self) -> list[Path]:
+        """
+        Get the list of files changed in the current commit or the current pullrequest
+        """
+        # Get the changes object
+        if self.job_config.trigger == JobConfigTriggerType.pull_request:
+            changes = self.copr_build_helper.pull_request_object.changes
```
betulependule: I double checked the definitions of GitHub and GitLab definitions of the MR / PR webhook events and  you are definitely right that the information related to what files were changed is not present.

Do I understand correctly that your current approach is the following?
1. In `parser.py`, once a PR / MR webhook event is received, use `ogr` to get a list of commits (cast as `CommitInfo`) contained in this PR / MR.
2. Inside the `AreFilesChanged` class, take the list of commits and aggregate the three lists of files `added`, `modified` and `removed`.

packit_service/worker/checker/copr.py
LecrisUT: Ok, I think I figured it out a bit. It should come from `ConfigFromEventMixin` and it would have either commit, or PR or equivalent event results. Commit I think it is pretty straightforward, but PRs would be tricky since it can also take in comments. Maybe we could ignore all non-push, non-pr-update events (i.e. make the checker always true then). Updated the PR with a quick sketch that I have so far in case you get some ideas around this as well :)

LecrisUT: Should be resolved now I think

packit_service/worker/checker/copr.py
```diff
@@ -138,3 +141,53 @@ def _pre_check(self) -> bool:
...
+        """
+        Get the list of files changed in the current commit or the current pullrequest
+        """
+        # Get the changes object
+        if self.job_config.trigger == JobConfigTriggerType.pull_request:
+            pr_event = self.data.to_event()
+            if not isinstance(pr_event, (github.pr.Action, gitlab.mr.Action)):
+                # TODO: What about comments?
+                raise NotImplementedError()
+            # TODO: How to handle PRs?
```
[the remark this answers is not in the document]
LecrisUT: Suggestion: Shall we just try-except this part and make it behave as previously until https://github.com/packit/ogr/pull/921 lands? Or we wait for that first?

Thing is I don't know how `raise` is handled and if it is safe to use.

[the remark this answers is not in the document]
LecrisUT: Oh neat, the `depends-on` should actually work according to zuul below :thinking:. I guess we can test it with that cross-dependency

Issue #2712, opened 2025-02-11
Implement installability check for Packit as a Fedora dist-git CI


lachmanfrantisek:
Run [installability check](https://github.com/fedora-ci/installability-pipeline) for successful Koji scratch builds on Fedora dist-git pull-requests.

You can also consider defining another plan and putting it e.g. [here](https://github.com/packit/tmt-plans). Maybe also making [the upstream one](https://gitlab.com/testing-farm/tests/-/blob/main/packit/install-and-verify.sh) usable both upstream and downstream.

*This is a part of the phase 2 of https://github.com/packit/packit-service/issues/2692*

keszybz:
At the risk of sounding like a scratched record, *please* make the installability check actually useful. Since you're implementing a new thing, it'd be great time to improve on what we had before.

Currently, it tries to install all packages in the build, which cannot work in many cases. https://pagure.io/fedora-ci/general/issue/448 has a long discussion and links to issues.

LecrisUT:
Re: installing all pqckages, probably it would be sufficient to make the artifacts empty for those tests like setting `skip_build`.

Another issue brought up was to use rhel builders for epel-9.

lbarcziova:
> Also installability tmt plan should be [this one](https://github.com/fedora-ci/installability-pipeline/blob/master/installability.fmf) actually. Regarding the link to install-and-verify.sh link

That one is also linked in the description (the repo).

When implementing this, consider implementing the retriggering also, or create a follow-up issue.

LecrisUT:
Most of the tmt plans in Fedora CI should be simple to port, since it only involves providing the appropriate Koji task ID and a few other environment variables. A bit unfortunate that you would have to navigate the implementation to find the necessary input, but some of them have it in the `prepare` step. It would be nice to consolidate the tests in a single repo, but better to defer that for after packit takes over Fedora CI.

nforro:
Just an update, I'm still actively working on this, hopefully a PR with the implementation will appear soon.

PR #2782, opened 2025-04-23
Implement installability check for Packit as a Fedora dist-git CI

nforro:
Fixes https://github.com/packit/packit-service/issues/2712.

RELEASE NOTES BEGIN

Installability check is now being run as part of CI on dist-git pull requests.

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/handlers/abstract.py
+++ b/packit_service/worker/handlers/abstract.py
@@ -47,11 +47,11 @@ MAP_REQUIRED_JOB_TYPE_TO_HANDLER: dict[JobType, set[type["JobHandler"]]] = defau
     set,
 )
 SUPPORTED_EVENTS_FOR_HANDLER: dict[type["JobHandler"], set[type["Event"]]] = defaultdict(set)
-SUPPORTED_EVENTS_FOR_HANDLER_FEDORA_CI: dict[type["JobHandler"], set[type["Event"]]] = defaultdict(
-    set
+SUPPORTED_EVENTS_FOR_HANDLER_FEDORA_CI: dict[type["FedoraCIJobHandler"], set[type["Event"]]] = (
+    defaultdict(set)
 )
 MAP_COMMENT_TO_HANDLER: dict[str, set[type["JobHandler"]]] = defaultdict(set)
-MAP_COMMENT_TO_HANDLER_FEDORA_CI: dict[str, set[type["JobHandler"]]] = defaultdict(set)
+MAP_COMMENT_TO_HANDLER_FEDORA_CI: dict[str, set[type["FedoraCIJobHandler"]]] = defaultdict(set)
 MAP_CHECK_PREFIX_TO_HANDLER: dict[str, set[type["JobHandler"]]] = defaultdict(set)
 
 
@@ -128,7 +128,7 @@ def reacts_to_as_fedora_ci(event: type["Event"]):
     ```
     """
 
-    def _add_to_mapping(kls: type["JobHandler"]):
+    def _add_to_mapping(kls: type["FedoraCIJobHandler"]):
         SUPPORTED_EVENTS_FOR_HANDLER_FEDORA_CI[kls].add(event)
         return kls
 
@@ -183,7 +183,7 @@ def run_for_comment_as_fedora_ci(command: str):
     ```
     """
 
-    def _add_to_mapping(kls: type["JobHandler"]):
+    def _add_to_mapping(kls: type["FedoraCIJobHandler"]):
         MAP_COMMENT_TO_HANDLER_FEDORA_CI[command].add(kls)
         return kls
 
@@ -463,3 +463,11 @@ class RetriableJobHandler(JobHandler):
 
     def run(self) -> TaskResults:
         raise NotImplementedError("This should have been implemented.")
+
+
+class FedoraCIJobHandler(JobHandler):
+    check_name: str = ""
+
+    @classmethod
+    def get_check_names(cls, service_config: ServiceConfig, metadata: EventData) -> list[str]:
+        return [cls.check_name] if cls.check_name else []
--- a/packit_service/worker/handlers/abstract.py
+++ b/packit_service/worker/handlers/abstract.py
@@ -224,6 +224,8 @@ class TaskName(str, enum.Enum):
     installation = "task.run_installation_handler"
     testing_farm = "task.run_testing_farm_handler"
     testing_farm_results = "task.run_testing_farm_results_handler"
+    downstream_testing_farm = "task.run_downstream_testing_farm_handler"
+    downstream_testing_farm_results = "task.run_downstream_testing_farm_results_handler"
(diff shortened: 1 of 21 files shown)


packit_service/worker/helpers/testing_farm.py
```diff
@@ -141,6 +141,285 @@ def parse_unknown_arguments(self, unknown_args: list[str]) -> None:
                 break
 
 
+class TestingFarmClient:
```
mfocko, line 146: Do you think it would make sense to take this out to a separate source file?

mfocko: I'm kinda for removing the `ServiceConfig` from the `TestingFarmClient` and pass directly API URL, token and ranch…

We should probably open at least an issue for that; I checked how is this client-logic handled in the TF CLI (to check if it's feasible to reuse it), but there are no abstractions :/

mfocko (approved):

packit_service/worker/helpers/testing_farm.py
nforro: > Do you think it would make sense to take this out to a separate source file?

Yes, I think so.

> I'm kinda for removing the `ServiceConfig` from the `TestingFarmClient` and pass directly API URL, token and ranch…

Makes sense if we separate it from the helpers.

lbarcziova: +1 for passing those directly, should make it more readable for usage of internal/public TF

packit_service/worker/helpers/testing_farm.py
```diff
@@ -1476,3 +1487,266 @@ def report_status_to_configured_job(
...
+        all_tests = list(FEDORA_CI_TESTS.keys())
+        if metadata.event_type != pagure.pr.Comment.event_type():
+            return all_tests
+        comment_command_prefix = service_config.comment_command_prefix
+        # TODO: remove this once Fedora CI has its own instances and comment_command_prefixes
+        # comment_command_prefixes for Fedora CI are /packit-ci and /packit-ci-stg
+        if comment_command_prefix.endswith("-stg"):
+            comment_command_prefix = "/packit-ci-stg"
+        else:
+            comment_command_prefix = "/packit-ci"
```
lbarcziova, line 1531:
```suggestion
        comment_prefix = (
          "/packit-ci-stg" if service_config.comment_command_prefix.endswith("-stg")
          else "/packit-ci"
      )
```

packit_service/worker/helpers/testing_farm.py
```diff
@@ -1476,3 +1487,266 @@ def report_status_to_configured_job(
...
+        commands = get_packit_commands_from_comment(
+            metadata.event_dict.get("comment"), comment_command_prefix
+        )
+        if not commands:
+            return []
+        if len(commands) > 1 and (tests := [t for t in all_tests if t == commands[1]]):
+            return tests
```
lbarcziova, line 1538:
```suggestion
        if len(commands) > 1 and commands[1] in all_tests:
            return [commands[1]]
```

packit_service/worker/handlers/abstract.py
```diff
@@ -463,3 +465,11 @@ def __init__(
 
     def run(self) -> TaskResults:
         raise NotImplementedError("This should have been implemented.")
+
+
+class FedoraCIJobHandler(JobHandler):
+    check_name: str = ""
+
+    @classmethod
+    def get_check_names(cls, service_config: ServiceConfig, metadata: EventData) -> list[str]:
```
lbarcziova, line 474: this is returning a list because of the future use case of running multiple testsuits?

packit_service/worker/helpers/testing_farm.py
```diff
@@ -1476,3 +1487,266 @@ def report_status_to_configured_job(
...
+        )
+
+    @implements_fedora_ci_test("installability")
+    def _payload_installability(self, compose: str) -> dict:
+        git_repo = "https://github.com/fedora-ci/installability-pipeline.git"
+        git_ref = (
+            commands.run_command(["git", "ls-remote", git_repo, "HEAD"], output=True)
+            .stdout.strip()
+            .split()[0]
+        )
```
lbarcziova, line 1645: wouldn't `master` here be enough?

lbarcziova (commented):
looks really good!

packit_service/worker/handlers/abstract.py
nforro: Yes, so far there is only one test/suite: https://github.com/packit/packit-service/blob/a8b2d13d55dc21591cc5e680b74d190090bfbccb/packit_service/worker/helpers/testing_farm.py#L1638

packit_service/worker/helpers/testing_farm.py
nforro: > Do you think it would make sense to take this out to a separate source file?

Where should it be? In `packit_service/worker/helpers`? That doesn't seem like a good fit. Perhaps directly in `packit_service`?

lbarcziova: to me, having it in helpers sounds ok, as it will still be used just by the TFHelper, right?

packit_service/worker/helpers/testing_farm_client.py
```diff
@@ -0,0 +1,278 @@
...
+        )
+
+    def send_testing_farm_request(
+        self,
+        endpoint: str,
+        method: Optional[str] = None,
+        params: Optional[dict] = None,
+        data: Optional[dict] = None,
+    ) -> RequestResponse:
+        method = method or "GET"
```
mfocko, line 72:
```suggestion
        method: str = "GET",
        params: Optional[dict] = None,
        data: Optional[dict] = None,
    ) -> RequestResponse:
```

packit_service/worker/helpers/testing_farm_client.py
```diff
@@ -0,0 +1,278 @@
...
+            return False
+        return True
+
+    @classmethod
+    def get_request_details(cls, request_id: str) -> dict[str, Any]:
+        """Testing Farm sends only request/pipeline id in a notification.
+        We need to get more details ourselves."""
+        service_config = ServiceConfig.get_service_config()
+        self = cls(
+            api_url=service_config.testing_farm_api_url,
```
mfocko, line 113: ~~I guess it's coming from the original code, but could you add a comment here? It feels like we can't tell apart…~~ oh never mind, the API is public for both public and RH ranch :facepalm:

guess it deserves a comment anyways :D

lbarcziova (approved):
🚀

packit_service/worker/helpers/testing_farm_client.py
nforro: @mfocko Is this ok? https://github.com/packit/packit-service/blob/4dd28421c37345ad558fc9b5d653f4b2dcac7a0e/packit_service/worker/helpers/testing_farm_client.py#L113

PR #2785, opened 2025-04-29
Don't build for targets from test jobs

### LINKED ISSUE #2669: Too many copr builds triggered when tests jobs are defined
Working on re-triggering of failed tests, I realized we are triggering non needed copr builds.

```
jobs:
- job: copr_build
  trigger: pull_request
  identifier: latest
  targets:
  - fedora-latest-x86_64

- job: copr_build
  trigger: pull_request
  targets:
  - fedora-rawhide-x86_64
```
The above snippet is taken from [this PR](https://github.com/packit/hello-world/pull/2119); one copr build for `rawhide` and one for `f41` with `latest` identifier should be enough, instead we are triggering even the `rawhide` build in the copr repo with `latest` identifier and  the `f41` build in the copr repo with no identifier.
This is caused by the definition of tests jobs (not listed in the snippet) for which we are looking in all defined targets:

https://github.com/packit/packit-service/blob/d6b6a6e5fc8e934ed3d71444e0b82fc8e980df42/packit_service/worker/helpers/build/build_helper.py#L94-L96

**There are no issues for any user but we are wasting resources.**

---

There is also another issue related with the wrong `build_targets` count.

https://github.com/packit/packit-service/blob/d6b6a6e5fc8e934ed3d71444e0b82fc8e980df42/packit_service/worker/jobs.py#L359-L360

**We are submitting the wrong number of build_targets (for too many times) to the pushgateway.**
When there are multiple copr jobs defined and there is also at least one test job defined, we are submitting for every copr job and every test job the sum of all defined targets (since the CoprBuildJobHelper is initialized even for the tests jobs).
I think this is the cause of our wrong metrics #2430 (we have many more queued tasks than started or finished).

lbarcziova:
These should rather be explicitly configured in the build jobs as well. Otherwise, this results in potentially too many Copr builds being submitted when there are multiple build jobs defined.
Fixes #2669

Before getting this to prod, we will need to open PRs to affected repositories to not break their workflows.

RELEASE NOTES BEGIN

In test jobs, it is now explicitly required to include the targets needed for running build jobs; they will no longer be automatically deduced to prevent inconsistencies in creating build and test jobs.

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/helpers/build/build_helper.py
+++ b/packit_service/worker/helpers/build/build_helper.py
@@ -86,16 +86,12 @@ class BaseBuildJobHelper(BaseJobHelper):
         """
         Return the configured targets for build job.
 
-        1. Use targets defined for build job and targets defined for test job.
+        1. Use targets defined for build job.
         2. Use "fedora-stable" alias if neither defined.
         """
         targets = set()
         if self.job_build:
             targets.update(self.job_build.targets)
-
-        for test_job in self.job_tests_all:
-            if not test_job.skip_build:
-                targets.update(test_job.targets)
 
         if (
             self.job_type_build == JobType.copr_build
--- a/tests/unit/test_build_helper.py
+++ b/tests/unit/test_build_helper.py
@@ -260,7 +260,7 @@ def _mock_targets(jobs, job, job_type):
                 ),
             ],
             (JobConfigTriggerType.pull_request, ProjectEventModelType.pull_request),
-            set(STABLE_VERSIONS),
+            {"fedora-stable"},
             id="test_with_targets",
         ),
         pytest.param(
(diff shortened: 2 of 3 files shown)


mfocko (approved):

mfocko:
I guess blocked on updating the affected configs, therefore adding the label

lbarcziova:
hi @martinpitt. We have been discussing other possibilities, as we realise this makes the setup a bit worse, but it looks this is most straightforward solution to the issue without creating another possible corner case issues. 

But thinking about your config, I think a separate `copr_build` job with using the aliases should work as well, I will play with that and push the changes to the PR. The `<<` syntax doesn't work as it is supported for maps and not lists.

mfocko:
I didn't expect this to get even worse :D

Issue #2713, opened 2025-02-11
Run package-specific tests on Fedora dist-git pull requests


lachmanfrantisek:
Similarly to the upstream testing, users can have a custom tmt setup for their project.

The task of this is to run custom tmt tests when they are present. Unlike with upstream, we need to run both default checks (e.g. installability) AND the package-specific ones.

*This is a part of the phase 3 of https://github.com/packit/packit-service/issues/2692*

PR #2792, opened 2025-05-06
Run custom tests on dist-git pull requests

nforro:
Fixes https://github.com/packit/packit-service/issues/2713.

---

A couple of notes:

- I've set initiator to `packit-fedora-ci` (thanks @mfocko for the suggestion), is that ok, should it be `packit` or `fedora-ci`?
- The check name is `custom test(s)`, the original Fedora CI calls it `dist-git test`, is that ok?

---

RELEASE NOTES BEGIN

Packit now runs custom TMT tests as part of CI on dist-git pull requests.

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/helpers/testing_farm.py
+++ b/packit_service/worker/helpers/testing_farm.py
@@ -1385,7 +1385,9 @@ class DownstreamTestingFarmJobHelper:
             msg = "We were not able to map distro to TF compose."
             return TaskResults(success=False, details={"msg": msg})
 
-        payload = FEDORA_CI_TESTS.get(test_run.data["fedora_ci_test"])(self, compose)
+        payload = FEDORA_CI_TESTS.get(test_run.data["fedora_ci_test"])(
+            self, distro=test_run.target, compose=compose
+        )
 
         endpoint = "requests"
 
@@ -1408,7 +1410,7 @@ class DownstreamTestingFarmJobHelper:
         )
 
     @implements_fedora_ci_test("installability")
-    def _payload_installability(self, compose: str) -> dict:
+    def _payload_installability(self, distro: str, compose: str) -> dict:
         git_repo = "https://github.com/fedora-ci/installability-pipeline.git"
         git_ref = (
             commands.run_command(["git", "ls-remote", git_repo, "HEAD"], output=True)
@@ -1416,13 +1418,13 @@ class DownstreamTestingFarmJobHelper:
             .split()[0]
         )
 
-        if self.koji_build.target == "rawhide":
+        if distro == "fedora-rawhide":
             # profile names are in "fedora-N" format
             # extract current rawhide version number from its candidate tag
             candidate_tag = self.koji_helper.get_candidate_tag("rawhide")
             profile = re.sub(r"f(\d+)(-.*)?", r"fedora-\1", candidate_tag)
         else:
-            profile = compose.lower()
+            profile = distro
 
         return {
             "api_key": self.service_config.testing_farm_secret,
--- a/packit_service/worker/helpers/testing_farm.py
+++ b/packit_service/worker/helpers/testing_farm.py
@@ -1343,7 +1343,7 @@ class DownstreamTestingFarmJobHelper:
 
     @staticmethod
     def get_check_name(test_name: str) -> str:
-        return f"Packit - {test_name} test"
+        return f"Packit - {test_name} test(s)"
 
     def report(
         self,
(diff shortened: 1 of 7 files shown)


packit_service/worker/helpers/testing_farm.py
```diff
@@ -1408,21 +1410,21 @@ def prepare_and_send_tf_request(
...
+    def _payload_installability(self, distro: str, compose: str) -> dict:
         git_repo = "https://github.com/fedora-ci/installability-pipeline.git"
         git_ref = (
             commands.run_command(["git", "ls-remote", git_repo, "HEAD"], output=True)
             .stdout.strip()
             .split()[0]
         )
 
-        if self.koji_build.target == "rawhide":
+        if distro == "fedora-rawhide":
```
mfocko, line 1428: Any specific reason why you switched? :eyes:

packit_service/worker/helpers/testing_farm.py
```diff
@@ -1459,6 +1459,49 @@ def _payload_installability(self, distro: str, compose: str) -> dict:
             },
         }
 
+    @implements_fedora_ci_test(
+        "custom",
+        skipif=lambda service_config, project, metadata: not project.get_file_content(
+            path=".fmf/version", ref=metadata.commit_sha
+        ),
```
mfocko, line 1466: Won't work, this raises

https://github.com/packit/ogr/blob/f68be67a30293c39970fc219feca1323dc831bd8/ogr/services/pagure/project.py#L474-L475

packit_service/worker/helpers/testing_farm.py
```diff
@@ -1459,6 +1459,49 @@ def _payload_installability(self, distro: str, compose: str) -> dict:
...
+                    },
+                    "artifacts": [
+                        {"id": self.koji_build.task_id, "type": "fedora-koji-build"},
+                    ],
+                    "tmt": {
+                        "context": {
+                            "distro": distro,
+                            "arch": "x86_64",
+                            "trigger": "commit",
+                            "initiator": "packit-ci",
```
mfocko, line 1492: As for the question in the description, I'm not sure if tmt/TF validates the `initiator` value, but IMO this should be eventually switched to `fedora-ci` and also Packit is CI anyways, so… maybe `packit-fedora-ci` would be good substitute for now

packit_service/worker/helpers/testing_farm.py
nforro: It makes no difference here, but it does below (lowercased compose might not always be correct, distro is a better bet), so the change here is just for consistency sake.

packit_service/worker/helpers/testing_farm.py
nforro: Oh, I must be blind, I thought I "copied" this:

https://github.com/packit/packit-service/blob/6e158b33e6dfe5c3a5c8bf2d8c57ea34d3a503c8/packit_service/worker/helpers/testing_farm.py#L658-L665

packit_service/worker/helpers/testing_farm.py
nforro: I don't think it validates it, but some test plans may rely on it, I think we have even some such condition in specfile (not to run certain tests on dist-git PRs, only in upstream CI). `packit-fedora-ci` sounds good.

mfocko (approved):

packit_service/worker/helpers/testing_farm.py
```diff
@@ -1450,6 +1459,60 @@ def _payload_installability(self, compose: str) -> dict:
...
+                    },
+                    "artifacts": [
+                        {"id": self.koji_build.task_id, "type": "fedora-koji-build"},
+                    ],
+                    "tmt": {
+                        "context": {
+                            "distro": distro,
+                            "arch": "x86_64",
+                            "trigger": "commit",
+                            "initiator": "packit-fedora-ci",
```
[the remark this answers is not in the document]
nforro: That was my concern as well, on the other hand until the original Fedora CI is deprecated we would have two services and no way to differentiate between them (but maybe that's not a concern).

@packit/the-packit-team WDYT? Should I change it to `fedora-ci`?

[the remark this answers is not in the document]
lbarcziova: +1 for changing to `fedora-ci` for the compatibility.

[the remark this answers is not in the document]
nforro: https://github.com/packit/packit-service/pull/2803

PR #2797, opened 2025-05-09
Fix/tf client

### LINKED ISSUE #2796: Switch `Authorization` header for TF instead of secrets in payload
After introducing Testing Farm client class, it has been discovered that we're actually not using the API token from the client itself, but still are passing it into the payload from outside.

This breaks the use-case of job canceling that relies only on passing the request ID and everything else is expected to be handled by the client class itself.

There is a question of how to handle this in the least intrusive and the most correct way, since `api_key` being part of the payload itself, would mean that the client class itself injects (i.e., overrides some parts of the payload) into payload which could lead to some unexpected behavior, e.g., API tokens for different ranches.

Additionally usage of `api_key` has been deprecated since May 2024, see [TFT-2533](https://issues.redhat.com/browse/TFT-2533). Therefore it would make the most sense to just switch to the new way instead of hacking around an existing solution which is to be deprecated anyways.

mfocko:
TODO:

- [ ] ‹fill in›

Fixes #2796

Related to #2796

### DIFF
--- a/packit_service/constants.py
+++ b/packit_service/constants.py
@@ -256,17 +256,22 @@ GITLAB_ISSUE = (
 DASHBOARD_JOBS_TESTING_FARM_PATH = "/jobs/testing-farm-runs"
 
 # https://docs.testing-farm.io/general/0.1/test-environment.html#_supported_architectures
-PUBLIC_TF_ARCHITECTURE_LIST: list[str] = [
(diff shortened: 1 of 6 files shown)


lbarcziova (approved):

lbarcziova (approved):
👍

PR #2798, opened 2025-05-12
Fix dist-git PR git URL passed to TF

nforro:
In Pagure a project web URL and git URL are different and web URL can not be used for cloning.

Related to https://github.com/packit/packit-service/issues/2713.


mfocko (approved):
Damn you Pagure

Issue #2777, opened 2025-04-11
Pre-release need to go against a side tag


nievesmontero:
The pre-releases of a package, usually beta releases, need to go against a side tag, since they usually have major changes that depend on many other packages. This means that they cannot be automatically and individually updated, they need to go all together against a side tag.

nforro:
IIUC, this is about implementing a run condition so that a different `koji_build`/`bodhi_update` job will run under different circumstancies (in this case version is/isn't a pre-release).

nforro:
We could implement this as a job-level action that would run as part of pre-check here: https://github.com/packit/packit-service/blob/11d4722765650d350cab89ba8ec80fcfd0736cfa/packit_service/worker/jobs.py#L705-L713

Git repo wouldn't be cloned at that time, but we could pass the event dict (or a sanitized subset of it) as an environment variable to the action.

PR #2810, opened 2025-05-21
Implement run-condition action

nforro:
Fixes https://github.com/packit/packit-service/issues/2777.

Merge after https://github.com/packit/packit/pull/2608.

---

For now it's enabled only for downstream jobs. It seems there are no test data for issue comments, so test cases for those are missing (I'll try to add them), but other cases should be covered, and I added two upstream cases as well. Eventually I'd like to cover everything.


packit_service/worker/checker/run_condition.py
```diff
@@ -0,0 +1,192 @@
...
+            from sandcastle.exceptions import SandcastleCommandFailed
+
+            if not actions_handler.has_action(ActionName.run_condition):
+                return True
+            try:
+                actions_handler.run_action(
+                    actions=ActionName.run_condition, env=self.common_env(version, extra_env)
+                )
+            except (PackitCommandFailedError, SandcastleCommandFailed):
+                return False
```
mfocko, line 181: oof, I hope this doesn't trip false positives from network flakes, cluster, etc.

mfocko (approved):

packit_service/worker/checker/run_condition.py
nforro: It could. `SandcastleCommandFailed` has a `rc` attribute, but that still doesn't have to be the actual exit code of the command. Maybe it would be safer to expect a specific stdout output from the action?

mfocko: I guess it would be way too much to expect from them to provide a normal type instead of raw JSON 🤦

packit_service/worker/checker/run_condition.py
```diff
@@ -0,0 +1,197 @@
...
+            github.check.PullRequest.event_type(),
+        ):
+            project = self.project.get_pr(int(self.data.pr_id)).source_project
+        elif self.data.event_type in (anitya.NewHotness.event_type(),):
+            project = self.service_config.get_project(
+                url=self.data.event_dict.get("distgit_project_url")
+            )
+            git_ref = self.data.tag_name
+            version = self.data.event_dict.get("version")
+        elif self.data.event_type in (
```
lbarcziova, line 122: this could be in theory also retrigger of `propose-downstream`, in which case the project should be probably the upstream one? I am not sure conditioning this only based on event type is the right way, but can't think of what would be better way for now (my first idea was based on the handler, but that would not be enough as well).

nforro: > perhaps the event doesn't have `dist_git_project_url` set in `propose-downstream` case?

Thinking about it, that would make zero sense. So I suppose the only way is to parse the comment.

lbarcziova: > Thinking about it, that would make zero sense. So I suppose the only way is to parse the comment.

So could we use the information about handler so that we don't duplicate the parsing etc?

nforro: In the end I decided to pass in task name as a string to avoid circular imports, WDYT?

packit_service/worker/checker/run_condition.py
```diff
@@ -0,0 +1,204 @@
...
+            pagure.pr.Comment.event_type(),
+            gitlab.mr.Comment.event_type(),
+            pagure.pr.Flag.event_type(),
+            github.check.PullRequest.event_type(),
+        ):
+            project = self.project.get_pr(int(self.data.pr_id)).source_project
+        elif self.data.event_type in (anitya.NewHotness.event_type(),):
+            project = self.service_config.get_project(
+                url=self.data.event_dict.get("distgit_project_url")
+            )
```
lbarcziova, line 125: is this actually needed? (isn't it just`self.project` also?)

packit_service/worker/checker/run_condition.py
lbarcziova: looks good 👍

packit_service/worker/checker/run_condition.py
nforro: Yes, `self.project` is the upstream project (associated with the issue).

lbarcziova: from [this](https://github.com/packit/packit-service/blob/main/packit_service/events/anitya/abstract.py#L48-L58) it looks like the switching happens in the event class, or am I missing something?

nforro: > from [this](https://github.com/packit/packit-service/blob/main/packit_service/events/anitya/abstract.py#L48-L58) it looks like the switching happens in the event class, or am I missing something?

Yes, but that's not propagated to the dict returned from `get_dict()` that is passed to the checker. But I stand corrected, `self.project` is not an upstream project in this case, it is `None`.

nforro: That makes me think if rewriting the code so that the processing is based on task name rather than event type and associated event objects are reconstructed from event data wouldn't make more sense, it could at least be more readable.

lbarcziova: might be, I am also still thinking if this couldn't be implemented on handlers as some class methods maybe, without hitting the circular imports for that.

nforro: I'm not sure the circular import can be avoided, unless the "checker" class itself is defined in the same module as handlers (or handler abstract classes).

PR #2813, opened 2025-05-27
fix(tf): differentiate ranches when canceling

### LINKED ISSUE #2808: Fix cancelling of TF requests to different ranches
Sentry Issue: [PCKT-002-PACKIT-SERVICE-9R3](https://red-hat-it.sentry.io/issues/6607978580/?referrer=github_integration)

```
Failed to cancel TF request xxxxx: {'message': 'Not authorized to perform this action'}
```

As we have discussed with @lbarcziova, it appears that filtering the candidates for cancelling based on just the commit SHA can yield also test runs that are not triggered on the same ranch (i.e., `public` and `redhat` (internal) ones), since the tokens are different, we cannot cancel `redhat` runs with `public` token and vice versa…

Reasoning for labels:
- `impact/high` -- as of now it's enabled globally, therefore affects everyone
- `gain/low` -- it just spams Sentry…

mfocko:
As we are given different API tokens for `public` and `redhat` ranch, it is not possible to cancel the running tests on public ranch using redhat token and vice versa, therefore we need to be able to differentiate between the ranches where the tests are being run.

Since we can trigger multiple targets on one ranch and they are all tied to one test job in Packit config, it is possible to keep `ranch` on the TF test group model.

It is also possible to deduce the ranch of the previous tests based on the URL (URL is kept on the target; Packit test target = TF pipeline).

When canceling Testing Farm runs, filter additionally by the ranch to not fail on the 403.

Fixes #2808

TODO:

- [ ] ‹fill in›

Fixes #2808

RELEASE NOTES BEGIN

Packit now supports automatic ordering of ☕ after all checks pass.

RELEASE NOTES END


nforro (approved):
LGTM

alembic/versions/43882376fe16_add_ranch_to_the_tf_groups.py
```diff
@@ -0,0 +1,142 @@
...
+def upgrade():
+    bind = op.get_bind()
+    session = orm.Session(bind=bind)
+
+    op.add_column("tft_test_run_groups", sa.Column("ranch", sa.String(), nullable=True))
+
+    # Populate the ranches
+    targets = select(TFTTestRunTargetModel)
+    for (target,) in session.execute(targets):
+        if target.web_url is None:
```
lbarcziova, line 129: just a nitpick, since we are settting this on the group level, we could skip if the group has the ranch already set?

lbarcziova (approved):
thanks!

PR #2823, opened 2025-07-01
Prepopulate fedora-distro-aliases cache when building worker image

nforro:
Fixes https://github.com/packit/packit/issues/2633.

### DIFF
--- a/files/docker/Dockerfile.worker
+++ b/files/docker/Dockerfile.worker
@@ -33,4 +33,7 @@ RUN git rev-parse HEAD >/.packit-service.git.commit.hash \
     && rm -rf /src/*
 # /src content is no longer needed, clean it for 'hardly'
 
+# prepopulate fedora-distro-aliases cache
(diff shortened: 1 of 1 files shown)


files/docker/Dockerfile.worker
```diff
@@ -33,4 +33,7 @@ RUN git rev-parse HEAD >/.packit-service.git.commit.hash \
     && rm -rf /src/*
 # /src content is no longer needed, clean it for 'hardly'
 
+# prepopulate fedora-distro-aliases cache
+RUN python3 -c 'from fedora_distro_aliases import get_distro_aliases; get_distro_aliases(cache=True)'
```
mfocko, line 37: This will fail with exit code 1 on exception, right?

mfocko (approved):

files/docker/Dockerfile.worker
nforro: Yes, and it does fail now that Bodhi is down.

PR #2824, opened 2025-07-03
fix(worker): fetch the host keys from admin.fedoraproject.org

mfocko:
As it appears, there is some ongoing issue with `ssh-keyscan` that blocks running this action. The host keys are also present at a specific subpage of admin.fedoraproject.org, so fetch them from there until this issue is resolved.

Also pods trying to retry "bootstrap" can be in issue because of the Fedora's DDoS protection.

I've also considered "baking in" the known hosts, but that could result in issues during the image build.

TODO:

- [ ] ‹fill in›


mfocko:
the prettier fails only in Zuul… fml…

nforro:
> the prettier fails only in Zuul… fml…

Yes, and not just in this repo: https://github.com/packit/packit.dev/pull/1053

nforro (approved):

Issue #2505, opened 2024-08-22
Respect number of items requested for API responses


lbarcziova:
E.g. for `copr-builds` endpoint, we skip items with particular statuses ([here](https://github.com/packit/packit-service/blob/e06642881d6e8d998b5f6a55e9440337ccbb68a9/packit_service/service/api/copr_builds.py#L39)) which may result in lower number of items in the API response than requested, sometimes even no returned items. We should fix this so that the number of items conforms. Check also other API endpoints and fix the occurrences, if present.

PR #2863, opened 2025-11-04
Filter out Copr builds without SRPM in SQL

m-blaha:
In addition to pagination slicing, the CoprBuildsList class uses Python code to filter out Copr builds that are waiting for an SRPM or whose SRPM build failed. This causes that the API in some cases can return fewer items than the user requested.

Moving the filter into SQL resolves the problem. The SQL filter relies on the fact that the build_id field is NULL until the build is actually created by submitting to Copr.

Resolves: https://github.com/packit/packit-service/issues/2505

### DIFF
--- a/tests_openshift/conftest.py
+++ b/tests_openshift/conftest.py
@@ -117,6 +117,7 @@ class SampleValues:
     copr_web_url = "https://copr.something.somewhere/123456"
     koji_web_url = "https://koji.something.somewhere/123456"
     srpm_logs = "some\nboring\nlogs"
+    build_start_time = datetime.datetime(2026, 1, 14, 10, 0, 0)
 
     # TFT
     pipeline_id = "123456"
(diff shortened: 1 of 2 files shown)


packit_service/service/api/copr_builds.py
```diff
@@ -35,15 +34,6 @@ def get(self):
...
-            if build_info.status == BuildStatus.waiting_for_srpm:
-                continue
-            if (
-                build_info.status == BuildStatus.failure
-                and not build_info.build_start_time
-                and not build_info.build_logs_url
-            ):
-                # SRPM build failed, it doesn't make sense to list this build
-                continue
             project_info = build_info.get_project()
```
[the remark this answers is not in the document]
m-blaha: This makes sense, but is out of scope here. If we wanted to optimize the code, it should be tracked as a separate issue.

lbarcziova (commented):
this approach looks good to me! Do you think we could add a testcase in https://github.com/packit/packit-service/blob/main/tests_openshift/database/test_models.py and/or https://github.com/packit/packit-service/blob/main/tests_openshift/service/test_api.py? These tests are not run in CI unfortunately (yet), but are good for this kind of changes; they require compose setup, there is [this](https://github.com/packit/packit-service/blob/847c18545b42b5508bb9684310d72ae3153bf315/Makefile#L158) Make target.

m-blaha:
Sure, will try!

m-blaha:
@lbarcziova I've  added the test

packit_service/models.py
```diff
@@ -2260,6 +2261,19 @@ def get_merged_chroots(
...
+                    # different builds together)
+                    CoprBuildTargetModel.build_id.isnot(None),
+                    # Exclude builds waiting for SRPM - not actual Copr builds yet
+                    CoprBuildTargetModel.status != BuildStatus.waiting_for_srpm,
+                    # Exclude SRPM build failures (no build_start_time/build_logs_url)
+                    or_(
+                        CoprBuildTargetModel.status != BuildStatus.failure,
+                        CoprBuildTargetModel.build_start_time.isnot(None),
+                        CoprBuildTargetModel.build_logs_url.isnot(None),
+                    ),
```
lbarcziova: I checked now how we actually handle SRPM failure, and [here](https://github.com/packit/packit-service/blob/994a28a65b24c0f163285142f50b66e723f1a133/packit_service/worker/handlers/copr.py#L433-L450) it looks like we don't touch the Copr build models at all, we only [change the status to cancelled](https://github.com/packit/packit-service/blob/main/packit_service/worker/helpers/build/copr_build.py#L1022-L1046), if the build is retriggered (by push or similar)

have you found some code that leaves the DB items in this state, or we could remove this and stick to only filtering based on status (probably add also BuildStatus.canceled)

m-blaha: @lbarcziova  I checked the production database and found many `copr_build_targets` records missing `start_time` and `build_logs_url` and marked as `failure`. They are coming at a rate of approximately 80 records per day. It appears there was no `copr.build.Start` event, only `copr.build.End` for these builds (I checked the logs). I'm not sure whether these events were lost or never issued from the copr side. Some of these builds correspond to srpm build failures (e.g. https://copr.fedorainfracloud.org/coprs/packit/osbuild-image-builder-frontend-3989/build/10010130/), but others are regular target build failures where the srpm succeeded (e.g. https://copr.fedorainfracloud.org/coprs/packit/systemd-systemd-40311/build/10010141/). Is this a known issue, or should I file a report?

PR #2876, opened 2025-11-19
Log Detective run database model

jpodivin:
Model for keeping track of Log Detective run requested by packit. 

The table has relations to both copr and koji builds, so we can handle both when the time comes. Single build can have multiple associated Log Detective runs, to account for possibility of run failing.

The mechanism for triggering Log Detective analysis for a build is not part of this PR. Neither is the collection and parsing of messages reporting analysis results. 

Due to inherently asynchronous nature of Log Detective, and latency, the request can not be submitted to Log Detective directly and awaited. Instead, the inteded behavior, is that a failed build will trigger a Log Detective analysis run, by sending a request to a thin interface [server](github.com/fedora-copr/logdetective-packit).

This will respond with a UUID of the Log Detective analysis, to indicate that the request was processed, and will submit logs for analysis by Log Detective itself, awaiting the result. The response of Log Detective server, once received by the interface, will be submitted to fedora message bus. This message will contain a `log_detective_response` field holding response of the Log Detective server, `result` field, indicating if the analysis was successful, `build_system` field, set to either `koji` or `copr` and a `target_build` field, with unique identifier of the build and `log_detective_analysis_id`.

The last field will be used to find the original record in the `log_detective_run` table.

The PR contains a commit from https://github.com/packit/packit-service/pull/2883, because alembic can't generate new revision cleanly.

This means that https://github.com/packit/packit-service/pull/2894 should be merged first, if possible.

Implements: 
https://issues.redhat.com/browse/LD-210

Merge after:

https://github.com/packit/packit-service/pull/2874

https://github.com/packit/packit-service/pull/2894


majamassarini (approved):
LGTM! I tried it locally and it worked on the production db.
Please **rebase** and add the **mergeit** label 🙂

I would also remove the RELEASE NOTES, because this is not yet a user facing change.

PR #2877, opened 2025-11-19
Revert "Synchronizing alembic versioning with existing code (#2874)"

majamassarini:
This reverts commit c05290159f61de85d5e840633c9ef9f6ac27a810, reversing
changes made to a42d5850044c26ee30a2bb78631b141ad4dd37a7.

staging error is:
sqlalchemy.exc.ProgrammingError: (psycopg2.errors.UndefinedObject) type "projecteventmodeltype" does not exist


jpodivin:
I don't like that we don't know what is going on here. And I'm concerned that when I generate revision for https://github.com/packit/packit-service/pull/2876 the change to type is going to show up again. But if it fails in stage, revert is a decent thing to do.

PR #2443, opened 2024-06-17
ci(tmt): run the tests on Testing Farm

mfocko:
TODO:

- [ ] Write new tests or update the old ones to cover new functionality.
- [ ] Update doc-strings where appropriate.
- [ ] Update or write new documentation in `packit/packit.dev`.
- [ ] ‹fill in›

Fixes

Related to

Merge before/after

RELEASE NOTES BEGIN

Packit now supports automatic ordering of ☕ after all checks pass.

RELEASE NOTES END

PR #2881, opened 2025-11-21
Adding workflow for tests

jpodivin:
Unit tests, integration tests and database tests are now executed as github action. The workflow is, intentionally, very similar to what developer would have to use. And it should stay that way if possible.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,40 @@
+name: Tests
+# This workflow is intended to follow testing as performed manually
+# by an actual developer. This includes using make and relying on docker
+# or podman CLI to build images.
+# The intent is to verify that developers can run these tests locally,
+# with no, or at worst minimal, modifications.
+# As such, all changes, that would lead to divergence with recommended
+# developer practice, must be justified in comments.
+# Divergence:
+# SOURCE_BRANCH variable value is set for each execution of make,
+# since the actions/checkout execution switches to a single commit.
+# COMPOSE variable value is set to `podman-compose`, since podman is
+# default for github runners
+
+on:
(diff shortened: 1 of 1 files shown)


mfocko:
As per the `discuss` label:

1. This feels like further fragmentation of the CI setup Packit has (more specifically: pre-commit.ci, Zuul, Packit itself where it's feasible (all the way to the Testing Farm))

2. In a sense it feels unnecessary, e.g., `make check-in-container` is what Zuul basically does.
   - except for `make check-db`… that part was deemed low-priority and was waiting for migration to TF

4. If possible, I'd prefer utilising Testing Farm as has been proposed in #2443, but
   - the intention of aforementioned PR was to unify the testing CI across Packit, but it doesn't allow for reverse-dependency testing of Service itself, because of [TFT-2521](https://issues.redhat.com/browse/TFT-2521)
   - there's been a considerable amount of work spent by @majamassarini on making the OpenShift setup “somewhat” work on Testing Farm; in the context of this PR we also discussed the possibility of using `podman-compose` on TF, which would probably require less effort to make it work in comparison to the OpenShift itself

5. I appreciate the _Build Test Image_ step, it would definitely make sense to me to:
   - run it for PRs, **if** any of the files related to the image build get changed, i.e., any of the `Containerfile`s and (most likely all of) `files/` subtree
   - run it as part of the merge queue **all the time**, though this suggestion will probably be the most problematic, as the setup, right now, depends on Zuul and merging is also managed via Zuul.

---

There's also a possibility of trying out the TF setup even if the rev-dep tests (from Packit and ogr) are not possible right now, making the `make check-db` work with TF and decide on the next steps later based on the outcome of this.

PR #2921, opened 2025-12-18
Add CODEOWNERS entries based on git blame analysis

lbarcziova:
Assign ownership for files with ≥60% contribution from a single author using git blame for current Packit team members.

Open for discussion, we can adjust the 60% threshold/methodology.

Assisted by: Cursor(Claude)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -13,7 +13,107 @@
 # the repo. Unless a later match takes precedence
 *	@packit/the-packit-team
 
+# Core Service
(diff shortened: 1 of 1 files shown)


nforro:
If it is possible to have multiple code owners per file, I would maybe try to have at least two, so there is a backup.
And I would try to factor in recency of the code, someone who's done 10% of changes in the last year will probably be more knowledgeable than someone with 60% done 4 years ago. But I could be wrong :sweat_smile:

PR #2933, opened 2026-01-09
Adding docstrings to more complex checkers.

jpodivin:
While working on code for Log Detective, I have noticed that most checkers don't have any documentation.
If they are short, the code is enough, but in some cases they can be complex and have side effects.

I have written some short explanatory docstrings for those checkers that seemed to need them,
and placed them on the class level, so they can be seen when someone hovers over the class name in IDE.
Technically, you could put them in the `_pre_check` method, but this way it's more user friendly.

I deliberately decided against describing operation of simple checkers and details of operation.
This PR is only focusing on parts that may give people trouble.

RELEASE NOTES BEGIN

Enhanced documentation of checker classes.

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/checker/distgit.py
+++ b/packit_service/worker/checker/distgit.py
@@ -307,7 +307,7 @@ class IsUpstreamTagMatchingConfig(Checker):
     or not among configured excluded tags.
 
     If the tag in event is `None`, for example in case of pull-from-upstream
-    retriggering the check passes.
+    retriggering, the check passes.
     """
 
     def pre_check(self) -> bool:


packit_service/worker/checker/distgit.py
```diff
@@ -30,6 +30,11 @@
 
 
 class LabelsOnDistgitPR(Checker, GetPagurePullRequestMixin):
+    """Verifies that state of labels on the PR matches the configuration.
+
+    The check passes also if the event is not a `pagure.push.Commit`,
+    or if no configuration exists."""
```
majamassarini:
```suggestion
    or if labels are not defined in the configuration."""
```

majamassarini (approved):
Thank you! I have just two small improvements.

Issue #2920, opened 2025-12-18
ELN scratchbuilds for rawhide PRs


hroncok:
### Description

Hi.

Zuul has the check-for-eln and eln-rpm-scratch-build jobs enabled on rawhide Pull Requests.

The purpose of the check-for-eln is to determine whether the package is part of ELN. When it is, the eln-rpm-scratch-build job is not skipped. 

The purpose of the eln-rpm-scratch-build job is to submit an ELN scratchbuild.

Mainly, this serves as a way to verify that a package that is present in ELN still builds fine on ELN. Either because of some differences in the environment or because of the %rhel/%fedora conditionals in the spec file. This is useful when:

 - Submitting a PR that changes some aspects of the build wrt ELN to verify such changes.
 - Regression tests for all other PRs to avoid accidentally breaking the build for ELN.

### Importance

This is a feature that is blocking us from stopping the use of Zuul.

Realistically, however, Zuul is unstable, and most of the time, this doesn't finish successfully.

A workaround is to submit such scratchbuidls manually -- but it's easy to forget, and I doubt many packagers would do that for regression testing.

### Workaround

- [x] There is an existing workaround that can be used until this is implemented.

PR #2937, opened 2026-01-12
Trigger ELN scratch builds from rawhide

nforro:
Fixes https://github.com/packit/packit-service/issues/2920.

### DIFF
--- a/packit_service/constants.py
+++ b/packit_service/constants.py
@@ -36,6 +36,11 @@ TESTING_FARM_EXTRA_PARAM_MERGED_SUBTREES = (
 )
 TESTING_FARM_ARTIFACTS_KEY = "artifacts"
 
+ELN_PACKAGE_LIST = "https://tiny.distro.builders/view-all-source-package-name-list--view-eln.txt"
+ELN_EXTRAS_PACKAGE_LIST = (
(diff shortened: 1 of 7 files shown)


packit_service/worker/jobs.py
```diff
@@ -558,6 +560,33 @@ def process_fedora_ci_jobs(self) -> list[TaskResults]:
...
+                            kwargs={
+                                "package_config": None,
+                                "job_config": None,
+                                "event": self.event.get_dict(),
+                                "rawhide_eln_build": True,
+                            },
+                        )
+                        celery_signature.apply_async()
+                        logger.debug(f"Celery signature sent for handler {handler_kls}.")
+
```
lbarcziova: to me this looks like a good fit for `pre-check` method, would it make sense to subclass the current handler for Koji scratch build and handle this on the `pre-check` level there rather than here?

lbarcziova (approved):
LGTM! I don't see any adjustments related to processing of the build status, so everything should work out of the box?

nforro:
> I don't see any adjustments related to processing of the build status, so everything should work out of the box?

Yes, but I want to verify that on stage. Tests triggering should work as well.

PR #2949, opened 2026-01-16
Avoid checking a stale relationship

majamassarini:


packit_service/models.py
```diff
@@ -3450,9 +3454,11 @@ def create(cls, run_model: "PipelineModel", ranch: str) -> "TFTTestRunGroupModel
                 new_run_model.copr_build_group = locked_run_model.copr_build_group
                 new_run_model.koji_build_group = locked_run_model.koji_build_group
                 new_run_model.test_run_group = test_run_group
+                new_run_model.test_run_group_id = test_run_group.id
```
nforro: Isn't this enough? Why keep the line above?

nforro: Sure, I get that, but my comment is about recording the relationship (in either branch). Both lines do effectively the same thing, don't they?

PR #2952, opened 2026-01-19
Use atomic update when adding a testing farm test

majamassarini:
To prevent race conditions.
Both threads see that the pipeline model has not yet been linked to the test. So both link it to the test. The latest one overwrites the previous value.

The with_for_update clause isn't enough.

- SELECT ... FOR UPDATE + Python check: Two separate operations (read, then check in Python, then write) → race condition possible
- Atomic UPDATE with WHERE clause: One operation that checks and updates atomically → no race condition

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -3447,32 +3447,40 @@ class TFTTestRunGroupModel(ProjectAndEventsConnector, GroupModel, Base):
             # Flush to get test_run_group.id before using it
             session.flush()
 
-            # lock the run_model in the DB by using SELECT ... FOR UPDATE
-            # all other workers accessing this run_model will block here
-            # until the context is exited
-            locked_run_model = (
-                session.query(PipelineModel).filter_by(id=run_model.id).with_for_update().first()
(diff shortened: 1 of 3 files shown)


packit_service/models.py
```diff
@@ -3447,32 +3447,41 @@ def create(cls, run_model: "PipelineModel", ranch: str) -> "TFTTestRunGroupModel
...
-            locked_run_model = (
-                session.query(PipelineModel).filter_by(id=run_model.id).with_for_update().first()
+            # Attempt to update the existing run_model if test_run_group_id is NULL
+            # This is an atomic operation at the DB level, preventing race conditions
+            # Only one thread will successfully update (updated_count=1)
+            # The other thread will get updated_count=0 and proceed to clone
+            updated_count = (
+                session.query(PipelineModel)
+                .filter_by(id=run_model.id, test_run_group_id=None)
+                .update({"test_run_group_id": test_run_group.id}, synchronize_session=False)
```
nforro: Could you explain what `synchronize_session=False` does and why is it used here? Ideally with a link to source.

nforro (approved):
🙏

PR #2962, opened 2026-01-22
Unifying time stamp checking of Testing Farm and Copr build models

jpodivin:
I have worked on resolving TODO in the `get_submitted_time_from_model` function of the `packit_service.models` module. The `CoprBuildTargetModel` now has a `submitted_time`, rather than `build_submitted_time` column, for consistency with other models that previously used the function.

Then I proceeded to do the same for other build target models, `SRPMBuildModel` and `VMImageBuildTargetModel`.

The mmd diagram was regenerated. The new diagram includes unrelated changes to the database schema, which were not used to update diagram due to omission. 

I deliberately avoided changing the API, so the `build_submitted_time` is still used by related responses.


lbarcziova (approved):
LGTM, thanks! I can see `build_submitted_time` is still used for `KojiBuildTargetModel` and `SRPMBuildModel`, but those are not used for the filtering method, so probably could be done as a followup, wdyt?

PR #2971, opened 2026-01-28
Trigger Log Detective for failed downstream Koji builds

Jany26:
This is a complementary handler to the one from https://github.com/packit/packit-service/pull/2905 by @jpodivin 

This handler is supposed to detect changes in (downstream) Koji tasks and if they signal that the build has failed, it creates a json payload, then sends it to a middle-man server (implementation here: https://github.com/fedora-copr/logdetective-packit, the URL itself has not yet been determined) which triggers Log Detective. Additionally, some DB entries for the Log Detective run are also created. Analysis results themselves are then handled in another class. 

RELEASE NOTES BEGIN

Packit can now trigger Log Detective when it detects a failed downstream Koji task.

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/handlers/koji.py
+++ b/packit_service/worker/handlers/koji.py
@@ -67,6 +67,7 @@ from packit_service.worker.handlers.mixin import (
 )
 from packit_service.worker.helpers.build.koji_build import KojiBuildJobHelper
 from packit_service.worker.helpers.fedora_ci import FedoraCIHelper
+from packit_service.worker.helpers.logdetective import LogDetectiveKojiTriggerHelper
 from packit_service.worker.helpers.sidetag import SidetagHelper
 from packit_service.worker.mixin import (
     ConfigFromEventMixin,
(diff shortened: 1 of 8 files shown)


packit_service/worker/handlers/logdetective.py
```diff
@@ -37,6 +43,147 @@
 logger = logging.getLogger(__name__)
 
 
+@reacts_to_as_fedora_ci(event=koji.result.Task)
```
majamassarini: I don't know whether it makes sense to have a new handler triggered by the same event, we already have the Koji handler, which reacts to it.

Personally, I would handle logdetective triggering inside the Koji handler, since it's tightly coupled with the Koji results. It doesn't make sense to me to redo the work already done there.

packit_service/worker/handlers/koji.py
```diff
@@ -157,6 +158,9 @@ def notify_about_failure_if_configured(
         self, packit_dashboard_url: str, external_dashboard_url: str, logs_url: str
     ): ...
 
+    @abstractmethod
```
lbarcziova: since this is abstract, I think you need to create an empty implementation like
```python
def trigger_log_detective_if_configured(self):
      pass
```
in `KojiTaskReportHandler` too

lbarcziova (approved):
LGTM, thanks a lot! I will wait a bit in case other team members want to have a look and merge later today

Jany26:
@lbarcziova done

PR #2972, opened 2026-01-29
Do not proceed if Log Detective analysis is still in progress

jpodivin:
The code for handling cases of Log Detective analysis in progress is not necessary. The counter for started analysis should be incremented when the analysis begins, not when it ends. The check for state was moved before attempted retrieval of analysis record, to spare us a database operation. 

Tests were slightly expanded, and docstring clarified.


lbarcziova (commented):
makes sense, lgtm

PR #2973, opened 2026-01-29
Improving mocks to properly cover new metric and get_rate_limit_remaining method

jpodivin:
Tests were broken following a modification of the source. Since these tests don't run in CI, issues like this are inevitable.

New mock ensures that the rate limit is properly returned, and that the new metric is handled, including assertion on increment.


majamassarini:
> Mocking the database would significantly limit usefulness of our tests. We absolutely do need it.

I don't understand this. You are already testing the database tables, with the tests in the `database` directory. 
What you are doing in the tests you have fixed here, from what I understand, is testing the logic of the handler. 
Personally, I would mock the db to be completely sure I am only testing the logic and nothing else - as we do in the test/integration directory.

However, I am fine with merging this as it is.

PR #2974, opened 2026-01-30
Add config options for Fedora CI to become default

### LINKED ISSUE #2910: Config switch for dist-git CI to become a default solution
We need to prepare Packit service for running as a default solution and also to let stage instance coexist with the  production.

The task is to have config option(s) to allow:

* **prod instance**: run by default, opt-out possible
* **stg instance**: manual by default, auto-trigger opt-in possible

lbarcziova:
Fixes #2910


majamassarini (approved):
LGTM thanks!

PR #2978, opened 2026-02-02
Fix oc exec syntax for newer Openshift CLI compatibility

lbarcziova:
Update all oc exec commands in documentation to use the required `--` separator, as the old syntax without it was removed.


majamassarini (approved):
👍

PR #2981, opened 2026-02-05
Differentiate between handler for upstream/downstream tests

### LINKED ISSUE #2977: Fedora CI reporting for rpmlint and rpminspect tests seems broken
In the ogr dist-git PRs created when releasing ogr version 0.60.1, it seems there is something wrong with reporting the succeeded status of `rmpinspect` and `rpmlint` tests.

We receive the messages for updating the status to success, and in the logs for the tasks there is nothing weird.
I supposed there could be a mismatch from the check status name, when we initially create it, and when we try to update it.
Asked cursor, it tells me:

> Mismatch: the check name includes target_branch, and it's set differently at creation vs update.
Initial creation (jobs.py line 481): target_branch=self.event.pull_request_object.target_branch
Status update (testing_farm.py line 1354): target_branch=self.koji_build.target

This is the proposed patch:

```
diff --git a/packit_service/worker/helpers/testing_farm.py b/packit_service/worker/helpers/testing_farm.py
index d8bda77c..d6db170e 100644
--- a/packit_service/worker/helpers/testing_farm.py
+++ b/packit_service/worker/helpers/testing_farm.py
@@ -1348,10 +1348,14 @@ class DownstreamTestingFarmJobHelper:
     @property
     def ci_helper(self) -> FedoraCIHelper:
         if not self._ci_helper:
+            # Use PR target_branch to match the check name created initially in jobs.py
+            # (which uses self.event.pull_request_object.target_branch)
+            # This ensures the check name is consistent for Pagure, which includes
+            # target_branch in the composed check name.
             self._ci_helper = FedoraCIHelper(
                 project=self.project,
                 metadata=self.metadata,
-                target_branch=self.koji_build.target,
+                target_branch=self.pr.target_branch,
             )
         return self._ci_helper
```

However I am not completely sure. It seems to happen always for rawhide, and sometimes for other target branches. And I am not able to explain it.

lbarcziova:
Without this, if the package uses Packit also for upstream tests, there can be a race condition resulting in not updating status in Pagure.

Fixes #2977

Assuming we will have separate deployments in future, I didn't create a separate column for flagging downstream/upstream test, just reused what we already have in DB. Let me know if you have other preference.

RELEASE NOTES BEGIN

We have fixed a race condition of not reporting the statuses of tests in dist-git properly.

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/checker/testing_farm.py
+++ b/packit_service/worker/checker/testing_farm.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: MIT
 
 import logging
+from typing import Optional
 
 from packit_service.constants import (
     DOCS_TESTING_FARM,
@@ -9,7 +10,8 @@ from packit_service.constants import (
     INTERNAL_TF_TESTS_NOT_ALLOWED,
     KojiTaskState,
 )
-from packit_service.events import gitlab
+from packit_service.events import gitlab, testing_farm
+from packit_service.models import TFTTestRunTargetModel
 from packit_service.worker.checker.abstract import (
     ActorChecker,
     Checker,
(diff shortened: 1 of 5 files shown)


nforro (approved):
LGTM, thanks!

PR #2982, opened 2026-02-06
Fix-copr-builds-race-condition

majamassarini:
It fixes the race condition triggered by the unit test created in the previous commit.

RELEASE NOTES BEGIN

We have fixed a race condition that prevented copr builds to being reported as complete.

RELEASE NOTES END

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -2035,24 +2035,140 @@ class CoprBuildGroupModel(ProjectAndEventsConnector, GroupModel, Base):
         return self.copr_build_targets
 
     @classmethod
-    def create(cls, run_model: "PipelineModel") -> "CoprBuildGroupModel":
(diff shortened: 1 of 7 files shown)


packit_service/models.py
```diff
@@ -2035,24 +2035,112 @@ def grouped_targets(self) -> list["CoprBuildTargetModel"]:
...
                 )
-                new_run_model.srpm_build = run_model.srpm_build
-                new_run_model.copr_build_group = build_group
+                new_run_model.srpm_build = db_run_model.srpm_build
+                new_run_model.copr_build_group_id = build_group.id
                 session.add(new_run_model)
-            else:
-                run_model.copr_build_group = build_group
-                session.add(run_model)
+                session.flush()
```
nforro: Why the flush?

PR #2988, opened 2026-02-10
Expose the deployment as a tmt context

### LINKED ISSUE #2987: Add a tmt context to distinguish between stg and prod
### Description

After we get the common tmt plans repository (https://github.com/packit/packit-service/issues/2914), it would be hard to adjust what plans to run. We could however pass some context to testing-farm, something like `state: stg` (better name needed), which we can use to mark some plans to be skipped.

### Importance

_No response_

### Workaround

- [ ] There is an existing workaround that can be used until this is implemented.

LecrisUT:
TODO:

Fixes #2987

RELEASE NOTES BEGIN

The packit deployment (`prod` or `stg`) is now exposed in the tmt context `deployment`

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/helpers/testing_farm.py
+++ b/packit_service/worker/helpers/testing_farm.py
@@ -580,6 +580,7 @@ class TestingFarmJobHelper(CoprBuildJobHelper):
                     "arch": arch,
                     "trigger": "commit",
                     "initiator": "packit",
+                    "deployment": self.service_config.deployment.name,
                 },
             },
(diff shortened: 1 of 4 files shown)


mfocko (approved):
I don’t think it is a good idea from the perspective of abstraction. IMO it should be irrelevant which deployment triggers the tests, but I guess we can go with it.

Sorry for the late review, there’s been some discussion on the issue and then it probably slipped through the cracks.

Issue #2986, opened 2026-02-10
Broken retries of Bodhi updates


nforro:
There are two related issues here:
1. It seems that recently Bodhi API endpoint started occasionally returning `OIDCClientError("Connection aborted")`. This exception is not handled in `PackitAPI.create_update()` so the task is not retried in the fully automated way without user notification.
2. Generic exceptions still trigger a retry, but the run status is kept in `retry` and that breaks assumptions for `BodhiUpdateTargetModel.get_all_successful_or_in_progress_by_nvrs()` which always succeeds in the new run (because `retry` is considered in progress), skipping the update.

PR #2997, opened 2026-02-15
Set the run status to `error` rather than `retry` on failure

nforro:
When a Bodhi update fails with uncommon exception, the Celery retry mechanism kicks in and the job is retried. However, a new target model and a new run is created for the retry, so it makes sense for the previous run to end in an `error` status rather than `retry`. Also, the `retry` status isn't considered as final by the NVR check, and it doesn't make sense to do so as in case of common exceptions it is guaranteed to transition to some other final status.

Fixes https://github.com/packit/packit-service/issues/2986.

Merge after https://github.com/packit/packit/pull/2688.


lbarcziova (approved):
LGTM

Issue #3004, opened 2026-02-18
Replace session.execute(select(...)) with session.query() in get_running()   methods


m-blaha:
### Description

Several get_running() classmethods use `session.execute(select(Model))` which returns single-element tuples, requiring awkward (build,) unpacking throughout callers.
Switching to session.query(Model) returns model instances directly, matching the pattern used by other query methods in models.py.

Affected methods:

- [ ] CoprBuildGroupModel.get_running()
- [ ] TFTTestRunGroupModel.get_running()
- [ ] LogDetectiveRunModel.get_running()

It's already used in https://github.com/packit/packit-service/pull/3000 for KojiBuildGroupModel.

### Benefit

Cleaner code.

### What is the impacted category (job)?

General

### Workaround

- [ ] There is an existing workaround that can be used until this feature is implemented.

### Participation

- [x] I am willing to submit a pull request for this issue. (Packit team is happy to help!)

PR #3018, opened 2026-02-24
Cleanup: get_running() methods return query

m-blaha:
Makes *.get_running() methods consistent with other model methods that
already return a query. Returns model instances directly instead of
single-element tuples, removing awkward (obj,) unpacking in
all callers.

Resolves: https://github.com/packit/packit-service/issues/3004

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -42,7 +42,6 @@ from sqlalchemy import (
     func,
     null,
     or_,
-    select,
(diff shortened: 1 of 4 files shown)


nforro (approved):
LGTM, but let's get an opinion from @mfocko who introduced the select in https://github.com/packit/packit-service/pull/2774.

PR #3028, opened 2026-02-27
A couple of fixes for Fedora CI (not only) @ `tests` namespace

nforro:


packit_service/worker/parser.py
```diff
@@ -1305,13 +1305,18 @@ def parse_data_from_testing_farm(
...
             }.get(result, summary)
         else:
-            artifact: dict = nested_get(env, "artifacts", 0, default={})
-            a_type: str = artifact.get("type")
-            if a_type == "fedora-copr-build":
-                copr_build_id = artifact["id"].split(":")[0]
-                copr_chroot = artifact["id"].split(":")[1]
+            artifacts = nested_get(env, "artifacts", default=[])
+            if len(artifacts) > 0:
+                artifact: dict = artifacts[0] or {}
```
majamassarini: I am curious why we are getting exactly 0 😅 maybe it is worth a comment? or a link to the documentation which explains this behavior?

majamassarini (approved):

packit_service/worker/parser.py
nforro: > maybe it is worth a comment? or a link to the documentation which explains this behavior?

I don't know, this simply fetches the request we have sent earlier, and that is constructed in the helpers:
https://github.com/packit/packit-service/blob/main/packit_service/worker/helpers/testing_farm.py

PR #3029, opened 2026-02-27
Move copr_build_handler to short running queue

### LINKED ISSUE #2995: Investigate and potentially move CoprBuildHandler to be run in short-running worker
Follow-up of the architecture discussion: this handler has been running in long-running because historically, it did also local building of SRPM and running user-actions. But now this is being done in Copr and this task should not really need to run in long-running worker. This might negatively affect the processing of how quickly we can process these tasks as long-running queue isn't concurrent. 

@majamassarini raised a concern about a possible long duration of this task, therefore we should first check how long it takes and what takes the longest (identify if something can be improved). If confirmed the long duration shouldn't be a frequent problem, we could switch to running this task in short-running workers as well.

majamassarini:
Looking at Grafana, the copr build handler tasks take less time than the copr build start/end handler tasks. The latter already are in the short running queue.
Also the highest peaks, around 12 minutes, all come from the copr build start/end handlers; the highest peak I saw for the copr build handler in the last 30 days was 4 minutes.

So I think this change is safe.

Fix #2995

### DIFF
--- a/packit_service/worker/tasks.py
+++ b/packit_service/worker/tasks.py
@@ -279,7 +279,7 @@ def run_copr_build_end_handler(event: dict, package_config: dict, job_config: di
     return get_handlers_task_results(handler.run_job(), event)
 
 
-@celery_app.task(bind=True, name=TaskName.copr_build, base=TaskWithRetry, queue="long-running")
+@celery_app.task(bind=True, name=TaskName.copr_build, base=TaskWithRetry)
 def run_copr_build_handler(
     self,
     event: dict,


lbarcziova (approved):
👍

lbarcziova:
is this waiting for something or can we merge so that we can see behaviour in staging for a few days?

majamassarini:
> is this waiting for something or can we merge so that we can see behaviour in staging for a few days?

sorry, I forgot about it, I was waiting  for the Monday deployment...

PR #3039, opened 2026-03-04
Remove hardcoded list of supported Fedora CI tests

### LINKED ISSUE #3038: Replace hardcoded values
The test types that can be triggered in Fedora CI are hardcoded in the parser meant for parsing Fedora CI retriggering comments.

https://github.com/packit/packit-service/blob/1d3ed538ebea6a66d7cd57fedd2f787ef9bd8925/packit_service/utils.py#L329

Additional support of new types of tests without updating this hardcoded list would result of retriggering of those tests not being possible. Remove the hardcoded values and use `DownstreamTestingFarmJobHelper.get_fedora_ci_tests()` instead.

betulependule:
The Fedora CI comment parser no longer uses a hardcoded list supported test targets. Now it relies on the `get_fedora_ci_tests` to retrieve all available test targets in the given context. The `implements_fedora_ci_test` has been overhauled to accept a list of Checkers rather than a singular Callable, making the code easier to read and easy to add additional Checkers. Two new Checkers `IsFMFConfigPresent` and `IsProjectOutsideOfTestsNamespace` have been added to replace previously used methods.

Fixes #3038

### DIFF
--- a/packit_service/worker/helpers/testing_farm.py
+++ b/packit_service/worker/helpers/testing_farm.py
@@ -1339,7 +1339,10 @@ class DownstreamTestingFarmJobHelper:
 
     @staticmethod
     def get_fedora_ci_tests(
-        service_config: ServiceConfig, project: GitProject, metadata: EventData
+        service_config: ServiceConfig,
+        project: GitProject,
+        metadata: EventData,
+        filter_specific_test: bool = True,
     ) -> list[str]:
         """
         Gets relevant Fedora CI tests registered using the `@implements_fedora_ci_test()` decorator.
@@ -1351,6 +1354,7 @@ class DownstreamTestingFarmJobHelper:
             service_config: Service config.
             project: Git project.
             metadata: Event metadata.
+            filter_specific_test: Whether to filter tests based on the command in user's comment.
 
         Returns:
             List of registered Fedora CI test names.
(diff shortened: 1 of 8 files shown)


packit_service/worker/jobs.py
```diff
@@ -281,6 +237,60 @@ def process_message(
...
+        try:
+            args = parser.parse_args(commands)
+            return ParsedComment(command=args.command, package=args.package)
+        except SystemExit:
+            # tests expect invalid syntax comments be ignored
+            logger.debug(
+                f"Comment {comment} uses unexpected syntax or contains unsupported commands. "
+                "It will be ignored.",
+            )
+            return ParsedComment()
```
nforro: Wouldn't some feedback be nice? Implementing a response comment should be simple, we already provide feedback for accepted retrigger tasks:
https://github.com/packit/packit-service/blob/1d3ed538ebea6a66d7cd57fedd2f787ef9bd8925/packit_service/worker/jobs.py#L1339

We could either forward the error from the `SystemExit` exception or use `parser.parse_known_args()` and report unknown arguments.

packit_service/worker/helpers/testing_farm.py
```diff
@@ -1303,31 +1308,47 @@ def get_fedora_ci_tests(
...
+            for name, (_, checkers) in FEDORA_CI_TESTS.items()
+            if not any(
+                checker(
+                    package_config=None,
+                    job_config=None,
+                    event=metadata.event_dict,
+                    task_name=None,
+                ).pre_check()
+                for checker in checkers
+            )
```
nforro: Isn't the logic reversed? If the check passes, the test should be available.

nforro (approved):
Nice, thanks!

packit_service/worker/helpers/testing_farm.py
betulependule: I think you're right. I'll go ahead and turn the `IsFMFConfigMissing` and `IsProjectInTestsNamespace` to `IsFMFConfigPresent` and `IsProjectOutsideOfTestsNamespace` and reverse the logic here. Thanks.

betulependule:
It seems that resolving merge conflicts introduced a new circular dependency. I'll try to resolve it.

nforro (approved):
Thanks!

PR #3058, opened 2026-03-16
Add --version argument to pull-from-upstream comment command

### LINKED ISSUE #3033: Support version argument when retriggering pull-from-upstream
In case of multiple release streams and pull-from-upstream configured for one of them (e.g. with `upstream_tag_include`/`upstream_tag_exclude` or `version_update_mask`), retriggering such job will work only if the configured release stream is the highest, because the version to update to is the highest retrieved from release-monitoring.org. I don't think it's possible to determine the correct stream from the configuration, so allowing users to specify the version when retriggering seems to be the only option to make it work for other streams.

grulja:
Allow users to specify a version string when retriggering via a dist-git PR comment.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

TODO:

- [ x] Write new tests or update the old ones to cover new functionality.

Fixes #3033

RELEASE NOTES BEGIN

Users can now specify a version string when retriggering `pull-from-upstream`  via a dist-git PR comment:
```
/packit pull-from-upstream --version 1.2.3
```
This overrides the version that Packit would otherwise resolve automatically.

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/mixin.py
+++ b/packit_service/worker/mixin.py
@@ -21,6 +21,7 @@ from packit_service.constants import (
     SANDCASTLE_LOCAL_PROJECT_DIR,
 )
 from packit_service.events.event_data import EventData
+from packit_service.utils import get_packit_commands_from_comment
 from packit_service.worker.helpers.job_helper import BaseJobHelper
 from packit_service.worker.reporting import BaseCommitStatus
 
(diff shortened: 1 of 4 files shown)


packit_service/worker/mixin.py
```diff
@@ -199,13 +199,17 @@ def clean_api(self) -> None:
 class GetSyncReleaseTagMixin(PackitAPIWithUpstreamMixin):
     _tag: Optional[str] = None
 
+    def get_version_from_comment(self) -> Optional[str]:
+        return None
```
nforro: It would probably make more sense if the implementation from `PullFromUpstreamHandler` was here instead.

Issue #3067, opened 2026-03-20
LogDetectiveKojiTriggerHelper is constructing invalid URL for Log Detective


jpodivin:
### What happened? What is the problem?

The helper uses top level Koji task to construct the URL. Whereas logs are available only for children tasks.

### What did you expect to happen?

URL should point to actual build logs.

### Steps to reproduce

```bash
1.Have a koji build fail.
2. Observe how Log Detective fails to get the log.
3.
```

### What is the impacted category (job)?

Other

### Workaround

- [ ] There is an existing workaround that can be used until this issue is fixed.

### Participation

- [x] I am willing to submit a pull request for this issue. (Packit team is happy to help!)

PR #3071, opened 2026-03-23
Trigger LD for the correct buildArch taskID

Jany26:
Instead of the parent task ID, we now pass the failed buildArch child task IDs.
This should prevent 404 errors when accessing logs at `https://kojipkgs.fedoraproject.org/work/tasks/YYYY/XXXXYYYY/build.log`.

Additionally, the CI reports are now consistent with other tasks, they include deployment and target-arch / chroot information.

TODO:

- added `rpm_build_task_states` dictionary for arch -> task state lookup
- we only trigger LD for failed buildArch subtasks
- there might be an edge case possible, where the parent task (scratch build) fails, but no failed buildArch task is found -> this happens for a failed SRPM build
- we might consider adding a fallback mechanism for this case where we would search for an SRPM build (the change is fairly simple -> we could add a 'srpm' to the `rpm_build_task_ids` and `rpm_build_task_states`)
- hopefully the LDRunModel and LDRunGroupModel are correctly used
- "target" field in the LDRunModel now contains not only build target, but also arch -> for distinguishing (perhaps this also needs a change into separate fields)

Fixes https://github.com/packit/packit-service/issues/3067

Related to https://github.com/packit/packit-service/pull/2971

RELEASE NOTES BEGIN

Packit now triggers LogDetective with the correct artifact URLs for failed downstream Koji scratch builds.
LogDetective CI jobs are now properly reported.

RELEASE NOTES END

### DIFF
--- a/packit_service/worker/handlers/logdetective.py
+++ b/packit_service/worker/handlers/logdetective.py
@@ -43,7 +43,7 @@ class DownstreamLogDetectiveResultsHandler(
 ):
     __test__ = False
     task_name = TaskName.downstream_log_detective_results
-    check_name = "Log Detective Analysis"
+    _check_name = "Log Detective analysis"
 
     def __init__(self, package_config: PackageConfig, job_config: JobConfig, event: dict):
         super().__init__(package_config, job_config, event)
@@ -132,8 +132,17 @@ class DownstreamLogDetectiveResultsHandler(
             self.branch_name = build.get_branch_name()
 
         url = build.web_url or ""
+        # LDRunModel.target is "target-arch" for Koji (i.e. rawhide-x86_64),
+        # for Copr it would be chroot which also includes arch information
+        self._ci_helper = FedoraCIHelper(
+            project=self.project,
+            metadata=self.data,
+            target_branch=log_detective_run_model.target or self.branch_name,
+        )
         self.report(
-            state=status, description=f"Log Detective analysis status: {self.status.value}", url=url
+            state=status,
+            description=f"Log Detective analysis status: {self.status.value}",
+            url=url,
         )
 
         if self.log_detective_response:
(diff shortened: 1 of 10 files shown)


tests/unit/test_logdetective_koji_helper.py
```diff
@@ -55,15 +55,17 @@ def mock_koji_task_failed_event():
...
-        target="fedora-rawhide-x86_64",
+        target="rawhide",
         build_model=mock_build_model,
+        rpm_build_task_ids={"x86_64": 12345},
+        rpm_build_task_states={"x86_64": 5},
     )
 
-    event.should_receive("get_koji_build_logs_url").with_args(12345).and_return(
-        "https://kojipkgs.fedoraproject.org//work/tasks/2345/12345/build.log"
+    event.should_receive("get_koji_build_task_all_logs_urls").with_args("x86_64").and_return(
```
lbarcziova: I wasn't able to find `get_koji_build_task_all_logs_urls` 🤔

Jany26: OH, you're right, forgot to fix this. This was from a previous attempt from accessing log names through the `event.info.children[].result.logs` but then I found out that some failed buildArch child tasks can have null result.

packit_service/worker/handlers/logdetective.py
```diff
@@ -132,8 +132,17 @@ def _run(self) -> TaskResults:
             self.branch_name = build.get_branch_name()
 
         url = build.web_url or ""
+        # target includes arch for Koji
+        check_name = (
+            f"{self.check_name} ({log_detective_run_model.target})"
+            if log_detective_run_model.target
+            else self.check_name
+        )
```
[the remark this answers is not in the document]
jpodivin: I believe we are already setting branch_name using either information from the PR, or from the `KojiBuildTagModel` or `GitBranchModel`. Wouldn't that be enough.

https://github.com/packit/packit-service/blob/23e89b56b0053a302d7a0abdf7fc4fa9ee04c388/packit_service/worker/handlers/logdetective.py#L129-L132

jpodivin:
The derivation of URL will be solved by https://github.com/packit/packit-service/pull/3075

Issue #2983, opened 2026-02-06
valkey-pvc requires periodic increases


majamassarini:
The valkey-pvc defined in the deployment has 1Gi of resources: requests: storage.                                                                                                                                                                                                       
Today Valkey stopped working, and the Packit stack was stuck because of Write error while saving DB to the disk (rdbSaveRio): No space left on device.                                                                                                                                   
                                                                                                                                                                                                                                                                                           
Looking at the valkey-pvc deployed in production**,** its size was already increased to 2Gi.                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                           
Today we increased it to 3Gi.                                                                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                           
We should investigate whether there is a safe way to clean data in there.                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                           
If not, we need at least some sort of alert to be able to increase the size before the Packit stack gets stuck.

PR #3085, opened 2026-04-01
Add periodic cleanup for orphaned Celery pidbox queues

majamassarini:
Problem: Celery workers create pidbox (control) reply queues for worker management commands (inspect, ping, stats, etc.). These queues accumulate when workers crash or restart improperly, leading to:
- 1,693+ orphaned *.reply.celery.pidbox keys in production
- Keys with no TTL (TTL = -1) that persist indefinitely

Root cause: Celery's Redis transport does not provide a native way to set TTL on pidbox reply queues when they're created. These are internal implementation details of Celery's broadcast/control mechanism, and there's no configuration option to automatically expire them.

Solution: Heartbeat cleanup task Since we cannot tell Celery to natively set TTL on pidbox messages, we implement a periodic heartbeat task that:
- Runs nightly at 12:30 AM via Celery beat
- Scans for *.reply.celery.pidbox keys without TTL
- Sets 1-hour expiration on orphaned queues
- Tracks total Redis keys via Prometheus for monitoring

Related to: https://github.com/packit/deployment/pull/701
Should fix: https://github.com/packit/packit-service/issues/2983

### DIFF
--- a/packit_service/celerizer.py
+++ b/packit_service/celerizer.py
@@ -6,7 +6,31 @@ from os import getenv
 from celery import Celery
 from lazy_object_proxy import Proxy
 
+from packit_service.constants import (
+    REDIS_DEFAULT_CELERY_BACKEND,
+    REDIS_DEFAULT_DB,
+    REDIS_DEFAULT_HOST,
+    REDIS_DEFAULT_PASSWORD,
+    REDIS_DEFAULT_PORT,
+)
 from packit_service.sentry_integration import configure_sentry
+
+
+def get_redis_config():
+    """
+    Get Redis connection configuration from environment variables.
+
+    Returns:
+        dict: Redis configuration with keys: host, password, port, db, celery_backend
+    """
+    return {
+        "host": getenv("REDIS_SERVICE_HOST", REDIS_DEFAULT_HOST),
+        "password": getenv("REDIS_PASSWORD", REDIS_DEFAULT_PASSWORD),
+        "port": getenv("REDIS_SERVICE_PORT", REDIS_DEFAULT_PORT),
+        "db": getenv("REDIS_SERVICE_DB", REDIS_DEFAULT_DB),
+        "celery_backend": getenv("REDIS_CELERY_BACKEND")
+        or getenv("REDIS_CELERY_BECKEND", REDIS_DEFAULT_CELERY_BACKEND),
+    }
 
 
 class Celerizer:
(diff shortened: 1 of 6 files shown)


packit_service/worker/tasks.py
```diff
@@ -994,3 +997,81 @@ def get_usage_statistics() -> None:
...
+
+                # Set TTL if key exists but has no expiry (TTL = -1)
+                if redis_client.ttl(key) == -1:
+                    redis_client.expire(key, REDIS_PIDBOX_TTL_SECONDS)
+                    keys_with_ttl_set += 1
+                    logger.debug(f"Set TTL on pidbox key: {key}")
+
+            # Break when cursor returns to 0 (full scan complete)
+            if cursor == 0:
+                break
```
mfocko:
```suggestion
        cursor = -1
        keys_processed = 0
        keys_with_ttl_set = 0
        pattern = "*.reply.celery.pidbox"

        # Break when cursor set to 0 (full scan complete)
        while cursor != 0:
            cursor, keys = redis_client.scan(
                cursor=cursor,
                match=pattern,
                count=100,
            )

            for key in keys:
                keys_processed += 1

                # Set TTL if key exists but has no expiry (TTL = -1)
                if redis_client.ttl(key) == -1:
                    redis_client.expire(key, REDIS_PIDBOX_TTL_SECONDS)
                    keys_with_ttl_set += 1
                    logger.debug(f"Set TTL on pidbox key: {key}")
```

no hard opinions on this :D

packit_service/celerizer.py
```diff
@@ -6,23 +6,43 @@
...
+    REDIS_DEFAULT_CELERY_BACKEND,
+    REDIS_DEFAULT_DB,
+    REDIS_DEFAULT_HOST,
+    REDIS_DEFAULT_PASSWORD,
+    REDIS_DEFAULT_PORT,
+)
 from packit_service.sentry_integration import configure_sentry
 
 
+def get_redis_config():
```
mfocko, line 19: I hope we're not going to fix the rest of the Celery's backlog on our own…

packit_service/worker/tasks.py
```diff
@@ -994,3 +997,81 @@ def get_usage_statistics() -> None:
...
+            )
+
+            for key in keys:
+                keys_processed += 1
+
+                # Set TTL if key exists but has no expiry (TTL = -1)
+                if redis_client.ttl(key) == -1:
+                    redis_client.expire(key, REDIS_PIDBOX_TTL_SECONDS)
+                    keys_with_ttl_set += 1
+                    logger.debug(f"Set TTL on pidbox key: {key}")
```
mfocko: Do we really need this? Given that it might spam around 2k lines on every run?

majamassarini: No, we don't need this, I will remove the logging line. However we collected 2000 keys in several weeks, I hope we collect much less in a day 🙂

packit_service/worker/tasks.py
majamassarini: Claude tells me that this change is wrong because Redis SCAN requires cursor to start at 0 - not -1. So I will leave it as it was.

PR #3092, opened 2026-04-01
fix: handle projects with multi-level namespaces

xingzihai:
## Summary
This PR fixes an issue with handling projects that have multi-level namespaces.

## Problem
The API routes were using `<namespace>` which doesn't match multi-level namespaces like `organization/team` or `org/suborg/project`.

## Solution
Changed the route pattern from `<namespace>` to `<path:namespace>` which allows matching namespaces with slashes (multi-level namespaces).

## Changes
- Updated API routes in `projects.py` to use `<path:namespace>`
- Updated API routes in `usage.py` to use `<path:namespace>`

## Testing
- Tested with single-level namespaces (existing behavior preserved)
- Tested with multi-level namespaces (now works correctly)

## Example
Before: `/github/org/team/repo` would fail to match
After: `/github/org/team/repo` correctly matches with namespace=`org/team`

### DIFF
--- a/packit_service/service/api/projects.py
+++ b/packit_service/service/api/projects.py
@@ -50,7 +50,7 @@ class ProjectsList(Resource):
         return resp
 
 
-@ns.route("/<forge>/<namespace>/<repo_name>")
+@ns.route("/<forge>/<path:namespace>/<repo_name>")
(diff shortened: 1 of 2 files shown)


mfocko (approved):
LGTM

Issue #2914, opened 2025-12-15
Create a Fedora-CI plan repository


lachmanfrantisek:
The goal of this is to have one repository with tmt plans to be run via Packit as a dist-git CI.

* Plan decision/management is done outside of Packit codebase.
* If available, this can be stored on Fedora-managed Forgejo instance.
* Checks can be either imported or stored directly in the repo.
* For now, whole Packit team should have access, later we can move this to a Fedora-CI SIG (once/when it's revamped.)
* The main (and only) audience is Packit dist-git PR CI.

Question(s) to resolve:
* Should we have a branching schema similar to Fedora packages?

LecrisUT:
Initial proof-of-concept in: https://forge.fedoraproject.org/ci/shared-tests. Naming and documentation needed.

After #2987 we can try to actually run them in the `stg` deployment.

PR #3097, opened 2026-04-08
Simplify fedora-ci jobs

LecrisUT:
As part of the cleanup of the current fedora-ci jobs https://forge.fedoraproject.org/ci/tickets/issues/553, the interface for all of these jobs would converge to that from packit, simplifying how the jobs are handled here.

TODO:

- [x] Wait for installability to be compatible https://github.com/fedora-ci/installability-pipeline/pull/63

### DIFF
--- a/packit_service/worker/helpers/testing_farm.py
+++ b/packit_service/worker/helpers/testing_farm.py
@@ -10,8 +10,8 @@ from typing import Any, Callable, Optional, Union
 from ogr.abstract import GitProject, PullRequest
 from ogr.utils import RequestResponse
 from packit.config import JobConfig, PackageConfig
-from packit.exceptions import PackitConfigException, PackitException
-from packit.utils import commands, nested_get
+from packit.exceptions import PackitConfigException
+from packit.utils import nested_get
 from packit.utils.koji_helper import KojiHelper
 
 from packit_service.config import Deployment, ServiceConfig
@@ -1459,49 +1459,15 @@ class DownstreamTestingFarmJobHelper:
     @implements_fedora_ci_test("installability")
     def _payload_installability(self, distro: str, compose: str) -> dict:
         git_repo = "https://github.com/fedora-ci/installability-pipeline.git"
-        git_ref = (
-            commands.run_command(["git", "ls-remote", git_repo, "HEAD"], output=True)
-            .stdout.strip()
-            .split()[0]
-        )
-
-        if distro == "fedora-rawhide":
-            # profile names are in "fedora-N" format
-            # extract current rawhide version number from its candidate tag
-            if not (candidate_tag := self.koji_helper.get_candidate_tag("rawhide")):
-                raise PackitException(f"Failed to get test profile for {distro}")
-            profile = re.sub(r"f(\d+)(-.*)?", r"fedora-\1", candidate_tag)
-        else:
-            profile = distro
-
-        # installability test requires a Koji build
-        # (it should be guaranteed by the `HasEventSuccessfulScratchBuild` checker)
-        assert self.koji_build
-
-        return {
-            "test": {
-                "tmt": {
-                    "url": git_repo,
-                    "ref": git_ref,
-                },
-            },
-            "environments": [
-                {
-                    "arch": "x86_64",
-                    "os": {"compose": compose},
-                    "variables": {
-                        "PROFILE_NAME": profile,
-                        "TASK_ID": self.koji_build.task_id,
-                    },
-                },
-            ],
-            "notification": {
-                "webhook": {
-                    "url": f"{self.api_url}/testing-farm/results",
-                    "token": self.service_config.testing_farm_secret,
-                },
+        git_ref = "master"
+        payload = self._get_tf_base_payload(distro, compose)
+        payload["test"] = {
+            "tmt": {
+                "url": git_repo,
+                "ref": git_ref,
             },
         }
+        return payload
 
     @implements_fedora_ci_test("rpminspect")
     def _payload_rpminspect(self, distro: str, compose: str) -> dict:
--- a/packit_service/worker/helpers/testing_farm.py
+++ b/packit_service/worker/helpers/testing_farm.py
@@ -10,8 +10,8 @@ from typing import Any, Callable, Optional, Union
 from ogr.abstract import GitProject, PullRequest
 from ogr.utils import RequestResponse
 from packit.config import JobConfig, PackageConfig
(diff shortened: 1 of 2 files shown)


packit_service/worker/helpers/testing_farm.py
```diff
@@ -1459,49 +1459,15 @@ def prepare_and_send_tf_request(
...
-                        "TASK_ID": self.koji_build.task_id,
-                    },
-                },
-            ],
-            "notification": {
-                "webhook": {
-                    "url": f"{self.api_url}/testing-farm/results",
-                    "token": self.service_config.testing_farm_secret,
-                },
+        git_ref = "master"
```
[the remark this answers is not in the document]
LecrisUT: Existing issue and not worth the refactor now. Would be obsoleted when these can be merged in the same repo

LecrisUT:
https://artifacts.dev.testing-farm.io/53027160-0548-4339-bf40-2983a0c026d2/ should be an equivalent job that goes through the new api

packit_service/worker/helpers/testing_farm.py
```diff
@@ -1525,49 +1525,15 @@ def prepare_and_send_tf_request(
...
-                    "ref": git_ref,
-                },
-            },
-            "environments": [
-                {
-                    "arch": "x86_64",
-                    "os": {"compose": compose},
-                    "variables": {
-                        "PROFILE_NAME": profile,
-                        "TASK_ID": self.koji_build.task_id,
```
mfocko, line 1560: Is this OK that you're switching from `TASK_ID` to `KOJI_TASK_ID`?

mfocko: Also it will be added as an artifact… therefore installed during provisioning

packit_service/worker/helpers/testing_farm.py
```diff
@@ -1525,49 +1525,15 @@ def prepare_and_send_tf_request(
...
-                    "url": git_repo,
-                    "ref": git_ref,
-                },
-            },
-            "environments": [
-                {
-                    "arch": "x86_64",
-                    "os": {"compose": compose},
-                    "variables": {
-                        "PROFILE_NAME": profile,
```
mfocko, line 1559: This will be dropped from the request.

mfocko (approved):
LGTM, as long as it works. Which I can't really tell. There's also no justification for the changes.

Could you please link the relevant ticket (and possibly also changes) _in the commit_? Also please squash, I'd prefer to keep this atomic (changes + tests).

LecrisUT:
> Could you please link the relevant ticket (and possibly also changes) _in the commit_?

Not sure which ticket to include, there are 3 of them(https://forge.fedoraproject.org/ci/tickets/issues/553, https://github.com/fedora-ci/installability-pipeline/pull/63, https://github.com/fedora-ci/installability-pipeline/pull/64)

> LGTM, as long as it works. Which I can't really tell.

Previous comment has a testing-farm job running with these variables

> There's also no justification for the changes.

Basically all Fedora-CI jobs will expect and use the set of variables that packit provides. This is a prerequisite for https://github.com/packit/packit-service/issues/2914 and https://github.com/packit/packit-service/pull/2990

packit_service/worker/helpers/testing_farm.py
LecrisUT: > Is this OK that you're switching from `TASK_ID` to `KOJI_TASK_ID`?

Yes, the whole point it to align with that format which was one in https://github.com/fedora-ci/installability-pipeline/pull/63

> Also it will be added as an artifact… therefore installed during provisioning

Oh good point. Should remove it from the other tests also. Surprised it didn't cause more issues like in rpminspect. (Edit: maybe because this is the only one with an actual `compose`)

packit_service/worker/helpers/testing_farm.py
LecrisUT: Yes, intentional, it is being removed from the installability test

packit_service/worker/helpers/testing_farm.py
LecrisUT: > > Also it will be added as an artifact… therefore installed during provisioning
> 
> Oh good point. Should remove it from the other tests also. Surprised it didn't cause more issues like in rpminspect. (Edit: maybe because this is the only one with an actual `compose`)

Added b53868b to resolve this thing

mfocko:
> Not sure which ticket to include, there are 3 of them(https://forge.fedoraproject.org/ci/tickets/issues/553, [fedora-ci/installability-pipeline#63](https://github.com/fedora-ci/installability-pipeline/pull/63), [fedora-ci/installability-pipeline#64](https://github.com/fedora-ci/installability-pipeline/pull/64))

Well, ideally all of them then :) sorry about annoying you with that, but it's better to have it in history than be in the dark when it's needed

> Previous comment has a testing-farm job running with these variables

cool, thanks!

> Basically all Fedora-CI jobs will expect and use the set of variables that packit provides. This is a prerequisite for #2914 and #2990

👍

packit_service/worker/helpers/testing_farm.py
```diff
@@ -1637,7 +1639,9 @@ def _get_tf_base_payload(self, distro: str, compose: Optional[str]) -> dict:
         artifacts = []
         if self.koji_build:
             variables["KOJI_TASK_ID"] = self.koji_build.task_id
-            artifacts.append({"id": self.koji_build.task_id, "type": "fedora-koji-build"})
+            # TODO: Drop this check with the switch to multihost-pipeline
+            if with_artifacts:
+                artifacts.append({"id": self.koji_build.task_id, "type": "fedora-koji-build"})
```
mfocko: If you're introducing `with_artifacts`, I would've expected it to drop the artifacts altogether, so something like

```py
"artifacts": (with_artifacts and artifacts) or [],
```

below, rather than conditionally extend the empty list 🤔

LecrisUT: > If you're introducing with_artifacts, I would've expected it to drop the artifacts altogether

Sure. My hope is that this will be short-lived and the artifacts will always be added after #3127 anyway. As long as we make some room for adding some additional artifacts when that would be possible.

LecrisUT:
Anything else needed for this?

mfocko:
> Anything else needed for this?

No, LGTM… if you want to keep the commits as “verified”, could you please rebase yourself?

mfocko (approved):

PR #3098, opened 2026-04-09
Always honor `upstream_tag_include`/`upstream_tag_exclude`

nforro:
Tag filtering doesn't happen when processing multi-release events and when retriggering pull-from-upstream. Fix that.


majamassarini (approved):
LGTM! Thanks

PR #3102, opened 2026-04-13
Do not cancel test jobs with no associated build on build completion

### LINKED ISSUE #3101: Tests canceled and not restarted after a force push
In this [PR](https://github.com/teemtee/tmt/pull/4765/checks?check_run_id=70606013982) @LecrisUT noticed, that after a force push Packit canceled a test but did not start it again. This happens randomly.

On the same day we had another report for [PR](https://github.com/teemtee/tmt/pull/4765/checks?check_run_id=70606013982). 

Strange thing is that all canceled jobs are the `sanity` check, and the problem seems to hit the `tmt` team more often (probably because of the many identifiers used).

A workaround exists because another force push usually makes all the tests run again.

nforro:
Fixes https://github.com/packit/packit-service/issues/3101.

### DIFF
--- a/packit_service/models.py
+++ b/packit_service/models.py
@@ -3670,6 +3670,7 @@ class TFTTestRunGroupModel(ProjectAndEventsConnector, GroupModel, Base):
         event_id: int,
         ranch: str,
         created_before: Optional[datetime] = None,
+        has_copr_build: Optional[bool] = None,
     ) -> Iterable["TFTTestRunTargetModel"]:
         """Get list of currently running Testing Farm runs for a given project
         object (e.g. a PR or branch).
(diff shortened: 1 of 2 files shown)


betulependule (approved):
lgtm

majamassarini (commented):
Maybe I am missing something. But why cancel tests at all when `skip_build: true`?
I think we shouldn't, because they won't be restarted.

nforro:
> I think we shouldn't, because they won't be restarted.

`/packit test` doesn't retrigger test jobs with `skip_build: true`?

majamassarini:
> `/packit test` doesn't retrigger test jobs with `skip_build: true`?

yes, it does. But we don't cancel tests only after this comment, right? Other comments like `/packit build`
 could also cancel the test and it will not start again, or am I wrong?

nforro:
> Other comments like `/packit build` could also cancel the test and it will not start again, or am I wrong?

As far as I can tell, `/packit build` cancels running builds and starts new ones, it doesn't touch running test jobs.

nforro:
And we definitely want to cancel running no-build tests jobs after they are retriggered (manually or by a PR event).

PR #3106, opened 2026-04-17
Submit more files to Log Detective for analysis

jpodivin:
References to multiple artifacts are now used when asking Log Detective for analysis.
Since we can not be completely sure about their availability, all URLs are checked using get request.
If no artifacts pass the check, the trigger fails and the information is logged.

In order to avoid references to files that can't be possibly analyzed by Log Detective, only URLs pointing to files of the `text/plain` type are allowed, all others are dropped from the final request.
This check should see more use as we interact with other build backends, and submit artifacts other than just logs.

New tests for the `verify_artifact` were introduced and existing tests adjusted.

Merge after
- https://github.com/fedora-copr/logdetective-packit/pull/36
- https://github.com/fedora-copr/logdetective/pull/492

RELEASE NOTES BEGIN

Packit now sends build.log, root.log and mock_output.log from failed Koji builds to Log Detective.

RELEASE NOTES END

### DIFF
--- a/packit_service/utils.py
+++ b/packit_service/utils.py
@@ -432,6 +432,12 @@ def get_user_agent() -> str:
 def check_url(url: str) -> bool:
     """
     Verify that URL is valid.
+
+    Args:
+        url: URL to be validated
+
+    Returns:
+        True if the URL can be parsed, contains both scheme and netloc
     """
 
     try:
@@ -444,6 +450,13 @@ def check_url(url: str) -> bool:
 def check_content_type(content_type: Union[str, None], url: str) -> bool:
     """
     Verify that content_type is text/plain and log failures.
+
+    Args:
+        content_type: Value of Content-Type header field
+        url: URL of the checked artifact
+
+    Returns:
+        True if content type is text/plain
     """
     if not content_type:
         logger.warning("No content-type reported for artifact at %s", url)
@@ -461,6 +474,12 @@ def verify_artifact(url: str) -> bool:
     Verify that URL points to an text/plain artifact that can be downloaded.
 
     Checks Content-Type in headers using GET request.
+
+    Args:
+        url: URL of the artifact to be verified
+
+    Returns:
+        True if the artifact satisfies all requirements
     """
 
     if not check_url(url=url):
--- a/packit_service/utils.py
+++ b/packit_service/utils.py
@@ -12,7 +12,8 @@ from io import StringIO
 from logging import StreamHandler
 from pathlib import Path
 from re import search
-from typing import Optional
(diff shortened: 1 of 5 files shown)


packit_service/utils.py
```diff
@@ -428,6 +429,63 @@ def get_user_agent() -> str:
...
+            headers={"User-Agent": get_user_agent()},
+            timeout=(10, 30),
+            stream=True,
+            allow_redirects=True,
+        ) as response:
+            if response.status_code != 200:
+                logger.warning(
+                    "Request for artifact: %s failed with: %s", url, response.status_code
+                )
+                return False
```
nforro, line 500: Can't we use HEAD?
```suggestion
        with requests.head(
            url=url,
            headers={"User-Agent": get_user_agent()},
            timeout=(10, 30),
            allow_redirects=True,
        ) as response:
            if response.status_code != 200:
                logger.warning(
                    "Request for artifact: %s failed with: %s", url, response.status_code
                )
                return False
```

nforro (approved):

packit_service/utils.py
jpodivin: It's a valid idea. I have initially tried it. But it has a tradeoff, in that we to be sure that it is supported.
Some servers may block it for various reasons, including security.

If we were to use it, we would have to implement a fallback check using GET. Which would look pretty much like the function does now.

packit_service/worker/helpers/logdetective.py
```diff
@@ -84,14 +85,30 @@ def trigger_log_detective_analysis_for_arch(self, arch: str) -> bool:
...
+        # Use only artifacts with valid URLs
+        for artifact in possible_koji_artifacts:
+            url = koji.result.KojiEvent.get_koji_build_logs_url(
+                rpm_build_task_id=build_arch_task_id,
+                koji_logs_url=self.koji_logs_url,
+                log_file=artifact,
             )
-        }
+            if verify_artifact(url):
+                artifacts[artifact] = url
```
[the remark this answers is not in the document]
jpodivin: I guess we could use it, although I'm not sure if it wouldn't be too convoluted.

packit_service/utils.py
nforro: Ok, if `verify_artifact()` is meant to be general and not Koji-specific then using GET makes sense.

lbarcziova (approved):
LGTM, please fix the failing tests in CI and we can merge this after.

jpodivin:
> LGTM, please fix the failing tests in CI and we can merge this after.

Done

PR #3110, opened 2026-04-23
Pass variant to TF to differentiate ELN builds from rawhide branch

### LINKED ISSUE #3055: What should `dist-git-branch` be for eln builds open on rawhide PRs?
This is related to the future added in https://github.com/packit/packit-service/issues/2920. Currently it is just `rawhide`, but should it be masked as `eln` instead?

The case for making it `eln` is that it would avoid a second explicit check of `distro` to figure out that we actually have an `eln` scratch-build and not a rawhide one. For example in rpminspect the `dist-git-branch` is used because it gives a much easier connection to `fedora-distro-aliases` and bodhi. I want to refactor all of the others as well to use such a logic as needed.

The case for keeping it `rawhide`... I don't really know.

nforro:
Fixes https://github.com/packit/packit-service/issues/3055.

### DIFF
--- a/packit_service/worker/helpers/testing_farm.py
+++ b/packit_service/worker/helpers/testing_farm.py
@@ -1590,6 +1590,17 @@ class DownstreamTestingFarmJobHelper:
         if self.koji_build:
             variables["KOJI_TASK_ID"] = self.koji_build.task_id
             artifacts.append({"id": self.koji_build.task_id, "type": "fedora-koji-build"})
+
+        context = {
+            "distro": distro,
+            "arch": "x86_64",
+            "trigger": "commit",
+            "initiator": "fedora-ci",
+            "dist-git-branch": dist_git_branch,
+        }
+        if dist_git_branch == "rawhide" and distro == "fedora-eln":
+            context["variant"] = "eln"
+
         return {
             "environments": [
                 {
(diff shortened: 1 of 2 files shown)


packit_service/worker/helpers/testing_farm.py
```diff
@@ -1579,6 +1579,17 @@ def _get_tf_base_payload(self, distro: str, compose: Optional[str]) -> dict:
...
             artifacts.append({"id": self.koji_build.task_id, "type": "fedora-koji-build"})
+
+        context = {
+            "distro": distro,
+            "arch": "x86_64",
+            "trigger": "commit",
+            "initiator": "fedora-ci",
+            "dist-git-branch": dist_git_branch,
+        }
+        if dist_git_branch == "rawhide" and distro == "fedora-eln":
```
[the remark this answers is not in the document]
nforro: Nope, the constant is used in different context, it's also `rawhide` only by coincidence.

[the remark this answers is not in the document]
nforro: Nope.

majamassarini, line 1601: I am not sure if the following could happen somehow, in that case I think this could be helpful.

```suggestion
        if distro == "fedora-eln":
            if dist_git_branch != "rawhide":
                logger.warning(f"Unexpected: ELN build from branch {dist_git_branch}, expected rawhide")
```

majamassarini (approved):
LGTM, thanks!

packit_service/worker/helpers/testing_farm.py
nforro: Yes, it can happen, and it is expected. You can have ELN builds either from `rawhide` or from `eln` branch. Warning about that makes no sense.

nforro: That's the point of the new variable - to make it clear that the ELN build (`eln` Koji build target) did not originate from `eln` branch (in all other cases one can assume that target name matches branch name).

PR #3111, opened 2026-04-26
Update job statuses when submitting TF request consistently fails

### LINKED ISSUE #3109: Stuck rpmlint tests
### Description

Looking at https://src.fedoraproject.org/rpms/rubygem-cucumber-core/pull-request/6 , it seems that the rpmlint test https://dashboard.packit.dev/jobs/testing-farm/1508544 has not be finished for ~2 days.

Trying to find out what might be the issue, looking at https://dashboard.packit.dev/projects/src.fedoraproject.org/rpms/rubygem-cucumber-core for test runs, there two successful (ATM) entries which corresponds with installability / rpm inspect. But there are also dozens of other entries without any detail ...

### Importance

_No response_

### Workaround

- [ ] There is an existing workaround that can be used until this is implemented.

nforro:
Fixes https://github.com/packit/packit-service/issues/3109.


majamassarini (commented):
I am wondering if it isn't better to use `BASE_RETRY_INTERVAL_IN_MINUTES_FOR_OUTAGES` instead of `DEFAULT_RETRY_BACKOFF`. 
The `DEFAULT_RETRY_BACKOFF` is 7 seconds, so we will retry after, 7, 14 and 28 seconds. Is this enough? The `BASE_RETRY_INTERVAL_IN_MINUTES_FOR_OUTAGES` is 60 seconds, so it should give more time to the service to recover.

majamassarini (approved):

nforro:
> I am wondering if it isn't better to use `BASE_RETRY_INTERVAL_IN_MINUTES_FOR_OUTAGES` instead of `DEFAULT_RETRY_BACKOFF`. The `DEFAULT_RETRY_BACKOFF` is 7 seconds, so we will retry after, 7, 14 and 28 seconds. Is this enough? The `BASE_RETRY_INTERVAL_IN_MINUTES_FOR_OUTAGES` is 60 seconds, so it should give more time to the service to recover.

The constant is used here to calculate roughly after what time will the job be retried, so it can be reported to users, we are not actually retrying anything here. Default Celery retry mechanism applies, if we want to change it, we have to do it in a different place and I believe that's out of scope of this PR.

nforro:
> we have to do it in a different place

Actually, I suppose we can do it here :slightly_smiling_face: I think it makes sense, let me do it.

majamassarini (approved):
Thanks!

PR #3112, opened 2026-04-27
Only cancel those running test jobs that will be retriggered

### LINKED ISSUE #3104: rebuild-failed and retest-failed cancel all running jobs
They shouldn't cancel anything - they should only restart failed jobs and ignore the rest.

nforro:
Consider test target and/or identifier when filtering test jobs to cancel, depending on the trigger.

Fixes: https://github.com/packit/packit-service/issues/3104


majamassarini (approved):
LGTM thanks!

PR #3113, opened 2026-04-27
Improve content of help message

betulependule:
The previous version of help messages referred to the `--package` parameter, which is not supported in Fedora CI and should not be included in the corresponding help message. Also, the notes in the epilog of the message weren't worded the best way. This is now fixed.

### DIFF
--- a/packit_service/constants.py
+++ b/packit_service/constants.py
@@ -34,12 +34,19 @@ HELP_COMMENT_EPILOG = (
     "**Documentation**: \n - {docs}"
 )
 HELP_NOTE = (
-    "If using Fedora CI, refer to `/packit-ci help` instead. "
-    "Refer to `/packit-ci-stg help` if using the staging instance."
+    "`/packit` commands interact with the opt-in packit jobs "
+    "(e.g. `propose-downstream`, `pull-from-upstream`). "
+    "If you are looking for help on Fedora CI jobs "
+    "(e.g. `scratch-build`, `test rpminspect`, etc.), "
+    "refer to `/packit-ci` or `/packit-ci-stg`."
 )
 HELP_NOTE_FEDORA_CI = (
-    "For default Packit behaviour, refer to `/packit help` instead. "
-    "Refer to `/packit-stg help` if using the staging instance."
+    "`/packit-ci` commands interact with Fedora CI jobs "
+    "(e.g. `scratch-build`, `test rpminspect`, etc.). "
+    "If you are looking for help on the opt-in packit jobs "
+    "(e.g. `propose-downstream`, `pull-from-upstream`), "
+    "refer to `/packit` or `/packit-stg`. "
+    "These assume a `.packit.yaml` configuration file is present in the repo."
 )
 
 KOJI_PRODUCTION_BUILDS_ISSUE = "https://pagure.io/releng/issue/9801"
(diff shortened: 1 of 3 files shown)


packit_service/worker/jobs.py
```diff
@@ -273,9 +273,8 @@ def parse_comment(
         try:
             args = parser.parse_args(commands)
             check_target = getattr(args, "check_target", None)
-            return ParsedComment(
-                command=args.command, package=args.package, check_target=check_target
-            )
+            package = getattr(args, "package", None)
```
mfocko, line 276: Have we actually hit a `TypeError` because of a `None` here?

mfocko (approved):
LGTM 🚀

packit_service/worker/jobs.py
mfocko: There should be a default specified, right?

betulependule: The package parameter is no longer specified for the Fedora CI comment parses, and I hit this error after removing it:

`E           AttributeError: 'Namespace' object has no attribute 'package'`

general comment parser: `args["package"]` should be None by default
Fedora CI parser: throws an AttributeError when trying to access it

That's why I added `package = getattr(args, "package", None)`. To prevent AttributeError.

packit_service/constants.py
```diff
@@ -34,12 +34,19 @@
     "**Documentation**: \n - {docs}"
 )
 HELP_NOTE = (
-    "If using Fedora CI, refer to `/packit-ci help` instead. "
-    "Refer to `/packit-ci-stg help` if using the staging instance."
+    "`/packit` commands interact with the opt-in packit jobs "
+    "(e.g. `propose-downstream`, `pull-from-upstream`). "
+    "If you are looking for help on Fedora CI jobs "
+    "(e.g. `scratch-build`, `test rpminspect`, etc.), "
+    "refer to `/packit-ci` or `/packit-ci-stg`."
```
nforro, line 41: Do these show up only in dist-git?

packit_service/constants.py
```diff
@@ -34,12 +34,19 @@
...
 )
 HELP_NOTE_FEDORA_CI = (
-    "For default Packit behaviour, refer to `/packit help` instead. "
-    "Refer to `/packit-stg help` if using the staging instance."
+    "`/packit-ci` commands interact with Fedora CI jobs "
+    "(e.g. `scratch-build`, `test rpminspect`, etc.). "
+    "If you are looking for help on the opt-in packit jobs "
+    "(e.g. `propose-downstream`, `pull-from-upstream`), "
+    "refer to `/packit` or `/packit-stg`. "
+    "These assume a `.packit.yaml` configuration file is present in the repo."
```
nforro, line 49: I would use either something like "Packit configuration file" or list all the possible filenames, I believe there are 4 (matching this regex: `\.?packit\.ya?ml`).

packit_service/constants.py
betulependule: This epilog is part of the comment triggered by `/packit help` (and `packit-stg help`). They show up anywhere where the `/packit help` command is called.

packit_service/constants.py
betulependule: Ok, will do a follow-up PR

packit_service/constants.py
nforro: Then I think this formulation would be rather confusing in upstream repos. There are a lot of projects that use Packit only as a CI and don't care/know about downstream jobs, so the examples mean nothing to them, Fedora CI is of course irrelevant upstream, and the opt-in also feels weird, they have already opted in to onboard Packit.

betulependule: So I suppose the best solution will be to only add this in downstream repos.

PR #3115, opened 2026-04-28
fix: drop guppy3 — segfaults on Python 3.14

TomasTomecek:
## Summary

- `guppy`'s `heapyc` C extension crashes with SIGSEGV under Python 3.14 due to incompatible weak reference internals (`PyWeakref_NewRef` in the C stack trace)
- Removes the `hp.heap()` profiling block from `run_n_clean` entirely — `log_memory_stats` already tracks RSS via `resource.getrusage` which is sufficient
- Removes the `guppy3` pip install from `install-deps-worker.yaml`
- Cleans up stale comments in `test_logdetective.py` that referenced the old workaround

Fixes a CI segfault in `tests/integration/test_dg_pr.py::test_downstream_koji_scratch_build[rawhide target branch]`.

## Test plan

- [x] Existing integration tests pass without the segfault

🤖 Generated with [Claude Code](https://claude.com/claude-code)

RELEASE NOTES BEGIN

Empty on purpose

RELEASE NOTES END


tests_openshift/service/test_logdetective.py
```diff
@@ -123,8 +123,6 @@ def test_logdetective_process_message(
     service_config.fedora_ci.enabled_projects = {logdetective_analysis_success_event["project_url"]}
     service_config.logdetective_enabled = True
 
-    # Set deployment to prod to disable guppy memory profiling logic
-    # which causes UnboundLocalError when guppy is missing.
     service_config.deployment = Deployment.prod
```
nforro: Is this still needed then?

nforro (approved):

majamassarini (approved):
Thanks a lot!

tests_openshift/service/test_logdetective.py
TomasTomecek: yeah, I was wondering too; let me try to remove it

PR #3116, opened 2026-04-29
Fix circular import preventing allowlist.py script execution

majamassarini:
The allowlist.py script failed with ImportError due to circular dependency.

Created packit_service/worker/helpers/testing_farm_mixin.py to separate Testing Farm-specific mixins from general handler mixins.

Extracted get_check_name_prefix() from FedoraCIJobHandler to packit_service/utils.py as a standalone utility function.

RELEASE NOTES BEGIN

Fix circular import that prevented the allowlist script from running.

RELEASE NOTES END

### DIFF
--- a/packit_service/utils.py
+++ b/packit_service/utils.py
@@ -1,6 +1,8 @@
 # Copyright Contributors to the Packit project.
 # SPDX-License-Identifier: MIT
 
+from __future__ import annotations
+
 import argparse
 import itertools
 import logging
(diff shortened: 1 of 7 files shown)


packit_service/utils.py
```diff
@@ -607,3 +613,8 @@ def break_lines_in_text(text: str, sep: str, max_line_length: int):
         split_text.append(line)
 
     return "\n".join(split_text)
+
+
+def get_check_name_prefix(service_config: ServiceConfig) -> str:
+    """Return the prefix for check names based on deployment environment."""
+    return "Packit-stg" if service_config.deployment == Deployment.stg else "Packit"
```
mfocko, line 620: I'm not sure if this is a safe assumption about the prefix, but OK

mfocko (approved):

PR #3121, opened 2026-05-06
Avoid race condition when reporting job statuses after retriggering

nforro:
Retriggering starts a new job run that shares the same status check with the one that is about to be canceled. Cancel request is sent to Copr/Koji/TF, status check is updated by the new job run, but then a build end event or TF webhook arrives and the status check is updated with a wrong status (canceled in this case, but this could happen on e.g. a build failure as well).
Prevent this by checking if there is a newer run of the same job before reporting status in the handlers.


majamassarini (approved):
Thanks!

PR #3127, opened 2026-05-14
Switch to using testing-farm multihost pipeline

LecrisUT:
TODO:

- [ ] Make sure relevant plans are compatible

### DIFF
--- a/packit_service/worker/helpers/testing_farm.py
+++ b/packit_service/worker/helpers/testing_farm.py
@@ -1652,7 +1652,12 @@ class DownstreamTestingFarmJobHelper:
         }
         return payload
 
-    def _get_tf_base_payload(self, distro: str, compose: Optional[str]) -> dict:
+    def _get_tf_base_payload(
+        self,
+        distro: str,
+        compose: Optional[str],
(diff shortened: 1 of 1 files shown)


mfocko:
Why do you want to turn it on by default? 👀

LecrisUT:
> Why do you want to turn it on by default? 👀

Because if `compose` is not set, testing-farm will force it to be `provision/container`, which does not work for `fedora-review`. For all other cases with compose not set, mostly because there should not be any difference between these two. When `compose` is set things are tricky because these differ in the artifact installation which I would like to migrate these to using, but the implementation is just arriving today so still need to try it out.

It is a draft until can confirm that it is stable for those simple tests, will post an example run when I have it

LecrisUT:
Yep, I've identified 2 bugs with multihost pipeline affecting these, so definitely not ready yet:
- imported jobs automatically translate to `provision/artemis` even though it has `provision/container`. Will affect `rpminspect` migration to shared-tests
- similarly `provision/virtual` is not translated to `provision/artemis`, affects `fedora-review`

PR #3131, opened 2026-05-20
Pin GitHub Actions to commit hashes for security

majamassarini:
Pin all actions to specific commit SHAs to prevent supply chain attacks and ensure reproducible builds.

Note: ossf-scorecard.yml already had actions pinned.

Assisted-By: Claude Sonnet 4.5 <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/check-release-notes.yml
+++ b/.github/workflows/check-release-notes.yml
@@ -10,6 +10,6 @@ on:
 jobs:
   check_release_notes:
     name: Notes are either written, or there are none
-    uses: packit/.github/.github/workflows/check-release-notes.yml@main
+    uses: packit/.github/.github/workflows/check-release-notes.yml@2837c96caf71966609451ad0323552ef4be11a23
     with:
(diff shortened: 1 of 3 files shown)


.github/workflows/check-release-notes.yml
```diff
@@ -10,6 +10,6 @@ on:
 jobs:
   check_release_notes:
     name: Notes are either written, or there are none
-    uses: packit/.github/.github/workflows/check-release-notes.yml@main
+    uses: packit/.github/.github/workflows/check-release-notes.yml@2837c96caf71966609451ad0323552ef4be11a23 # main
```
nforro: Assuming `main` is a branch and not a tag, I don't think it makes sense to reference it here, it can change at any time.

nforro (approved):

.github/workflows/check-release-notes.yml
nforro: I didn't notice this in the other PR(s), but the same applies there as well of course.

PR #3133, opened 2026-05-26
Add tests for handling sources in vcs

majamassarini:
Add a unit test verifying that a propose_downstream JobConfig built with the documented workaround (files_to_sync + a post-modifications git-add action) is accepted by ProposeDownstreamJobHelper and that the action survives into job_config.actions.

Assisted-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/tests/unit/test_propose_downstream.py
+++ b/tests/unit/test_propose_downstream.py
@@ -3,8 +3,10 @@
 
 import pytest
 from flexmock import flexmock
+from packit.actions import ActionName
 from packit.config import CommonPackageConfig, JobConfig, JobType, PackageConfig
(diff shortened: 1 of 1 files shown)


nforro (changes requested):
I don't really see a point in this test. I would either drop it entirely or change it to test something meaningful.

majamassarini:
> I don't really see a point in this test. I would either drop it entirely or change it to test something meaningful.

It does only test that the action defined by the user and the listed files are passed to the packit layer, where the behavior is tested. But I am ok with dropping it, I will close this PR.

PR #3147, opened 2026-07-01
Fix multi-release retries

nforro:
Related to https://github.com/packit/specfile/pull/542.

RELEASE NOTES BEGIN

When a package is configured to consume mutiple release streams and release-monitoring.org detects multiple releases at the same time, if a pull-from-upstream job run for one released version fails, only that one is retried and the rest are unaffected and properly processed. Before this fix it could happen that a failed job run for one released version could prevent other released versions from being processed at all.

RELEASE NOTES END


majamassarini (approved):
LGTM thanks!

PR #3148, opened 2026-07-02
Install gnome-srpm-macros on workers

nforro:
Fixes https://github.com/packit/packit/issues/2732.
Related to https://github.com/packit/sandcastle/pull/221.

### DIFF
--- a/files/install-deps-worker.yaml
+++ b/files/install-deps-worker.yaml
@@ -23,6 +23,7 @@
           - python3-sentry-sdk
           - python3-syslog-rfc5424-formatter # logging to Splunk
           - dnf-utils
+          - gnome-srpm-macros
           - make
           # for pip-installing sandcastle from git repo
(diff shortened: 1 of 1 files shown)


lbarcziova (approved):

mcrha:
There is something fishy about this, because the package should be brought in with `redhat-rpm-config` package - you probably do not use new enough version of the `redhat-rpm-config` in the workers, maybe? It has it for few months, for f43 (since 343-13), f44 (since 344-2), f45/rawhide (since 344-2), all since end of April or beginning of May, according to respective package ChangeLog entries shown in koji.

nforro:
@mcrha This has been reverted: https://github.com/packit/packit-service/pull/3150

mcrha:
ok :) I received a notification about this change only today, through #2732, which led me here. I'm sorry for the noise.

PR #3151, opened 2026-07-08
Expose error messages from Log Detective to API

Jany26:
Exposes optional `error_msg` field present in messages from fedora-messaging which contain why the Log Detective analysis failed. `error_msg` is also introduced to the DB model.

Based on `status` field, one of `log_detective_response` and `error_msg` will be set and the other will be null.

Merge this before the Log Detective related changes in Dashboard.

Examples of JSON from fedmsg: [failed LD run](https://apps.fedoraproject.org/datagrepper/v2/id?id=e902470c-f4cf-46e9-93b2-405177c64c1d&is_raw=true&size=extra-large), [successful LD run](https://apps.fedoraproject.org/datagrepper/v2/id?id=77ab85c9-8512-4691-b0c9-be6520946b5e&is_raw=true&size=extra-large)

RELEASE NOTES BEGIN

`LogDetectiveResult` now contains `error_msg` in case of error (for better dashboard observability).

RELEASE NOTES END


tests/unit/events/test_logdetective.py
```diff
@@ -247,6 +252,62 @@ def test_logdetective_run_success(
...
+        ),
+        (
+            None,
+            "Server error '500 Internal Server Error'",
+            LogDetectiveResult.error,
+            {"error_msg": "Server error '500 Internal Server Error'"},
+        ),
+    ],
+    ids=["stores_response", "stores_error_msg"],
+)
```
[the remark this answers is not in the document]
Jany26: overkill

packit_service/worker/handlers/logdetective.py
```diff
@@ -151,6 +152,10 @@ def _run(self) -> TaskResults:
             log_detective_run_model.set_log_detective_response(
                 self.log_detective_response, self.status
             )
+        elif self.error_msg:
+            log_detective_run_model.set_log_detective_response(
+                {"error_msg": self.error_msg}, self.status
+            )
```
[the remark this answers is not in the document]
Jany26: should not happen, but okay

packit_service/events/logdetective.py
```diff
@@ -40,6 +40,7 @@ def __init__(
         commit_sha: str,
         pr_id: Optional[int] = None,
         identifier: Optional[str] = None,
+        error_msg: Optional[str] = None,
```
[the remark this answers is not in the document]
Jany26: ok, but I think there are more optional fields than that, but response is probably the most notable one

packit_service/events/logdetective.py
```diff
@@ -31,7 +31,7 @@ def event_type(cls) -> str:
     def __init__(
         self,
         target_build: str,
-        log_detective_response: dict,
+        log_detective_response: Optional[dict],
```
jpodivin: This will also need a `None` default.

packit_service/events/logdetective.py
```diff
@@ -31,7 +31,7 @@ def event_type(cls) -> str:
     def __init__(
```
jpodivin, line 31: Is there is a situation when both `log_detective_response`v and ` error_msg` can be `None`? Or a situation when both are not `None`? Either way, the expected usage should be documented in a docstring, because the signature is getting a bit too ambiguous.

packit_service/worker/handlers/logdetective.py
```diff
@@ -151,6 +152,10 @@ def _run(self) -> TaskResults:
             log_detective_run_model.set_log_detective_response(
                 self.log_detective_response, self.status
             )
+        elif self.error_msg or self.status == LogDetectiveResult.error:
```
jpodivin: Can there be truthy `self.error_msg` without  `self.status == LogDetectiveResult.error`?

jpodivin (changes requested):

packit_service/worker/handlers/logdetective.py
Jany26: I don't think so, it's just something gemini flagged, so I guess better be safe than sorry.

packit_service/events/logdetective.py
Jany26: documenting this is a good point

packit_service/events/logdetective.py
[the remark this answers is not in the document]
Jany26: unifying the function typehints with fedmsg schema and logdetective-packit server is a matter for a separate PR though

Jany26:
Based on discussion with Jirka, we'll try to match the structure of `event_dict` and the final API dict as close as possible to the fedmsg structure (so separate `error_msg` and `log_detective_response` fields).

jpodivin (approved):
I like this. The sooner we get this in the better.

packit_service/events/logdetective.py
```diff
@@ -31,16 +31,24 @@ def event_type(cls) -> str:
...
         self,
         target_build: str,
-        log_detective_response: dict,
         status: LogDetectiveResult,
         build_system: LogDetectiveBuildSystem,
         log_detective_analysis_start: datetime,
         log_detective_analysis_id: str,
         project_url: str,
         commit_sha: str,
+        log_detective_response: Optional[dict] = None,
```
mfocko, line 40: Since you switched this to `Optional`, do the consumers expect the `None`?

tests/unit/events/test_logdetective.py
```diff
@@ -202,13 +211,26 @@ def test_logdetective_run_success(
...
-        ),
-    ).once()
+    mock_timestamp = datetime(2024, 1, 1, 12, 0, tzinfo=timezone.utc).replace(tzinfo=None)
+
+    if status_str == "complete":
+        run_model.should_receive("set_log_detective_response").with_args(
+            handler.log_detective_response,
+            status=handler.status,
+            log_detective_analysis_start=mock_timestamp,
+        ).times(int(status_str == "complete"))
```
mfocko:
```suggestion
        ).once()
```

packit_service/events/logdetective.py
Jany26: i'm not sure i follow, but result messages from fedmsg can either contain the error_msg field only or the log_detective_response field only

tests/unit/events/test_logdetective.py
Jany26: :+1:

packit_service/events/logdetective.py
mfocko: ~~Originally it was just a `dict`… right now it can be `None`, is this reflected where the event is used?~~

checked the diff again, it is…

mfocko (approved):

Issue #2887, opened 2025-11-27
Make Packit easier to contribute


lachmanfrantisek:
People have a tough time to contribute to Packit, mainly the service part. Let's collect tasks to improve the situation. (Anyone's suggestions are welcome.)

nforro:
> 1. I see that the contributing guidelines contain some passages (like in [Running packit-service locally](https://github.com/packit/packit-service?tab=contributing-ov-file#running-packit-service-locally)) which require you to have access to secrets (which only the Packit team should have access to). I thought anyone outside the team should be able to contribute, yet the guidelines don't seem to reflect that.

That's not true, the instructions state:
> You also need to populate `secrets/packit/dev/` manually, for instructions see [deployment repo](https://github.com/packit/deployment/tree/main/secrets).

Where you can read:
> Either:
> * generate them with `make generate-local-secrets` or
> * if you're a Packit team member with access to Bitwarden vault, download `stg/` secrets (`DEPLOYMENT=stg make download-secrets`) and copy into `dev/`

How usable the generated secrets are, that's another question. And perhaps something to improve.

---

> 2\. How often do we even use [Building images locally](https://github.com/packit/packit-service?tab=contributing-ov-file#building-images-locally) and [Running packit-service locally](https://github.com/packit/packit-service?tab=contributing-ov-file#running-packit-service-locally)?

All the time? 😅 How else would you want to do non-trivial changes to the service?

---

> 3\. Outside of the guidelines themselves, what I most struggled with personally is that the code of packit-service is quite complex with a lot of technologies being used (Redis queue, Celery, Sentry etc.) and there is no proper documentation regarding the code besides docstrings (as far as I know)

That's a good point, and I remember struggling with this as well, notably Celery and how the task distribution/processing works. That could be documented/explained better.

> 3\. What I've personally had the most trouble with was following the data flow (for example how the data is processed when a Github event happens, how SteveJobs processes it and how it, along with all necessary metadata, eventually gets to an appropriate handler).

Also a good point, though I feel the source code is quite self-explanatory and doesn't take that long to grasp. That doesn't mean more documentation or some high level overview wouldn't help.

PR #3167, opened 2026-08-20
Deploying packit-service via quadlets

LecrisUT:
The goal of such a setup is to have a more composable deployment setup than the docker-compose setup. This makes use of the systemd drop-in configuration files to be able to select the components that you want, e.g. `dev` being a setup that would use the current sources and pick up the live edits. In principle the user would simply run
```console
$ systemctl --user daemon-reload
$ systemctl --user start packit-service
```
(after setting up systemd to pick up these quadlets file, see `quadlet/README.md` for more info)

TODO:

- [x] Document basic quadlet setup
- [ ] Define all necessary quadlet files


LecrisUT:
CC @FrostyX you would probably be interested to try this out (once we figure out the actual components). Could also be interesting for copr to check

FrostyX:
Hello @LecrisUT, thank you for trying to address the issue.
I haven't used quadlets for development yet, so I don't know how comfortable it is in comparison with `docker-compose` but if I remember correctly, my biggest problem when trying to run Packit locally was setting up credentials. There was too many of them and IIRC it wasn't well documented where should I get them from.

For local deployment (not any public instance), my wish would be:

- Generate as many credentials on the fly as you can, or commit preconfigured credentials in the git repo. For example our Copr development stack has many `1234` passwords in git
- Simplify the infrastructure. For example, in Copr the signing of packages is configurable, so in the past our development stack had completely disabled signing, and therefore we didn't have to run one whole service and manage its credentials
- Complain about missing credentials only when I actually need them. Let's say my local Packit instance tries to submit a Copr build but I didn't configure my token yet

With these, I should be able to do `docker-compose up` (or its quadlet alternative) and have the whole stack running. Requiring literally zero configuration on my side. And whenever I get a failure because of missing credentials, I will know exactly what they are, because they will be related to a specific feature that I am working on.

One additional thing that makes our life very easy when developing Copr is having scripts that do a "oneshot" version of our daemons. For example, we have an action dispatcher and build dispatcher which are services running on the background and they are spawning a new process for every action or build and then processing it. But we also have scripts that take an action ID or a build ID as their input and they process just this one single action or build. Therefore you can easily work on the build pipeiline and you don't even need any servers running.

I have no idea how many of these points are applicable to Packit but I can't see a reason why they shouldn't be.

LecrisUT:
> I haven't used quadlets for development yet, so I don't know how comfortable it is in comparison with `docker-compose`

From playing around with it, I find it useful if you want to make adjustments to it, e.g. I can more easily thinker with a reverse-proxy setup without breaking the whole compose file, or add/remove secrets that I do not use.

> but if I remember correctly, my biggest problem when trying to run Packit locally was setting up credentials. There was too many of them and IIRC it wasn't well documented where should I get them from.

Indeed that's what I'm chipping away at next. I am trying to remove each secret requirements and see if it can still work or patch it so that it can. So far I've removed just the SSL requirement and at least it spun up, but I didn't give it any payloads to see if it breaks somewhere further down. I want to do that part next, but would need to figure out how to mock some payloads or mock some other interaction with it.

> One additional thing that makes our life very easy when developing Copr is having scripts that do a "oneshot" version of our daemons. For example, we have an action dispatcher and build dispatcher which are services running on the background and they are spawning a new process for every action or build and then processing it. But we also have scripts that take an action ID or a build ID as their input and they process just this one single action or build. Therefore you can easily work on the build pipeiline and you don't even need any servers running.

Hmm, that sounds interesting. From what I can read in the deployment, the only roadblock for this is that the packit service is split into the service (web api) and the workers. Otherwise if we could interface directly with `celery` on the worker, then we could have a similar thing. But I have not worked with celery before so I am not sure how feasible that is.

betulependule:
Hi @LecrisUT, thank you for working on this. My understanding of quadlets is limited, but it looks good so far (at least on surface level). I'll try to follow your progress on this PR to get familiar with new changes as you work on it.

I agree with @FrostyX that setting up credentials is probably the biggest pain point when trying to get a deployment of Packit running locally. Personally, it took me hours to get it running and then some more additional hours to be able to send fedmsg payloads to it from the terminal (thanks to @nforro for helping me get this working a couple of months back :pray:).

We have an epic relevant to this, but unfortunately, it hasn't been a priority to us lately: https://github.com/packit/packit-service/issues/2887

quadlets/base/packit-service.container
```diff
@@ -0,0 +1,4 @@
+[Container]
+ContainerName=packit-service
+Image=quay.io/packit/service:prod
+User=1024
```
betulependule, line 4: I do wonder whether this is going to work as-is. Whenever I need to run a local deployment (using the docker-compose setup), I need to open `docker-compose.yaml` and make sure all instances of `user: "1024"` are changed from 1024 to the number output by `id -u`.

The deployment doesn't start properly without it (at least for me).

PR #3169, opened 2026-08-24
Properly parse Koji side-tag build events

nforro:
Fixes https://red-hat-it.sentry.io/issues/7670177728.

### DIFF
--- a/packit_service/worker/parser.py
+++ b/packit_service/worker/parser.py
@@ -1516,10 +1516,35 @@ class Parser:
         #       "rawhide",
         #       {}
         #     ],
+        #
+        # or:
+        #
+        # "request": [
+        #     "git+https://src.fedoraproject.org/rpms/forgejo.git#ced1aa24b245770d46e72e14d18b323aba3dbf3f",
+        #     {
+        #         "_tagging_tasks_index": {},
+        #         "base_tag": "eln-build",
+        #         "initial_build_ids": [
+        #             3085529,
+        #             3085527
+        #         ],
+        #         "name": "eln-build-side-148578"
+        #     },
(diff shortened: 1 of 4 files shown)


nforro:
> * **[null handling]** `packit_service/worker/parser.py:1545` — When `fedora_target` is a dict (side-tag build), `fedora_target.get("base_tag")` will return `None` if the `base_tag` key is absent. The subsequent `fedora_target.removesuffix("-candidate")` would raise `AttributeError`. While known side-tag payloads always include `base_tag`, adding a default value would guard against unexpected dict formats.
>   Remediation: Use `fedora_target = fedora_target.get("base_tag", "")` or add a guard that logs a warning and returns `None` if `base_tag` is missing.

I think it's better if it appears in Sentry, should it ever happen (which it shouldn't :slightly_smiling_face:).

> * **[test structure consistency]** `tests/unit/events/test_koji.py:314` — The new `test_parse_koji_build_event_completed_side_tag` omits the `packages_config` assertion block present in sibling tests (`_rawhide`, `_f36`, `_epel8`). Consider adding it for consistency, though omitting it is a reasonable scoping choice for this bug fix.

That's on purpose, I think that assertion is rather questionable in the other test(s) too.

> * **[scope-adjacent-fix]** `packit_service/worker/parser.py:1561` — The typo fix (`namespace="rmps"` → `namespace="rpms"`) is adjacent to the side-tag parsing fix. The correction is valid and appropriate to include alongside the primary bug fix.

This was a good catch, I just hope it won't have any consequences :sweat_smile:

betulependule (approved):
LGTM, thanks 🙏

PR #3172, opened 2026-08-26
Support identifiers when selecting OpenScanHub base builds

rgetz:
Select a commit-triggered Copr build with the same configured identifier when determining the base build for an OpenScanHub differential scan.

Retain the existing fallback to the first matching base build when no identifier match is found.

This is important when multiple base builds are configured for the same branch but use different configurations. Without identifier-aware selection, an OpenScanHub differential scan can compare a pull request build against the wrong base build, resulting in an inaccurate comparison between different build configurations.

Supporting identifiers allows each differential scan job to use the corresponding base build and keeps the comparison between equivalent configurations.

Related: packit/packit#2763

TODO:

### DIFF
--- a/packit_service/worker/helpers/open_scan_hub.py
+++ b/packit_service/worker/helpers/open_scan_hub.py
@@ -208,10 +208,13 @@ class CoprOpenScanHubHelper(OpenScanHubHelper):
         """
         Find the job in the config that can provide the base build for the scan
         (with `commit` trigger and same branch configured as the target PR branch).
+        Prefer a base build with the same identifier as the configured job,
+        falling back to the first matching base build.
         """
         base_build_job = None
         target_branch = self.copr_build_helper.pull_request_object.target_branch
         default_branch = self.copr_build_helper.project.default_branch
+        configured_identifier = self.copr_build_helper.job_config.identifier
 
         for job in self.copr_build_helper.package_config.get_job_views():
             if (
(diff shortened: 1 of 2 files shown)


packit_service/worker/helpers/open_scan_hub.py
```diff
@@ -222,8 +225,11 @@ def find_base_build_job(self) -> Optional[JobConfig]:
                     or (not job.branch and default_branch == target_branch)
                 )
             ):
-                base_build_job = job
-                break
+                if configured_identifier and job.identifier == configured_identifier:
+                    return job
+
+                if base_build_job is None:
+                    base_build_job = job
```
nforro, line 232: This makes it fall back to the last matching base build, not the first one, or am I missing something?

betulependule: I think the `is None` check makes it so that the first matching job is assigned to `base_build_job`, and it's never overwritten by any other matching job.

betulependule (approved):
Looks good to me. Thank you for the contribution. 🙏 I'll go ahead and create a PR to update the docs today.

packit_service/worker/helpers/open_scan_hub.py
nforro: Oh, right, of course, sorry :slightly_smiling_face:

Issue #3173, opened 2026-08-26
Fix `MRTarget` equality comparison


rgetz:
### What happened? What is the problem?

Not when using things, but when I was looking at the code base, I noticed the `MRTarget.__eq__()` implementation in [`packit_service/config.py`](https://github.com/packit/packit-service/blob/main/packit_service/config.py#L78-L93) appears to contain a typo (copy/paste issue?) that causes the branch comparison to always evaluate as true.

The current implementation is:

```python
return self.repo == other.repo and self.branch == self.branch
```

The second comparison should presumably compare against `other.branch` (or be removed since it's always `true`).

Either tests of:
```python
def test_mr_target_equality():
    target = MRTarget("packit-service/src/.+", "rawhide")

    assert target == MRTarget("packit-service/src/.+", "rawhide")
    assert target != MRTarget("packit-service/src/.+", "c9s")
    assert target != MRTarget("other/src/.+", "rawhide")
```

or by augmenting the [existing tests](https://github.com/packit/packit-service/blob/main/tests/unit/test_config.py) to add something like:
```python
assert MRTarget("packit-service/src/.+", "not-c9s") not in config.gitlab_mr_targets_handled
```
I think both would fail with the current broken implementation and pass after the fix.

### Steps to reproduce

```bash

```

### What is the impacted category (job)?

General

### Workaround

- [ ] There is an existing workaround that can be used until this issue is fixed.

### Participation

- [ ] I am willing to submit a pull request for this issue. (Packit team is happy to help!)

betulependule:
Fixed in https://github.com/packit/packit-service/pull/3175

PR #3175, opened 2026-08-26
Fix MRTarget equality to compare branch values

rgetz:
As described in #3173

`MRTarget.__eq__()` compared `self.branch` with itself instead of comparing it with `other.branch`, causing targets with the same repository but different branches to compare as equal.

Correct the comparison and add a regression test verifying that `gitlab_mr_targets_handled` membership respects the configured branch.

This ensures GitLab MR target matching does not incorrectly treat different branch patterns as the same target.

The `MRTarget.__eq__()` issue being corrected here has been present since the original implementation was introduced in [2021](https://github.com/packit/packit-service/commit/4d72a8101fe2082fa54f5e44c386aee325fbb5d9). It was noticed during a review of the code rather than as the result of a reported bug or observed production failure.

This correction aligns `MRTarget` equality with the behavior described by the original commit message: both the repository and branch are part of the target definition.

No specific production issue or crash has been attributed to this issue.

### DIFF
--- a/packit_service/config.py
+++ b/packit_service/config.py
@@ -90,7 +90,7 @@ class MRTarget(NamedTuple):
         if not isinstance(other, MRTarget):
             raise NotImplementedError()
 
-        return self.repo == other.repo and self.branch == self.branch
+        return self.repo == other.repo and self.branch == other.branch
 
 
 class ServiceConfig(Config):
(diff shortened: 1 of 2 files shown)


betulependule (approved):
Looks good to me, thank you 🙏

PR #3184, opened 2026-09-03
fullsend: review on-demand only

TomasTomecek:
TODO:

- [ ] Let's see if fullsend likes this change :)

RELEASE NOTES BEGIN

Internal change

RELEASE NOTES END


.github/workflows/fullsend.yaml
```diff
@@ -33,15 +33,22 @@ on:
...
     types: [submitted]
 
 jobs:
   dispatch:
     if: >-
-      github.event_name != 'issue_comment'
-      || github.event.comment.user.type != 'Bot'
+      (github.event_name != 'issue_comment'
+        || (github.event.comment.user.type != 'Bot'
+          && github.event.comment.body != '/fs-review'))
```
betulependule, line 47: Now that I think about it, do we want to somehow restrict who can trigger a review from fullsend (similarly how it's done for stop-fix bellow)? I'm thinking how likely it is that someone would use their non-bot account to spam '/fs-review' across PRs in all our repos where fullsend is enabled (could cost us a lot).

TomasTomecek: thank you for reminding me this Bety, it was definitely something I wanted to look into

TomasTomecek: good news! fullsend 0.40 already has this built in (thanks to Codex for checking it):

Fullsend v0.40 already restricts slash commands based on GitHub repository permissions, so we do not need an additional GitHub Actions check to prevent outside contributors from invoking agents.

| Command | Required repository role |
| --- | --- |
| `/fs-review`, `/fs-triage` | `triage` or higher |
| `/fs-code`, `/fs-fix`, `/fs-retro`, `/fs-prioritize` | `write`, `maintain`, or `admin` |

Bot comments are rejected. The `ready-for-review` label trigger is safe as well: applying a PR label requires write access, and our workflow additionally ignores bot-applied labels.

One terminology caveat: Fullsend does not use GitHub’s `CONTRIBUTOR` association. That only indicates someone has contributed before; it is not an authorization role. Fullsend uses repository collaborator permissions instead, which is the appropriate boundary here.

betulependule (approved):
Looks good to me, thanks 🙏

mfocko (approved):
LGTM, thanks a lot.

You’re updating the workflows, need to merge manually.

PR #3186, opened 2026-09-13
Show a command's sub-arguments with /packit help <command>

### LINKED ISSUE #3073: Add arguments to the `/packit help` command
As of now, Packit responds to the `/packit help` command in the comments of issues and PRs. It generates a response message containing the usage and a list of supported arguments.

However, it shows only top-level arguments available, so the user wouldn't be informed about sub-arguments such as the `--commit` sub-argument, which can be used following the `copr-build` argument like `/packit copr-build --commit 123456`.

The help command should support listing sub-argument. For example, the command to retrieving the sub-arguments of the `copr-build` job would be:

`/packit help build`

The implementation should be simple. The `GitCommentHelpHandler` would need to be editted. The `format_help` method can still be used, but it has to be called on the relevant parser object (the relevant sub-parser of the top-level parser). For example, the "copr-build" subparser would need to be retrieved in the case of sub-arguments relevant to `/packit copr-build`.

In case an argument doesn't support any sub-argument (for example `rebuild-failed`), Packit should still respond to `/packit help rebuild-failed`, but should make it clear to the user that there are no sub-arguments available.

burakeyler:
Fixes #3073

## What

`/packit help <command>` (also `/packit-stg`, `/packit-ci`, `/packit-ci-stg`) now replies with the help of that one command:

```
/packit help build
usage: /packit copr-build [--commit COMMIT] [--release RELEASE]

options:
  --commit COMMIT    run Copr build jobs configured with the commit trigger
  --release RELEASE  run Copr build jobs configured with the release trigger
```

```
/packit help rebuild-failed
/packit rebuild-failed has no sub-arguments.
```

- `packit_service/utils.py`: new `get_subcommand_help(parser, command)` finds the parser's subparsers action and calls `format_help()` on the relevant sub-parser, as suggested in the issue. Aliases (`build` → `copr-build`) resolve to the same parser. Returns `None` for an unknown command.
- `GitCommentHelpHandler.run`: reads the command tokens with the existing `get_packit_commands_from_comment(comment, prog)`; for `help <command>` it uses the sub-command help, falling back to the full help for anything unrecognised. The epilog is appended as before.

### One related change

Sub-parsers created by `add_parser` get `-h/--help` by default even though the top-level parser uses `add_help=False`. That would have shown `-h, --help  show this help message and exit` in every command's reply (meaningless in a comment), and it made "no sub-arguments" undetectable since every command had at least `-h`. Both `add_subparsers` calls now pass a `parser_class` that creates sub-parsers with `add_help=False` and the same `RawTextHelpFormatter`. The top-level `/packit help` output is unchanged.

## Tests

Five new tests in `tests/unit/test_comment_parsers.py`: sub-arguments listed (and no `--help`), alias resolves, a command without sub-arguments, unknown command → `None`, and the Fedora CI parser.

`tests/unit/test_comment_parsers.py`: 40 passed (35 existing + 5 new), `ruff check` / `ruff format --check` clean.

Caveat: I'm on Windows, where the full dependency set (Kerberos etc.) doesn't install, so I ran this file with the two parser fixtures provided directly rather than through `tests/unit/conftest.py`, and did not run the integration tests for the help handler (`tests/integration/test_pr_comment.py`). CI should cover those.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


tests/unit/test_comment_parsers.py
```diff
@@ -310,3 +315,38 @@ def test_test_eln_branch_comment_fedora_ci(comment_parser_fedora_ci):
...
     assert args.check_target == "eln"
+
+
+def test_subcommand_help_lists_sub_arguments():
+    help_text = get_subcommand_help(get_comment_parser(prog="/packit"), "copr-build")
+
+    assert help_text.startswith("usage: /packit copr-build")
+    assert "--help" not in help_text
+    assert "--commit" in help_text
+    assert "--release" in help_text
```
betulependule, line 326: I noticed that none of the tests try to run `parse_args` like the previous tests do. So I tried adding:

```
commands = get_packit_commands_from_comment("/packit help build", packit_comment_command_prefix)

args = comment_parser.parse_args(commands)
assert args.command == "help"
```

And here is a snippet of the error that the test raised:

> /usr/lib64/python3.14/argparse.py:2012: in parse_args
>     self.error(msg)
>         args       = Namespace(package=None, command='help')
>         argv       = ['build']
>         msg        = 'unrecognized arguments: build'
>         namespace  = None
>         self       = ArgumentParser(prog='python3 -m pytest', usage=None, description=None, formatter_class=<class 'argparse.RawTextHelpFormatter'>, conflict_handler='error', add_help=False)
>

This isn't a big deal, because `arg_parse()` is called in `SteveJobs.parse_comment()`, which these kinds of commands ("/packit-stg help test" and similar) never reach. However, I would still prefer to edit the parsers to actually support sub-commands for the help command so that if anyone ever reads the code definition of the parser, they will know that commands such as "/packit-stg help test" are supported.

An argument with `nargs='?'` should be added to the help subparser:
https://github.com/packit/packit-service/blob/c1f019130b4e81a0c366a4886230a0115967eee8/packit_service/utils.py#L265-L268

Same for the help subparser in `get_comment_parser_fedora_ci()`.

packit_service/worker/handlers/forges.py
```diff
@@ -373,10 +374,15 @@ def run(self) -> TaskResults:
         parser = parser_func(prog=prog, description=HELP_COMMENT_DESCRIPTION)
         epilog = HELP_COMMENT_EPILOG.format(note=epilog_note, docs_url=docs_url)
 
+        # `/packit help build` shows the sub-arguments of one command; anything
+        # it does not recognise falls back to the full help.
+        help_text = parser.format_help()
```
betulependule, line 379: I noticed, when running this on a local deployment, that when I tried to run this handler for "/packit-ci-stg help test", the usage generated was as follows:

```
usage: /packit-ci-stg test [--target [{eln,rawhide}]] [test_identifier]

positional arguments:
  test_identifier       specific type of tests to run

options:
  --target [{eln,rawhide}]
                        target for which to trigger tests
```

`test_identifier` should include a list of supported values, but doesn't list them at the moment. This is because the Fedora CI parser is currently initialized without them. Previously, this handler didn't need to initialize the parser with these values because it didn't support showing the sub-commands (so these values wouldn't be shown).

Above, `parser_func(prog=prog, description=HELP_COMMENT_DESCRIPTION)` is called and it needs to be initialized with `supported_test_types` in case of the Fedora CI parser. Perhaps use `**kwargs` and set `supported_test_types` in the two if branches where the comments starts with `fedora-ci` and `fedora-ci-stg`? There are many ways of doing this, I think.

The list can be retrieved like this:

https://github.com/packit/packit-service/blob/c1f019130b4e81a0c366a4886230a0115967eee8/packit_service/worker/jobs.py#L273-L278

Perhaps add a new method to `GitCommentHelpHandler`, which calls the code above and returns `supported_test_types`?

betulependule, line 379: I'm also thinking whether we could also mention the expected format of the `targets` argument when "/packit help test" (or "/packit-stg help test") is used. [As our docs state](https://packit.dev/docs/retriggering#running-tests-with-builds-from-another-pull-request), the expected format is `<namespace>/<repo>#<pr_id>`, so it would be nice to mention it in the usage.

This is the current output when "/packit-stg help test" is used:
```
usage: /packit-stg test [--commit COMMIT] [--release RELEASE]
                        [--identifier IDENTIFIER] [--labels LABELS]
                        [--env ENV]
                        [targets ...]

positional arguments:
  targets               reference(s) to PR(s) in different repositories containing builds to test


options:
  --commit COMMIT       run tests configured with the commit trigger
  --release RELEASE     run tests configured with the release trigger
  --identifier, --id, -i IDENTIFIER
                        identifier of job for which to run tests
  --labels LABELS       comma-separated list of labels identifying tests to run
  --env ENV             environment variables
```

betulependule (commented):
Hello, thank you for taking an interest in Packit and opening this PR. I tested this on a local deployment and the changes are functional.  I left a couple of comments regarding small things I believe could make a small improvement. Overall, this PR looks good. I didn't find any major issues.

burakeyler:
Thanks for testing this on a local deployment and for the detailed feedback, @betulependule!

Addressed the first point in d871154: the `help` subparser in both `get_comment_parser()` and `get_comment_parser_fedora_ci()` now accepts an optional `command_arg` positional (`nargs=?`), named `command_arg` rather than `command` to avoid clobbering the top-level subparsers' `dest=command`. Added `parse_args()` round-trip tests for `/packit help build` and `/packit-ci help test`.

Verified with a standalone argparse harness (unable to run the full suite in this environment — see below) that:
- `/packit help build` -> `command='help'`, `command_arg='build'`
- `/packit help` (no sub-arg) -> `command_arg=None`
- `/packit build --commit abc` and other existing commands are unaffected
- `/packit-ci help test` -> `command='help'`, `command_arg='test'`
- `/packit-ci test rpmlint` still resolves `test_identifier` correctly

`ruff check` and `ruff format --diff` are clean on both changed files. I could not run the actual pytest suite here since `packitos` (a core dependency) needs `krb5.h`/kerberos headers to build and this sandbox has no Docker to use `make build-test-image` / `check-in-container` as CONTRIBUTING.md describes — please have CI confirm.

On your other two points — passing `supported_test_types` into the Fedora CI parser used for `/packit-ci help test`, and mentioning the `targets` format in the `test` help text — I'd rather not guess at the design here given you flagged multiple ways to do it (e.g. new `GitCommentHelpHandler` method vs. `**kwargs`, and where exactly to source `supported_test_types` from `jobs.py`). Happy to take a stab at either if you can confirm the approach you'd prefer, or I can leave those as a intentionally separate follow-up.

betulependule:
@burakeyler, thank you for the latest changes. They look ok.

When it comes to the `supported_test_types`, I was thinking of adding a method like this one to `GitCommentHelpHandler` (note that the variables are named a bit different in `GitCommentHelpHandler` than in the context of the code that I linked before):

```
    def get_fedora_ci_tests(self) -> list[str]:
        return DownstreamTestingFarmJobHelper.get_fedora_ci_tests(
            self.service_config,
            self.data.project,
            self.data,
            filter_specific_tests=False,
        )
```

Then we can use `kwargs` to pass `supported_test_types` to the parser, which needs this list, like this:
```
        kwargs: dict[str, Any] = {}

        if comment.startswith("/packit-ci-stg"):
            prog = HELP_COMMENT_PROG_FEDORA_CI_STG
            parser_func = get_comment_parser_fedora_ci
            epilog_note = self.get_epilog_note_fedora_ci()
            docs_url = DOCS_URL_FEDORA_CI
            kwargs["supported_test_types"] = self.get_fedora_ci_tests()
        elif comment.startswith("/packit-ci"):
            prog = HELP_COMMENT_PROG_FEDORA_CI
            parser_func = get_comment_parser_fedora_ci
            epilog_note = self.get_epilog_note_fedora_ci()
            docs_url = DOCS_URL_FEDORA_CI
            kwargs["supported_test_types"] = self.get_fedora_ci_tests()
        elif comment.startswith("/packit-stg"):
            prog = HELP_COMMENT_PROG_STG
            parser_func = get_comment_parser  # type: ignore[assignment]
            epilog_note = self.get_epilog_note()
            docs_url = DOCS_URL
        else:
            prog = HELP_COMMENT_PROG
            parser_func = get_comment_parser  # type: ignore[assignment]
            epilog_note = self.get_epilog_note()
            docs_url = DOCS_URL

        # Use determined parameters to create parser and epilog
        parser = parser_func(prog=prog, description=HELP_COMMENT_DESCRIPTION, **kwargs)
```

The `forges.py` file would also need this import somewhere at the top: `from packit_service.worker.helpers.testing_farm import DownstreamTestingFarmJobHelper`.

And as for the `targets` format, that would just involve making a small change in `get_comment_parser()`. Maybe just add the expected format mention in brackets? I'm fine with other solutions.

```
    test_parser.add_argument(
        "targets",
        nargs="*",
        help="reference(s) to PR(s) in different repositories containing builds to test \n"
        "(expected format: <namespace>/<repo>#<pr_id>)",
    )
```

burakeyler:
Thanks for the detailed suggestions, @betulependule — pushed 6cf13c7 implementing both.

- Added `GitCommentHelpHandler.get_fedora_ci_tests()` exactly as you sketched, and pass its result as `supported_test_types` into `get_comment_parser_fedora_ci()` for both the `/packit-ci` and `/packit-ci-stg` branches (not `/packit`/`/packit-stg`, which use the non-Fedora-CI parser).
- Added the `<namespace>/<repo>#<pr_id>` format hint to the `targets` argument's help text in `get_comment_parser()`.

Verified with a standalone harness (`get_comment_parser_fedora_ci(prog="/packit-ci", supported_test_types=["rpmlint","rpminspect"])` then `get_subcommand_help(parser, "test")`) that `/packit-ci help test` now renders the positional as `{rpmlint,rpminspect}` instead of the generic `test_identifier`, and that the `targets` help text carries the new format note.

Also added `test_pr_comment_help_handler_fedora_ci_test_choices` to `tests/integration/test_pr_comment.py`, mocking `DownstreamTestingFarmJobHelper.get_fedora_ci_tests` and asserting the posted help comment for `/packit-ci help test` contains the mocked test type names.

I could not run the actual pytest suite or mypy here — `packit_service/events/koji/tag.py` pulls in `koji`, which imports the Unix-only `pwd`/`resource` stdlib modules and ultimately `gssapi`, which requires a full MIT Kerberos for Windows install (system-level, not something I want to do in this sandbox); mypy here also errors inside an unrelated `click` dependency when checking against the pinned Python 3.9 target. `ruff check` and `ruff format --diff` are clean on all three changed files. Please have CI confirm the actual test run.

mfocko (commented):
LGTM. Thanks a lot for the contribution <3

I leave the merging to @betulependule once she decides it’s ready to merge 🫡

packit_service/worker/handlers/forges.py
```diff
@@ -371,7 +375,7 @@ def run(self) -> TaskResults:
             docs_url = DOCS_URL
 
         # Use determined parameters to create parser and epilog
-        parser = parser_func(prog=prog, description=HELP_COMMENT_DESCRIPTION)
+        parser = parser_func(prog=prog, description=HELP_COMMENT_DESCRIPTION, **kwargs)
```
betulependule, line 387: I didn't realize previously that calling `self.get_fedora_ci_tests()` in the if branches above would result in an API call whenever "/packit-ci(-stg) help" is called (and not just "/packit-ci(-stg) help test").

Let's make sure to only call it when necessary. Perhaps add these lines of code above line 377.

```
requested = get_packit_commands_from_comment(comment, prog)
requested_subcommand = requested[1] if len(requested) > 1 and requested[0] == "help" else None
if requested_subcommand == "test" and parser_func is get_comment_parser_fedora_ci:
    kwargs["supported_test_types"] = self.get_fedora_ci_tests()
```

This way, `kwargs["supported_test_types"] = self.get_fedora_ci_tests()` can be removed from the if branches above. The current line 384 can then be removed and 385 can be simplified to:

```
if requested_subcommand:
```

Hopefully this should be the last change request from me. :) I tested the code on a local deployment this morning and everything looks good otherwise.

burakeyler:
Good catch, thank you — pushed a83ed7d.

The requested subcommand is now worked out before the parser is built, and `supported_test_types` is only passed when the comment is `/packit-ci(-stg) help test`, so a plain `/packit-ci help` costs no API call. The two `kwargs["supported_test_types"] = ...` lines in the prefix branches are gone, and the later block is down to `if requested_subcommand:`, as you suggested.

Added `test_pr_comment_help_handler_fedora_ci_skips_test_lookup`, which asserts `get_fedora_ci_tests` is **never** called for `/packit-ci help`, next to the existing test that checks the choices do show up for `/packit-ci help test`. `ruff check` and `ruff format` are clean.

Thanks also @mfocko for the review.

betulependule (approved):
LGTM. Thank you. Could you please just clean up the commits? I'll be happy to merge it afterwards. I feel like some of the commits can be squished, but I'll leave it up to you. I also noticed that commit 6cf13c792b8e90c2c3a2c8b22a1617f3b2afcb63 includes the change adding `<namespace>/<repo>#<pr_id>`, which is unrelated to the Fedora CI parser like the commit title would make it seem, so it might be good to reword it a little if you're going to keep the commit.

burakeyler:
Cleaned up the history as requested — squashed the 6 commits down to 3, and reworded the one you flagged so its title matches its content:

1. `Show a command's sub-arguments with /packit help <command>` — the core feature, the mypy typing fix, and the optional `command_arg` positional, all folded together (they're one change: making `/packit help <command>` work).
2. `Document the targets argument's expected PR reference format` — pulled the `<namespace>/<repo>#<pr_id>` hint out on its own, since it isn't part of the Fedora CI parser work.
3. `Pass Fedora CI test choices into the /packit-ci help parser` — the Fedora CI test-type wiring, the test-stub fix, and the lazy-lookup optimization, squashed together.

No code changes — the final tree is byte-identical to the previous head (`a83ed7d`); this is a pure history rewrite (force-pushed). `ruff check` on the touched files is clean.

betulependule:
Hello @burakeyler, thank you for all your work. I did some final testing on a local deployment and everything looks good now, so I'll rebase the branch on main and merge afterwards. Thank you once again.

Issue #2969, opened 2026-01-28
Unify the process of parsing comments


betulependule:
There is a redundancy in the parsing of comments at the moment. Comments are first parsed by `SteveJobs`. Historically, this parsing was always done to retrieve the triggering command (copr-build, test, propose-downstream...) and all other parameters were disregarded. 

`ArgumentParser` from `argparse` is now being used by `SteveJobs`. This change has been added because of a new argument (`--package`) that needs to be processed earlier then all of the other arguments, which meant that the initial parsing would no longer be as trivial as before.  The currently used checks the syntax of comments, processes all arguments and (wastefully) discards most arguments. That is because individual handlers are implemented to parse the original raw comment once again to retrieve these arguments later on (once these handlers are run).

It would be benefical to, instead of parsing the comments inside the parsers, to simply store the comment parsed by `ArgumentParser` and pass these parsed arguments to the handlers. This change would lead to deleting the parsing functions inside handlers which base their behavior based on arguments (like test, build and more). The `ParsedComment` class should be used to store the parsed comment.

betulependule:
Related issue: https://github.com/packit/packit-service/issues/3185

PR #3188, opened 2026-09-18
Support re-ordering of `/packit-ci test` arguments to ensure consistency with the help command's output

betulependule:
Currently, users are able to use the `/packit-ci help test` command, which outputs the usage of `/packit-ci test` (implemented in https://github.com/packit/packit-service/pull/3186). It starts like this:

```
usage: /packit-ci test [--target [{eln,rawhide}]]
                       [{installability,rpminspect,rpmlint,
                       rmdepcheck,license-validate}]
```

With the addition of this functionality of the help command, users could end up being confused when trying to run `/packit-ci test --target eln rpmlint` and similar commands. That is because Packit currently supports only the following order of arguments: `/packit-ci test rpmlint --target eln`.

This PR is a quick fix so that we avoid having confused users. It just edits the parsing of comments in `DownstreamTestingFarmJobHelper` so that the order of arguments doesn't matter. The parsing is duplicated (has been like this for a long time), so for the future, it would be ideal to remove this duplication and unify parsing as described in this issue: https://github.com/packit/packit-service/issues/2969. I expect that work related to this (pre-existing) technical debt to be non-trivial.


lbarcziova (approved):
looks good!

PR #3191, opened 2026-09-22
Add endpoint for displaying multiple groups of log detective runs

Jany26:
Adding `/api/log-detective/groups?page=X&per_page=Y` endpoint (works similarly to `/api/log-detective/`) that shows multiple **groups** of Log Detective runs

This is to improve the dashboard readability, so that we can show the Log Detective runs from the same trigger on the same line.

I opted to not modify the original endpoint, but dashboard is probably the only consumer.

Merge before https://github.com/packit/dashboard/pull/544

RELEASE NOTES BEGIN

Exposing an endpoint to get a list of groups of Log Detective runs.

RELEASE NOTES END


jpodivin (approved):

mfocko (approved):
LGTM

